PR #3, opened 2019-11-14
Add initial support for python-language-server

clason:
Get it working; can be improved!

### DIFF
--- a/lua/nvim_lsp/pyls.lua
+++ b/lua/nvim_lsp/pyls.lua
@@ -10,6 +10,7 @@ skeleton.pyls = {
     filetypes = {"python"};
     root_dir = function() return cwd end;
     log_level = lsp.protocol.MessageType.Warning;
+    settings = {};
   };
   -- on_new_config = function(new_config) end;
   -- on_attach = function(client, bufnr) end;
@@ -18,11 +19,47 @@ skeleton.pyls = {
 https://github.com/palantir/python-language-server
 
 python-language-server, a language server for Python
-]];
+
+the following settings (with default options) are supported:
+```lua
+settings = {
+  pyls = {
+    enable = true;
+    trace = { server = "verbose"; };
+    commandPath = "";
+    configurationSources = { "pycodestyle" };
+    plugins = {
+      jedi_completion = { enabled = true; };
+      jedi_hover = { enabled = true; };
+      jedi_references = { enabled = true; };
+      jedi_signature_help = { enabled = true; };
+      jedi_symbols = {
+        enabled = true;
+        all_scopes = true;
+      };
+      mccabe = {
+        enabled = true;
+        threshold = 15;
+      };
+      preload = { enabled = true; };
+      pycodestyle = { enabled = true; };
+      pydocstyle = {
+        enabled = false;
+        match = "(?!test_).*\\.py";
+        matchDir = "[^\\.].*";
+      };
+      pyflakes = { enabled = true; };
+      rope_completion = { enabled = true; };
+      yapf = { enabled = true; };
+    };
+  };
+};
+```
+    ]];
     default_config = {
       root_dir = "vim's starting directory";
     };
   };
-}
+};
 
 -- vim:et ts=2 sw=2
--- a/lua/nvim_lsp/pyls.lua
+++ b/lua/nvim_lsp/pyls.lua
@@ -1,0 +1,28 @@
+local skeleton = require 'nvim_lsp/skeleton'
+local util = require 'nvim_lsp/util'
+local lsp = vim.lsp
+
+local cwd = vim.loop.cwd()
+
+skeleton.pyls = {
+  default_config = {
+    cmd = {"pyls"};
+    filetypes = {"python"};
+    root_dir = function() return cwd end;
+    log_level = lsp.protocol.MessageType.Warning;
+  };
+  -- on_new_config = function(new_config) end;
+  -- on_attach = function(client, bufnr) end;
+  docs = {
+    description = [[
+https://github.com/palantir/python-language-server
+
+python-language-server, a language server for Python
+]];
+    default_config = {
+      root_dir = "vim's starting directory";
+    };
+  };
+}
+
+-- vim:et ts=2 sw=2
--- a/README.md
+++ b/README.md
@@ -55,17 +55,9 @@ In progress:
 
 From vim:
 ```vim
-call nvim_lsp#texlab({})
-call nvim_lsp#gopls({})
-
-" These are still TODO, but will be done.
-call nvim_lsp#clangd({})
-call nvim_lsp#ccls({})
-call nvim_lsp#tsserver({})
-
-" Or using a dynamic name.
 call nvim_lsp#setup("texlab", {})
 call nvim_lsp#setup("gopls", {})
+call nvim_lsp#setup("pyls", {})
 ```
 
 From Lua:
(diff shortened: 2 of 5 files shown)


lua/nvim_lsp/pyls.lua
```diff
@@ -0,0 +1,28 @@
...
+local util = require 'nvim_lsp/util'
+local lsp = vim.lsp
+
+local cwd = vim.loop.cwd()
+
+skeleton.pyls = {
+  default_config = {
+    cmd = {"pyls"};
+    filetypes = {"python"};
+    root_dir = function() return cwd end;
```
norcalli: Does pyls prefer to have one per project or one per editor?

clason: That's a good question -- I don't know... It's unclear from their repo. 

Any way for me to easily test this? (i.e., what should I try to see if it breaks?)

lua/nvim_lsp/pyls.lua
```diff
@@ -0,0 +1,28 @@
...
+  -- on_new_config = function(new_config) end;
+  -- on_attach = function(client, bufnr) end;
+  docs = {
+    description = [[
+https://github.com/palantir/python-language-server
+
+python-language-server, a language server for Python
+]];
+    default_config = {
+      root_dir = "vim's starting directory";
```
norcalli: Do you think it would be a good idea to explain why we made this choice? i.e. that it makes only one server per neovim session?

lua/nvim_lsp/pyls.lua
```diff
@@ -0,0 +1,28 @@
...
+
+local cwd = vim.loop.cwd()
+
+skeleton.pyls = {
+  default_config = {
+    cmd = {"pyls"};
+    filetypes = {"python"};
+    root_dir = function() return cwd end;
+    log_level = lsp.protocol.MessageType.Warning;
+  };
```
norcalli: We may want to include the default settings available so that users know what they can share. We could do this either via putting those default settings, or putting it in the docs below? https://github.com/neoclide/coc.nvim/wiki/Language-servers#python

lua/nvim_lsp/pyls.lua
clason: To be honest, I've never worried about that -- I just started the server as advised in vim-lsp docs and went about my day. (But then, I'm only working with single-file scripts.)

norcalli:
Thanks :) Just a few questions which I'm also unsure of myself. We'll get this merged either way, because it's good for now.

lua/nvim_lsp/pyls.lua
clason: Oh, I haven't seen this. That's a lot...

Either way is fine with me; what would you prefer?

clason:
Literally took me only a minute, thanks to your templates :)

lua/nvim_lsp/pyls.lua
norcalli: I think it'll work fine either way, but if you do goto definition, it might consider two different directories as part of the same project. Probably fine, though. We'll change it if someone suggests otherwise.

lua/nvim_lsp/pyls.lua
norcalli: I do want users to be aware of the options. I always find it annoying when I don't know what I can set, so I think we should provide whatever info we can. My only hesitation on including settings was if they end up breaking between versions, but I think that it's probably fine. Then again, settings specified by the language server act as a complete override, which may surprise users who have home dir configs.

Let's put it in the docs, and try to keep it to as few lines as possible. if you just collapse the larger arrays

          "jedi_references": {
            "enabled": true
          },

into

          "jedi_references": { "enabled": true },

It'll look fine. I'd also double check that all the options you can find are included on a best effort basis, because that coc.nvim doc could be outdated.

norcalli: Oh and try to convert the json syntax into lua syntax.

```
{commandPath = "",configurationSources = { "pycodestyle" },enable = true,plugins = {jedi_completion = {enabled = true},jedi_hover = {enabled = true},jedi_references = {enabled = true},jedi_signature_help = {enabled = true},jedi_symbols = {all_scopes = true,enabled = true},mccabe = {enabled = true,threshold = 15},preload = {enabled = true},pycodestyle = {enabled = true},pydocstyle = {enabled = false,match = "(?!test_).*\\.py",matchDir = "[^\\.].*"},pyflakes = {enabled = true},rope_completion = {enabled = true},yapf = {enabled = true}},trace = {server = "verbose"}}
```

here's a starter. Just put something like "available settings" and then finally add `settings = {};` as a place holder in the default_config so that users know that's where they can put settings.

norcalli:
@clason yeah that was my goal :D thanks for being the first

lua/nvim_lsp/pyls.lua
clason: Wouldn't lua syntax be 
```lua
settings = {
       pyls = {
         enable = true;
         trace = { server = "verbose"; };
         commandPath = "";
         configurationSources = [ "pycodestyle" ];
         plugins = {
           jedi_completion = { "enabled" = true; };
           jedi_hover = { "enabled" = true; };
           jedi_references = { "enabled" = true; };    
           jedi_signature_help = { "enabled" = true; };
           jedi_symbols = {
             enabled = true;
             all_scopes = true;
           };
           mccabe = {
             enabled = true;
             threshold = 15;
           };
           preload = { enabled = true; };
           pycodestyle = { enabled = true; };
           pydocstyle = {
             enabled = false;
             match = "(?!test_).*\\.py";
             matchDir = "[^\\.].*";
           };
           pyflakes = { enabled = true; };
           rope_completion = { enabled = true; };
           yapf = { enabled = true; };
         };
       };
```

clason: I'd just dump that in the doc for easier cut&paste.

norcalli: Yeah, by syntax, I meant table syntax. I just meant to give you a quick starting point because I can quickly convert things, and I left indentation up to you :P 

Except you have some mistakes since that's not valid Lua for the `"enabled"` and the array style in `configurationSources`.

I literally evaluated `vim.fn.json_decode(...)` and pasted the result for you to hopefully save time.

lua/nvim_lsp/pyls.lua
clason: The only difference is setting the root dir, right? Not sure what the canonical choice would be there.

lua/nvim_lsp/pyls.lua
clason: Nice ;) But doing it by hand, I learn something. (For example, not to forget half of them...)

I can't find a proper documentation for pyls :/

norcalli:
Don't forget to regen the docs so you can preview the README.

clason:
Hmm, that's not doing anything for some reason....

lua/nvim_lsp/pyls.lua
clason: @norcalli where exactly should I put that? in the docstring between ``[[ ]]``, or in the `default_config` below?

norcalli:
@clason yeah for me too....

norcalli:
@clason K. I got it to work. Apply this patch/make these changes:

```diff
diff --git a/scripts/docgen.lua b/scripts/docgen.lua
index 7b33854..a2583a3 100644
--- a/scripts/docgen.lua
+++ b/scripts/docgen.lua
@@ -1,3 +1,4 @@
+require 'nvim_lsp'
 local skeleton = require 'nvim_lsp/skeleton'
 local inspect = vim.inspect
 
@@ -38,7 +39,7 @@ local function indent(n, s)
   return table.concat(lines, '\n')
 end
 
-local writer = io.popen("cat README_preamble.md - > README.md  common-lsp-docs.md", "w")
+local writer = io.popen("cat README_preamble.md - > README.md", "w")
 
 for k, v in pairs(skeleton) do
   local tconf = v.template_config
```

Then do `nvim-master -u NONE +'set rtp+=$PWD' +'luafile scripts/docgen.lua' +q`

norcalli:
@clason you had it right the first time. Putting it in `default_config.settings` will make people think we send those by default. I like the original note you had.

clason:
Ah, that gets me a `settings = table: 0x406791d0` ;) I'll revert to what I had before, also for the reason you note.

clason:
d'oh!

clason:
First try \o/ :wink:

lua/nvim_lsp/pyls.lua
```diff
@@ -10,19 +10,55 @@ skeleton.pyls = {
...
     root_dir = function() return cwd end;
     log_level = lsp.protocol.MessageType.Warning;
+    settings = {};
   };
   -- on_new_config = function(new_config) end;
   -- on_attach = function(client, bufnr) end;
   docs = {
     description = [[
-https://github.com/palantir/python-language-server
+    https://github.com/palantir/python-language-server
```
norcalli: One final nit, I swear:

Remove the indentation on this docs string and use code fence for the settings part, e.g.

    ````
    settings = {
      ...
    }
    ```

clason: No problem! Sorry for the back-and-forth. Are the four backticks at the start on purpose? Should I give a `lua` specifier?

norcalli: no it was because I was a dumb, and yeah a lua specifier should be fine.

At least now, you probably know what to do for other servers if you decide to add them :P and I can point people at your pyls.lua and say "just copy this style."

norcalli:
Thanks :)

PR #5, opened 2019-11-14
Initial add of elmLS support

megalithic:
I'm not able to fully test this at the moment. These are, roughly, the settings I'm using from coc.nvim.

PR #9, opened 2019-11-15
Add ElmLS

norcalli:
Continuation of #5 

closes #5


norcalli:
Still TODO would be to check for updates and warn the user if there are updates when an elm server is started.

PR #10, opened 2019-11-15
feat(lsp): add rust-analyzer

fannheyward:


norcalli:
This is a little too minimal for what we are trying to provide. For instance, rust-analyzer has settings that aren't included or described here: https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user.

I'd like to see if this can be filled out with more information that's available. See some of the other implementations like elmls or pyls for an example of documentation of settings.

And if you aren't comfortable enough with Lua to add that, then at least the research on what commands rust analyzer provides, what edge cases there are, the settings, etc would be good to provide here and someone else can take over.

fannheyward:
Yes, this is a very initial support for RA, I'll add some settings and docs later.

For commands, here is a list has built with coc.nvim extension, https://github.com/fannheyward/coc-rust-analyzer#commands, something are client side only, I'll try to implement if I can.

fannheyward:
@norcalli I've added some settings from RA docs.

README.md
```diff
@@ -35,12 +35,12 @@ Implemented:
 - [elmls](https://github.com/neovim/nvim-lsp#elmls)
 - [gopls](https://github.com/neovim/nvim-lsp#gopls) (has some errors)
 - [pyls](https://github.com/neovim/nvim-lsp#pyls)
+- [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer)
```
norcalli: This should go in README_preamble.md. README.md is autogenerated.

Also it is meant to be a link to the section in the README, not a link to the project. To be honest, I'll probably auto generate this part eventually.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -0,0 +1,49 @@
+-- Author: Heyward Fann <fannheyward@gmail.com>
```
norcalli: This is unnecessary since we're all going to maintain this file going forward.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -0,0 +1,49 @@
...
+https://rust-analyzer.github.io
+
+Bringing great IDE experience to the Rust programming language.
+
+the following settings (with default options) are supported:
+```lua
+settings = {
+  ["rust-analyzer"] = {
+    trace = { server = "verbose"; };
+    excludeGlobs = {};
```
norcalli: There are many more flags available:

rust-analyzer.highlightingOn: enables experimental syntax highlighting
rust-analyzer.enableEnhancedTyping: by default, rust-analyzer intercepts Enter key to make it easier to continue comments. Note that it may conflict with VIM emulation plugin.
rust-analyzer.raLspServerPath: path to ra_lsp_server executable
rust-analyzer.enableCargoWatchOnStartup: prompt to install & enable cargo watch for live error highlighting (note, this does not use rust-analyzer)
rust-analyzer.excludeGlobs: a list of glob-patterns for exclusion (see globset docs for syntax). Note: glob patterns are applied to all Cargo packages and a rooted at a package root. This is not very intuitive and a limitation of a current implementation.
rust-analyzer.useClientWatching: use client provided file watching instead of notify watching.
rust-analyzer.cargo-watch.command: cargo-watch command. (e.g: clippy will run as cargo watch -x clippy )
rust-analyzer.cargo-watch.arguments: cargo-watch check arguments. (e.g: --features="shumway,pdf" will run as cargo watch -x "check --features="shumway,pdf"" )
rust-analyzer.cargo-watch.ignore: list of patterns for cargo-watch to ignore (will be passed as --ignore)
rust-analyzer.trace.server: enables internal logging
rust-analyzer.trace.cargo-watch: enables cargo-watch logging
RUST_SRC_PATH: environment variable that overwrites the sysroot
rust-analyzer.featureFlags -- a JSON object to tweak fine-grained behavior:

fannheyward:
There are many setting flags available in rust-analyzer, but I can't make them works with nvim-lsp. For example:

1. `highlighting` related configs are not works on vim/neovim, it's not nvim-lsp's error, but vim/neovim limitation. RA will add additional highlight and inlay hint inside text, which is not support by vim/neovim. https://github.com/fannheyward/coc-rust-analyzer/issues/8
2. `publishDecorations` is used by inlay hint, same as 1
3. `cargo-watch` related configs are LSP client side configs, RA will not use them. `cargo-watch`  will run a progress alongside RA server, this does not use RA
4. `featureFlags/excludeGlobs/useClientWatching` configurations are used by RA server, but it's not working by `settings` section.

So I just close this PR by now.

PR #14, opened 2019-11-15
Texlab settings

clason:
Adds the complete set of texlab settings.

Also makes a small change to the docgen script to declare fenced code as lua (vim colors the `s:` in `Commands:`).


norcalli:
Thanks :)

clason:
oops, I missed two settings:
```lua
        forwardSearch = {
          command = {""},
        },
        lint = {
          onSave = true
        }
```
(although I strongly believe the latter should be `false`...)

Should I make another pull request?

PR #15, opened 2019-11-15
Add bash language server

DanielVoogsgerd:
Short implementation for the bash language server.
Feedback is welcome.


norcalli:
Since I'm about to improve installation support, I'll add that myself. Otherwise, the PR looks good, except that instead of modifying `README.md`, you should modify `README_preamble.md`, but that seems to confuse just about everyone, so that's not really your fault :sweat_smile: 

Thanks!

DanielVoogsgerd:
That would make sense. I just copied what the others were doing.

Maybe it's a good idea to make a short "guide" as to how to contribute.

norcalli:
Did you see CONTRIBUTING.md?

DanielVoogsgerd:
Apologies, completely missed it.

Nice job in general btw!

PR #16, opened 2019-11-15
Add Javascript support

jssee:
Initially meant to add this to the previous pull request but it got merged. I could see an argument for making this optional, since someone might want to use an different server client for their javascript files, but im not sure how much of a concern that should be. open to hear suggestions


norcalli:
Handled in #17. My opinion is that the filetypes it uses are shown in the README and it's very easy to override. I think most people using tsserver don't also use another server (I didn't).

PR #17, opened 2019-11-16
Redo installation.

norcalli:
Servers which want to be auto installed should specify
skeleton[name].install()
and it will be automatically added to the list of installable servers.

- Add :LspInstall and :LspInstallInfo
- Auto generate docs for servers with .install() available.
- Add util.npm_installer
- Refactor utf8 capabilities common config into a single function


lua/nvim_lsp/elmls.lua
```diff
@@ -1,59 +1,27 @@
...
-
-util.create_module_commands("elmls", global_commands)
+local installer = util.npm_installer {
+  server_name = server_name;
+  packages = { "elm", "elm-test", "elm-format", "@elm-tooling/elm-language-server" };
+  binaries = {bin_name, "elm", "elm-format", "elm-test"};
+}
 
+local default_capabilities = lsp.protocol.make_client_capabilities()
+default_capabilities.offsetEncoding = {"utf-8", "utf-16"}
```
liskin, line 16: After this PR, `default_capabilities` isn't used anywhere. That might be a mistake. Perhaps `capabilities = default_capabilities` should be added to `default_config`?

PR #18, opened 2019-11-16
Add support flow language server

h-michael:
Since Flow is installed not for global but for each project, we will not provide an install command.

ref: https://flow.org/en/docs/install/, https://flow.org/en/docs/config/


norcalli:
If you've tested it and it works, then it looks good to me.

h-michael:
It works :)

PR #21, opened 2019-11-16
Add support for haskell-ide-engine

czwinzscher:
This adds support for haskell-ide-engine (https://github.com/haskell/haskell-ide-engine).
Everything works fine except that i get this error message on startup:
2019-11-16-182509_958x111_scrot
But as i said the lsp still works and i also found this: https://github.com/haskell/haskell-ide-engine#parse-errors-file-state-going-out-of-sync which could be related.


norcalli:
@czwinzscher that might be our bad. I think it was expecting an empty object, not an array. We can fix that on master soon.

norcalli:
@czwinzscher have you tested that the settings are properly reflected? I just want to make sure that it should go in init_options and not settings.

Also, this is not required for this PR, but do you think that HIE could be installed in a local directory for execution?

Also, you should merge master and revert any changes to README.md. README is auto generated.

czwinzscher:
> @czwinzscher have you tested that the settings are properly reflected? I just want to make sure that it should go in init_options and not settings.

Yes i tested it and it does have to be in init_options. Also [here](https://github.com/neoclide/coc.nvim/wiki/Language-servers#Haskell) you can see that coc also uses init options.
 
> Also, this is not required for this PR, but do you think that HIE could be installed in a local directory for execution?

Well hie has a build script that automatically installs it to  ~/.local/bin.

> Also, you should merge master and revert any changes to README.md. README is auto generated.

Okay i reverted it.

PR #22, opened 2019-11-16
Add jdtls support

DanielVoogsgerd:
Simple implementation for jdtls (Java). Install instructions are lacking, since none seem to be mentioned on the language server page and I install it myself via the AUR.


lua/nvim_lsp/jdtls.lua
```diff
@@ -0,0 +1,30 @@
...
+local cwd = vim.loop.cwd()
+
+local server_name = "jdtls"
+local bin_name = "jdtls"
+
+skeleton[server_name] = {
+  default_config = {
+    cmd = {"jdtls"};
+    filetypes = {"java"};
+    root_dir = function() return cwd end;
```
justinmk: this will always return the CWD found when `jdtls.lua` was first evaluated. Is that intentional? Instead this could be:

    root_dir = vim.loop.cmd

then the actual-current directory will be returned, after user does `:cd`.

DanielVoogsgerd: No, I just did not think or it. I like this better indeed. I'll change it in a minute.

DanielVoogsgerd: Ghhm, vim.loop.cmd seems to be nil. Unfortunately, I'm quite foreign to Lua and to vim scripting in general. Unfortunately I do not have the time right now to learn vim scripting right now. So if anybody could either fix it or tell me what should work, that would be great!

norcalli: It's `vim.loop.cwd` as in `w` not `m` :P

@justinmk The pattern of always returning a single value is to enforce that even if the cwd changes, only one instance of the server is spawned. Basically there's 2 major patterns for servers:

1. spawn a new server for each project root
2. start just one server for each neovim session

I was thinking of making this a util function called `util.once(value)` which returns the same value forever.

DanielVoogsgerd: Ah, that makes sense. Again both vim scripting and Lua are foreign to me. I just love Language server and this implementation, so I'll gladly help wherever I can.

Do I understand correctly that maybe it's not the best idea then to change the root dir if the cwd changes?
If this is the case. Do you have any further remarks on the MR? :)

norcalli: Yup that would not be good. So basically just answer the question: how many instances of the server do you want? If the answer is 1 per neovim then keep things as is (and merge origin/master).

Otherwise use something like a `rootPattern` you can see in some other lsp files.

lua/nvim_lsp/jdtls.lua
```diff
@@ -0,0 +1,30 @@
...
+local lsp = vim.lsp
+
+local cwd = vim.loop.cwd()
+
+local server_name = "jdtls"
+local bin_name = "jdtls"
+
+skeleton[server_name] = {
+  default_config = {
+    cmd = {"jdtls"};
```
norcalli: might as well use `bin_name` here if you define it above, or remove `local bin_name` above.

norcalli:
@DanielVoogsgerd if you decide that one per neovim is the right approach, then use `util.once` that I just added. See here for an example: https://github.com/neovim/nvim-lsp/pull/31/files

norcalli:
@DanielVoogsgerd oh yeah and you can just revert any changes to README_preamble. That part you changed (adding the server) is now auto generated.

norcalli:
@DanielVoogsgerd I made the changes recommended just to save you some effort, but the question still remains of whether `jdt` expects a server per project or one per editor instance.

lua/nvim_lsp/jdtls.lua
DanielVoogsgerd: One per instance seems good to me. Maybe altering the root pattern might be a good idea to do eventually, but this might be different for different frameworks and build systems. I guess this is a fine start.

DanielVoogsgerd:
@norcalli apologies. You seem to have more effort in fixing the MR than I'm actually providing :sweat_smile:

DanielVoogsgerd:
@norcalli good to know. I'll make sure to that one right the next one :wink:

DanielVoogsgerd:
@norcalli good to know. I'll make sure to do that one right the next ls 😉

norcalli:
@DanielVoogsgerd I didn't choose anything, I just cleaned it up based on the existing PR to match refactoring that I was doing at the moment.

I am leaving it up to you because I don't use jdt to decide what the best solution is for how to spawn it/root_dir. To be perfectly clear: **I have not made any decisions on this PR**. I'm still waiting for an answer on what the correct solution is and for you to implement it so that I can merge it.

DanielVoogsgerd:
> @DanielVoogsgerd I didn't choose anything, I just cleaned it up based on the existing PR to match refactoring that I was doing at the moment.

Ah, that I missed. The API is changing multiple times a day so it seems. That's fine, I was just curious

> I am leaving it up to you because I don't use jdt to decide what the best solution is for how to spawn it/root_dir. To be perfectly clear: **I have not made any decisions on this PR**. I'm still waiting for an answer on what the correct solution is and for you to implement it so that I can merge it.

Let's keep it calm. Right now I'm trying out setting the root dir with `util.root_pattern`. But JDT seems not to like it due to overlapping workspaces. Keep in mind that I'm just working on it in the spare free time that I can find. I can imagine that it seems slow compared to full-time development, but that's just the way it is. If that is a problem feel free to close the MR. I do realise that refactoring my MR is more effort than I'm providing with this MR.

teasp00n:
I think a single instance of `jdtls` per project is the right approach (so shared across neovim instances if I understood that correctly).

Also you can tell `jdtls` where to create the workspace folder so it doesn't pollute the filesystem with them.

The relevant parts of my `init.vim`:

```
let g:data_dir = expand("~/.cache/jdtls/").fnamemodify(getcwd().'/', ":h:t")
let g:LanguageClient_serverCommands = {
      \ 'java': ['jdtls', '-data', g:data_dir],
      \ }
```
That snippet is for LanguageClient-neovim just adjust it accordingly.

DanielVoogsgerd:
@teasp00n. Last commit did just that, but I think that the workspace should be located in the project root of the project. Right next to a src and bin directory.  I think this is also an approach that is quite compatible with eclipse. the thing that is missing is that is should be inside a workspace directory. This commit makes the project root itself the workspace. Which isn't terrible as far as I understand, but with the time constraints and limited understanding of Lua I got this is how far I got last night.

pwntester:
Just out of curiosity, do code action such as add imports work out of the box? We required some tweaking in vim-lsc to adapt a non-spec-compliant response (see: https://github.com/natebosch/vim-lsc/issues/88)

justinmk:
If someone can confirm that this works we can just merge it. @DanielVoogsgerd are you using this successfully?

DanielVoogsgerd:
It somewhat works. But jdtls seemed to be an odd Language Server, so it's definitely not a perfect or clean implementation. Some additional work will have to be done for code actions and maybe somebody has a better idea of where to store the jdtls directory. Since writing java is mostly code generation and since code actions aren't yet working I am not using it at the moment.

@jesse-ba

PR #30, opened 2019-11-17
Add workspace/didChangeConfiguration support.

norcalli:
- Add client.workspace_did_change_configuration()
- Notify settings on init.


norcalli:
Users have been complaining about settings not working and this might help that (such as #10 )

PR #31, opened 2019-11-17
Fix docs and add util.once

norcalli:


lua/nvim_lsp/pyls.lua
```diff
@@ -2,13 +2,11 @@ local skeleton = require 'nvim_lsp/skeleton'
...
 local lsp = vim.lsp
 
-local cwd = vim.loop.cwd()
-
 skeleton.pyls = {
   default_config = {
     cmd = {"pyls"};
     filetypes = {"python"};
-    root_dir = function() return cwd end;
+    root_dir = util.once(vim.loop.cwd());
```
justinmk: `once()` seems a bit special-case. The old code was clearer.

Is this called `lazy()` in other languages?

norcalli: I called it `once()` because it implies its usage, but you're right that the actual function would be typically called `always()` or `lazy()`. They're usually aliases for eachother.

And I would suggest that if it was clearer, then you wouldn't have suggested using `vim.loop.cwd` in the other PR :P

We can rename it to something else, but I think that the function itself is useful.

norcalli: I think I'd go with `always()`

norcalli: Oh I see, you're talking about a different function, which is `lazy(vim.loop.cwd)` which accepts a function instead. Yeah I could see that being slightly more useful if the user cds, but usually if you're using this option, then the server doesn't seem to care what rootUri/root_dir is at all

justinmk: > if it was clearer, then you wouldn't have suggested using `vim.loop.cwd` in the other PR :P

In that case I understand what it was doing, but not the intention. I would still have questioned the intention (until it was explained).

PR #33, opened 2019-11-17
Change root_dir for single instance servers

norcalli:


justinmk:
"always" makes no sense at all to me. "once" makes more sense, though I suspect this will not be used much, which is why I wondered if we need it.

norcalli:
We could change it entirely to `vim.loop.os_homedir` which is a static value.

norcalli:
Yeah that's much better.

PR #36, opened 2019-11-19
Add css lsp

mhartington:
Hello! 

I started the process of adding LSP configuration for CSS/LESS/SCSS based on this server

https://github.com/vscode-langservers/vscode-css-languageserver-bin

### DIFF
--- a/lua/nvim_lsp/cssls.lua
+++ b/lua/nvim_lsp/cssls.lua
@@ -11,11 +11,15 @@ local installer = util.npm_installer {
   binaries = {bin_name};
 }
 
+local root_pattern = util.root_pattern("package.json")
+
 skeleton[server_name] = {
   default_config = util.utf8_config {
     cmd = {bin_name, "--stdio"};
     filetypes = {"css", "scss", "less"};
-    root_dir = util.root_pattern("package.json");
+    root_dir = function(fname)
+      return root_pattern(fname) or vim.loop.os_homedir()
+    end;
     log_level = lsp.protocol.MessageType.Warning;
     settings = {
       css = { validate = true },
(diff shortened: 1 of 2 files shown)


lua/nvim_lsp/cssls.lua
```diff
@@ -0,0 +1,56 @@
...
+  server_name = server_name;
+  packages = { "vscode-css-languageserver-bin" };
+  binaries = {bin_name};
+}
+
+skeleton[server_name] = {
+  default_config = util.utf8_config {
+    cmd = {bin_name, "--stdio"};
+    filetypes = {"css", "scss", "less"};
+    root_dir = util.root_pattern("package.json");
```
jssee: I think perhaps  `root_dir = vim.loop.os_homedir;` makes more sense here in case a user is writing CSS outside of a node project, which is not uncommon.

norcalli: You could do an or condition. rootpattern or os_homedir

mhartington: something like...

```lua
root_dir = util.path.exists(util.root_pattern("package.json")) ? util.root_pattern("package.json") : vim.loop.os_homedir;
```


Not too familiar with lua, so not sure if that is correct of not.

norcalli: In lua terms, you can save the root pattern in a variable above and capture it in a function for root_dir

so

```lua

local root_pattern = util.root_pattern("package.json")

skeleton.cssls = {
  root_dir = function(fname)
    return root_pattern(fname) or vim.loop.os_homedir()
  end
}
```

mhartington: Today I learned...

Will update.

norcalli:
Thanks for using the installer too! This looks great!

PR #40, opened 2019-11-20
Replace sh with bash (fix #39)

### LINKED ISSUE #39: `set -o pipefail` fails in ubuntu
`sh` on Ubuntu points to `/bin/dash`, which has no `-o` option and fails all `LspInstall` commands.

e-kwsm:


norcalli:
As mentioned in #39, we could just get rid of `-o pipefail` and just use `set -e`. `set -e` should be available for every `sh`.

norcalli:
Closing this in favor of removing pipefail. If there are more problems we can consider reopening. Thanks anyway @e-kwsm!

PR #42, opened 2019-11-21
[RFC] Fix iterate_parents infinite loop on Windows

erw7:


norcalli:
Nice! Thanks. I admit windows testing is sorely lacking.

PR #48, opened 2019-11-23
[RFC] Support npm with config 'global=true'

zhaozg:
~~But bad performance when nvim start.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ that config.
 - [elmls](#elmls)
 - [flow](#flow)
 - [fortls](#fortls)
+- [gdscript](#gdscript)
 - [ghcide](#ghcide)
 - [gopls](#gopls)
 - [hie](#hie)
@@ -1506,6 +1507,22 @@ require'nvim_lsp'.fortls.setup{}
     settings = {
       nthreads = 1
     }
+```
+
+## gdscript
+
+https://github.com/godotengine/godot
+
+Language server for GDScript, used by Godot Engine.
+
+
+```lua
+require'nvim_lsp'.gdscript.setup{}
+
+  Default Values:
+    cmd = { "nc", "localhost", "6008" }
+    filetypes = { "gd", "gdscript3" }
+    root_dir = <function 1>
 ```
 
 ## ghcide


lua/nvim_lsp/util.lua
```diff
@@ -306,8 +307,17 @@ function M.npm_installer(config)
     post_install_script = {config.post_install_script, 's', true};
   }
 
-  local install_dir = M.path.join(base_install_dir, config.server_name)
-  local bin_dir = M.path.join(install_dir, "node_modules", ".bin")
+  local _, global = vim._system('npm config get global')
+  local install_dir, bin_dir
+  if global then
```
norcalli: This check will always succeed because `global` is a string

norcalli:
So I'm not sure I understand the purpose of this PR. Is the `global` key a standard thing or is just something that you are using to indicate to use a different directory?

If this isn't a standard npm flag, then I recommend just setting these yourself in the `on_new_config`.

Also, as far as performance goes, you're right the binary is quite slow to execute. You should make them lazy, at least, in case you have multiiple servers because this is executing for each of your npm based servers.

Here's a function you can use to do that:

```lua
local function lazy(fn)
  validate { fn = {fn, 'f'} }
  local value
  return function()
    if value == nil then value = fn() end
  return value
  end
end

local npx_global_config = M.lazy(function()
  local has_global = vim.trim(io.popen("npm config get global"):read("*a"))
  if has_global == "true" then
    return {
      has_global = true;
      install_dir = vim.trim(io.popen("npm prefix"):read("*a"));
      bin_dir = vim.trim(io.popen("npm bin"):read("*a"));
    }
  end
  return {}
end)
```

but as it is, I don't think I'll merge this PR.

zhaozg:
Ok, Let me explain.

I use nodejs/npm as tools, so I change [global](https://docs.npmjs.com/misc/config#global) to true, that make packages are installed into the prefix folder instead of the current working directory. But when global is true, LspInstall and LspInstallInfo fail. so I make the PR.

Please think about it, and very thanks for your great work for nvim.

zhaozg:
If need, visit here: https://github.com/zhaozg/nvim-lsp/tree/global

PR #49, opened 2019-11-23
Add sumneko_lua installation and util.sh helper.

norcalli:
This one is a weird one. By default there won't be a command available.

Also I added the option to download the vspackage for docgen. We'll probably use this a lot more than package.json because it's the source of truth for VSCode extensions. Also some people (like sumneko) generate their package.json, so it's not always available.


lua/nvim_lsp/sumneko_lua.lua
```diff
@@ -0,0 +1,104 @@
...
+		if not util.has_bins("sh", "chmod", "unzip") then
+			error('Need the binaries "sh", "chmod", "unzip" to install this')
+			return
+		end
+		local script = [=[
+set -e
+# Install ninja if not available.
+which ninja >/dev/null || {
+	test -x ninja || {
+		curl -LO https://github.com/ninja-build/ninja/releases/download/v1.9.0/ninja-linux.zip
```
zhaozg: Other platforms? windows, macos

norcalli: @zhaozg I could do mac and I guess windows has its subsystem. I'll consider adding those.

PR #57, opened 2019-11-27
Add Microsoft Python Language Server support

mjlbach:
This is a little rough, but it adds autocompletion/linting support as well as auto install of dotnet/the language server for a fixed version (latest). Most other LSP clients parse an XML endpoint to determine the latest version and pull; also this currently doesn't have support for installing on windows. I'm still not sure if I've completely enabled all of the LSP features.


lua/nvim_lsp/mpls.lua
```diff
@@ -0,0 +1,143 @@
...
+
+# clone project
+if [ "$(uname)" == "Darwin" ]
+  then 
+    curl -fLO https://pvsc.azureedge.net/python-language-server-stable/Python-Language-Server-osx-x64.0.4.114.nupkg
+elif [ "$(uname)" == "Linux" ]
+  then
+    curl -fLO https://pvsc.azureedge.net/python-language-server-stable/Python-Language-Server-linux-x64.0.4.114.nupkg
+fi
+unzip Python-Language-Server-osx-x64.0.4.114.nupkg
```
hujianxin: just unzip for osx?

mjlbach: whoops! fixed. Honestly, I don't like the hardcoded curl commands anyways... I need to figure out how to replicated the xml parsing used by the emacs/vs-code versions

lua/nvim_lsp/mpls.lua
```diff
@@ -0,0 +1,145 @@
...
+    if install_info.is_installed then
+      print(name, "is already installed")
+      return
+    end
+    if not (util.has_bins("curl")) then
+      error('Need "curl" to install this.')
+      return
+    end
+    if not util.has_bins("sh", "chmod", "unzip") then
+      error('Need the binaries "sh", "chmod", "unzip" to install this')
```
justinmk: does it really need to be a bash script? `curl` is shipped with Neovim on Windows (plus Windows 10 includes `curl` anyways, these days)

justinmk:
There's a lot of shell-isms here, which could be avoided by just using Lua (including all Vimscript functions available from Lua via `vim.fn`).

No reason to block this meanwhile, but hoping it can be improved.

lua/nvim_lsp.lua
```diff
@@ -11,6 +11,7 @@ require 'nvim_lsp/gopls'
 require 'nvim_lsp/hie'
 require 'nvim_lsp/leanls'
 require 'nvim_lsp/pyls'
+require 'nvim_lsp/mpls'
```
justinmk: Unless `mpls` is an existing convention, wouldn't it  be more intuitive to call this `mspyls`? 

- We already have `pyls`, so calling this `_pls` is needless variation.
- `ms` is a common abbreviation of "Microsoft"

mjlbach: MPLS was in my head from reading this [doc](https://github.com/microsoft/python-language-server/blob/7d0e3eb4752296fea66c92fa2722f77c1c9820ab/Using_in_sublime_text.md) in Microsoft's repo  when trying to get the language server running. I agree with you though, and will rename to pyls_ms.

lua/nvim_lsp/mpls.lua
mjlbach: Which component? Downloading an extracting the nupkg? Or using Microsoft's bash script to install .net core (Which maybe we should just assume people will handle installing themselves?) I used the sumneko_lua install as inspiration, but I'm happy to eliminate shell commands in favor of doing as much as possible within lua.

justinmk: Everything. I don't see any parts of this PR where a shell is needed. It would be unfortunate if `pyls_ms` didn't actually install for Windows users ...

mjlbach: Ok! I'll work on porting it over to lua and adding windows compatibility

mjlbach:
Of note, this also fails like pyls on indexing numpy... Here is the [log](https://www.dropbox.com/s/rl4jua8hzk1j3p6/vim-lsp.log?dl=0)

lua/nvim_lsp/pyls_ms.lua
```diff
@@ -0,0 +1,140 @@
+local skeleton = require 'nvim_lsp/skeleton'
+local util = require 'nvim_lsp/util'
+local lsp = vim.lsp
+
+local name = "pyls_ms"
+
+local function get_active_interpreter()
+  local f = io.popen("which python 2>&1")
```
justinmk: instead of `popen` with `2>&1`, could use the vim function `exepath`:

    local f = vim.fn.exepath('python')

justinmk:
> Here is the [log](https://www.dropbox.com/s/rl4jua8hzk1j3p6/vim-lsp.log?dl=0), and this is my best attempt to format it [here](https://www.dropbox.com/s/2abv7bl2w8lbn79/vim-lsp-formatted.log?dl=0)

looks like a bunch of docstrings were dumped to stdout?

lua/nvim_lsp/pyls_ms.lua
mjlbach: Fixed, thanks for pointing me to the function

mjlbach:
Yeah that was my impression. Microsoft has reported issues with indexing numpy before, but nothing like this (See: https://github.com/microsoft/python-language-server/issues/1414). It's unclear if vscode python has some specific workaround for numpy, which shows reasonable completion results on my mac.

I'd consider tensorflow a non-trivial dependency, and it works for that:

Screen Shot 2019-11-28 at 8 39 05 PM

I can file an issue on Microsoft's repo, unless I'm missing some special initialization option.

lua/nvim_lsp/pyls_ms.lua
```diff
@@ -0,0 +1,133 @@
...
+    end
+    local script = [=[
+set -e
+# Install dotnet if not available.
+which dotnet  >/dev/null || {
+  test -x dotnet || {
+    curl -L https://dot.net/v1/dotnet-install.sh | sh
+  }
+  export PATH=$PWD:$PATH
+}
```
justinmk: I'm not sure it's reasonable for us to install dotnet, that would be like installing python. I would suggest showing an error and mention `curl -L https://dot.net/v1/dotnet-install.sh | sh` in the docs.

lua/nvim_lsp/pyls_ms.lua
```diff
@@ -0,0 +1,133 @@
...
+
+# clone project
+if [ "$(uname)" == "Darwin" ]
```
justinmk: - `vim.fn.has('mac')` (or 'unix' for Unix, or 'win32' for Windows)
- the `curl` steps can be `vim.fn.system('curl ...')`

Then the shell script basically isn't needed. `unzip` isn't available on Windows, but at least if someone comes along wanting to get it working on Windows, there's much less of a wall staring at them :)

lua/nvim_lsp/pyls_ms.lua
mjlbach: Fixed

justinmk:
> I can file an issue on Microsoft's repo, unless I'm missing some special initialization option.

They would want to know how the server is init'd and what requests we're sending I guess.

lua/nvim_lsp/pyls_ms.lua
mjlbach: Ok, I think this is improved in my latest version but I'll need to test on windows. The only thing that I see is guaranteed on (modern) Windows would be powershell's Expand-Archive.

justinmk: > powershell's Expand-Archive.

sure, that's fine. Running powershell scripts can be tricky though, but there's a trick: piping the command through stdin avoids powershell complaints about untrusted scripts: https://github.com/glacambre/firenvim/commit/22db6ad01cb40f2500f85678fa869df095b19c6f#diff-1711f9eb5662d11dbf436e757decb205R284

justinmk:
This looks good, thanks for seeing it through! If you tested that the latest version (including install) works, we can merge it and improve Windows support after.

mjlbach:
It seems to install fine on Windows/Mac/Linux, but runs into problems executing dotnet from within neovim for Windows `A fatal error was encountered. The library 'hospolicy.dll' required to execute the application was not found`.

Everything runs fine on mac/linux. For future work:
[ ] - Automatically pull the latest version of pyls_ms (maybe we should add LspUpgrade in addition to LspInstall)
[ ] - Debug whatever is going on with top level numpy imports (strangely submodules like numpy.linalg show completion fine, so maybe this is a broader LS problem)
[ ] - Fix execution of dotnet on Windows
[ ] - Fix setting appropriate root directory (or throw error so language server doesn't silently fail in non-root repositories)

README.md
```diff
@@ -1436,6 +1437,38 @@ nvim_lsp#setup("pyls", {config})
...
+
+https://github.com/Microsoft/python-language-server
+
+`Microsoft Language Server for Python `, a language server for Python.
+    
+Requires [.NET Core](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) to run. On Linux or macOS:
+
+```bash
+curl -L https://dot.net/v1/dotnet-install.sh | sh
+```
```
justinmk: this got lost in the auto-update https://github.com/neovim/nvim-lsp/commit/aa187f1afb5ee1c7cf55089f8c8dc7da44c931c6 

I think it should go in the `docs` ?

mjlbach: Ah sorry, I didn't read the docgen section. I can make a PR to fix this.

mjlbach:
What's even stranger, is that the hover/documentation commands work for top-level modules like np.dot/np.abs that fail to show up in the completion results. I'm wondering if this is a bug in the LSP implementation since it seems to work fine in lsp-python-ms and VScode

PR #61, opened 2019-12-03
Add intelephense (PHP)

deevus:
This is just a modified `tsserver` script since `intelephense` is in NPM and is started the same way. 

It works, however at least in my setup I get the following error when opening a PHP file:

```
Error executing vim.schedule lua callback: ...r/neovim/HEAD-b1e4ec1/share/nvim/runtime/lua/vim/lsp.lua:430: RPC InternalError message =  "Request initialize failed with message: Cannot read property 'dataPaths' of undefined"
```

Not sure if this is an issue with neovim or Intelephense. 

Also, Intelephense has a premium mode which can be enabled using the `licenseKey` initialization option. See here: https://github.com/bmewburn/intelephense-docs

Any ideas how that would be configured?

### DIFF
--- a/lua/nvim_lsp/intelephense.lua
+++ b/lua/nvim_lsp/intelephense.lua
@@ -17,7 +17,16 @@ skeleton[server_name] = {
     filetypes = {"php"};
     root_dir = util.root_pattern("composer.json");
     log_level = lsp.protocol.MessageType.Warning;
-    settings = {};
+    settings = {
+      intelephense = {
+        environment = {
+          documentRoot = ""
+        }
+      }
+    };
+    init_options = {
+      licenceKey = ""
+    }
   };
   on_new_config = function(new_config)
     local install_info = installer.info()
--- a/lua/nvim_lsp/intelephense.lua
+++ b/lua/nvim_lsp/intelephense.lua
@@ -23,16 +23,6 @@ skeleton[server_name] = {
       return util.path.is_descendant(cwd, root) and cwd or root;
     end;
     log_level = lsp.protocol.MessageType.Warning;
-    settings = {
-      intelephense = {
-        environment = {
-          documentRoot = "";
-        };
-      };
-    };
-    init_options = {
-      licenceKey = "";
-    };
   };
   on_new_config = function(new_config)
     local install_info = installer.info()
@@ -59,7 +49,19 @@ npm install -g intelephense
       on_init = [[function to handle changing offsetEncoding]];
       capabilities = [[default capabilities, with offsetEncoding utf-8]];
       init_options = [[{
-        licenceKey = ""
+        storagePath = Optional absolute path to storage dir. Defaults to os.tmpdir().
+        globalStoragePath = Optional absolute path to a global storage dir. Defaults to os.homedir().
+        licenceKey = Optional licence key or absolute path to a text file containing the licence key.
+        clearCache = Optional flag to clear server state. State can also be cleared by deleting {storagePath}/intelephense
+        -- See https://github.com/bmewburn/intelephense-docs#initialisation-options
+      }]];
+      settings = [[{
+        intelephense = {
+          files = {
+            maxSize = 1000000;
+          };
+        };
+        -- See https://github.com/bmewburn/intelephense-docs#configuration-options
       }]];
     };
   };
@@ -68,4 +70,3 @@ npm install -g intelephense
 skeleton[server_name].install = installer.install
 skeleton[server_name].install_info = installer.info
 -- vim:et ts=2 sw=2
-
--- a/lua/nvim_lsp/intelephense.lua
+++ b/lua/nvim_lsp/intelephense.lua
@@ -1,0 +1,53 @@
+local skeleton = require 'nvim_lsp/skeleton'
+local util = require 'nvim_lsp/util'
+local lsp = vim.lsp
+
+local server_name = "intelephense"
+local bin_name = "intelephense"
+
+local installer = util.npm_installer {
+  server_name = server_name;
+  packages = { "intelephense" };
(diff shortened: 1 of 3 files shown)


justinmk:
> Also, Intelephense has a premium mode which can be enabled using the `licenseKey` initialization option. See here: https://github.com/bmewburn/intelephense-docs
> Any ideas how that would be configured?

Spelled `licenceKey` . I guess we would need to expose this: https://github.com/neovim/neovim/blob/3aa95ef27eb8817bc2ceddb6caf2b209e00a5e8d/runtime/lua/vim/lsp.lua#L388

norcalli:
@justinmk that's already exposed via `config.before_init`, but it sounds like it would be done with `config.init_options`. Sometimes LSPs are inconsistent in using the right terminology, so I would do it either way.

@deevus I recommend reading the docs on `:h lsp.start_client()` to understand more about init params.

deevus:
Just defining `init_options` was enough to get past the `dataPaths` error.

deevus:
Working with hover:
image

deevus:
Would the convention be to call this `phpls` rather than `intelephense`?

norcalli:
@deevus depends how many phpls's there are. In some cases, they are the only one, so something like phpls is inoffensive. But if there are multiple, then you might as well name it something more unique.

But really, there's no strict convention. In this case, since it has such a unique name, I'd probably go with `intelephense`. Usually the servers don't have a name, they're just `$LANG-language-server`.

deevus:
In PHP projects a `composer.json` could exist at multiple levels within a project, but they shouldn't all be considered a root, however `util.server_per_root_dir_manager` creates a new client if I open a subdirectory that has another `composer.json`.

deevus:
```vim
call nvim_lsp#setup("intelephense", {'init_options': {'licenceKey': '****'}})
```

This appears to work for passing `licenceKey`. I say "appears to" because I initially used the setting `intelephense.licenceKey` but it's deprecated, and I'm not sure if Intelephense caches the key somewhere.

deevus:
> In PHP projects a `composer.json` could exist at multiple levels within a project, but they shouldn't all be considered a root, however `util.server_per_root_dir_manager` creates a new client if I open a subdirectory that has another `composer.json`.
> 
> This causes a bunch of definitions to be missing since it should still be using the project root.

Having read the discussion in #22, using `cwd` seems to makes sense and I think 1:1 instances of `intelephense` and `neovim` also makes sense. Perhaps it could be more nuanced and use `cwd` as default, however if you access a file outside of the tree of `cwd` then it creates a new instance using `root_pattern`?

justinmk:
> ```viml
> call nvim_lsp#setup("intelephense", {'init_options': {'licenceKey': '****'}})
> ```

Avoid the Vimscript wrapper, it will be removed. Use Lua instead, see `:help lua-heredoc`.

justinmk:
shall we merge this or still working on it?

deevus:
@justinmk I'll make a few more tweaks and let you know

deevus:
@justinmk I think it's good now. Please review my `root_dir` implementation.

Aside: I tried to migrate my lsp setup code to lua but `luafile` kept saying "file not found" when trying to reference a lua file in the same directory.

lua/nvim_lsp/intelephense.lua
```diff
@@ -11,22 +11,38 @@ local installer = util.npm_installer {
   binaries = {bin_name};
 }
 
+local function is_inside_cwd(path, cwd)
```
justinmk: I would call this `util.path.is_descendant(root, path)`, it doesn't really have anything to do with CWD.

OTOH we should probably mimic Python's `commonpath()` function. https://stackoverflow.com/a/37923477/152142

lua/nvim_lsp/intelephense.lua
```diff
@@ -11,22 +11,38 @@ local installer = util.npm_installer {
...
   default_config = util.utf8_config {
     cmd = {bin_name, "--stdio"};
     filetypes = {"php"};
-    root_dir = util.root_pattern("composer.json");
+    root_dir = function (pattern)
+      local cwd  = vim.loop.cwd();
+      local root = util.root_pattern("composer.json", ".git")(pattern);
+
+      -- prefer cwd if it is inside root
+      return is_inside_cwd(root, cwd) and cwd or root;
```
norcalli: I don't really understand this logic. This would spawn a different server for every directory inside of the project. Is that intentional? Do be aware that root_dir doesn't modify the vim working directory, it just informs how to spawn a new server.

lua/nvim_lsp/intelephense.lua
deevus: Yeah you're right. In the context I was thinking it was about CWD but you could pass any path.

lua/nvim_lsp/intelephense.lua
deevus: The intention is to create an instance at the CWD, and use the default `root_pattern` otherwise. If you're viewing a file that is a descendant of CWD, it uses the same instance. If you open a file that is not a descendant of CWD, it uses `root_pattern` to find the project directory.

lua/nvim_lsp/intelephense.lua
deevus: I have refactored/extracted `is_inside_cwd` into `util.path.is_descendant`.

deevus:
@justinmk I think this is good to go

deevus:
@justinmk Actually something has changed in the framework (I assume)... The intelephense server is throwing lots of errors about undefined types.

deevus:
Nevermind... had to clear the intelephense cache

deevus:
:open_hands:

PR #71, opened 2019-12-15
Enable to install sumneko_lua in macOS

delphinus:
This enables to install `lua-language-server` in macOS in addition to Linux.

### DIFF
--- a/lua/nvim_lsp/sumneko_lua.lua
+++ b/lua/nvim_lsp/sumneko_lua.lua
@@ -3,18 +3,34 @@ local util = require 'nvim_lsp/util'
 local vim = vim
 
 local name = "sumneko_lua"
+local bin_name = "lua-language-server"
 
 local function make_installer()
   local P = util.path.join
   local install_dir = P{util.base_install_dir, name}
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/sumneko_lua.lua
```diff
@@ -3,18 +3,34 @@ local util = require 'nvim_lsp/util'
 local vim = vim
 
 local name = "sumneko_lua"
+local bin_name = "lua-language-server"
 
 local function make_installer()
   local P = util.path.join
```
justinmk: strange alias... oh well.

justinmk:
Thanks @delphinus ! As a followup, would be nice to remove the bash script, there is no need for bash. It can be done in Lua instead, see e.g. `pyls_ms.make_installer()`

delphinus:
@justinmk Thanks for merge. And I tried on `pyls_ms` and saw successfully the binary to be installed. But Neovim has shown no progress and been stuck while it was working.

Instead, the current logic here with bash scripts are nice to see the progress and Neovim seems not to be stuck. I think it has better UX than the `pyls_ms` way.

PR #75, opened 2019-12-17
Add metals

fgborges:
Hello, I added a recipe for `metals`. Cheers!


lua/nvim_lsp/metals.lua
```diff
@@ -0,0 +1,77 @@
...
+    cmd = {bin_name};
+    filetype = {"scala", "sbt"};
+    root_dir = util.root_pattern("build.sbt");
+    log_level = lsp.protocol.MessageType.Warning;
+    settings = {};
+  };
+  on_new_config = function(config)
+    installer.configure(config)
+  end;
+  docs = {
```
h-michael:
```suggestion
  docs = {
  	vscode = "scalameta.metals";
  	package_json = "https://raw.githubusercontent.com/scalameta/metals-vscode/master/package.json";
```

If the language server has vscode extension, we can auto generate README.md by writing like this.

fgborges: Thank you for your comments. I added these information.

h-michael:
@fgborges 
Install is succeeded.
```
  metals = {
    cmd = { "/home/h-michael/.cache/nvim/nvim_lsp/metals/metals" },
    install_dir = "/home/h-michael/.cache/nvim/nvim_lsp/metals",
    is_installed = "file"
  },
```
BTW I'm not familiar with Scala, but is there always a build.sbt in the project root?

norcalli:
@h-michael I used scala, and yeah basically, any Scala project will have a build. It's much like Rust where you could technically use it outside of that, but most people use Cargo.

norcalli:
I'm going to test it on one of my scala projects and then merge. We really should try to set up automated tests for things which have LspInstall.

lua/nvim_lsp/metals.lua
```diff
@@ -0,0 +1,79 @@
...
+      return
+    end
+    if not (util.has_bins("java")) then
+      error('Need "JDK" to install this.')
+      return
+    end
+    local coursier_bin = install_dir .. "/coursier"
+    local download_cmd = string.format("curl -fLo %s --create-dirs https://git.io/coursier", coursier_bin)
+    local chmod_cmd = string.format("chmod +x %s", coursier_bin)
+    local install_cmd = string.format("%s bootstrap --java-opt -Xss4m --java-opt -Xms100m --java-opt -Dmetals.client=coc.nvim org.scalameta:metals_2.12:0.7.6 -r bintray:scalacenter/releases -r sonatype:snapshots -o %s -f", coursier_bin, metals_bin)
```
ckipp01: You may want to take a quick look at to make sure you want `Dmetals.client=coc.nvim` since that's determining some of the functionality and the way we return stuff to the client. `vim-lsc` is also an option for now which might be closer depending on how you are handing messages coming from the server. If need be, we can totally also create a new case specifically for `nvim.lsp` or something if you need other settings. If it's just additional settings, then you can just use an existing and pass them in here as well.

Here is a reference to where these are read on the server side.
https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala#L111-L132

lua/nvim_lsp/metals.lua
```diff
@@ -0,0 +1,79 @@
...
+local bin_name = "metals"
+
+local function make_installer()
+  local install_dir = util.path.join{util.base_install_dir, server_name}
+  local metals_bin = util.path.join{install_dir, bin_name}
+  local X = {}
+  function X.install()
+    local install_info = X.info()
+    if install_info.is_installed then
+      print(name, "is already installed")
```
ckipp01: Out of curiosity, how to updates work? If a user has an old version installed, this will just tell them an old version is installed? Would they have to remove it, and then issue another install to get the new version installed?

lua/nvim_lsp/metals.lua
```diff
@@ -0,0 +1,79 @@
...
+  	vscode = "scalameta.metals";
+  	package_json = "https://raw.githubusercontent.com/scalameta/metals-vscode/master/package.json";
+    description = [[
+https://scalameta.org/metals/
+
+Scala language server with rich IDE features.
+`metals` can be installed via `:LspInstall metals`.
+]];
+    default_config = {
+      root_dir = [[util.root_pattern("build.sbt")]];
```
ckipp01: One other thing we might want to think about is by having this, we are only going to be triggering on sbt builds. However, we also can look for `build.sc` files for Mill builds, `pom.xml` files for Maven builds and `build.gradle` for gradle builds.

You can find some more information about the build tools we support [here](https://scalameta.org/metals/docs/build-tools/overview.html)

lua/nvim_lsp/metals.lua
ckipp01: Never mind about my question with configuration options, I'm not seeing the info about settings in `default_config`

ckipp01: I'd recommend replacing the `0.7.6` here with a variable that is either defaulting to the latest stable or pulling in `metals.serverVersion` from the settings object.

lua/nvim_lsp/metals.lua
justinmk: > Would they have to remove it, and then issue another install to get the new version installed?

yes

lua/nvim_lsp/metals.lua
fgborges: I will use `latest.release` parameter provided by coursier

https://github.com/coursier/coursier/blob/510ca407c060d23519592be74d20f70433d68f0d/modules/coursier/shared/src/test/scala/coursier/ResolveTests.scala#L152-L160

fgborges: Hmm, I do not want to add `-Dmetals.client` because the builtin LSP support is still under development, but someone will add this option in the future when builtin LSP support becomes stable.

lua/nvim_lsp/metals.lua
fgborges: Thank you for good review. I will add them.

fgborges:
I have applied every suggestion to my request. Please check it.

README.md
```diff
@@ -1362,6 +1363,66 @@ require'nvim_lsp'.leanls.setup{}
...
+  Default: `"0.7.6"`
+
+</details>
+
+```lua
+require'nvim_lsp'.metals.setup{}
+
+  Default Values:
+    cmd = { "metals" }
+    filetype = { "scala", "sbt" }
```
ckipp01: One last though. You probably don't need `sbt` in here. I'm assuming the fast majority of people will purposefully have `*.sbt` files marked as `scala`, which is recommended to do.

fgborges: I see. Thanks.

fgborges: I have updated it in metals.lua. The bot will be update the README automatically.

nabezokodaikon:
A character string that is not related to completion will be entered.

スクリーンショット 2020-01-18 16 28 43

fgborges:
@nabezokodaikon Your problem is not for this request because I provided only configuration to interact with metals and we cannot change any behaviors from settings. I mean that there is no abilities to modify messages in this change to solve your problem. Please send it to the upstream, neovim/neovim

nabezokodaikon:
@fgborges Thanks for the reply. I'm sorry.

tssm:
[This seems to be the issue](https://twitter.com/mattn_jp/status/1221067443839258626?s=21), may be useful

PR #76, opened 2019-12-18
Add ocamlls

fgborges:
I added [ocaml-language-server](https://github.com/ocaml-lsp/ocaml-language-server). Cheers!


h-michael:
Thank you for your PR :)

There seems to be [an official language server](https://github.com/ocaml/ocaml-lsp), is ocaml-language-server better than that?

fgborges:
Yes, I also prefer the official one, but the project still younger than the unofficials.
I keep a name `ocaml_lsp` for the official language server.  That's why, I set `ocamlls` be the older unofficials.

PR #98, opened 2020-01-19
Overwrite only when the value of dst is falsy

delphinus:
I found this when I attempt to set `init_options` for pyls_ms.

```lua
require'nvim_lsp'.pyls_ms.setup{
  init_options = {
    interpreter = {
      properties = {
        InterpreterPath = '/usr/bin/python3';
        Version = '3.6.9';
      }
    }
  }
}
```

But with this, `new_config` still has the default value from [here][].

[here]: https://github.com/neovim/nvim-lsp/blob/7a15a52c0a7d735625ac73dc4d8efe70c5e99707/lua/nvim_lsp/pyls_ms.lua#L99-L111

### DIFF
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -47,7 +47,7 @@ function M.tbl_deep_extend(dst, ...)
       if type(v) == 'table' and not vim.tbl_islist(v) then
         dst[k] = M.tbl_deep_extend(dst[k] or vim.empty_dict(), v)
       else
-        dst[k] = v
+        dst[k] = dst[k] or v
       end
     end
   end


lua/nvim_lsp/util.lua
```diff
@@ -47,7 +47,7 @@ function M.tbl_deep_extend(dst, ...)
       if type(v) == 'table' and not vim.tbl_islist(v) then
         dst[k] = M.tbl_deep_extend(dst[k] or {}, v)
       else
-        dst[k] = v
+        dst[k] = dst[k] or v
```
justinmk: This doesn't seem like a universally correct choice. Maybe `tbl_deep_extend` needs a `behavior` parameter?

But the old behavior seems questionable too: I would not expect "extend" to overwrite each item in a list. 

Hope  @norcalli can clarify

h-michael:
fixed #178

PR #99, opened 2020-01-23
Bump pyls_ms version

mjlbach:
Bump pyls_ms to version 0.5.30

### DIFF
--- a/README.md
+++ b/README.md
@@ -1996,7 +1996,7 @@ This server accepts configuration via the `settings` key.
 
   Default: `true`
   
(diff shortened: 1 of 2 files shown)


justinmk:
Ah, the life of an auto-installer :)

This is why one should never try to auto-install anything: it always ends up being a package manager..

Thanks!

mjlbach:
Unfortunately until we have a convenient way of XML parsing in neovim-lua (or better yet, microsoft exposes a nice endpoint for retrieving the latest binary), I'll have to manually update it.

PR #109, opened 2020-02-02
 lazy require language server modules

### LINKED ISSUE #80: lazy load  configs
I've tried to configure nvim-lsp in a similar way I do coc. I try to avoid loading the plugin unless I'm in a file type that uses it. Here is what I'm doing right now:

```vim
" In plugin/lsp.vim
autocmd FileType python,rust call config#lsp#init()

" In autoload/config/lsp.vim
function! config#lsp#init() abort
      packadd nvim-lsp
      nnoremap <silent> ;dc <cmd>lua vim.lsp.buf.declaration()<CR>
      nnoremap <silent> ;df <cmd>lua vim.lsp.buf.definition()<CR>
      nnoremap <silent> ;h  <cmd>lua vim.lsp.buf.hover()<CR>
      nnoremap <silent> ;i  <cmd>lua vim.lsp.buf.implementation()<CR>
      nnoremap <silent> ;s  <cmd>lua vim.lsp.buf.signature_help()<CR>
      nnoremap <silent> ;td <cmd>lua vim.lsp.buf.type_definition()<CR>
    if &filetype ==# 'rust'
        lua require'nvim_lsp'.rust_analyzer.setup{}
    elseif &filetype ==# 'python'
        lua require'nvim_lsp'.pyls_ms.setup{ log_level = 2 }
    endif
    setlocal omnifunc=v:lua.vim.lsp.omnifunc
endfunction
```
I think the reason this doesn't work is because nvim-lsp is setting filetype autcmds. Is there a way to do what I want?

If I put nvim-lsp in the `start` package directory, it doubles my vim load time, so I really don't want to load it unless I need it.

h-michael:
fix #80
And we are no longer bothered by conflicting in nvim_lsp.lua.

## startup benchmark

before
```
234.395   .dein >
	222.443   plugin/nvim_lsp.vim
	2.056     plugin/fzf.vim
	1.992     plugin/abolish.vim

...
```

after
```
14.848    .dein >
	2.628     plugin/nvim_lsp.vim
	2.140     plugin/abolish.vim
	2.043     plugin/fzf.vim
...
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,8 +64,6 @@ Example:
 
     configs.texlab.buf_build = buf_build
 
-Finally, add a `require 'nvim_lsp/SERVER_NAME'` line to `lua/nvim_lsp.lua`.
-
 ## Auto-install
 
 Configs may optionally provide `install()` and `install_info()` functions. This
--- a/lua/nvim_lsp.lua
+++ b/lua/nvim_lsp.lua
@@ -1,32 +1,5 @@
 local configs = require 'nvim_lsp/configs'
 
-require 'nvim_lsp/bashls'
-require 'nvim_lsp/ccls'
-require 'nvim_lsp/clangd'
-require 'nvim_lsp/cssls'
-require 'nvim_lsp/dockerls'
-require 'nvim_lsp/elmls'
-require 'nvim_lsp/flow'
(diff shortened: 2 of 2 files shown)


justinmk:
LGTM! @norcalli any objection?

PR #110, opened 2020-02-03
docgen.lua: require language server modules

h-michael:
In # 109, when loading nvim_lsp.lua, all the language server lua modules were not loaded,
so the language server module was missing from the configs module referenced in docgen.
So require the language server module in docgen.lua.
configs.lua and utils.lua are also required, but they have no effect as they do not add themselves to the configs module.

### DIFF
--- a/scripts/docgen.lua
+++ b/scripts/docgen.lua
@@ -222,6 +222,14 @@ local function make_lsp_sections()
 end
 
 local function make_implemented_servers_list()
+  -- load language server modules
+  local pfile = io.popen('ls lua/nvim_lsp/')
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks! Didn't see this, sorry. Let's go with https://github.com/neovim/nvim-lsp/pull/111 since it uses `glob()` . We should always avoid platform-specific stuff like `ls` where possible. Even if docgen.sh is not currently cross-platform.

h-michael:
Yeah, I didn't want to use `ls` either. There was a glob in the vim module. I remember :)

PR #120, opened 2020-02-07
[RDY] checkhealth: add checkhealth nvim_lsp command

teto:
Only check command for now.

This was much harder than expected as it's dynamically typed, the generation of the config and the start of the client is intertwined, and the logic for merging configs is very good but not the easiest to grasp.

It needs https://github.com/neovim/neovim/pull/11832 to work.

And gives this output:
```
health#nvim_lsp#check
  - INFO: Checking pyls
  - ERROR: The given command "python" is not executable.
  - INFO: Checking ccls
  - INFO: Checking lua_lsp
  - INFO: Checking ghcide
  - ERROR: The given command "ghcide" is not executable.
  - INFO: Checking texlab
  - ERROR: The given command "texlab" is not executable.
```
It's primitive but it's easy to improve. just check the attributes of new_config.

### DIFF
--- a/autoload/health/nvim_lsp.vim
+++ b/autoload/health/nvim_lsp.vim
@@ -1,0 +1,4 @@
+function! health#nvim_lsp#check()
+  call health#report_start('Checking language server protocol configuration')
+  lua require 'nvim_lsp/health'.check_health()
+endfunction
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -93,7 +93,8 @@ function configs.__newindex(t, config_name, config_def)
       end
       M.manager = nil
     end
-    local manager = util.server_per_root_dir_manager(function(_root_dir)
+
+    local make_config = function(_root_dir)
       local new_config = vim.tbl_extend("keep", {}, config)
       -- Deepcopy anything that is >1 level nested.
       new_config.settings = vim.deepcopy(new_config.settings)
@@ -146,8 +147,14 @@ function configs.__newindex(t, config_name, config_def)
               ))
         end
       end)
+
+      new_config.root_dir = _root_dir
       return new_config
-    end)
+    end
+
+    local manager = util.server_per_root_dir_manager(function(_root_dir)
+        return make_config(_root_dir)
+      end)
 
     function manager.try_add()
       local root_dir = get_root_dir(api.nvim_buf_get_name(0), api.nvim_get_current_buf())
@@ -159,6 +166,7 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     M.manager = manager
+    M.make_config = make_config
   end
 
   function M._setup_buffer(client_id)
--- a/lua/nvim_lsp/health.lua
+++ b/lua/nvim_lsp/health.lua
@@ -1,0 +1,22 @@
+local M = {}
+function M.check_health()
+  local configs = require 'nvim_lsp/configs'
+
+  for _, top_level_config in pairs(configs) do
+    -- the folder needs to exist
+    local new_config = top_level_config.make_config(".")
+
+    local cmd, _ = vim.lsp._cmd_parts(new_config.cmd)
+    if not (vim.fn.executable(cmd) == 1) then
+      vim.fn['health#report_error'](
+        string.format("%s: The given command %q is not executable.", new_config.name, cmd)
+      )
+    else
+      vim.fn['health#report_info'](
+        string.format("%s: configuration checked.", new_config.name)
+      )
+    end
+  end
+end
+
+return M
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -208,7 +208,7 @@ end)()
 
 -- Returns a function(root_dir), which, when called with a root_dir it hasn't
 -- seen before, will call make_config(root_dir) and start a new client.
-function M.server_per_root_dir_manager(make_config)
+function M.server_per_root_dir_manager(_make_config)
   local clients = {}
   local manager = {}
 
@@ -218,8 +218,7 @@ function M.server_per_root_dir_manager(make_config)
     -- Check if we have a client alredy or start and store it.
     local client_id = clients[root_dir]
     if not client_id then
-      local new_config = make_config(root_dir)
-      new_config.root_dir = root_dir
+      local new_config = _make_config(root_dir)
       new_config.on_exit = M.add_hook_before(new_config.on_exit, function()
         clients[root_dir] = nil
       end)
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+tab_width = 8
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[{Makefile,**/Makefile,runtime/doc/*.txt}]
+indent_style = tab
+indent_size = 8


autoload/health/nvim_lsp.vim
```diff
@@ -0,0 +1,20 @@
+
+function! health#nvim_lsp#check()
+	call health#report_start('Checking language server protocol configuration')
+lua << EOF
+	local configs = require 'nvim_lsp/configs'
+	local lsp = require 'vim.lsp'
+
+	for name, top_level_config in pairs(configs) do
+		-- the folder needs to exist, => use stdpath ?
+		local new_config = top_level_config.gen_config("/tmp")
```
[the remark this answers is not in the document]
teto: at first I had used "/fake/path" but the validator checks the folder exists so I was thinking of using `stdpath('cache')` to avoid creating a temporary directory just for checkhealth.  Question is does `stdpath('cache')` always exist ?

teto:
> Can this healthcheck also check for cmds overwritten by the user?

yes the whole make_config merges various defaults with the user config. That's the whole point.

autoload/health/nvim_lsp.vim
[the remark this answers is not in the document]
h-michael: > Question is does stdpath('cache') always exist ?

At least it didn't exist in my environment.
If `stdpath('config')`, init.vim will be placed, so it may not exist in many environments.

@justinmk Is there any good workaround?

teto:
I've tried to address your comments.

autoload/health/nvim_lsp.vim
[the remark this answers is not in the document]
justinmk: > Question is does `stdpath('cache')` always exist ?

Would need to create it here. Is there something wrong with that?

Nvim could add e.g. `ensurecache()`, but that's not different than `mkdir(stdpath('cache'), 'p')`.  Though I suppose Nvim could do this once at startup, then plugins can reasonably assume it exists...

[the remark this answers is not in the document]
teto: I've been using `.`

teto:
I've written a more elegant version. I just need to test and I push in 10 mn.

autoload/health/nvim_lsp.vim
```diff
@@ -0,0 +1,8 @@
+let s:FILE_DIR = fnamemodify(expand("<sfile>"), ':h') . '/'
+let s:FILE_DIR = substitute(s:FILE_DIR, '\\', '/', 'g')
+
+function! health#nvim_lsp#check()
+	call health#report_start('Checking language server protocol configuration')
+	execute 'luafile '.s:FILE_DIR.'check.lua'
```
justinmk: if `autoload/health/check.lua` were moved to `lua/nvim_lsp/health.lua`, then this could be:
```suggestion
	require('nvim_lsp.health').nvim_lsp_check_health()
```

And then the workaround in `.luacheckrc` would not be needed.

Is there a reason for the current approach?

autoload/health/check.lua
```diff
@@ -0,0 +1,14 @@
+function nvim_lsp_check_health()
+	local configs = require 'nvim_lsp/configs'
```
justinmk: we always use 2-space indent.

autoload/health/nvim_lsp.vim
teto: you are right, thats what I've done afterwards but I haven't had the opportunity to test it yet. my system is rebuilding.

lua/nvim_lsp/configs.lua
```diff
@@ -146,6 +147,14 @@ function configs.__newindex(t, config_name, config_def)
...
       end)
+
+      new_config.root_dir = _root_dir
+      return new_config
+    end
+
+    local manager = util.server_per_root_dir_manager(
+    function(_root_dir)
+      local new_config = make_config(_root_dir)
       return new_config
```
justinmk: strange indent

autoload/health/check.lua
```diff
@@ -0,0 +1,14 @@
...
+
+	for _, top_level_config in pairs(configs) do
+		-- the folder needs to exist
+		local new_config = top_level_config.make_config(".")
+		vim.fn['health#report_info']("Checking "..new_config.name)
+
+		local cmd, _ = vim.lsp._cmd_parts(new_config.cmd)
+		if not (vim.fn.executable(cmd) == 1) then
+			vim.fn['health#report_error'](string.format("The given command %q is not executable.", cmd))
```
justinmk: 1. brevity
2. the "variable" part of messages should go at the end after a colon, when possible

```suggestion
			vim.fn['health#report_error'](string.format("%s: command is not executable: %q", new_config.name, cmd))
```

autoload/health/check.lua
```diff
@@ -0,0 +1,14 @@
+function nvim_lsp_check_health()
+	local configs = require 'nvim_lsp/configs'
+
+	for _, top_level_config in pairs(configs) do
+		-- the folder needs to exist
+		local new_config = top_level_config.make_config(".")
+		vim.fn['health#report_info']("Checking "..new_config.name)
```
justinmk: this is noise. instead, show a success message for non-error cases.

autoload/health/nvim_lsp.vim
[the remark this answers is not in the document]
justinmk: isn't it better to use cache or data dirs for their intended purpose?

autoload/health/check.lua
teto: It's difficult to guarantee the success with basic or even more exhaustive checks. 
I just wanted the user to see their config was at least found. So maybe instead: `Found config for X` or `No obvious error for X`.

justinmk: no need to overthink it. "success" is the opposite of "error".

teto: I chose `configuration checked`

teto:
Ready. I took the liberty to copy neovim's .editorconfig to prevent future wrong tab/space usage.

lua/nvim_lsp/configs.lua
```diff
@@ -93,7 +93,8 @@ function configs.__newindex(t, config_name, config_def)
       end
       M.manager = nil
     end
-    local manager = util.server_per_root_dir_manager(function(_root_dir)
+
+    local make_config = function(_root_dir)
```
h-michael:
```suggestion
    local make_config = function(root_dir)
```

lua/nvim_lsp/configs.lua
```diff
@@ -146,8 +147,15 @@ function configs.__newindex(t, config_name, config_def)
...
+
+      new_config.root_dir = _root_dir
       return new_config
-    end)
+    end
+
+    local manager = util.server_per_root_dir_manager(function(_root_dir)
+        local new_config = make_config(_root_dir)
+        return new_config
+      end)
```
h-michael:
```
    local manager = util.server_per_root_dir_manager(function(root_dir)
      return make_config(root_dir)
    end)
```

lua/nvim_lsp/util.lua
```diff
@@ -208,7 +208,7 @@ end)()
 
 -- Returns a function(root_dir), which, when called with a root_dir it hasn't
 -- seen before, will call make_config(root_dir) and start a new client.
-function M.server_per_root_dir_manager(make_config)
+function M.server_per_root_dir_manager(_make_config)
```
h-michael:
```suggestion
function M.server_per_root_dir_manager(make_config)
```

lua/nvim_lsp/util.lua
```diff
@@ -218,8 +218,7 @@ function M.server_per_root_dir_manager(make_config)
     -- Check if we have a client alredy or start and store it.
     local client_id = clients[root_dir]
     if not client_id then
-      local new_config = make_config(root_dir)
-      new_config.root_dir = root_dir
+      local new_config = _make_config(root_dir)
```
h-michael:
```suggestion
      local new_config = make_config(root_dir)
```

lua/nvim_lsp/health.lua
```diff
@@ -0,0 +1,22 @@
...
+
+    local cmd, _ = vim.lsp._cmd_parts(new_config.cmd)
+    if not (vim.fn.executable(cmd) == 1) then
+      vim.fn['health#report_error'](
+        string.format("%s: The given command %q is not executable.", new_config.name, cmd)
+      )
+    else
+      vim.fn['health#report_info'](
+        string.format("%s: configuration checked.", new_config.name)
+      )
```
h-michael: I think this is better, what do you think? cc: @justinmk 
```
      vim.fn['health#report_ok'](
        string.format("%s: The given command %q is executable.", new_config.name, cmd)
      )
```

lua/nvim_lsp/configs.lua
```diff
@@ -146,8 +147,15 @@ function configs.__newindex(t, config_name, config_def)
               ))
         end
       end)
+
+      new_config.root_dir = _root_dir
```
h-michael: Why did you move this process here? I don't think I need to move.

lua/nvim_lsp/health.lua
```diff
@@ -0,0 +1,22 @@
+local M = {}
+function M.check_health()
+  local configs = require 'nvim_lsp/configs'
+
+  for _, top_level_config in pairs(configs) do
+    -- the folder needs to exist
+    local new_config = top_level_config.make_config(".")
```
h-michael: If you do not [move this process](https://github.com/neovim/nvim-lsp/pull/120/files#diff-bb1a5a30518e3cb9d6bb43af411ddbf5R151), it should work without any arguments.

lua/nvim_lsp/util.lua
teto: This avoids collisions with the function `make_config`. I prefer it the current way,

lua/nvim_lsp/configs.lua
teto: because I use to call `validate_client_config` and it would complain about a missing root_dir. make_config should generate a full config and not a half decent one. Because I don't use `validate_client_config` anymore, I can now pass a fake path (lsp will validate it only upon starting lsp) so it solves the issue of having a valid path. Thanks for making me notice.

lua/nvim_lsp/health.lua
teto: If we add more checks later (I am sure we will), it's best to have a not too detailed message for success.

teto:
thanks for the reviews

Issue #94, opened 2020-01-15
[rust_analyzer] initalizationOptions causing ra_lsp_server an error


KokaKiwi:
Because of how vim `json_encode` transform vim data into JSON data, the lua value `{}` which is the default value for `init_options` is being encoded into an empty JSON array.
But `ra_lsp_server` expect to receive a JSON map for decoding the init options and emit an error: `failed to deserialize config: invalid length 0, expected struct ServerConfig with 12 elements` when receiving an empty initializationOptions by vim.lsp.
Hopefully rust-analyzer handle this error by just logging it and sticking with default values, but still the error message appears, which can be actually confusing.

justinmk:
> lua value `{}` which is the default value for `init_options` is being encoded into an empty JSON array.

This can be fixed by setting it to `vim.empty_dict()`. Just need someone to send a PR doing that.

KokaKiwi:
I think default to that would be good, as the LSP spec says `initOptions` has to be a dictionary anyway, so even if implementations actually mainly don't mind about this, defaulting to what the specs say seems to be the correct way of handling this imho

matu3ba:
Changing in lua/nvim_lsp/skeleton.lua
<code>init_options = {};</code> to
<code>init_options = vim.empty_dict();</code>
did **not solve the issue**.
I directly changed the git folder in plugged and used  
`lua require'nvim_lsp'.rust_analyzer.setup{}` in *init.vim*.

`call nvim_lsp#setup("rust_analyzer", {})` returns `unknown function`, so it should be removed from the docs.
What am I missing here?

anderejd:
@matu3ba Did you get it to work? If yes, I would very much appreciate a full example of how to do it.

matu3ba:
Sorry, somehow it works. I was very confused by the message, but the message shortly thereafter disappears.

matu3ba:
With the described 1-line [PR](https://github.com/neovim/nvim-lsp/pull/107) added, it works.

justinmk:
> `call nvim_lsp#setup("rust_analyzer", {})` returns `unknown function`, so it should be removed from the docs.

@matu3ba where are you seeing that? It was removed from all docs that I know of, weeks ago.

matu3ba:
@justinmk you are correct.

tuxsoy:
Can someone please double check? I still have this issue.

justinmk:
@tuxsoy always need to know exact version of Nvim and this plugin.

[...]

PR #121, opened 2020-02-11
Use vim.empty_dict() instead of {}

h-michael:
fix: #94
rust-analyzer parses initializeOptions according to the type definition.
Therefore, an empty dictionary fails because it does not satisfy the type definition.
So, set the default value of ServerConfig type to initializeOptions.

https://github.com/rust-analyzer/rust-analyzer/blob/d8d8c20077702b8537086d49914d02654a46ebc5/crates/ra_lsp_server/src/main.rs#L65-L79
https://github.com/rust-analyzer/rust-analyzer/blob/dc48f89581843248660ceb755bb20469ab6ac0c9/crates/ra_lsp_server/src/config.rs#L18-L49

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -26,7 +26,7 @@ function configs.__newindex(t, config_name, config_def)
 
   local default_config = tbl_extend("keep", config_def.default_config, {
     log_level = lsp.protocol.MessageType.Warning;
-    settings = {};
+    settings = vim.empty_dict();
     init_options = vim.empty_dict();
     callbacks = {};
   })
@@ -95,7 +95,7 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     local make_config = function(_root_dir)
-      local new_config = vim.tbl_extend("keep", {}, config)
+      local new_config = vim.tbl_extend("keep", vim.empty_dict(), config)
       -- Deepcopy anything that is >1 level nested.
       new_config.settings = vim.deepcopy(new_config.settings)
       util.tbl_deep_extend(new_config.settings, default_config.settings)
@@ -153,8 +153,8 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     local manager = util.server_per_root_dir_manager(function(_root_dir)
-        return make_config(_root_dir)
-      end)
+      return make_config(_root_dir)
+    end)
 
     function manager.try_add()
       local root_dir = get_root_dir(api.nvim_buf_get_name(0), api.nvim_get_current_buf())
--- a/lua/nvim_lsp/rust_analyzer.lua
+++ b/lua/nvim_lsp/rust_analyzer.lua
@@ -8,7 +8,6 @@ configs.rust_analyzer = {
     filetypes = {"rust"};
     root_dir = util.root_pattern("Cargo.toml");
     log_level = lsp.protocol.MessageType.Warning;
-    settings = {};
   };
   docs = {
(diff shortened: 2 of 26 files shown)


justinmk:
That seems brittle, and defeats the purpose of loose formats like JSON. @matklad is this expected behavior of rust-analyzer: that clients have to send all fields in init_options?

h-michael:
@justinmk initializeOptions keis optional and value is any type.
So how to handle this option is different from each servers.
So we might remove initializeOptions key if value is empty dict or nil.

matklad:
Nope, that would be a bug. We intend to handle empty object correctly, we use serde default:

```
#[serde(rename_all = "camelCase", default)]
```

But #94 says that we see an empty **array** here, and not an empty object, and that shape mismatch is I think worth complaining about?

matklad:
Anyway, I think it makes sense to double check on our side how do we handle empty objects, and to improve the error message to include the original json, thanks for the ping!

justinmk:
> But #94 says that we see an empty **array** here, and not an empty object, and that shape mismatch is I think worth complaining about?

Yes, we are sending an object instead of array since https://github.com/neovim/nvim-lsp/commit/b4f4b72a50983464864810e642e0b70e0b987e6f . 

> Anyway, I think it makes sense to double check on our side how do we handle empty objects, and to improve the error message to include the original json, thanks for the ping!

Thanks!

matklad:
I've just checked, we do handle `{}`. Here's log from the corresponding VS Code invocation: https://gist.github.com/matklad/86adaeb6a52feb87cb5ceb850ed14810.

I've also pushed https://github.com/rust-analyzer/rust-analyzer/pull/3102 which should include the original JSON in the error message, which should help with debugging the issue.

h-michael:
@matklad Thank you for your quickly fixing. :)

matklad:
@h-michael could you try wit this new version of rust-analyzer and paste the error message you get? I hope it sheds the light on the issue.

h-michael:
@matklad 
Sorry, I misunderstood.
This error didn't seem to be the response of initialize.
https://gist.github.com/h-michael/deb0052a00456709bcb2f557c94bf083

justinmk:
> "[ERROR ra_lsp_server] Failed to deserialize config: invalid length 0, expected struct ServerConfig with 12 elements; []\n"

That [] suggests that somewhere we're still sending an array ?

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -8,7 +8,23 @@ configs.rust_analyzer = {
     filetypes = {"rust"};
     root_dir = util.root_pattern("Cargo.toml");
     log_level = lsp.protocol.MessageType.Warning;
-    settings = {};
```
justinmk: does this help?

    settings = vim.empty_dict();

h-michael:
I think https://github.com/neovim/neovim/pull/11864 will solve this.

lua/nvim_lsp/configs.lua
```diff
@@ -94,7 +94,7 @@ function configs.__newindex(t, config_name, config_def)
       M.manager = nil
     end
     local manager = util.server_per_root_dir_manager(function(_root_dir)
-      local new_config = vim.tbl_extend("keep", {}, config)
+      local new_config = vim.tbl_extend("keep", vim.empty_dict(), config)
```
justinmk: Nice catch!

h-michael:
@matklad That was our problem. Thank you for helping us debug.

PR #123, opened 2020-02-12
Add js-langserver support

mwilliammyers:
For a minimal test:

1. Make sure you have a `lua require'nvim_lsp'.js_langserver.setup{}` or similar in your `init.vim`
1.  `mkdir test && cd test`
1.  Create a `package.json`:
    ```json
    {
      "name": "test",
      "version": "0.0.1",
      "description": "test lsp",
      "main": "index.js",
      "keywords": [],
      "author": "",
      "license": "ISC",
      "devDependencies": {
        "eslint": "^6.8.0"
      },
      "eslintConfig": {
        "env": {
          "browser": true,
          "commonjs": true,
          "es6": true
        },
        "extends": "eslint:recommended",
        "globals": {
          "Atomics": "readonly",
          "SharedArrayBuffer": "readonly"
        },
        "parserOptions": {
          "ecmaVersion": 2018
        },
        "rules": {}
      }
    }
    ```
1.  `npm i` (You might be able to use a globally installed version of `eslint`?)
1.  install `js-langserver`:
    ```
    npm i -g https://github.com/tbodt/js-langserver/tarball/d59d7d8
    ```
    or run `:LspInstall js_langserver` and (most likely) you will need to add a `.tern-project` that looks something like:
    ```json
    {
      "plugins": {
        "node": {}
      }
    }
    ```
1.  Create an `index.js`:
    ```javascript
    console.log(x);
    ```

You should see an error: `'x' is not defined`

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,7 @@ that config.
 - [gopls](#gopls)
 - [hie](#hie)
 - [intelephense](#intelephense)
+- [js_langserver](#js_langserver)
 - [jsonls](#jsonls)
 - [julials](#julials)
 - [leanls](#leanls)
@@ -1225,6 +1226,30 @@ require'nvim_lsp'.intelephense.setup{}
     log_level = 2
     on_init = function to handle changing offsetEncoding
     root_dir = root_pattern("composer.json", ".git")
+```
+
+## js_langserver
+
+https://github.com/tbodt/js-langserver
+
+`js-langserver` can be installed via `:LspInstall js-langserver` or by yourself with `npm`:
+```sh
+npm install -g js-langserver
+```
+
+Can be installed in Nvim with `:LspInstall js_langserver`
+
+```lua
+require'nvim_lsp'.js_langserver.setup{}
+
+  Default Values:
+    capabilities = default capabilities, with offsetEncoding utf-8
+    cmd = { "js-langserver", "--stdio" }
+    filetypes = { "javascript", "javascriptreact", "javascript.jsx" }
+    log_level = 2
+    on_init = function to handle changing offsetEncoding
+    root_dir = root_pattern("package.json")
+    settings = {}
 ```
 
 ## jsonls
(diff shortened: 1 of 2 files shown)


h-michael:
@mwilliammyers Thank you for your PR.
There is no policy that we only want to support a specific language server, but I don't think this language server is very good.
This language server has the eslint version locked into the language server itself.
In the development of JavaScript, it is common to specify eslint in package.json for each project, so I think that the way used by this language server does not match most users.

mwilliammyers:
Yeah, I understand your concern. It isn't a very popular LS but it is one of the few (if not only) that supports all of the (or even some of the basic) features you would expect from an LSP for *pure* Javascript (as opposed to Typescript).

Personally, I think it would be useful to have; better to have a functional LS that has some issues than none at all. We could work with them upstream to better integrate with the js ecosystem by not shipping a bundled eslint etc. We could also mention the eslint issue in the docs somewhere?

mwilliammyers:
I just double checked and this LS uses [semver caret ranges](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) for all its dependencies. For example, [here](https://github.com/tbodt/js-langserver/blob/d59d7d8487a893b45e75d69a03d4fdbb4e07e0e3/package.json#L10) a range is used to specify that you must have a locally installed ESLint version satisfying `>=4.19.1 <5.0.0` otherwise npm will download and install it for you. So you could `npm i -D js-langserver` and use it locally in a project or globally and use your own version of ESLint etc.

That being said, I understand it is not a very popular but it is the only one I could find for pure Javascript that has ESlint + Tern support (see [this](https://github.com/tbodt/js-langserver/#obligatory-stupid-arbitrary-feature-matrix-table-thing) comparison).

I can keep using it locally though if you don't want to add it...

justinmk:
seems reasonable but is tern EOL (IIUC, this LS depends on tern)?

mjlbach:
Seems like https://github.com/tbodt/js-langserver isn't actively maintained, and the dependencies are growing out of date. Feel free to reopen if this changes in the future.

PR #138, opened 2020-02-21
Add CodeQL support

pwntester:
Adds support for CodeQL language server

If you want to try, you will need to install codeql-cli and ql queries. Follow instructions [here](https://help.semmle.com/codeql/codeql-cli/procedures/get-started.html#download-the-codeql-cli-zip-package)

you need to pass a `search_path` argument pointing to your QL repo. eg:

``` lua
    require'nvim_lsp'.codeqlls.setup{
        settings = {
            search_path = {'~/codeql-home/codeql-repo'};
        };
    }
```

Create a directory and place a `qlpack.yml` file with:

``` yaml
name: test
version: 0.0.0
libraryPathDependencies: codeql-java
```

In the same directory write a java query `test.ql`, set filetype to `codeql` (I will soon publish a plugin for that) and write your query in it:

```
import java

from Method m
select m.getName()
```

You should get lintering and autocompletion.

### DIFF
--- a/lua/nvim_lsp/codeqlls.lua
+++ b/lua/nvim_lsp/codeqlls.lua
@@ -24,7 +24,14 @@ configs[server_name] = {
         };
     };
     docs = {
-        description = [[test]]
+        description = [[
+Reference:
+https://help.semmle.com/codeql/codeql-cli.html
+
+Binaries:
+https://github.com/github/codeql-cli-binaries
+        ]];
+        package_json = "https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/package.json";
     };
     on_new_config = function(config)
         if config.settings.search_path ~= nil and config.settings.search_path ~= '' then
--- a/lua/nvim_lsp/codeqlls.lua
+++ b/lua/nvim_lsp/codeqlls.lua
@@ -5,8 +5,10 @@ local server_name = "codeqlls"
 
 local root_pattern = util.root_pattern("qlpack.yml")
 
+-- vim.empty_dict()
 configs[server_name] = {
     default_config = {
+        cmd = {"codeql", "execute", "language-server", "--check-errors", "ON_CHANGE", "-q"};
         filetypes = {'codeql'};
         root_dir = function(fname)
             return root_pattern("qlpack.yml") or util.path.dirname(fname)
@@ -19,10 +21,11 @@ configs[server_name] = {
             }}
         end;
         settings = {
-            search_path = vim.empty_dict()
+            search_path = vim.empty_dict() 
         };
     };
     docs = {
+        package_json = "https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/package.json";
         description = [[
 Reference:
 https://help.semmle.com/codeql/codeql-cli.html
@@ -30,17 +33,19 @@ https://help.semmle.com/codeql/codeql-cli.html
 Binaries:
 https://github.com/github/codeql-cli-binaries
         ]];
-        package_json = "https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/package.json";
+        default_config = {
+            settings = {
+                search_path = [[list containing all search paths, eg: '~/codeql-home/codeql-repo']];
+            };
+        };
     };
     on_new_config = function(config)
-        if config.settings.search_path ~= nil and config.settings.search_path ~= '' then
+        if type(config.settings.search_path) == 'table' and not vim.tbl_isempty(config.settings.search_path) then
             local search_path = "--search-path="
             for _, path in ipairs(config.settings.search_path) do
                 search_path = search_path..vim.fn.expand(path)..":"
             end
             config.cmd = {"codeql", "execute", "language-server", "--check-errors", "ON_CHANGE", "-q", search_path}
-        else
-            config.cmd = {"codeql", "execute", "language-server", "--check-errors", "ON_CHANGE", "-q"}
         end
     end;
 }
--- a/lua/nvim_lsp/codeqlls.lua
+++ b/lua/nvim_lsp/codeqlls.lua
@@ -21,7 +21,7 @@ configs[server_name] = {
             }}
         end;
         settings = {
-            search_path = vim.empty_dict() 
+            search_path = vim.empty_dict()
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/codeqlls.lua
```diff
@@ -0,0 +1,46 @@
...
+        end;
+        log_level = vim.lsp.protocol.MessageType.Warning;
+        before_init = function(initialize_params, config)
+            initialize_params['workspaceFolders'] = {{
+                    name = 'workspace',
+                    uri = initialize_params['rootUri']
+            }}
+        end;
+        settings = {
+            search_path = vim.empty_dict()
```
h-michael: This should be `{}`. `vim.empty_dic()` acts as empty dictionary.
But I think you want search_path to be `list`.

lua/nvim_lsp/codeqlls.lua
```diff
@@ -0,0 +1,46 @@
...
+        };
+    };
+    docs = {
+        description = [[
+Reference:
+https://help.semmle.com/codeql/codeql-cli.html
+
+Binaries:
+https://github.com/github/codeql-cli-binaries
```
h-michael: I think we should also mention how to use on_new_config

lua/nvim_lsp/codeqlls.lua
```diff
@@ -0,0 +1,46 @@
...
+        ]];
+        package_json = "https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/package.json";
+    };
+    on_new_config = function(config)
+        if config.settings.search_path ~= nil and config.settings.search_path ~= '' then
+            local search_path = "--search-path="
+            for _, path in ipairs(config.settings.search_path) do
+                search_path = search_path..vim.fn.expand(path)..":"
+            end
+            config.cmd = {"codeql", "execute", "language-server", "--check-errors", "ON_CHANGE", "-q", search_path}
```
h-michael: What about doing this?
In this case, the user can decide the bin itself.

```lua
default_config = {
  ...;
  cmd = {"codeql", "execute", "language-server", "--check-errors", "ON_CHANGE", "-q"}
  ...;
};
on_new_config = function(config)
  ...
    config.cmd = tbl.insert(config.cmd,  search_path)
  ...
end
```

lua/nvim_lsp/codeqlls.lua
```diff
@@ -0,0 +1,46 @@
...
+Reference:
+https://help.semmle.com/codeql/codeql-cli.html
+
+Binaries:
+https://github.com/github/codeql-cli-binaries
+        ]];
+        package_json = "https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/package.json";
+    };
+    on_new_config = function(config)
+        if config.settings.search_path ~= nil and config.settings.search_path ~= '' then
```
h-michael:
```suggestion
        if type(config.settings.search_path) == 'table' and not vim.tbl_isempty(config.settings.search_path)
```

lua/nvim_lsp/codeqlls.lua
pwntester: Sure will add

lua/nvim_lsp/codeqlls.lua
pwntester: thanks!

lua/nvim_lsp/codeqlls.lua
pwntester: :+1:

lua/nvim_lsp/codeqlls.lua
pwntester: For whatever reason, using `{}` breaks it. But you are right, it should be an empty list? is there any similar `vim.empty_list()`

h-michael: What breaks mean?
Would you show me something like error messages?

justinmk: > is there any similar `vim.empty_list()`

no, that would be identical to `{}`

pwntester: But some reason, print statements do no not work within the `on_new_config` callback (or in any other callback as far as I know), so Im not able to debug why `search_path` is not accounted for when usin `{}` instead of an empty dict.

I have a simple codeql query like:

```
import java
import foo
```

If `search_path` is not accounted for, both lines will contain errors. If `search_path` is correctly passed to the language server, only `import foo` will get an error.

So using an empty dict, I get the expected error, but when using an empty list I get also an error on `import java` meaning that search path was not updated.

Is there a way to debug callbacks? 
Thanks

pwntester: I was able to see what was happening by logging to a file rather than using `print`. So when I use `vim.empty_dict()`, `settings.search_path` from the user config overrides the default config. But if I use `{}` instead, `config.settings` at the beggining of the `on_new_config` callback is `{ ["search_path"] = { } ,} `

So it seems like when using `{}`, user config do not override default config

pwntester: So apparently the bug is un `nvim_lsp/util.tbl_deep_extend`:

```
dst[k] = v
```

`v` comes from default config and `dst` is the user config. In that line, the user config is always overriden with default config, no matter the contents.

It can be reproduced with:

```
function test()
    user_config = { ["search_path"] = { "~/codeql-home/codeql-repo"}}
    print(dump(user_config))
    default_config = { ["search_path"] = { } ,}
    print(dump(default_config))
    new_config = util.tbl_deep_extend(user_config, default_config)
    print(dump(new_config))
end
```

pwntester: Added a new commit that fixes the issue for me, but not sure if something else is needed (eg: extend lists?)

h-michael: May be related with #98

h-michael:
@pwntester deep_extend is good for Nvim, so I made PR. Wait for a little.
https://github.com/neovim/neovim/pull/11969

pwntester:
@h-michael PR rebased, can we merge it now? Thanks

h-michael:
@pwntester ty :) merged.

Issue #139, opened 2020-02-21
CI failure


justinmk:
master is failing: https://github.com/neovim/nvim-lsp/runs/461200554

```
Error detected while processing command line:
E5113: Error while calling lua chunk: /home/runner/work/nvim-lsp/nvim-lsp/lua/nvim_lsp/dartls.lua:19: [LSP] Could not find Dart SDK root path
##[error]Process completed with exit code 1.
```

PR #140, opened 2020-02-22
gh actions: fix docgen action

h-michael:
ref: #139

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -7,6 +7,13 @@ jobs:
     runs-on: [ubuntu-latest]
     if: github.ref != 'master'
     steps:
+    - name: Setup Dart
+      run: |
+        sudo apt install apt-transport-https
+        sudo sh -c 'wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
+        sudo sh -c 'wget -qO- https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
+        sudo apt update
+        sudo apt install dart
     - uses: actions/checkout@v1
     - run: date +%F > todays-date
     - name: Restore cache for today's nightly.


justinmk:
Why do we require dart?

h-michael:
Because the dart config does a little hacky for the cmd detect.
https://github.com/neovim/nvim-lsp/blob/master/lua/nvim_lsp/dartls.lua#L8-L35

This is irrelevant because those who do not use the dart language server do not need the dart language server in the first place, but ci failed because docgen.lua required all configs.

In the case of CI, it was possible to add code to avoid these, but it was more appropriate in this case to adjust the environment of CI so that CI works rather than adding code for CI Changed the CI settings to think.

justinmk:
Hmm. But that means `dartls.lua` fails in an unconventional way, compared to other configs. Normally this is covered by `:checkhealth` which checks for things explicitly.

 `dartls.lua` is the only config that throws an error if it can't decide a `cmd`. It should just return `snapshot` instead of throwing an error. 

```diff
diff --git a/lua/nvim_lsp/dartls.lua b/lua/nvim_lsp/dartls.lua
index e93937286f3b..14e2c64f8fcd 100644   
--- a/lua/nvim_lsp/dartls.lua
+++ b/lua/nvim_lsp/dartls.lua
@@ -27,11 +27,7 @@ local analysis_server_snapshot_path = function()
     snapshot = snapshot:gsub("/", "\\")
   end
 
-  if vim.fn["filereadable"](snapshot) == 1 then
-    return snapshot
-  else
-    error("[LSP] Could not find analysis server snapshot")
-  end
+  return snapshot  -- May not exist.
 end
 
 configs[server_name] = {
```

If we want to give more-immediate feedback about missing things when user enables a config, we should do that intentionally for all configs, by similar logic as done in `:checkhealth`. That's a different topic though.

h-michael:
> If we want to give more-immediate feedback about missing things when user enables a config, we should do that intentionally for all configs, by similar logic as done in :checkhealth.

I agree with it. This logic is added before adding healthcheck.
I'll fix dart config.

Issue #136, opened 2020-02-21
checkhealth complains "Vim(lua):E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/lsp.lua:137: cmd type must be list."


mcepl:
I was sent here from https://github.com/neovim/neovim/issues/11919:

- `nvim --version`: own build from e8f160c82
- `vim -u DEFAULTS` (version: ) behaves differently? NA (doesn't have built-in LSP client)
- Operating system/version: openSUSE/Tumbleweed
- Terminal name/version: gnome-terminal-3.34.2-3.1.x86_64
- `$TERM`: vte-256color

### Steps to reproduce using `nvim -u NORC`

Run ``:checkhealth``

### Actual behaviour
```
health#nvim_lsp#check                                                           
========================================================================        
  - ERROR: Failed to run healthcheck for "nvim_lsp" plugin. Exception:          
    function health#check[21]..health#nvim_lsp#check, line 2                    
    Vim(lua):E5108: Error executing lua
    /usr/share/nvim/runtime/lua/vim/lsp.lua:137: cmd type must be list.
```
The rest of the output is green.

### Expected behaviour

All green.

justinmk:
it should show the config that actually failed... cc @teto 

@mcepl does it happen with `nvim -u NORC`? please share your lsp config.

mcepl:
> it should show the config that actually failed... cc @teto
> 
> @mcepl does it happen with `nvim -u NORC`? please share your lsp config.

It doesn’t show with ``-u NORC``, but then it doesn’t show anything about ``health#nvim_lsp#check``. Isn’t it switched off as well, LSP being still a plugin?

```
health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: RelWithDebInfo

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $VTE_VERSION='5803'
  - INFO: $COLORTERM='truecolor'

health#provider#check
========================================================================
## Clipboard (optional)
  - OK: Clipboard tool found: xclip

## Python 2 provider (optional)
  - WARNING: No Python executable found that can `import neovim`. Using the first available executable for diagnostics.
  - ERROR: Python provider error:
    - ADVICE:
      - provider/pythonx: Could not load Python 2:
[...]

PR #141, opened 2020-02-23
healthcheck: use pcall for config.cmd error

h-michael:
ref: #136

### DIFF
--- a/lua/nvim_lsp/health.lua
+++ b/lua/nvim_lsp/health.lua
@@ -4,17 +4,23 @@ function M.check_health()
 
   for _, top_level_config in pairs(configs) do
     -- the folder needs to exist
-    local new_config = top_level_config.make_config(".")
+    local config = top_level_config.make_config(".")
 
-    local cmd, _ = vim.lsp._cmd_parts(new_config.cmd)
-    if not (vim.fn.executable(cmd) == 1) then
+    local status, cmd = pcall(vim.lsp._cmd_parts, config.cmd)
+    if not status then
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/health.lua
```diff
@@ -2,19 +2,25 @@ local M = {}
...
-    local new_config = top_level_config.make_config(".")
+    local config = default_config.make_config(".")
 
-    local cmd, _ = vim.lsp._cmd_parts(new_config.cmd)
-    if not (vim.fn.executable(cmd) == 1) then
+    local status, cmd = pcall(vim.lsp._cmd_parts, config.cmd)
+    if not status then
       vim.fn['health#report_error'](
-        string.format("%s: The given command %q is not executable.", new_config.name, cmd)
+        string.format("%s: config.cmd, %s, is expected 'table', but got %s", config.name, vim.inspect(config.cmd), type(config.cmd))
```
teto: this may hide a real error, pcall should return the error message (`cmd`) if an error happens so use that in the message instead of assuming the error.

lua/nvim_lsp/health.lua
```diff
@@ -2,19 +2,25 @@ local M = {}
 function M.check_health()
   local configs = require 'nvim_lsp/configs'
 
-  for _, top_level_config in pairs(configs) do
+  for _, default_config in pairs(configs) do
```
teto: why rename top_level_config ?

h-michael:
@teto ty :)

PR #142, opened 2020-02-23
dartls: remove calling error in detecting cmd

h-michael:
ref: #140


lua/nvim_lsp/dartls.lua
```diff
@@ -11,12 +11,10 @@ local find_dart_sdk_root_path = function()
   elseif vim.fn["executable"]("flutter") == 1 then
     local flutter_path = vim.fn["resolve"](vim.fn["exepath"]("flutter"))
     local flutter_bin = vim.fn["fnamemodify"](flutter_path, ":h")
-    local dart_sdk_root_path = flutter_bin.."/cache/dart-sdk/bin/dart"
-    if vim.fn["executable"](dart_sdk_root_path) == 1 then
-      return dart_sdk_root_path
-    end
+    return flutter_bin.."/cache/dart-sdk/bin/dart"
+  else
+    return ''
```
justinmk: would be better to return some reasonable default here, or at least `unknown`. Empty strings are harder to debug and result in strange error messages.

h-michael: In the case of Dart language server, if we cannot get here, we cannot use it without fail. Here, I will return a null character, and I will implement healthcheck displaying a warning if this is a null character.
What do you think?

justinmk: You mean nil? Nul char would be weird, let's avoid surprises.

h-michael: Oh sorry I mean ''.

h-michael: This commit shows I want to say. df4087fa4d175582e5be6317a9165fb834de63c5

justinmk:
> healthcheck: add unique for each language server healthcheck

That's going to add a lot of code, and increases the cost for people to create new configs.

We can avoid it. Healthcheck should figure stuff out based on the info that is already in configs.

lua/nvim_lsp/dartls.lua
justinmk: That commit looks like something else. Just return `unknown`, or empty string, or `nil`. Let's not overcomplicate it.

h-michael:
> That's going to add a lot of code, and increases the cost for people to create new configs.

I think this(healthcheck function) is not required.
As this time, some servers have special requirements, so I think it would be more convenient to have such a thing.

h-michael:
@justinmk gentle ping

h-michael:
@tjdevries Would you check this PR?

PR #144, opened 2020-02-23
Change root_dir of pyls to CWD to prevent pyls from traversing HOME directory

### LINKED ISSUE #118: Rename is not working for python-language-server
Subj. Logs show nothing at all, even with log level set to 4.

e-kwsm:
[The doc](https://github.com/neovim/nvim-lsp/blob/be0f71585eb3185b214c84c643e43e9647ef67d1/README.md#pyls) says the root_dir of pyls is "vim's starting directory" but actually it is home directory.

I tried to fix #118 and noticed that rope performs rename on [files under home directory](https://github.com/python-rope/rope/blob/0.16.0/rope/refactor/rename.py#L85-L86). This significantly harms performance.

### DIFF
--- a/lua/nvim_lsp/pyls.lua
+++ b/lua/nvim_lsp/pyls.lua
@@ -1,11 +1,14 @@
 local configs = require 'nvim_lsp/configs'
+local util = require 'nvim_lsp/util'
 local lsp = vim.lsp
 
 configs.pyls = {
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks. I'm confused about why home dir was chosen for *any* of these configs. ..

Issue #127, opened 2020-02-17
[gopls] delay diagnostics or not run them in insert mode


urandom:
Is there any way to force a particular lsp server to delay its diagnostics, or not run them during insert mode?

gopls is tends to be quite chatty when i'm writing and i see a lot of temporary diagnostics virtual texts. Worse yet, at some point in time, the following messages (in a different buffer) start being printed out, causing pauses in the ui:

LSP[gopls] 2020/02/17 09:01:01 failed to fix AST: unable to parse defer or go from *ast.BadStmt: no defer or go statement found

urandom:
sorry about the bug label, this is more of a question/documentation issue

lithammer:
Ideally I would like an option to suspend all `textDocument/publishDiagnostics` communication in insert mode (i.e. both inbound and outbound). Or at the very least add some kind of debounce.

It's possible to mess around with a custom `vim.lsp.callbacks['textDocument/publishDiagnostics']` callback to remove most of the noise. But that just means throwing away the response, the LSP will still perform a lot of (needless) processing on every keypress. Which can be a huge drain on CPU and battery.

h-michael:
@urandom Would you show me a sample codes or project?

urandom:
@h-michael 
It seems that the code really doesn't matter. Any partial statement will obviously trigger diagnostics, and some will cause that blocking error as well, though I'm not exactly sure what causes the later.

I've noticed it mostly happening when I start writing struct literals, like: 
https://github.com/globusdigital/feature-toggles/blob/7ab6dd1a839c3a58354f95fb689d4cc558232de9/toggle/toggle_test.go#L37

the cursor position would be at:
{name:|CURSOR|}
and the error will popup, freezing nvim for a second.

though it also sometimes occurs when i write if statements.

norcalli:
@lithammer I would ignore the cost of the processing on the keypress if you're using an LSP at all on a battery and just disable it if that is a concern. There's no way to use an LSP without sending the textDocument updates, which, currently, send the whole text document due to a bug preventing incremental updates in the core LSP (but to be fair, a lot of LSP servers don't support incremental as well).

So I'd say the cost of parsing the json is pretty insignificant compared to the cost of displaying virtual text and updating it etc, so a custom callback which returns early if you're in insert mode for the LSP server of your choice (or all of them) is a solid idea.

[...]

PR #145, opened 2020-02-24
window/showMessage: the display level can be set for 'window/showMess…

h-michael:
related with #127

- add message_level
- remove `log_level = lsp.protocol.MessageType.Warning` from each server configs. These are not needed because they are already set in default_config.

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -4,7 +4,6 @@ local tbl_extend = vim.tbl_extend
 
 local configs = {}
 
-
 function configs.__newindex(t, config_name, config_def)
   validate {
     name = {config_name, 's'};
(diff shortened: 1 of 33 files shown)


lua/nvim_lsp/configs.lua
```diff
@@ -46,6 +46,14 @@ function configs.__newindex(t, config_name, config_def)
       end
     end
 
+    config.callbacks["window/showMessage"] = function(err, method, params, client_id)
+      if params and params.type <= config.message_level then
+        -- TODO(ashkan) remove this after things have settled.
+        assert(lsp.callbacks, "Update to Nvim HEAD. This is an incompatible interface.")
+        lsp.callbacks[method](err, method, params, client_id)
```
justinmk: what kind of strings will `method` be? should we validate it ? if `lsp.callbacks[method]` is potentially nil, we should show a meaningful error message .

h-michael: I fix it.

PR #146, opened 2020-02-24
init html lsp

mhartington:


lua/nvim_lsp/html.lua
```diff
@@ -0,0 +1,61 @@
...
+
+`html-languageserver` can be installed via `:LspInstall html` or by yourself with `npm`:
+```sh
+npm install -g vscode-html-languageserver-bin
+```
+]];
+    default_config = {
+      root_dir = [[root_pattern("package.json")]];
+      on_init = [[function to handle changing offsetEncoding]];
+      capabilities = [[default capabilities, with offsetEncoding utf-8]];
```
h-michael: We don't need to add this. Because offsetEncoding is not official protocol. That is only for clang now.

lua/nvim_lsp/html.lua
```diff
@@ -0,0 +1,61 @@
...
+local root_pattern = util.root_pattern("package.json")
+
+configs[server_name] = {
+  default_config = util.utf8_config {
+    cmd = {bin_name, "--stdio"};
+    filetypes = {"html"};
+    root_dir = function(fname)
+      return root_pattern(fname) or vim.loop.os_homedir()
+    end;
+    log_level = lsp.protocol.MessageType.Warning;
```
h-michael: This value is same as default log_level. So we don't need to add this line.

lua/nvim_lsp/html.lua
```diff
@@ -0,0 +1,61 @@
...
+    description = [[
+https://github.com/vscode-langservers/vscode-html-languageserver-bin
+
+`html-languageserver` can be installed via `:LspInstall html` or by yourself with `npm`:
+```sh
+npm install -g vscode-html-languageserver-bin
+```
+]];
+    default_config = {
+      root_dir = [[root_pattern("package.json")]];
```
h-michael:
```suggestion
      root_dir = [[root_pattern(fname) or vim.loop.os_homedir()]];
```

mhartington:
Thank you for the feedback @h-michael

PR #147, opened 2020-02-24
lua language server: add a more detailed description of cmd

h-michael:
ref: #136


teto:
cant we instead just use a binary name ? like sumneko ?

h-michael:
> cant we instead just use a binary name ? like sumneko ?

I think the problem is not limited to this language server, but the thing that we use our own abbreviations for all language servers instead of official names.

PR #148, opened 2020-02-25
MessageType: add None to MessageType and rename MessageType to Severity

h-michael:
Add None because the user has no way to hide logMessage and showMessage.
ref: #127

### DIFF
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -7,6 +7,8 @@ local fn = vim.fn
 
 local M = {}
 
+M.Severity = vim.tbl_extend("keep", lsp.protocol.MessageType, { None = 0 })
+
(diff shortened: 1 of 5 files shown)


lua/nvim_lsp/util.lua
```diff
@@ -7,6 +7,8 @@ local fn = vim.fn
 
 local M = {}
 
+M.Severity = vim.tbl_extend("keep", lsp.protocol.MessageType, { None = 0 })
```
justinmk, line 10: not in favor of this. Adding new types that have nearly the same semantics as other types is a cognitive burden.

justinmk:
> Add None because the user has no way to hide logMessage and showMessage.

hmm. In Vim itself, there's no way to hide `:echo` messages either, at least not for normal editor usage. Is this really important?

PR #149, opened 2020-02-25
default_config: let user specify default_config yourself.

h-michael:
If users want to change something like log_level for all servers, changing log_level for all servers is a bit tedious.
So we can make this easier by overriding util.default_config.

I will add documentation if we can agree with this policy.

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -23,13 +23,7 @@ function configs.__newindex(t, config_name, config_def)
 
   local M = {}
 
-  local default_config = tbl_extend("keep", config_def.default_config, {
-    log_level = lsp.protocol.MessageType.Warning;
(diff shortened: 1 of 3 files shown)


lua/nvim_lsp/configs.lua
```diff
@@ -23,13 +23,8 @@ function configs.__newindex(t, config_name, config_def)
...
   local M = {}
 
-  local default_config = tbl_extend("keep", config_def.default_config, {
-    log_level = lsp.protocol.MessageType.Warning;
-    message_level = lsp.protocol.MessageType.Warning;
-    settings = vim.empty_dict();
-    init_options = vim.empty_dict();
-    callbacks = {};
-  })
+  vim.validate{ default_config = { util.default_config, 't' } }
```
teto: isn't it going to be validated anyway ?

PR #153, opened 2020-03-01
Fix clangd issue on Windows

erw7:
Fixed the following error when editing a non-existent file such as `:e test.c` on Windows after #114.

```
E5108: Error executing lua shared.lua:426: root_dir: expected directory, got test.c
```

Note: However, I feel that `clangd` is not working properly just because no error occurs (If I start editing a file that does not exist, completion will not work).


h-michael:
thanks :)

PR #183, opened 2020-03-29
Automatically fetch latest version of pyls_ms

mjlbach:
Not sure if this is the best approach, but this eliminates the need to manually bump the version.

### DIFF
--- a/lua/nvim_lsp/pyls_ms.lua
+++ b/lua/nvim_lsp/pyls_ms.lua
@@ -10,6 +10,17 @@ local function get_python_version()
   return l:match("^Python%s*(...).*%s*$")
 end
 
+local function get_latest_pyls()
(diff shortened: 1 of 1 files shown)


h-michael:
@mjlbach Awesome! I'll check it later.

@yaegassy Would you try this?

mjlbach:
@h-michael we should consider adding an LspUpdate mechanism. Currently we have no way to overwrite previously installed language servers other than deleting from cache.

yaegassy:
It works fine!
My [pull request](https://github.com/neovim/nvim-lsp/pull/181) is closed.

mjlbach:
I should have done this from the beginning but couldn't figure out a good way to parse the xml when I originally made the config.

h-michael:
>  we should consider adding an LspUpdate mechanism. Currently we have no way to overwrite previously installed language servers other than deleting from cache.

Yes, I'll impement it :)

Issue #191, opened 2020-04-20
Add ability to conditionally add language server configuration if not defined


cideM:
```
NVIM v0.5.0-445-gca02db1f9
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-41e98c7d-18cc-46e0-90e2-721b8f0d476a

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
```

- nvim-lsp version(commit hash): `b6dc2ae6a95051e93486224481b80cb5b446dcc3`

- `:checkhealth` result
```
health#targets#check
========================================================================
  - WARNING: Conflicting mapping found:
    ab → <Plug>(textobj-sandwich-auto-a)
    b → {'pair': [{'c': ')', 'o': '('}, {'c': ']', 'o': '['}, {'c': '}', 'o': '{'}]}
  - WARNING: Conflicting mapping found:
    ib → <Plug>(textobj-sandwich-auto-i)
    b → {'pair': [{'c': ')', 'o': '('}, {'c': ']', 'o': '['}, {'c': '}', 'o': '{'}]}

health#nvim_lsp#check
========================================================================
## Checking language server protocol configuration
  - INFO: dhall: configuration checked.
[...]

PR #192, opened 2020-04-21
use pcall for requiring a lsp module

h-michael:
fix: #191

### DIFF
--- a/lua/nvim_lsp.lua
+++ b/lua/nvim_lsp.lua
@@ -74,7 +74,7 @@ end
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
-    require('nvim_lsp/'..k)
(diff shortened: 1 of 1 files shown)


teto:
the error is then silent which makes debugging harder. maybe checkhealth catches it though

tjdevries:
We should check the result of pcall, and if it is not OK, then warn the user that it failed. It is good not to crash though :)

h-michael:
I think we should not this `pcall` result.
Because I added this `require` by this PR #109.
This `require` is only for loading existing modules.

justinmk:
yeah this doesn't seem right, and would have made the problem in https://github.com/neovim/nvim-lsp/issues/191 much harder to find. It would silently do nothing instead of showing an error.

PR #197, opened 2020-04-22
Make installer for sumneko_lua more robust

dlukes:
- I had to install clang for the installation to complete -> check for
  that up front
- quote PATH update (especially on WSL, PWD and PATH will easily contain
  whitespace)

### DIFF
--- a/lua/nvim_lsp/sumneko_lua.lua
+++ b/lua/nvim_lsp/sumneko_lua.lua
@@ -39,8 +39,8 @@ local function make_installer()
       error('Need either "ninja" or "curl" (to download ninja) to install this.')
       return
     end
-    if not util.has_bins("sh", "chmod", "unzip") then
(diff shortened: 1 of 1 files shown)


tjdevries:
Does sumneko always need clang to compile or can it use something else?

dlukes:
I tried this on WSL2/Ubuntu with gcc installed, but it errored out instructing me to install clang. Grepping through the sumneko/lua-language-server checkout, I see lines like these:

```
lm.gcc = 'clang'
lm.gxx = 'clang++'
```

(Not familiar enough with the codebase and lua in general to draw any strong conclusions, but on the face of it, it indeed looks like clang specifically is needed.)

tjdevries:
Awesome, thanks!

tjdevries:
(I'll merge in a bit, just have to confirm one thing first)

PR #198, opened 2020-04-22
LspInstall: --no-package-lock --no-save --production

theJian:
Get rid of missing package.json warnings when installing npm packages.

Also eliminating package-lock.json.

### DIFF
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -392,7 +392,7 @@ function M.npm_installer(config)
     set -e
     mkdir -p "{{install_dir}}"
     cd "{{install_dir}}"
-    npm install {{packages}} --no-package-lock --no-save --production --silent
+    npm install {{packages}} --no-package-lock --no-save --production
     {{post_install_script}}
     ]]):gsub("{{(%S+)}}", install_params)
     cmd:write(install_script)
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -392,7 +392,7 @@ function M.npm_installer(config)
     set -e
     mkdir -p "{{install_dir}}"
     cd "{{install_dir}}"
-    npm install {{packages}}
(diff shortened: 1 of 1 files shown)


tjdevries:
@theJian you tested installing a fresh server w/ the new command?

@teto LGTM, I don't really know much about NPM though.

teto:
I don't want to comment on issues related to 3rd party software since I dont believe nvim-lsp should do it. I don't know much about the npm ecosystem, silencing warnings in general looks like a bad idea or may mean we don't follow the normal guidelines.

tjdevries:
Ah, good point about silent. Why do you want it to run silent @theJian ?

theJian:
@tjdevries Because there was some output from npm when I ran :LspInstall vimls. Npm complains missing package.json and that really doesn't matter since our purpose is fetching npm package.

theJian:
@tjdevries I tested installing vimls. It looks good. There is no warning output if installing success. I think the warning output from npm does not help but confuse users.

tjdevries:
@theJian Neither I nor @teto are the normal maintainers here, so we're working to figure out exactly how we want to handle these kinds of PRs right now. After we figure that out, we'll give you more information about what we'll do next (or merge it, if that's the case)

lua/nvim_lsp/util.lua
```diff
@@ -392,7 +392,7 @@ function M.npm_installer(config)
     set -e
     mkdir -p "{{install_dir}}"
     cd "{{install_dir}}"
-    npm install {{packages}}
+    npm install {{packages}} --no-package-lock --no-save --production --silent
```
justinmk, line 395:
```suggestion
    npm install {{packages}} --no-package-lock --no-save --production
```

justinmk:
merged without `--silent`. It's conventional in vim ecosystem to show shell output, see e.g. fugitive. And ignoring warnings/errors in general tends to be a bad tradeoff.

PR #199, opened 2020-04-22
Show correct lua syntax for configuring rust-analyzer

matklad:
Hi! It took me a while (knowing neither lua not vimscript) to figure out how to pass settings to rust-analyzer, so I thought it would be nice to document it. 

I am not sure if this change will actually be reflected in readme. Running `nvim -R -Es +'set rtp+=$PWD' +'luafile scripts/docgen.lua'` did nothing for me, so I am not sure...

### DIFF
--- a/lua/nvim_lsp/rust_analyzer.lua
+++ b/lua/nvim_lsp/rust_analyzer.lua
@@ -6,6 +6,9 @@ configs.rust_analyzer = {
     cmd = {"rust-analyzer"};
     filetypes = {"rust"};
     root_dir = util.root_pattern("Cargo.toml", "rust-project.json");
+    settings = {
+      ["rust-analyzer"] = {}
+    }
   };
   docs = {
     package_json = "https://raw.githubusercontent.com/rust-analyzer/rust-analyzer/master/editors/code/package.json";


tjdevries:
We do have some info in the README already:

```
## Usage

Each config provides a `setup()` function, to initialize the server with
reasonable defaults and some server-specific things like commands or different
diagnostics.

    vim.cmd('packadd nvim-lsp')
    require'nvim_lsp'.<config>.setup{name=…, settings = {…}, …}

```

But if you could figure out a way to make it more obvious that `settings` is where you pass that info, that'd be great and I can review that as well.

lithammer:
Personally, my problem isn't so much `settings` itself, but how its content is supposed to be laid out. To continue with rust-analyzer, according to its [`package.json`](https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/package.json) it has a configuration called `rust-analyzer.diagnostics.enable`. So what am I suppose to put in `settings`?

```lua
settings = {
  ['rust-analyzer.diagnostics.enable'] = true,
  ['rust-analyzer'] = {
    'diagnostics.enable' = true
  },
  ['rust-analyzer'] = {
    diagnostics = {
      enable = true
    }
  }
}
```

Or is that different for every language server?

matklad:
To clarify, the specific stumbling block for me was that the correct syntax is `["rust-analyzer"] = {` and not `rust-analyzer = `. That's the main point of confusion I aim to solve with this PR.

matklad:
>To continue with rust-analyzer, according to its package.json it has a configuration called rust-analyzer.diagnostics.enable. 

This is indeed confusing. VS Code stores user config in a flat form. In `settings.json`, one would write

```json
"rust-analyzer.diagnostics.enable": true,
```

However, when VS Code passes the settings to the editor, it makes them nested, so the JSON expected on the server is (roughly, as we ask for `rust-analyzer` section specifically)

```json
"rust-analyzer": {
  "diagnostics": {
    "enable": true
  }
}
```

h-michael:
@matklad 
It's certainly annoying to be flooded with user questions in Lua's syntax.
How about referring to "docs = {description = {..." as follows:

```lua
docs = {
  ...
  description = [[
...

If you want a specific setting, please write as follows.

require'nvim_lsp'.rust_analyzer.setup{
  settings = {
    ['rust-analyzer'] = {
      diagnostics = {
        enable = true;
      }
    }
  }
}
  ]]
}

```

justinmk:
seems reasonable. is there also a `rust-client` key involved? let's start with this anyway

PR #201, opened 2020-04-22
Leave nvim_lsp.util.sh output buffer open on error

### LINKED ISSUE #196: Can't inspect LspInstall errors
Because the `sh` utility function always closes the output buffer:

https://github.com/neovim/nvim-lsp/blob/39b743b9b36be26a5e58278d7608fe456a0ad3f8/lua/nvim_lsp/util.lua#L430-L432

Maybe only call `bwipeout` if the shell script exited successfully? Or if that's too complicated, leave it around for the user to delete? It's better than hiding useful feedback on error imho :)

dlukes:
Fixes #196.

If the child process exits with a non-zero exit code or receives a
non-zero signal, leave the buffer with the output open so that the user
has a chance to figure out what went wrong.

The `update_chunk` function is not modified, just moved up a bit, so
that `uv.spawn`'s `on_exit` callback can capture it and use it to
append the exit code and signal to the output buffer.

### DIFF
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -417,30 +417,6 @@ function M.sh(script, cwd)
   local stdout = uv.new_pipe(false)
   local stderr = uv.new_pipe(false)
 
-  -- luacheck: no unused
-  local handle, pid
-  handle, pid = uv.spawn("sh", {
-    stdio = {stdin, stdout, stderr};
-    cwd = cwd;
-  }, function()
-    stdin:close()
-    stdout:close()
-    stderr:close()
-    handle:close()
-    vim.schedule(function()
-      api.nvim_command("silent bwipeout! "..bufnr)
-    end)
-  end)
-
-  -- If the buffer closes, then kill our process.
-  api.nvim_buf_attach(bufnr, false, {
-    on_detach = function()
-      if not handle:is_closing() then
-        handle:kill(15)
-      end
-    end;
-  })
-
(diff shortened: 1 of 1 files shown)


dlukes:
By the way, is there a way to make the output buffer behave more like a terminal, probably leveraging the builtin terminal? As it stands, using e.g. `curl` with a progress bar appends all the progress bar states one after the other, separated by `^M`s.

If that's not possible, maybe it would be a better idea to direct the output to the same place that running external commands like `:!curl example.com` goes to? That has the appropriate behavior out of the box.

tjdevries:
If you'd like to figure it out (since I don't know) it'd be a welcome improvement. if not, I think we can merge this as it's a strict improvement to how it was before.

dlukes:
I just thought I’d ask because I have no idea what the best way to achieve this would be, but if you don’t have any immediate pointers, then maybe this PR could be merged for a start, and I’ll see if I can figure out a way to address more terminal-like behavior on my own? :)

dlukes:
So #208 is my attempt to address both of these issues at once. I think the desired UX is mostly there, it gives `util.sh` the flexibility of a full terminal emulator, and it only closes the buffer on exit 0. The downside is that it has to use a global VimL function as the `on_exit` callback (see the PR for details why), which is ugly and potentially brittle.

If you (the maintainers) feel that's not such a big problem though, or that it just needs a little bit of work in order to be made more resilient, then I think we can close this and focus on #208 instead :)

PR #204, opened 2020-04-23
Change filetype for codeql

pwntester:


h-michael:
@pwntester I think we need only 56ddadd commit.

pwntester:
For some reason, when I push my branch with only one commit, it create this PR with all commits :(
Can you cherry pick it or merge the PR since the final result will be the same

PR #205, opened 2020-04-23
Added support for the kotlin-language-server

thchha:
This PR adds support for the kotlin-language-server found at fwcd's [repository](https://github.com/fwcd/kotlin-language-server).

> Please provide guidance about further steps I should consider to make this contribution pass.

// edit:
Unfortunately I did not manage to rebase an fix for recover the lint test of the CI-pipe (trailing whitespaces). Should I rebase (squash all of my commits) from the previous commit (#d16b49e)? I picked #5e5c8b5 of my branch. Result is displayed below..

### DIFF
--- a/lua/nvim_lsp/kotlin_language_server.lua
+++ b/lua/nvim_lsp/kotlin_language_server.lua
@@ -6,7 +6,6 @@
 --- 	https://github.com/udalov/kotlin-vim (recommended)
 --- 	Note that there is no LICENSE specified yet.
 
-local vim = vim	-- from global table
 local util = require 'nvim_lsp/util'
 local configs = require 'nvim_lsp/configs'
 
--- a/lua/nvim_lsp/kotlin_language_server.lua
+++ b/lua/nvim_lsp/kotlin_language_server.lua
@@ -1,0 +1,58 @@
+--- default config for gradle-projects of the
+--- kotlin-language-server: https://github.com/fwcd/kotlin-language-server
+---
+--- This server requires vim to be aware of the kotlin-filetype.
+--- You could refer for this capability to:
+--- 	https://github.com/udalov/kotlin-vim (recommended)
+--- 	Note that there is no LICENSE specified yet.
+
+local vim = vim	-- from global table
+local util = require 'nvim_lsp/util'
+local configs = require 'nvim_lsp/configs'
+--- kotlin-language-server supports multiple types of projects.
+--- The default works with single- and multi-module gradle projects.
+local function get_project_root()
+  return util.root_pattern("settings.gradle")
+end
+
+configs.kotlin_language_server = {
+  default_config = util.utf8_config {
+	filetypes = { "kotlin" };
+	root_dir = get_project_root();
+	log_level = vim.lsp.protocol.MessageType.Warning;
+  };
+  docs = {
+	vscode = "0.5.2";
+	description = [[
+A kotlin language server which was developed for internal usage and
+released afterwards. Maintaining is not done by the original author,
+but by fwcd.
+
+It is builded via gradle and developed on github.
+Source and additional description:
+	https://github.com/fwcd/kotlin-language-server
+]];
+	default_config = {
+	  root_dir = [[root_pattern("settings.gradle")]];
+	  capabilities = [[
+	  smart code completion,
+	  diagnostics,
+	  hover,
+	  document symbols,
+	  definition lookup,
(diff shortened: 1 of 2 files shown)


lua/nvim_lsp/kotlin_language_server.lua
```diff
@@ -0,0 +1,58 @@
...
+end
+
+configs.kotlin_language_server = {
+  default_config = util.utf8_config {
+	filetypes = { "kotlin" };
+	root_dir = get_project_root();
+	log_level = vim.lsp.protocol.MessageType.Warning;
+  };
+  docs = {
+	vscode = "0.5.2";
```
h-michael, line 25:
```suggestion
	package_json = "https://raw.githubusercontent.com/fwcd/vscode-kotlin/master/package.json";
```

lua/nvim_lsp/kotlin_language_server.lua
```diff
@@ -0,0 +1,58 @@
...
+local function get_project_root()
+  return util.root_pattern("settings.gradle")
+end
+
+configs.kotlin_language_server = {
+  default_config = util.utf8_config {
+	filetypes = { "kotlin" };
+	root_dir = get_project_root();
+	log_level = vim.lsp.protocol.MessageType.Warning;
+  };
```
h-michael, line 23: `log_level` is the same as the default. So you don't need to redefine.
```suggestion

configs.kotlin_language_server = {
  default_config = util.utf8_config {
    filetypes = { "kotlin" };
    root_dir = util.root_pattern("settings.gradle");
  };
```

lua/nvim_lsp/kotlin_language_server.lua
```diff
@@ -0,0 +1,58 @@
...
+		The MIT License (MIT)
+
+		Copyright (c) 2016 George Fraser
+		Copyright (c) 2018 fwcd
+
+	  ]];
+	};
+  };
+}
+
```
h-michael, line 58:
```suggestion
-- vim:et ts=2 sw=2
```

thchha:
Thank you for the feedback and fixes! I intended to bundle your suggestions but you can see the result.. Anyways, the server is running.
The linked [json](https://raw.githubusercontent.com/fwcd/vscode-kotlin/master/package.json) includes the configuration-options. I will append them into the README in an additional commit.

h-michael:
@thchha 
And I have a question.
This setting applies utf8 to the language server.
Where can I see if kotlin-language-server is using utf8?

thchha:
@h-michael 
[Spec 3.15.X](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#contentPart) describes that the only permitted encoding type is utf-8. 
This server builds on top of [lsp4j](https://github.com/eclipse/lsp4jhttps://github.com/eclipse/lsp4j) which is an implementation of this spec. It uses a generic implementation of jsonrpc "2.0", referenced [here](https://github.com/eclipse/lsp4j/blob/master/documentation/jsonrpc.md). This source specifies JSON [RPC4627](https://www.ietf.org/rfc/rfc4627.txt).
> 3.  Encoding
   JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8.
   Since the first two characters of a JSON text will always be ASCII
   characters [RFC0020], it is possible to determine whether an octet
   stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking
   at the pattern of nulls in the first four octets.
           00 00 00 xx  UTF-32BE
           00 xx 00 xx  UTF-16BE
           xx 00 00 00  UTF-32LE
           xx 00 xx 00  UTF-16LE
           xx xx xx xx  UTF-8

So the transport-protocol supports more encodings but the server-spec.

Within the sources of fwcd nor at lsp4j appears to be a direct specification (besides the actual encoding of the source-files) of the encoding. Should I file an issue at these repositories for further clarifications and notify here again?

h-michael:
@thchha 
Our "utf8" config does not point out that.
LSP deals character offset with UTF-16.
But some servers like "clangd" provide UTF8 option.
https://clangd.llvm.org/extensions.html#utf-8-offsets
Our "utf8" config is for using that option.
So if kotlin-language-server does not provide that option, we should not set "utf8" config.
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocuments

thchha:
Thanks for the fast clarification. I will look into your resources and commit an appropriate update!

Issue #207, opened 2020-04-24
Send nvim_lsp.util.sh output to a more terminal-like destination


dlukes:
Following up on the discussion in #201, currently, `nvim_lsp.util.sh` sends its output to a regular buffer, but it would be nice if it used a more terminal-like destination instead. As it stands, using e.g. `curl` with a progress bar appends all the progress bar states one after the other, separated by `^M`s.

I can think of two ways to achieve that:

1. Use the same destination as `:!` uses. That doesn't seem to be a regular buffer though (is it the builtin pager?), so it affords no control over how much screen real estate it should take up, and there may be no way of automatically dismissing the output on successful exit.
2. Using the builtin terminal with `termopen()`. That looks promising, except it currently seems to be [a pain to specify Lua callbacks](https://github.com/neovim/neovim/issues/7607), so it may need some VimL wrapping.

dlukes:
Ironically, that neovim/neovim#7607 issue I linked is gravitating towards a solution where a Lua version of `jobstart()` would basically wrap `vim.loop.spawn`, which is exactly the solution currently used by `nvim_lsp.util.sh` :)

dlukes:
Also, I tried setting the `buftype` of the output buffer to `terminal` with `vim.api.nvim_buf_set_option(bufnr, "buftype", "terminal")`, just to see if that would work (it would be the simplest, smallest change), but that doesn't seem to be allowed (it fails with `E474: Invalid argument`).

dlukes:
So #208 is my attempt to address this. I think the desired UX is mostly there, simply courtesy of the builtin terminal. The downside is that it has to use a global VimL function as the `on_exit` callback (see the PR for details why), which is ugly and potentially brittle.

PR #208, opened 2020-04-24
Use builtin terminal for util.sh output

dlukes:
Addresses #207. Supersedes #201.

Using `termopen()`, `util.sh` gets full terminal functionality for free:

- progress bars leveraging `\r` and other more fancy output types (e.g.
  curses UIs)
- interactive prompts
- canceling with `Ctrl-C`
- etc.

The drawback is that currently (see https://github.com/neovim/neovim/issues/7607),
the `on_exit` callback for `termopen()`/`jobstart()` *has* to be a VimL
function, which means a global VimL function has to be created
on-the-fly with a name which hopefully won't clobber anything that's
already defined.

Like #201, this commit makes sure that the output buffer remains open if
the script exits with anything else than 0 (so that the user can figure out the
cause of the error).

### DIFF
--- a/lua/nvim_lsp/util.lua
+++ b/lua/nvim_lsp/util.lua
@@ -409,58 +409,29 @@ function M.npm_installer(config)
 end
 
 function M.sh(script, cwd)
-  api.nvim_command("10new")
   assert(cwd and M.path.is_dir(cwd), "sh: Invalid directory")
-  local winnr = api.nvim_get_current_win()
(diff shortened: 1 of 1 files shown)


dlukes:
One way to get rid of that ugly VimL callback would be to always leave it up to the user to dismiss the terminal buffer, even with exit 0. Then the callback wouldn't be needed.

justinmk:
Nice change, let's try it.

Issue #200, opened 2020-04-22
Feature: Provide "Update" callback for servers


tjdevries:
Servers are constantly evolving and getting better, we should make it easy to update (even if it's just delete the folder and install again) for servers.

@teto Thoughts?

teto:
I am thoroughy against it xD It's the role of a package manager to keep the software up to date. I imagine this will be an endless source of bug reports/pull requests and we just don't have the resources to deal with all the edgecases, bump our software to deal with all the langague server updates

tjdevries:
Maybe I didn't make myself clear. I don't think WE should write complicated update scripts. However, we provide "Install" scripts and then after you install them, it's difficult to figure out what should be done to update the server. For example, update for any of the JS servers would probably just be "cd $install_location; npm update" or something like that.

DanCardin:
fwiw, i just went looking for how to do this and found this issue.

I would be happy with `:LspUninstall foo` (basically just deletes the folder given to me by `:LspInstallInfo`), or `:LspReinstall` (uninstall + install).

theJian:
uninstall + install would be an easy way to implement this. Providing an update command means nvim-lsp have to know how every package manager updates their packages, even it is simple like "cd $install_location; npm update".

thchha:
I just started contributing, but when adding support for the first server, I also stumbled upon this functionality. I do not like the fact that we even bother with installing language servers. There are already plugin-managers a user can opt-in to. I think the installation is not transparent (but convenient) to the user, updating is not taken care of and the additional complexity when fully supported is not neccessary. 
This repo calls itself a place of common configuration (which in the end are utilities for binding to nvim, right?) and the functionality for installing and updating across different build systems is in conflict with common idioms like YAGNI, KISS and the unix philosophy of doing only one thing, but this to the most effective extend.

So my proposal is the following: 
- Remove all functionalities for installing servers.
- Optionally provide an option for external plugin-managers to hook (read inject) into this plugin instead.
> For example a plugin should be able to require the 'nvim-lsp'-module and inject the server. Lua-plugins are not fully integrated into neoim yet, right?

teto:
@norcalli @h-michael I am of the opinion of @thchha : lets remove LspIntall/Uninstall. But you are the main maintainers here so that's your call, I just know I would not like to deal with such code.

h-michael:
I personally like to remove the functionality of the installer.
[...]

PR #209, opened 2020-04-25
add in the ability to reuse cs if already installed

ckipp01:
This pr adds in a check to see if cs ([the Coursier native launcher](https://get-coursier.io/docs/cli-installation)) is already installed. Using the native launch is faster than using the jar based launcher. Plus a lot of scala users may have this installed already, so there is no need to install Coursier again. However, if it's not found, it just downloads it as it would have before.

I also updated the config in info to match the actual root patterns that are set in the default config.

### DIFF
--- a/lua/nvim_lsp/metals.lua
+++ b/lua/nvim_lsp/metals.lua
@@ -21,12 +21,18 @@ local function make_installer()
       error('Need "JDK" to install this.')
       return
     end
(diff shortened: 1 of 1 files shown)


ckipp01:
After reading #200, if the install functionality goes away, this can obviously just be ignored and closed.

justinmk:
no harm in merging it meanwhile.

PR #210, opened 2020-04-27
Add tsconfig.json and .git to tsserver root_dir

theJian:


justinmk:
I feel like this was discussed before and rejected in general. OTOH it requires the current filetype to be typescript, so it's reasonable...

theJian:
@justinmk I see no harm to use `.git` as root_dir fallback for all filetypes. Do you like to open a new issue? I like to know what other people think of it.

lithammer:
> I feel like this was discussed before and rejected in general.

Maybe you're thinking of the home directory? 

Isn't [`tsconfig.json`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) a better root marker than `.git`? Although I guess both could be present.

gbrlsnchs:
Just a heads-up: when using TSServer for JavaScript, it's also possible to have a `jsconfig.json` instead of `tsconfig.json`.

theJian:
I don't want to do too many things in this PR. Adding .git to root_dir is because there is plenty of project don't use npm, including projects exist before npm. Searching package.json is insufficient to determine the project root in that case.

PR #221, opened 2020-05-05
Remove unnecessary init_options for Metals.

ckipp01:
As of the 0.9.0 release of Metals, a bug was fixed where previously all
of the `InitializationOptions` needed to be specified if you set any in
order to not have a null pointer exception. This is fixed now so all of
the default have been removed, and only the necessary ones remain.

### DIFF
--- a/lua/nvim_lsp/metals.lua
+++ b/lua/nvim_lsp/metals.lua
@@ -67,21 +67,9 @@ configs[server_name] = {
     message_level = vim.lsp.protocol.MessageType.Log;
     init_options = {
         statusBarProvider = "show-message",
-        didFocusProvider = false,
-        slowTaskProvider = false,
-        inputBoxProvider = false,
-        quickPickProvider = false,
-        executeClientCommandProvider = false,
-        doctorProvider = "html",
-        isExitOnShutdown = false,
         isHttpEnabled = true,
         compilerOptions = {
-          isCompletionItemDetailEnabled = true,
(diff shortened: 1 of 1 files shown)


h-michael:
@ckipp01 Would you show me where to find the release notes and corrections?

ckipp01:
> @ckipp01 Would you show me where to find the release notes and corrections?

Sure. 

- The original issue I mentioned that caused me to put them all in here is located here: https://github.com/scalameta/metals/issues/1668

- The pr that fixes this issue can be found here: https://github.com/scalameta/metals/pull/1674

- And the release that contains this can be found here: https://scalameta.org/metals/blog/2020/05/04/lithium.html

ckipp01:
> @ckipp01 Thanks :) LGTM
> However, for users who do not have Metals up to date, it is a destructive change, so I think it is good to merge it after a while, what do you think?

I actually think we should be fine to merge. During this last release cycle of Metals we also first added the ability to set these features via `InitializationOptions` (previously they were done with `ClientExperimentalCapabilities` or server properties). So 0.9.0 is actually the first stable version where setting `InitializationOptions` will work. I only added these in there earlier for those that were setting the version to the latest snapshot and wanted to utilize it. So in theory, this will have no effect on those using the last stable, since the `InitializationOptions` won't be taken into account.

So I think we should be fine.

h-michael:
I see. Thanks for the detailed explanation!

PR #238, opened 2020-05-17
feat(angular-ls): add angular language service

steelsojka:
This integrates the angular language server with nvim-lsp.

This may need some help with some of "probe" directories and cmd setting. So any help is appreciated but this does work in an angular project!

### DIFF
--- a/lua/nvim_lsp/angularls.lua
+++ b/lua/nvim_lsp/angularls.lua
@@ -1,0 +1,72 @@
+local configs = require 'nvim_lsp/configs'
+local util = require 'nvim_lsp/util'
+
+local server_name = 'angularls'
+local bin_name = server_name
+local install_loc = util.base_install_dir .. '/' .. server_name
+local script_loc = install_loc .. '/node_modules/@angular/language-server/index.js'
+local bin_loc = install_loc .. '/node_modules/.bin/angularls'
+
+local installer = util.npm_installer {
+  server_name = server_name;
+  packages = { '@angular/language-server' };
+  binaries = { bin_name };
+  -- angular-language-service doesn't expose a binary, so we create an execution wrapper.
+  post_install_script =
+    'echo "#! /bin/sh\n' .. 'node ' .. script_loc .. ' \\$*' .. '" > ' .. bin_loc .. '\n' ..
+    'chmod +x ' .. bin_loc;
+}
+
+-- Angular requires a node_modules directory to probe for @angular/language-service and typescript
+-- in order to use your projects configured versions.
+-- This defaults to the vim cwd, but will get overwritten by the resolved root of the file.
+local function get_probe_dir(root_dir)
+  local project_root = util.find_node_modules_ancestor(root_dir)
+
+  return project_root and (project_root .. '/node_modules') or ''
+end
+
+local default_probe_dir = get_probe_dir(vim.fn.getcwd())
+
+configs[server_name] = {
+  default_config = {
+    cmd = {
+      bin_loc,
+      '--stdio',
+      '--tsProbeLocations', default_probe_dir,
+      '--ngProbeLocations', default_probe_dir
+    };
+    filetypes = {'typescript', 'html', 'typescriptreact', 'typescript.tsx'};
+    -- Check for angular.json or .git first since that is the root of the project.
+    -- Don't check for tsconfig.json or package.json since there are multiple of these
+    -- in an angular monorepo setup.
+    root_dir = util.root_pattern('angular.json', '.git');
+  };
+  on_new_config = function(new_config, new_root_dir)
+    local new_probe_dir = get_probe_dir(new_root_dir)
+
+    -- We need to check our probe directories because they may have changed.
+    new_config.cmd = {
+      bin_loc,
+      '--stdio',
+      '--tsProbeLocations', new_probe_dir,
+      '--ngProbeLocations', new_probe_dir
+    }
+  end;
+  docs = {
+    description = [[
+https://github.com/angular/vscode-ng-language-service
+
+`angular-language-server` can be installed via `:LspInstall angularls`
+
+If you prefer to install this yourself you can through npm `npm install @angular/language-server`.
+Be aware there is no global binary and must be run via `node_modules/@angular/language-server/index.js`
+    ]];
+    default_config = {
+      root_dir = [[root_pattern("angular.json", ".git")]];
+    };
+  }
+}
+
+configs[server_name].install = installer.install
+configs[server_name].install_info = installer.info
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -126,10 +126,10 @@ function configs.__newindex(t, config_name, config_def)
 
       add_callbacks(new_config)
       if config_def.on_new_config then
-        pcall(config_def.on_new_config, new_config)
+        pcall(config_def.on_new_config, new_config, _root_dir)
       end
       if config.on_new_config then
-        pcall(config.on_new_config, new_config)
+        pcall(config.on_new_config, new_config, _root_dir)
       end
 
       new_config.on_init = util.add_hook_after(new_config.on_init, function(client, _result)
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ nvim_lsp.SERVER.setup{config}
     the second parameter instead. Useful for doing buffer-local setup.
 
   {on_new_config}
-    `function(new_config)` will be executed after a new configuration has been
+    `function(new_config, new_root_dir)` will be executed after a new configuration has been
     created as a result of {root_dir} returning a unique value. You can use this
     as an opportunity to further modify the new_config or use it before it is
     sent to |vim.lsp.start_client()|.
@@ -241,6 +241,7 @@ The following LSP configs are included. Follow a link to find documentation for
 that config.
 
 - [als](#als)
+- [angularls](#angularls)
 - [bashls](#bashls)
 - [ccls](#ccls)
 - [clangd](#clangd)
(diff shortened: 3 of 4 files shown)


lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+local function get_probe_dir()
+  local project_root = util.find_node_modules_ancestor(vim.fn.getcwd())
+
+  return project_root and (project_root .. '/node_modules') or ''
+end
+
+local default_prob_dir = get_probe_dir()
+
+configs[server_name] = {
+  name = 'angular',
```
h-michael, line 29: This may not be needed.
```suggestion
```

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
+local configs = require 'nvim_lsp/configs'
+local util = require 'nvim_lsp/util'
+
+local server_name = 'angularls'
```
h-michael, line 4: File name should be the same with the server name.
Would you rename `angular.lua` to `angularls.lua`?

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+    'chmod +x ' .. bin_loc
+}
+
+local function get_probe_dir()
+  local project_root = util.find_node_modules_ancestor(vim.fn.getcwd())
+
+  return project_root and (project_root .. '/node_modules') or ''
+end
+
+local default_prob_dir = get_probe_dir()
```
h-michael, line 26:
```suggestion
local default_probe_dir = get_probe_dir()
```

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+
+configs[server_name] = {
+  name = 'angular',
+  default_config = {
+    cmd = {
+      bin_loc,
+      '--stdio',
+      '--tsProbeLocations', default_prob_dir,
+      '--ngProbeLocations', default_prob_dir
```
h-michael, line 35:
```suggestion
      '--tsProbeLocations', default_probe_dir,
      '--ngProbeLocations', default_probe_dir
```

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+      local install_info = installer.info()
+      local new_prob_dir = get_probe_dir()
+
+      if install_info.is_installed then
+        -- We need to check our probe directories because they may have changed.
+        new_config.cmd = {
+          bin_loc,
+          '--stdio',
+          '--tsProbeLocations', new_prob_dir,
+          '--ngProbeLocations', new_prob_dir
```
h-michael, line 49:
```suggestion
          '--tsProbeLocations', new_probe_dir,
          '--ngProbeLocations', new_probe_dir
```

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+
+`angular-language-server` can be installed via `:LspInstall angularls`
+
+If you prefer to install this yourself you can through npm `npm install @angular/language-server`.
+Be aware there is no global binary and must be run via `node_modules/@angular/language-server/index.js`
+    ]],
+    default_config = {
+      root_dir = [[root_pattern("angular.json", "tsconfig.json", "package.json", ".git")]],
+      on_init = [[function to handle changing offsetEncoding]],
+      capabilities = [[default capabilities, with offsetEncoding utf-8]]
```
h-michael, line 66:
```suggestion
```

lua/nvim_lsp/angular.lua
```diff
@@ -0,0 +1,72 @@
...
+      bin_loc,
+      '--stdio',
+      '--tsProbeLocations', default_prob_dir,
+      '--ngProbeLocations', default_prob_dir
+    },
+    filetypes = {'typescript', 'html', 'typescriptreact', 'typescript.tsx'},
+    root_dir = util.root_pattern('angular.json', 'tsconfig.json', 'package.json', '.git'),
+    on_new_config = function(new_config)
+      local install_info = installer.info()
+      local new_prob_dir = get_probe_dir()
```
steelsojka, line 41: @h-michael Do you know if there is a way to get the new resolved root directory in the `on_new_config` hook? I think using the `vim.fn.getcwd()` isn't quite the right way here.

lua/nvim_lsp/angularls.lua
```diff
@@ -0,0 +1,69 @@
...
+      bin_loc,
+      '--stdio',
+      '--tsProbeLocations', default_probe_dir,
+      '--ngProbeLocations', default_probe_dir
+    },
+    filetypes = {'typescript', 'html', 'typescriptreact', 'typescript.tsx'},
+    root_dir = util.root_pattern('angular.json', 'tsconfig.json', 'package.json', '.git'),
+    on_new_config = function(new_config)
+      local install_info = installer.info()
+      local new_probe_dir = get_probe_dir()
```
steelsojka, line 40: @h-michael  Do you know if there is a way to get the new resolved root directory in the `on_new_config` hook? I think using the `vim.fn.getcwd()` isn't quite the right way here.

h-michael: I think in this case(node project), probe directory equal the directory where `node_modules` exists.
So I think that that is `util.root_pattern('package.json').."/node_modules"`.

steelsojka: Actually this doesn't work because we need to know the resolved root directory that caused this hook to execute. For example, I have two angular projects and I open index.html in one and index.html in the other. The probe directories always point to the initial cwd. 

I updated this and added an argument to the `on_new_config` hook that passes the newly resolved root directory. This seemed like the only way to get that newly resolved root to generate the new command and doesn't break any existing functionality.

lua/nvim_lsp/configs.lua
```diff
@@ -110,10 +110,10 @@ function configs.__newindex(t, config_name, config_def)
 
       add_callbacks(new_config)
       if config_def.on_new_config then
-        pcall(config_def.on_new_config, new_config)
+        pcall(config_def.on_new_config, new_config, _root_dir)
```
steelsojka, line 113: This change makes it possible to use the newly resolved root directory in any config transformations that are needed and doesn't cause any breaking changes.

lua/nvim_lsp/angularls.lua
h-michael: @steelsojka
Could you give me a little more detail?
Guess your example, is it a project with multiple node_modules in a mono repo?

steelsojka: In an angular specific mono repo you only have one node_modules folder but multiple `package.json` files.

You can ignore my comment above. The issue I was running into is solved by providing the new root dir (from the `root_dir` function) to the `on_new_config` hook.

h-michael: I don't know why we should add root_dir argument.
Does it mean that we need to add "root_dir" argument because the root_dir may be different depending on the files opened in the same project?

steelsojka: The problem is we need the config to generate the new command. You open a file from two separate projects in the same  vim instance. We get a command for each project because the roots are different. We need to set the probe directories based off of that new root dir.

Now the problem with the current implementation is we don't have access to that new root dir. You can't use `new_config.root_dir` because it is set AFTER the hooks are invoked. So the solution is you either pass it as an argument to the hook OR you assign `root_dir` to `new_config` prior to calling the `on_new_config` callback. I'm fine with either solution, but we need some way to access the new root directory.

h-michael: > You open a file from two separate projects in the same vim instance. We get a command for each project because the roots are different. We need to set the probe directories based off of that new root dir.

Our built-in language server client can't you multiple same language server in same Nvim instance.
`lsp.foo.setup()` function is set one-shot autocmd to start a language server.

steelsojka: Looking at the code, it looks like it can. Basically if the `root_dir` function returns a new root directory that hasn't had a client created for that root, it starts a new client. I think that is one of the purposes of the `root_dir` function and the `server_per_root_dir_manager`.

https://github.com/neovim/nvim-lsp/blob/master/lua/nvim_lsp/util.lua#L246

So when a new config is generated, we need to have that new root directory to generate the new command for the new server that is being spawned.

steelsojka: @h-michael Is the point above make sense? Is there anything else required for this to go in?

h-michael:
@steelsojka 
Sorry for the late reply, could you wait until the weekend?

steelsojka:
@h-michael No problem. There's no rush.

justinmk:
are you using this and it's working? if so let's merge it

steelsojka:
I was using this for a bit, but had to disable it. At the time I put in this pull request, diagnostics couldn't be published from multiple clients with the same file. For example, most devs would have the tsserver and angularls running for a single TS file. The diagnostics from one overwrote the other. I thought there was work being done to fix it. But this should be good to go from a language server perspective. I'll rebase and test it again.

steelsojka:
@justinmk The server works after rebasing, but we run into a problem with LSP in TS files. Since tsserver and angularls are attached to the same buffer, but each server has different capabilities, we get this error constantly.
Screen Shot 2020-08-31 at 7 20 28 AM

As far as the server is concerned, it works great.

steelsojka:
I believe this PR would fix the errors. https://github.com/neovim/neovim/pull/12764

mhartington:
looks likes neovim/neovim#12764 has been merged in

steelsojka:
https://github.com/neovim/neovim/pull/12655 This was the other blocker but is now resolved. Angular server and TS server both publish TS diagnostics. They used to overwrite each other. So I don't believe anything else is blocking this.

PR #243, opened 2020-05-19
Add configuration for elixir-ls

mortepau:
Add support for Elixir language server.

If you run `:LspInstaller elixirls` it clones the elixir-lsp/elixir-ls repo and builds the binary. It is only tested on Ubuntu 19.10, so should probably be tested on other environments. I'm not completely sure if the Windows install works. 

There is an error when trying to open an elixir-file if the LSP's cmd isn't configured correctly, I think that is caused by the autocmd created by M.setup in configs.lua.

### DIFF
--- a/lua/nvim_lsp/elixirls.lua
+++ b/lua/nvim_lsp/elixirls.lua
@@ -3,21 +3,21 @@ local util = require 'nvim_lsp/util'
 
 local server_name = "elixirls"
 local bin_name = "elixir-ls"
-local bin
+local cmd = "language_server"
+
+if vim.fn.has('mac') == 1 or vim.fn.has('unix') == 1 then
+    cmd = cmd..".sh"
+elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
+    cmd = cmd..".bat"
+else
+    error("System is not supported, try to install manually.")
+    return
+end
 
 local function make_installer()
     local P = util.path.join
     local install_dir = P{util.base_install_dir, server_name}
-    local git_dir = P{install_dir, bin_name}
-
-    if vim.fn.has('mac') == 1 or vim.fn.has('unix') == 1 then
-        bin = P{git_dir, "release", "language_server.sh"}
-    elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
-        bin = P{git_dir, "release", "language_server.bat"}
-    else
-        error("System is not supported, try to install manually.")
-        return
-    end
+	local cmd_path = P{install_dir, bin_name, "release", cmd}
 
     local X = {}
     function X.install()
@@ -51,9 +51,9 @@ local function make_installer()
 
     function X.info()
         return {
-            is_installed = util.path.exists(bin);
+            is_installed = util.path.exists(cmd_path);
             install_dir = install_dir;
-            cmd = { bin };
+            cmd = { cmd_path };
         }
     end
 
@@ -70,7 +70,7 @@ local installer = make_installer()
 
 configs[server_name] = {
   default_config = {
-    -- cmd = {bin};
+    cmd = { cmd };
     filetypes = {"elixir", "eelixir"};
     root_dir = function(fname)
         return util.root_pattern("mix.exs", ".git")(fname) or vim.loop.os_homedir()
--- a/lua/nvim_lsp/elixirls.lua
+++ b/lua/nvim_lsp/elixirls.lua
@@ -38,7 +38,7 @@ local function make_installer()
             # clone project
             git clone https://github.com/elixir-lsp/elixir-ls
             cd elixir-ls
-            
+
             # fetch dependencies and compile
             mix deps.get && mix compile
 
--- a/lua/nvim_lsp/elixirls.lua
+++ b/lua/nvim_lsp/elixirls.lua
@@ -70,7 +70,7 @@ local installer = make_installer()
 
 configs[server_name] = {
   default_config = {
-    cmd = {bin};
+    -- cmd = {bin};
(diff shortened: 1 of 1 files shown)


h-michael:
@skbolton Would you try this?

skbolton:
@h-michael Trying now

skbolton:
Yeah seems to be working (MacOS Catalina). Only thing I have noticed so far is I can't get hover to actually pull and documentation for things.

h-michael:
@skbolton Thanks!
Does that problem occurs with coc.nvim?

skbolton:
Yeah I usually have problem with docs. I will dig into the code and see if I can assist on this

lua/nvim_lsp/elixirls.lua
```diff
@@ -0,0 +1,110 @@
...
+        end
+    end
+    return X
+end
+
+local installer = make_installer()
+
+configs[server_name] = {
+  default_config = {
+    -- cmd = {bin};
```
h-michael, line 73: I think that we had better define default cmd for users using binary on PATH.

mortepau: That is true. I'm not entirely sure where that should be, but the commented out version pointed to a path in the installation directory. Is that a reasonable place?

lua/nvim_lsp/elixirls.lua
```diff
@@ -0,0 +1,110 @@
...
+    local git_dir = P{install_dir, bin_name}
+
+    if vim.fn.has('mac') == 1 or vim.fn.has('unix') == 1 then
+        bin = P{git_dir, "release", "language_server.sh"}
+    elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
+        bin = P{git_dir, "release", "language_server.bat"}
+    else
+        error("System is not supported, try to install manually.")
+        return
+    end
```
h-michael, line 20: @mortepau https://github.com/neovim/nvim-lsp/pull/243/files#r427125879
How about this?

```suggestion
local server_name = "elixirls"
local bin_name = "elixir-ls"
local cmd = "language_server"

if vim.fn.has('mac') == 1 or vim.fn.has('unix') == 1 then
    cmd = cmd..".sh"
elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
    cmd = cmd..".bat"
else
    error("System is not supported, try to install manually.")
    return
end

local function make_installer()
    local P = util.path.join
    local install_dir = P{util.base_install_dir, server_name}
	local cmd_path = P{install_dir, bin_name, "release", cmd}
```
and replace bin to cmd_path.

mortepau: @h-michael Yes. That looks cleaner in my opinion.

lua/nvim_lsp/elixirls.lua
mortepau, line 73:
```suggestion
    cmd = { cmd_path };
```
https://github.com/neovim/nvim-lsp/pull/243/files#r427172387

lua/nvim_lsp/elixirls.lua
```diff
@@ -0,0 +1,110 @@
...
+        ]=]
+        vim.fn.mkdir(install_info.install_dir, "p")
+        util.sh(script, install_info.install_dir)
+    end
+
+    function X.info()
+        return {
+            is_installed = util.path.exists(bin);
+            install_dir = install_dir;
+            cmd = { bin };
```
mortepau, line 56:
```suggestion
            is_installed = util.path.exists(cmd_path);
            install_dir = install_dir;
            cmd = { cmd_path };
```
https://github.com/neovim/nvim-lsp/pull/243/files#r427172387

lua/nvim_lsp/elixirls.lua
h-michael: No, I think this should be "cmd". If users put bin at directory through global path, cmd is executable.

mortepau: I'm not entirely sure if this is correct, but wouldn't this just wrap the cmd string which holds the binary and not the complete path?

h-michael: Yes, AFAIK, elixr's ls formula was to run this script instead of the binary, so if I put something in the path, wouldn't it be this script?

mortepau: Yes, I just forgot about how global path works for a second...

mortepau:
@skbolton https://github.com/neovim/nvim-lsp/pull/243#issuecomment-630561029
Regarding your issue with hover, you need to save your changes so that elixir-ls can recompile the project. As far as I know, it gets all of the documentation after compiling it. For me it works with:
```
" Add nvim-lsp as a plugin
set rtp+=$HOME/src/nvim-lsp

lua << END
    local nvim_lsp = require 'nvim_lsp'

    nvim_lsp.elixirls.setup{}
END

" Find elixir files and give them elixir filetype
au BufRead,BufNewFile *.ex,*.exs set filetype=elixir

" Use LSP omni-completion in elixir files.
autocmd FileType elixir setlocal omnifunc=v:lua.vim.lsp.omnifunc

" Add keybindings for LSP
nnoremap <silent> gd <cmd>lua vim.lsp.buf.definition()<CR>
nnoremap <silent> gr <cmd>lua vim.lsp.buf.references()<CR>
nnoremap <silent> K <cmd>lua vim.lsp.buf.hover()<CR>
nnoremap <silent> gsd <cmd>lua vim.lsp.buf.document_symbol()<CR>
nnoremap <silent> gsw <cmd>lua vim.lsp.buf.workspace_symbol()<CR>
```

PR #245, opened 2020-05-21
feature(jdtls): add jdt ls for java

steelsojka:


lua/nvim_lsp/jdtls.lua
```diff
@@ -0,0 +1,139 @@
...
+end
+
+local installer = make_installer()
+
+configs[server_name] = {
+  default_config = {
+    filetypes = { "java" };
+    root_dir = util.root_pattern('.git');
+    init_options = {
+      workspace = vim.loop.os_homedir();
```
h-michael, line 101: I don't write Java.
@mfussenegger Do you think this is appropriate?

steelsojka: I don't write it "full time" (maintain a REST service layer) but this would be the eclipse workspace... which I'm not sure where that is for people. It can be changed during `setup` but if there is a better default, I'm all for it.

mfussenegger: According to https://github.com/eclipse/eclipse.jdt.ls/wiki/Running-the-JAVA-LS-server-from-the-command-line the `workspace` option doesn't exist. There would be `workspaceFolders`, but that is optional. 

But the `-data` argument has to point to the workspace folder. Which I guess is done here:  https://github.com/neovim/nvim-lsp/pull/245/files#diff-eaf8d0e99cf1a581b37366966939af30R82

Not sure if `os_homedir` is a good choice. The eclipse.jdt.ls wiki says:

> Choosing a value for -data: the value for your data directory, should be the directory where your active workspace is, and you wish for the java langserver to add in its default files. Should also be the absolute path to this directory, ie., /home/username/workspace


I personally have set it to `~/.local/share/eclipse`, and I even choose a different sub-folder for each project, because I've otherwise had problems when switching between different projects or working on 2 in parallel in the past, but since this is only a default, going with `$HOME/workspace` may be a good enough choice?

steelsojka: That sounds good to me. `workspace` is just what i called the option for setup. If you look at the configure method it sets it to `-data`. I feel like putting `data` as an option might be a little to vague.

bryall:
I am having trouble getting this working
* if jdtls is not installed the X.get_launcher() fails and will not let you install jdtls
* after I got past that, the lsp starts and then dispalys the following error:
"Error executing vim.schedule lua callback: /usr/local/share/nvim/runtime/lua/vim/lsp/util.lua:941: Is a directory"

steelsojka:
@bryall I moved the launcher call out of the installer info, realizing that the installer uses that internally BEFORE actually installing anything.

As far as the lua error, do you have your setup and the file you were working with?

bryall:
the lua error was from the "--data" option for me, unsure why, commentint it out and I get no error

update did fix the install issues

I think the rest of the problems I am seeing are jdtls specific issues,

steelsojka:
@bryall Did you set the `workspace` (-data) option yourself or use the default `~/workspace`?

lua/nvim_lsp/jdtls.lua
```diff
@@ -0,0 +1,138 @@
...
+    init_options = {
+      workspace = path.join { vim.loop.os_homedir(), "workspace" };
+      jvm_args = {};
+      os_config = nil;
+    };
+    callbacks = {
+      -- Due to an invalid protocol implementation in the jdtls we have to
+      -- conform these to be spec compliant.
+      -- https://github.com/eclipse/eclipse.jdt.ls/issues/376
+      ['textDocument/codeAction'] = function(a, b, actions)
```
steelsojka, line 108: @mfussenegger I noticed your plugin here https://github.com/mfussenegger/nvim-jdtls (which looks fantastic) 

Do you think this should only be responsible for spawning the server and leave the rest of the logic to plugins like yours? (should I remove this callback?)

mfussenegger: It would make sense to me.

If `nvim-lsp` is up for adding optional dependencies for language specific plugins I could also imagine adding some more functions to `nvim-jdtls`, so that `nvim-lsp` could do a `pcall(require, 'jdtls')` and then optionally use these.

For example, in my dotfiles I already do something like this:

```
    config['init_options'] = {
        settings = {
            java = {
              ...
            };
        };
        bundles = bundles;
        extendedClientCapabilities = require('jdtls').extendedClientCapabilities;
    }
```

RishabhRD:
What's the status of this pull request. It seems to work fine. Why it's not being merged to master? Issues referenced are closed .

steelsojka:
I've been using this configuration in my local config daily since I submitted this PR and it's works for me with my setup. I have NOT tested this with java versions < 8. I also don't develop Java full time (20% of the time), so if any Java gurus have any improvements they would like to add, please submit a review.

RishabhRD:
I have also tested it. It works all the time. I found it better than coc-java. As coc-java startup time was terrible. If there is any more issue please reference it so that it can be resolved.

PR #246, opened 2020-05-22
flow: use npx instead of expecting a npm script

ybc37:
Hey!

I noticed the config for the flow language server expects a specific npm script: `npm run flow lsp`. Wouldn't it be better to not rely on this npm script and use npx instead: `npx flow lsp`. This would also mean that flow will be installed if not present in the current project (see https://github.com/npm/npx#description).

Tbh, I'm not familiar with flow and how projects use it. I'm just looking at one project without that npm script and tested it with npx. If there are reason why it's done this way, feel free to close this PR, of course.

### DIFF
--- a/lua/nvim_lsp/flow.lua
+++ b/lua/nvim_lsp/flow.lua
@@ -3,7 +3,7 @@ local util = require 'nvim_lsp/util'
 
 configs.flow = {
   default_config = {
-    cmd = {"npm", "run", "flow","lsp"};
+    cmd = {"npx", "flow","lsp"};
     filetypes = {"javascript", "javascriptreact", "javascript.jsx"};
     root_dir = util.root_pattern(".flowconfig");
   };
@@ -19,7 +19,7 @@ https://flow.org/en/docs/install/
 See below for lsp command options.
 
 ```sh
-npm run flow lsp -- --help
+npx flow lsp --help
 ```
     ]];
     default_config = {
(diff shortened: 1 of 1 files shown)


h-michael:
This is official recomendation.
https://flow.org/en/docs/install/

If you use npx, you use latest version.
But you should use same version as .flowconfig.

ybc37:
Hey @h-michael,

I see the install docs more like example usage. There are projects using flow, but not having a `flow` npm script, e.g. [yarn](https://github.com/yarnpkg/yarn/blob/master/package.json), [jitsi-meet-electron](https://github.com/jitsi/jitsi-meet-electron/blob/master/package.json) and I guess others.

>If you use npx, you use latest version.

I don't think so. The [readme of `npx`](https://github.com/npm/npx#description) states:

> Executes <command> either from a local node_modules/.bin, or from a central cache, installing any packages needed in order for <command> to run.

So it first tries to use local installed `flow`, which is the version defined in `package.json` (and the one that should be in `.flowconfig`).

Though I see it could be confusing/wrong if the local packages aren't installed yet (`npm i`/`yarn` in the project dir). Then it would install (or fall back to) `flow` in the central cache and indeed use the latest version.

If `npx` is not an option here, maybe just a notice in the docs, that the default `cmd` for the flow config requires a `flow` npm script?

ybc37:
I changed the command to use `npx --no-install`, so it doesn't try to install `flow` if it's not yet installed. I don't think nvim-lsp should trigger a installation of `flow`, but rely on already installed `flow`.

h-michael:
@ybc37 Good! Thanks :)

PR #253, opened 2020-05-27
feat(javascript/typescript) add javascript-typescript-langserver

### LINKED ISSUE #250: typescript-language-server no longer maintained
It looks like the [typescript-language-server](https://github.com/theia-ide/typescript-language-server) library is no longer actively maintained ([source](https://github.com/theia-ide/typescript-language-server/issues/141)).

There's another language server that supports both javascript and typescript that seems to be more active ([javascript-language-server](https://github.com/sourcegraph/javascript-typescript-langserver)). Would that be a better option?

lynndylanhurley:
The current typescript / javascript language server ([typescript-language-server](https://github.com/theia-ide/typescript-language-server)) is no longer actively maintained ([source](https://github.com/theia-ide/typescript-language-server/issues/141)).

This PR adds support for [javascript-typescript-langserver](https://github.com/sourcegraph/javascript-typescript-langserver), which is the most active js / ts language server I could find.

Fixes #250.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,7 @@ that config.
 - [hie](#hie)
 - [html](#html)
 - [intelephense](#intelephense)
+- [javascript_typescript_langserver](#javascript_typescript_langserver)
 - [jsonls](#jsonls)
 - [julials](#julials)
 - [kotlin_language_server](#kotlin_language_server)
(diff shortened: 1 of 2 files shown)


glepnir:
I think this is not a solution to the problem because javascript-typescript-langserver is only updating dependencies for a long time without fixing the problem. And it has some memory leaks...

lynndylanhurley:
I've been switching between javascript-typescript-langserver and typescript-language-server for the past week and both seem to cause high CPU load. Also vim freezes with either one enabled. @glepnir what do you recommend for typescript LSP?

Lazerbeak12345:
Actually, typescript-language-server is being maintained now, so perhaps this isn't needed... ([source](https://github.com/theia-ide/typescript-language-server/issues/141#issuecomment-706584854))

DonnieWest:
https://github.com/sourcegraph/javascript-typescript-langserver#%EF%B8%8F-this-project-is-no-longer-maintained

Unfortunately, `javascript-typescript-langserver` is no longer maintained. I can confirm though that `typescript-language-server` is maintained :P

mjlbach:
@lynndylanhurley Can we close this given https://github.com/theia-ide/typescript-language-server is actively maintained?

PR #258, opened 2020-06-02
Fix julials configuration

kdheepak:
This PR updates julials to a working configuration. It appears adding `using SymbolServer` is important to prevent segfaults.

See https://github.com/julia-vscode/LanguageServer.jl/issues/735 for more information.


cmcaine:
I think your commit is not clean: it has changes to jedi and dart, too. Might need a rebase :)

cmcaine:
Oh, sorry, it's the bot's commit that is not clean. I don't know what that's about, so I'll bow out.

kdheepak:
I branched from the latest `master`, and double checked as well when I saw the bot's commit :) I should have posted here though. I'm not sure what is going on with the bot.

kdheepak:
The bot updates the README based on the options in the `package.json` from respective vscode plugins. I didn't know that, that's pretty neat! It was looking at [this file](https://raw.githubusercontent.com/Dart-Code/Dart-Code/master/package.json) and that has [changed](https://github.com/Dart-Code/Dart-Code/commit/2e709d945d464feb60d49ffdf2ecd7b914c30e55#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) in the last week or so since the [README.md](https://github.com/neovim/nvim-lsp/blob/54f4f823a27ffa8f9e6c308d550888f3ae21373e/README.md) was last updated.

kdheepak:
So this PR is ready to be reviewed.

lua/nvim_lsp/julials.lua
```diff
@@ -5,10 +5,18 @@ configs.julials = {
...
     cmd = {
         "julia", "--project", "--startup-file=no", "--history-file=no", "-e", [[
-        using LanguageServer;
         using Pkg;
-        server = LanguageServer.LanguageServerInstance(stdin, stdout, false, dirname(Pkg.Types.Context().env.project_file));
-        server.runlinter = true; run(server);
+        Pkg.instantiate()
+        using LanguageServer; using SymbolServer;
+        depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
+        project_path = dirname(Pkg.Types.Context().env.project_file)
```
KristofferC, line 12: Is this the same as `Base.active_project()`?

kdheepak: Ah, yes, I could use that instead then.

kdheepak: Is that what you would suggest?

KristofferC: If it works, yes. It seems better than digging so many layers into the Pkg internals.

h-michael: @kdheepak Mention me if you decide what to do here.

kdheepak: @KristofferC, I made the change you suggested and it works. So I'll resolve this conversation.

kdheepak:
@h-michael I have a question here. 

The previous implementation of the configuration was installing the Julia LanguageServer and SymbolServer packages in the global Julia environment. This is not recommended by the LanguageServer developers. 

Julia has good support for environments using built in package manager. After discussing with other Julia users, I've modified the PR above to add an environment that is specified by a `JuliaProject.toml` file. And I've modified the lsp client to start the Julia process and run the LanguageServer code using this specified environment file.

My question is, is this okay? Where should one put supplementary files to the lua files. Right now I put the `JuliaProject.toml` file in the `lua/nvim_lsp` folder. If you have a better suggestion let me know and I'll modify this PR.

kdheepak:
Thanks for the comment @clason! I think making that configurable as part of a user setting is straightforward. But we still need to ship the `JuliaProject.toml` file as part of this repo, and I'm wondering where in this repo should we put this file.

kdheepak:
Okay, I've updated it such that it uses `~/.cache/nvim/nvim_lsp/julials`. I've also removed the `JuliaProject.toml` file. I think users can see the documentation and manually configure this environment however they want.

clason:
I was about to write something like that :) One small nit: I wouldn't hard code the path but refer to the XDG compliant variables (e.g., `vim.fn.stdpath("cache")`).

kdheepak:
I've used this

```
environment_directory = util.path.join(util.base_install_dir, "julials")
```

I've hardcoded it in the docstring though. Maybe I should change it there was well?

clason:
> I've hardcoded it in the docstring though. Maybe I should change it there was well?

Well, the doc string should correspond to what actually happens :) Yes, better replace it with that, otherwise it will be confusing on Windows... (If you look at `util.lua`, it's indeed `stdpath('cache')/nvim_lsp`.)

kdheepak:
Okay great! Thanks for all the comments! I've updated it so that docstring uses the same `environment_directory` as the install and cmd script. I've also tested it locally and it works as expected.

clason:
(but my user name isn't `runner` ;))

kdheepak:
Okay, dependencies are being found correctly now as well! Thanks to @Non-jedi for all the help figuring this out.

kdheepak:
With the exception of autocomplete, everything is working as expected for me across multiple projects. This should be good to do :) cc @h-michael

lua/nvim_lsp/julials.lua
```diff
@@ -22,23 +32,24 @@ configs.julials = {
 https://github.com/julia-vscode/julia-vscode
 `LanguageServer.jl` can be installed via `:LspInstall julials` or by yourself the `julia` and `Pkg`:
 ```sh
-julia -e 'using Pkg; Pkg.add("LanguageServer")'
+julia --project=]] .. environment_directory .. [[ -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
```
h-michael, line 35: https://github.com/neovim/nvim-lsp/pull/258/files#diff-04c6e90faac2675aa89e2176d2eec7d8R2236

The document will be generated based on the environment in which the generating script was executed.
Since every user's environment is different, you have to ask the user ":lua print(require'nvim _lsp'.util.path.join(require'nvim_lsp'.util .base_install_dir, "julials"))" to check the path Encourage them to do so.

kdheepak: I added a comment in this description guiding users that want to install it manually.

kdheepak:
I've made the change you requested @h-michael and rebased on top of the latest master as well.

h-michael:
Ty :)

PR #262, opened 2020-06-04
Add CMake language server

Shatur95:
~To test it you need the latest master, see https://github.com/regen100/cmake-language-server/issues/16~


Shatur95:
New version of CMake language server is released, you can install it from pip and this PR will work.

PR #263, opened 2020-06-04
Improve clangd configuration

Shatur95:
I added support for [this](https://clangd.llvm.org/extensions.html#switch-between-sourceheader) extension. Very convenient feature to work with C / C++.
Made by analogy with [texlab](https://github.com/neovim/nvim-lsp/blob/master/lua/nvim_lsp/texlab.lua)

### DIFF
--- a/lua/nvim_lsp/clangd.lua
+++ b/lua/nvim_lsp/clangd.lua
@@ -1,6 +1,7 @@
 local configs = require 'nvim_lsp/configs'
 local util = require 'nvim_lsp/util'
 
+-- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/clangd.lua
```diff
@@ -1,6 +1,16 @@
 local configs = require 'nvim_lsp/configs'
 local util = require 'nvim_lsp/util'
 
+local function switch_source_header(bufnr)
```
h-michael, line 4:
```suggestion
-- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
local function switch_source_header(bufnr)
```

h-michael:
Thanks :)

PR #267, opened 2020-06-07
Add coursier candidate, fix folder creation when empty

landerlo:
Fix a couple of problems with the installation. The standard executable is coursier, cs is suggested as alias, which won't be necessarily found depending on the $SHELL, etc. Now we check for both. It also fixes issue where mkdir would fail if parents missing in case coursier wasn't downloaded.
/cc @ckipp01

### DIFF
--- a/lua/nvim_lsp/metals.lua
+++ b/lua/nvim_lsp/metals.lua
@@ -27,17 +27,23 @@ local function make_installer()
       error('Need "JDK" to install this.')
       return
     end
-    local coursier_exe
-    if not (util.has_bins("cs")) then
+
+    local coursier_exe = nil
+    if util.has_bins("cs") then
+      coursier_exe = "cs"
+    elseif util.has_bins("coursier") then
+      coursier_exe = "coursier"
+    end
+    if not coursier_exe then
       coursier_exe = install_dir .. "/coursier"
       local download_cmd = string.format("curl -fLo %s --create-dirs https://git.io/coursier-cli", coursier_exe)
       local chmod_cmd = string.format("chmod +x %s", coursier_exe)
       vim.fn.system(download_cmd)
       vim.fn.system(chmod_cmd)
     else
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/metals.lua
```diff
@@ -27,17 +27,23 @@ local function make_installer()
       error('Need "JDK" to install this.')
       return
     end
-    local coursier_exe
-    if not (util.has_bins("cs")) then
+
+    local coursier_exe = nil
+    if util.has_bins("coursier") then
```
ckipp01, line 32: I think we should probably reverse this check to look for `cs` first. I a lot of scenarios, if the user has `coursier`, they will have `cs`, and it's more preferable to use the native `cs`. I'd rather use the `coursier` exec as a backup. Is there a reason you have them in this order?

lua/nvim_lsp/metals.lua
```diff
@@ -27,17 +27,23 @@ local function make_installer()
...
+    if not coursier_exe then
       coursier_exe = install_dir .. "/coursier"
       local download_cmd = string.format("curl -fLo %s --create-dirs https://git.io/coursier-cli", coursier_exe)
       local chmod_cmd = string.format("chmod +x %s", coursier_exe)
       vim.fn.system(download_cmd)
       vim.fn.system(chmod_cmd)
     else
-      os.execute("mkdir " .. install_dir)
-      coursier_exe = "cs"
+      os.execute("mkdir -p " .. install_dir)
```
ckipp01, line 44: Ah, good catch 👍

lua/nvim_lsp/metals.lua
landerlo: ok, I've changed the order now

lua/nvim_lsp/metals.lua
justinmk: can use `vim.fn.mkdir(...,'p')` instead

Issue #264, opened 2020-06-06
Custom commands don't seem to be added


wbthomason:
- `nvim --version`:
```
NVIM v0.5.0-547-gdd4018947
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -D_FORTIFY_SOURCE=2 -march=native -O2 -pipe -fno-plt -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/yaytmp-1000/neovim-git/src/build/config -I/tmp/yaytmp-1000/neovim-git/src/neovim-git/src -I/usr/include -I/tmp/yaytmp-1000/neovim-git/src/build/src/nvim/auto -I/tmp/yaytmp-1000/neovim-git/src/build/include
Compiled by wil@loki

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
- nvim-lsp version(commit hash): a7fdf26
- `:checkhealth` result:
<details>
<pre>
health#completion_nvim#check
========================================================================
## general
  - OK: neovim version is supported

## completion source
  - OK: all completion source are valid

## snippet source
  - INFO: You haven't setup any snippet source.

health#nvim_lsp#check
========================================================================
## Checking language server protocol configuration
  - INFO: html: configuration checked.
  - INFO: texlab: configuration checked.
  - INFO: jsonls: configuration checked.
  - INFO: clangd: configuration checked.
  - INFO: ocamllsp: configuration checked.
  - INFO: julials: configuration checked.
  - INFO: bashls: configuration checked.
  - INFO: cssls: configuration checked.
  - INFO: sumneko_lua: configuration checked.
  - INFO: ghcide: configuration checked.
  - INFO: vimls: configuration checked.
[...]

PR #268, opened 2020-06-09
fix the problem that commads cannot be registered in custom setup

h-michael:
resolve: https://github.com/neovim/nvim-lsp/issues/264


h-michael:
@wbthomason Does this solve #264 ?

wbthomason:
@h-michael It seems to, yes. Thanks!

PR #280, opened 2020-06-16
change Julia LS to use the new dedicated neovim LS wrapper LSPNeovim.jl

ExpandingMan:
This change is to make nvim-lsp use the package [LSPNeovim.jl](https://github.com/ExpandingMan/LSPNeovim.jl) by default.  This simplifies the interface and resolves some issues, see discussion [here](https://github.com/ExpandingMan/LSPNeovim.jl/issues/1).

From conversations with others from the Julia community involved with LanguageServer.jl (@kdheepak , @davidanthoff, @non-Jedi), it seems there is consensus we should do this.  If you are involved with LanguageServer.jl please thumbs up/down this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2274,14 +2274,10 @@ require'nvim_lsp'.jsonls.setup{}
 ## julials
 
 https://github.com/julia-vscode/julia-vscode
-`LanguageServer.jl` can be installed via `:LspInstall julials` or by yourself the `julia` and `Pkg`:
-```sh
-julia --project=/home/runner/.cache/nvim/nvim_lsp/julials -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
-```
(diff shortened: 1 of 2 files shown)


davidanthoff:
I can't opine on this, I'm not familiar enough with the details of how nvim is integrated with the Julia language server, sorry :)

justinmk:
- what is https://github.com/ExpandingMan/LSPNeovim.jl doing that can't be done here instead (in Lua)? Then users can avoid an extra step.
- we need to get out of the auto-install business here ( https://github.com/neovim/nvim-lspconfig/issues/334 ) so is there a different way users could install LSPNeovim.jl (a julia package manager or whatever), that can be mentioned in the docs here instead of `:LSPInstall` ?

mjlbach:
I've updated the Julia start commands and it seems to work well now, I would recommend creating an nvim-julia plugin if you'd like to add more advanced configuration or use wrappers around julials

PR #292, opened 2020-07-01
pyls_ms - Remove call to get Python interpreter and version.

kosayoda:
By default, the Microsoft Python Language Server already gets an existing Python interpreter and version from $PATH if the setting is an empty string or null.

[src/LanguageServer/Impl/Implementation/Server.cs](https://github.com/microsoft/python-language-server/blob/838ba78e00173d639bd90f54d8610ec16b4ba3a2/src/LanguageServer/Impl/Implementation/Server.cs#L143-L157)
```c#
var interpeterPath = initializationOptions?.interpreter.properties?.InterpreterPath;
Version version = null;


if (!string.IsNullOrWhiteSpace(interpeterPath)) {
	Version.TryParse(initializationOptions?.interpreter.properties?.Version, out version);
} else {
	var fs = _services.GetService<IFileSystem>();
	var exePath = PathUtils.LookPath(fs, "python");
	if (exePath != null && TryGetPythonVersion(exePath, out version)) {
		_log?.Log(TraceEventType.Information, Resources.UsingPythonFromPATH);
		interpeterPath = exePath;
	} else {
		interpeterPath = null;
	}
}

```
Thus, looking for `python` and launching the interpreter to get the version on our end is redundant and time-consuming.

### DIFF
--- a/lua/nvim_lsp/pyls_ms.lua
+++ b/lua/nvim_lsp/pyls_ms.lua
@@ -3,13 +3,6 @@ local util = require 'nvim_lsp/util'
 
 local name = "pyls_ms"
 
-local function get_python_version()
-  local f = io.popen("python --version 2>&1") -- runs command
-  local l = f:read("*a") -- read output of command
-  f:close()
-  return l:match("^Python%s*(...).*%s*$")
-end
-
 local function get_latest_pyls()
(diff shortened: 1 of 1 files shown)


tjdevries:
This makes sense to me, what do you think @h-michael

mjlbach:
I implemented this solely because the implementation in emacs lsp-mode passed the same initialization params, so if this has been thoroughly tested I don’t see why not to merge this.

PR #296, opened 2020-07-04
Allow to use filename patterns on root_pattern

adelarsq:
This pull request will allow to use filename patterns on `root_pattern` function. So will possible to use something like:

```lua
root_dir = util.root_pattern("*.sln");
```


adelarsq:
Waiting https://github.com/neovim/nvim-lsp/pull/291#issuecomment-663804618 to be merged, which have the same change.

adelarsq:
Thanks @justinmk 🥳

PR #301, opened 2020-07-11
Added optional extra parameter to on_new_config method for the root directory

svermeulen:
For some lsp configs, it's helpful to have access to the root directory that was chosen for the new config.  For example, it is common for python to use a `.venv` directory at the root of the project that includes the python interpreter that should be used for the project, which needs to be specified to the server at initialization time.

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -126,10 +126,10 @@ function configs.__newindex(t, config_name, config_def)
 
       add_callbacks(new_config)
       if config_def.on_new_config then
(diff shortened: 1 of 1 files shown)


justinmk:
this looks like a good change and maybe resolves https://github.com/neovim/nvim-lspconfig/issues/320 .

might be worth adding a test for this in https://github.com/neovim/nvim-lspconfig/blob/master/test/nvim_lsp_spec.lua

h-michael:
This was resolved with this https://github.com/neovim/nvim-lspconfig/pull/238.
Thank you for your PR.

PR #307, opened 2020-07-19
[RFC] don't start up LSP for 'nofile' buffers

beardedsakimonkey:
high level goal is to allow users some way to set filetype for a buffer in cases where they don't want to start up an LSP server, e.g. in a fuzzy finder preview

cc @tjdevries

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -171,6 +171,9 @@ function configs.__newindex(t, config_name, config_def)
     end)
 
     function manager.try_add()
+      if vim.bo.buftype == 'nofile' then
+        return
+      end
       local root_dir = get_root_dir(api.nvim_buf_get_name(0), api.nvim_get_current_buf())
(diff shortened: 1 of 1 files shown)


tjdevries:
(Sorry I didn't see this earlier)

What do you think about doing this on neovim core? Is there any reason to ever load up the LSP for `nofile` buffers?

We might want to make this never happen.

beardedsakimonkey:
I guess that seems like a resonable thing to do, but not sure how it could be implemented nicely.  It doesn't really make sense for the check to be in `vim.lsp.start_client()` since that function isn't meant to be associated with any buffer.  You could add the check in `vim.lsp.buf_attach_client()`, but that is called after `start_client()`, so you'd need to `stop_client()`.  That seems a lot less desirable than this solution, which avoids starting the client altogether.

tjdevries:
OK, I think the reasoning makes sense there. We can always upstream an additional thing if we want to.

@norcalli / @h-michael are either of you good with this?

PR #313, opened 2020-07-29
tsserver binary was moved to typescript package

juangabrielr4:


gbrlsnchs:
TSServer itself doesn't implement the Language Server protocol. `typescript-language-server` is a wrapper around it that implements such protocol. It will use a local `tsserver` installed in a project `node_modules` or fallback to a global one.

PR #314, opened 2020-07-30
[elixirls] change root dir for umbrella project

Stanislav-Lapata:
@mortepau 
What do you think about it?
This code fix "go to definition"
before
```
find . -name '*elixir_ls*'
./.elixir_ls
./apps/app_name1/.elixir_ls
./apps/app_name2/.elixir_ls
./apps/app_name3/.elixir_ls
./apps/app_name4/.elixir_ls
```
after
```
find . -name '*elixir_ls*'
./.elixir_ls
```

### DIFF
--- a/lua/nvim_lsp/elixirls.lua
+++ b/lua/nvim_lsp/elixirls.lua
@@ -73,7 +73,7 @@ configs[server_name] = {
     cmd = { cmd };
     filetypes = {"elixir", "eelixir"};
     root_dir = function(fname)
-        return util.root_pattern("mix.exs", ".git")(fname) or vim.loop.os_homedir()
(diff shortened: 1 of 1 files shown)


mortepau:
Hi @Stanislav-Lapata 
I'm sorry for the late reply, I haven't checked my notifications in forever...
I'm not really up to par on umbrella projects but is the mix.lock file exclusive to the top umbrella project folder, or can it appear in any of the children's applications. Another thing to which might prove troublesome is that the mix.lock file wouldn't be generated unless you actually include any dependencies in the project, or am I wrong here? If this is the case, I think it is safer to keep mix.exs, even though it adds multiple definitions.

mjlbach:
Closing because I think mix.exs is correct, and you can override root_dir for specific projects.

PR #317, opened 2020-08-04
Add installer for clangd

rickysaurav:
* Add support for installing clangd via github releases


justinmk:
Really appreciate your effort. However, this is a lot of code for installing something that really should be done by the OS package manager. `force_symlink()` in particular is a red flag.

In general we will likely remove the "installer" feature, it's just too much scope for us to deal with, and invites all of the challenges of a package manager.

PR #321, opened 2020-08-16
Autoinstall rust-analyzer

PatOConnor43:
:wave: Hi, I noticed that rust-analyzer didn't have an auto-installer so I took a shot at it. I included some half baked windows logic, but I don't have a machine to test it on :sweat_smile: . I'd appreciate any feedback to get that working.

I also included auto-updating of the server as new releases get published. The user will be prompted with `vim.fn.inputlist` to select if they want to update or not.

There are also a handful of errors that are thrown if you are calling `nvim_lsp.rust_analyzer.setup({})` without the binary actually being there. The prompt to update will still show though and you can download the binary from there.

I'd also _like_ to stop and restart the client if we install this way but I was having trouble getting set up as well.

### DIFF
--- a/lua/nvim_lsp/configs.lua
+++ b/lua/nvim_lsp/configs.lua
@@ -171,6 +171,9 @@ function configs.__newindex(t, config_name, config_def)
     end)
 
     function manager.try_add()
+      if vim.bo.buftype == 'nofile' then
+        return
+      end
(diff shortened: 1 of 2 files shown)


PatOConnor43:
Haha whoops, messed up a rebase.

PR #322, opened 2020-08-16
Add auto-install for rust-analyzer

PatOConnor43:
:wave: Hi, I noticed that rust-analyzer didn't have an auto-installer so I took a shot at it. I included some half baked windows logic, but I don't have a machine to test it on :sweat_smile: . I'd appreciate any feedback to get that working.

I also included auto-updating of the server as new releases get published. The user will be prompted with `vim.fn.inputlist` to select if they want to update or not.

There are also a handful of errors that are thrown if you are calling `nvim_lsp.rust_analyzer.setup({})` without the binary actually being there. The prompt to update will still show though and you can download the binary from there.

I'd also _like_ to stop and restart the client if we install this way but I was having trouble getting set up as well.

### DIFF
--- a/lua/nvim_lsp/rust_analyzer.lua
+++ b/lua/nvim_lsp/rust_analyzer.lua
@@ -1,11 +1,88 @@
 local configs = require 'nvim_lsp/configs'
 local util = require 'nvim_lsp/util'
+local server_name = 'rust_analyzer'
+local release_url = 'https://api.github.com/repos/rust-analyzer/rust-analyzer/releases/latest'
+local postfix = '-linux'
+local bin = 'rust-analyzer'
+
+if vim.fn.has('mac') == 1 then
+  postfix = '-mac'
+elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
+  postfix = '-windows.exe'
+end
+
+local function make_installer()
+  local I = {}
+  local P = util.path.join
+  local install_dir = P{util.base_install_dir, server_name}
+	local cmd_path = P{install_dir, bin}
+
+  local function getReleaseJson()
+    local json_string = vim.fn.system(string.format('curl -s "%s"', release_url), install_dir)
+    return vim.fn.json_decode(json_string)
+  end
+
+  function I.install()
+    local install_info = I.info()
+    vim.fn.mkdir(install_info.install_dir, "p")
+    local json = getReleaseJson()
+
+    local download_url = ''
+    local tag_name = json.tag_name
+    for _, value in pairs(json.assets) do
+      if (value ~= nil and string.match(value.browser_download_url, postfix..'$')) then
+        download_url = value.browser_download_url
+      end
+    end
+    if (download_url == '') then
+      print('Could not find download url. Aborting.')
+      return
+    end
+    util.sh(
+      string.format(
+        'echo "Starting download: %s" && ' ..
+        'curl -L "%s" > rust-analyzer && ' ..
+        'chmod 755 rust-analyzer && ' ..
+        'echo "%s" > .rust-analyzer-tag',
+      download_url, download_url, tag_name), install_info.install_dir)
+  end
+
+  function I.info()
+        return {
+            is_installed = util.path.exists(cmd_path);
+            install_dir = install_dir;
+            cmd = { cmd_path };
+        }
+  end
+
+  function I.configure(_)
+    local install_info = I.info()
+    local tag = vim.fn.system('cat .rust-analyzer-tag', install_info.install_dir)
+    local json = getReleaseJson()
+    if (json.tag_name == tag) then
+      return
+    end
+    local options = {"There is a new version of rust-analyzer. Would you like to update?", "1. Update", "2. Cancel"}
+    local choice = vim.fn.inputlist(options)
+    if (choice == 1) then
+      I.install()
+    else
+      print('Cancelling')
+    end
+  end
+  return I
+end
+
+local installer = make_installer()
 
 configs.rust_analyzer = {
   default_config = {
-    cmd = {"rust-analyzer"};
+    cmd = installer.info().cmd;
     filetypes = {"rust"};
     root_dir = util.root_pattern("Cargo.toml", "rust-project.json");
+    on_new_config = function(config)
+      installer.configure(config)
+    end;
   };
   docs = {
     package_json = "https://raw.githubusercontent.com/rust-analyzer/rust-analyzer/master/editors/code/package.json";
@@ -18,7 +95,12 @@ See [docs](https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user#
     ]];
     default_config = {
       root_dir = [[root_pattern("Cargo.toml", "rust-project.json")]];
+      on_new_config = function(config)
+        installer.configure(config)
+      end;
     };
   };
 };
+configs.rust_analyzer.install = installer.install
+configs.rust_analyzer.install_info = installer.info
 -- vim:et ts=2 sw=2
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+  postfix = '-windows.exe'
+end
+
+local function make_installer()
+  local I = {}
+  local P = util.path.join
+  local install_dir = P{util.base_install_dir, server_name}
+	local cmd_path = P{install_dir, bin}
+
+  local function getReleaseJson()
```
lithammer, line 20: Use snake_case. I'm not convinced it's a good name either since json is just an intermediate format, it doesn't return it. Maybe `get_release_info()`?

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+    local json_string = vim.fn.system(string.format('curl -s "%s"', release_url), install_dir)
+    return vim.fn.json_decode(json_string)
+  end
+
+  function I.install()
+    local install_info = I.info()
+    vim.fn.mkdir(install_info.install_dir, "p")
+    local json = getReleaseJson()
```
lithammer, line 28: Same here, it's not json. So maybe `release_info`?

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+  end
+
+  function I.configure(_)
+    local install_info = I.info()
+    local tag = vim.fn.system('cat .rust-analyzer-tag', install_info.install_dir)
+    local json = getReleaseJson()
+    if (json.tag_name == tag) then
+      return
+    end
+    local options = {"There is a new version of rust-analyzer. Would you like to update?", "1. Update", "2. Cancel"}
```
lithammer, line 65: Since I manage the installation of rust-analyzer myself, I assume I will get nagged about this every time I start Neovim? Obviously I could just change the config so it doesn't call `installer.info()`, but still... I'd rather not have to.

And what happens if I don't have internet access for some reason, will it barf errors?

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+      download_url, download_url, tag_name), install_info.install_dir)
+  end
+
+  function I.info()
+        return {
+            is_installed = util.path.exists(cmd_path);
+            install_dir = install_dir;
+            cmd = { cmd_path };
+        }
+  end
```
lithammer, line 56: Wrong indentation.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+    local tag_name = json.tag_name
+    for _, value in pairs(json.assets) do
+      if (value ~= nil and string.match(value.browser_download_url, postfix..'$')) then
+        download_url = value.browser_download_url
+      end
+    end
+    if (download_url == '') then
+      print('Could not find download url. Aborting.')
+      return
+    end
```
lithammer, line 40: Not sure why you're bothering with this since the download URL can be computed given the version tag and platform (i.e. mac, linux  etc):

```
https://github.com/rust-analyzer/rust-analyzer/releases/download/$VERSION/rust-analyzer-$PLATFORM
```

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+local server_name = 'rust_analyzer'
+local release_url = 'https://api.github.com/repos/rust-analyzer/rust-analyzer/releases/latest'
+local postfix = '-linux'
+local bin = 'rust-analyzer'
+
+if vim.fn.has('mac') == 1 then
+  postfix = '-mac'
+elseif vim.fn.has('win32') == 1  or vim.fn.has('win64') == 1 then
+  postfix = '-windows.exe'
+end
```
lithammer, line 12: You can omit the comparison with `1`, it's redundant in Lua.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,88 @@
...
+    local json = getReleaseJson()
+    if (json.tag_name == tag) then
+      return
+    end
+    local options = {"There is a new version of rust-analyzer. Would you like to update?", "1. Update", "2. Cancel"}
+    local choice = vim.fn.inputlist(options)
+    if (choice == 1) then
+      I.install()
+    else
+      print('Cancelling')
```
lithammer, line 70: I'd say drop this print.

PatOConnor43:
Thanks for all the feedback, this is the first lua code I've written. I'm getting this fixed up now.

lua/nvim_lsp/rust_analyzer.lua
justinmk: redundant how? Only `false` and `nil` are falsy in Lua. `0` is truthy, that's why we compare to `1` since `has()` returns 0 or 1.

lua/nvim_lsp/rust_analyzer.lua
PatOConnor43: This is a really good point. Do you have a recommended way to resolve this? If a user points to a non-default  rust-analyzer command we can disable the auto-updating? Maybe include a specific flag to disable auto-updating?

I did check without internet and that seemed fine.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,86 @@
...
+local bin = 'rust-analyzer'
+
+if vim.fn.has('mac') then
+  postfix = '-mac'
+elseif vim.fn.has('win32') or vim.fn.has('win64') then
+  postfix = '-windows.exe'
+end
+
+local function make_installer()
+  local I = {}
```
justinmk, line 15: why not just use a `do` block with local variables, instead of this OOP thing?

lua/nvim_lsp/rust_analyzer.lua
lithammer: You're absolutely right. My bad! Ignore this advice.

lua/nvim_lsp/rust_analyzer.lua
```diff
@@ -1,11 +1,86 @@
...
+local server_name = 'rust_analyzer'
+local release_url = 'https://api.github.com/repos/rust-analyzer/rust-analyzer/releases/latest'
+local postfix = '-linux'
+local bin = 'rust-analyzer'
+
+if vim.fn.has('mac') then
+  postfix = '-mac'
+elseif vim.fn.has('win32') or vim.fn.has('win64') then
+  postfix = '-windows.exe'
+end
```
lithammer, line 12: I was wrong about comparing to `1` being redundant, sorry...

PatOConnor43: No problem :+1:

lua/nvim_lsp/rust_analyzer.lua
PatOConnor43: Do you have an example of what you mean? I'm pretty new to lua? I was looking at the [elixirLS](https://github.com/neovim/nvim-lsp/blob/master/lua/nvim_lsp/elixirls.lua#L17) file for inspiration.

lua/nvim_lsp/rust_analyzer.lua
PatOConnor43: I changed this in d1bac250. Thanks!

lua/nvim_lsp/rust_analyzer.lua
lithammer: > Do you have a recommended way to resolve this?

I'm afraid not.

> If a user points to a non-default rust-analyzer command we can disable the auto-updating? Maybe include a specific flag to disable auto-updating?

I suppose either of those will work. But I think the latter will be less error prone since it doesn't have to figure oute what a "non-default" rust-analyzer command is. The question is if it should default to on or off?

> I did check without internet and that seemed fine.

👍

justinmk:
Thanks very much for your effort. I think we need to get out of the business of installing things. It's too much scope. 

in the configs where we have  `make_installer` , 70-80% of the code ends up dealing with install logic. That isn't really the point of this repo , it's a distraction.

PR #325, opened 2020-08-24
doc: document {cmd}, {on_new_config}

acalhoon:
Full disclosure -- I had never written any Lua before setting out to add this
functionality. I'll gladly make the changes more idiomatic if this patch is
deemed worthy enough to enter the main line.

I wanted this behavior to deal with multiple `compile_commands.json`
files produced for a multi-package CMake project I work on.

gen_cmdline_args is an optional function that can be provided in the setup table
for a given configuration. When non-nil the function takes the full path of the
active file and will return a table of strings which are appended to the cmd
table prior to launching the LSP server.

By dynamically generating command line arguments users can customize LSP server
configuration based on the project of interest. A more concrete example is that
it allows clangd users to specify the `--compile-commands-dir=` argument when
the `compile_commands.json` file is not directly in the build path for the
project (which is not unusual for CMake based projects) or when multiple
`compile_commands.json` files exist at the same level of a build tree, e.g.,
`~/megaproject/build/subproject1/compile_commands.json` and
`~/megaproject/build/subproject2/compile_commands.json`

A potential deficiency of this patch is that gen_cmdline_args is called even if no new
LSP server client would be kicked off as part of loading the particular file. It'd probably
be ideal to defer this until just before the start_client() call, but that seemed more
invasive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,11 @@ nvim_lsp.SERVER.setup{config}
   The `config` parameter has the same shape as that of
   |vim.lsp.start_client()|, with these additions and changes:
 
+  {cmd}
+    Required for all servers.
+    Table containing the server daemon executable name as the first element and
+    optional additional commandline arguments as subsequent elements.
+
   {root_dir}
     Required for some servers, optional for others.
     Function of the form `function(filename, bufnr)`.
@@ -194,6 +199,32 @@ nvim_lsp.SERVER.setup{config}
       - |nvim_lsp.util.find_git_root()|
       - |nvim_lsp.util.find_node_modules_root()|
       - |nvim_lsp.util.find_package_json_root()|
+
+  {gen_cmdline_args}
+    Optional for all servers.
+    Function of the form `function(filename, bufnr)`.
+    Called on new candidate buffers being attached-to.
+    Returns a table of strings that augment the commandline arguments in {cmd}.
+
+    If a non-empty table is returned, the element values are appended, in order,
+    to the {cmd} parameter before launching the server daemon.
+
+    If the parameter is set to nil, or an empty table is returned,
+    the {cmd} parameter is left unmodified.
+
+    Example: When running a [clangd](#clangd) language server, this function can
+    be useful to provide the location of the `compile_commands.json` using the
+    `--compile-commands-dir` argument, especially when the file is located in a
+    shared build location.
+
+    gen_cmdline_args = function(fname)
+        local root_path = util.root_pattern("CMakeLists.txt")(fname)
+        local build_path = string.gsub(root_path, "src", "build")
+
+        local args = {}
+        table.insert(args, ('--compile-commands-dir=%s'):format(build_path)
+        return args
+    end;
 
   {name}
     Defaults to the server's name.
--- a/README.md
+++ b/README.md
@@ -172,11 +172,6 @@ nvim_lsp.SERVER.setup{config}
   The `config` parameter has the same shape as that of
   |vim.lsp.start_client()|, with these additions and changes:
 
-  {cmd}
-    Required for all servers.
-    Table containing the server daemon executable name as the first element and
-    optional additional commandline arguments as subsequent elements.
-
   {root_dir}
     Required for some servers, optional for others.
     Function of the form `function(filename, bufnr)`.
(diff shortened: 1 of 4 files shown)


tjdevries:
Can you add some docs for this? I think it's a good idea. Perhaps adding in an example here (of the code you would use) to utilize this  feature would be helpful as well to chat about it.

acalhoon:
I'm happy to.

acalhoon:
I just added some documentation and a crude example.

Also added a tiny bit of unrelated documentation WRT the {cmd} parameter as I felt that added more context.

README.md
```diff
@@ -172,6 +172,11 @@ nvim_lsp.SERVER.setup{config}
   The `config` parameter has the same shape as that of
   |vim.lsp.start_client()|, with these additions and changes:
 
+  {cmd}
+    Required for all servers.
+    Table containing the server daemon executable name as the first element and
+    optional additional commandline arguments as subsequent elements.
```
justinmk, line 178: thanks for documenting this. i think the readme.md changes need to go in https://github.com/neovim/nvim-lspconfig/blob/4e73c29aaf03451d5efcdf6c4747bac7c0dc1b00/scripts/README_template.md .

readme.md is auto-generated from it.

(future: we could avoid this confusion perhaps by generating to a separate file and linking to it from readme.md ...)

lua/nvim_lsp/configs.lua
```diff
@@ -124,6 +126,14 @@ function configs.__newindex(t, config_name, config_def)
         }
       })
 
+      -- Dynamically generate command-line arguments based on the active buffer path
+      if gen_cmdline_args ~= nil then
+        local args = gen_cmdline_args(api.nvim_buf_get_name(0), api.nvim_get_current_buf())
+        for i = 1, #args do
+          table.insert(new_config.cmd, args[i])
```
justinmk, line 133: could use `vim.list_extend(new_config.cmd, args)` to avoid the for-loop here

justinmk:
is the main purpose here to update `cmd` per-buffer? then i'm not sure we need this new `gen_cmdline_args` thing. how hard is it for users to instead add a buffer event handler that sets `cmd` ?

acalhoon:
The idea is something like that, yes. The fully qualified path name for the file being loaded into the buffer determines which command line arguments are used when launching the language server.

You'll have to forgive me, but my experience with vimscript is pretty superficial. Is what you're suggesting possible? Probably. What I don't have a sense for is what that looks like in practice. If it means multiple setup/configs need to be maintained in separate places, then that seems less optimal.

If the buffer event approach proves to be a better approach, then I suggest an edit to the readmes (general or specific to clangd) that explains how one would setup these events to modify the command arguments per buffer.

justinmk:
oh my question wasn't rhetorical, I actually don't know :) 

You've added a new hook here, but there is an existing hook `on_new_config` , does it not working for this purpose? See for example https://github.com/neovim/nvim-lspconfig/pull/301 which adds a parameter to `on_new_config` so that handlers can get the resolved root.

So I wonder if users can specify a `on_new_config` and modify the `new_config.cmd` (i.e. what you've done here with `table.insert(new_config.cmd, args[i])`.

Alternatively the plugin could check if `cmd` is a function, and if it is, then call it. That allows the user to do something like

```lua
cmd = function()
  return {'foo', '-x', '-y'}
end

acalhoon:
I think you may be right about on_new_config solving this problem. I over looked that option. Let me have a go at solving my issue with that approach.

acalhoon:
Also, a function based command wouldn't easily solve my problem because I would need the root directory to make the modification to the arguments

acalhoon:
Confirmed. With the addition of #301, which is a more general solution to the problem, gen_cmdline_args becomes unnecessary. I think some examples in the documentation would be very useful.

justinmk:
so let's remove the gen_cmdline_args from this PR but keep the doc you added for `{cmd}`, and also stick in an example . Just paste whatever you did in your personal config.

acalhoon:
I believe I have the documentation updated in the correct place and I have removed any dependence on #301 in my example. That said, if that PR ends up merged, then the documentation should follow suit by using the root_dir passed as a parameter rather than recalculating it.

acalhoon:
Not rushing anyone -- just making sure I'm not leaving anything hanging here.

mjlbach:
Can you rebase on master? @justinmk @tjdevries Do you both agree this is a good addition to the documentation?

mjlbach:
This is out of date now, I believe the info is covered in the advanced_readme

PR #330, opened 2020-08-27
Run lint job as sudo to make CI pass again

theHamsta:
At least 2 of 3

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,4 +18,4 @@ jobs:
         sudo luarocks install luacheck
     - name: Run luacheck
       run: |
-        make lint
+        sudo make lint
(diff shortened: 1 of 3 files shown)


justinmk:
Thanks. Looks like CI is ok though now?

justinmk:
weird, seems like CI is having a problem again

theHamsta:
The problem on macos seems to be that Cmake gets installed twice. Maybe just remove int from the brew install command?

theHamsta:
Strange: even though all the checks passed. They get only executed before the docgen commit. Maybe an additional trigger for the workflows is necessary now?

justinmk:
Thanks, this is helpful!

Side note: do we really actually need a macos build for this repo? Seems overwrought.

PR #332, opened 2020-08-30
Check flutter path for dart sdk before checking dart path

himansingh241:
Since most of the time user just install the flutter sdk and the dart path is present in `~/flutter/bin/dart` it would be better to check first for the flutter path since even if a user has flutter path set the elseif part will be ignored every time.

### DIFF
--- a/lua/nvim_lsp/dartls.lua
+++ b/lua/nvim_lsp/dartls.lua
@@ -5,12 +5,12 @@ local server_name = "dartls"
 local bin_name = "dart"
 
 local find_dart_sdk_root_path = function()
-  if vim.fn["executable"]("dart") == 1 then
-    return vim.fn["resolve"](vim.fn["exepath"]("dart"))
-  elseif vim.fn["executable"]("flutter") == 1 then
+  if vim.fn["executable"]("flutter") == 1 then
     local flutter_path = vim.fn["resolve"](vim.fn["exepath"]("flutter"))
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/dartls.lua
```diff
@@ -5,12 +5,12 @@ local server_name = "dartls"
...
 local find_dart_sdk_root_path = function()
-  if vim.fn["executable"]("dart") == 1 then
-    return vim.fn["resolve"](vim.fn["exepath"]("dart"))
-  elseif vim.fn["executable"]("flutter") == 1 then
+  if vim.fn["executable"]("flutter") == 1 then
     local flutter_path = vim.fn["resolve"](vim.fn["exepath"]("flutter"))
     local flutter_bin = vim.fn["fnamemodify"](flutter_path, ":h")
     return flutter_bin.."/cache/dart-sdk/bin/dart"
+  elseif vim.fn["executable"]("dart") == 1 then
+    return vim.fn["resolve"](vim.fn["exepath"]("dart"))
```
justinmk, line 13: for all of these cases, is there a reason we aren't using `vim.fn.foo()` instead of `vim.fn["foo"]()` ?

himansingh241: I just changed the order it was like this previously. Don't know if there is any reason for that.

PR #335, opened 2020-08-31
add host pid dependecy to omnisharp command

klepz:


lua/nvim_lsp/omnisharp.lua
```diff
@@ -5,7 +5,7 @@ local bin_name = 'run'
 
 local function make_installer()
   local install_dir = util.path.join{util.base_install_dir, server_name}
-
+  local pid = vim.api.nvim_call_function('getpid', {})
```
justinmk, line 8: can use `vim.fn.getpid()` (see `:help vim.fn`).

PR #336, opened 2020-08-31
Add a default config for the haskell language server.

SaxJ:
Hi folks! First PR here.
Added a simple config for the Haskell Language Server (https://github.com/haskell/haskell-language-server).
Avoided installation as it gets pretty messy.

### DIFF
--- a/lua/nvim_lsp/hls.lua
+++ b/lua/nvim_lsp/hls.lua
@@ -1,7 +1,6 @@
 local configs = require 'nvim_lsp/configs'
 local util = require 'nvim_lsp/util'
 
-
 configs.hls = {
   default_config = {
     cmd = {"haskell-language-server-wrapper", "--lsp"};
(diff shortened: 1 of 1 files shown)


justinmk:
do we need https://github.com/haskell/haskell-ide-engine then? i.e. can we remove this : https://github.com/neovim/nvim-lspconfig#hie

joaopizani:
AFAIK `haskell-language-server` (`AKA `hls`) is supossed to be the integration point of `haskell-ide-engine` (AKA `hie`) with other language servers, and it's the project around which the Haskell community is unifying.  However, `hls` and `hie` will continue to co-exist side-by-side, with `hie` getting mostly bug fixes and new features being added to `hls`.

So I think it makes sense to keep both configurations here, let people choose what they want, and only remove `hie` later when it becomes clearly outdated.

SaxJ:
I agree with @joaopizani.
`hls` does not yet completely replace `hie`, so I think it makes sense to keep it for now.

PR #345, opened 2020-09-03
Switch to Hashicorp's official Terraform language server

lithammer:
I guess this is potentially controversial and a new config could be created instead (this can be closed in that case). The previous `juliosueiras/terraform-lsp` project has this to say in its readme:

> The aim to have a unified lsp for terraform in the future, but for now there is two concurrent development with collabration to each other, this repo is aim for more experimental features, and the terraform-ls is aim for stableness

Hashicorp's seems to be more actively developed nowadays as well.

https://github.com/hashicorp/terraform-ls

### DIFF
--- a/README.md
+++ b/README.md
@@ -992,6 +992,14 @@ This server accepts configuration via the `settings` key.
 
   Whether to consider files ending \'\_test\.dart\' that are outside of the test folder as tests\. This should be enabled if you put tests inside the \'lib\' folder of your Flutter application so they will be run with \'flutter test\' and not \'flutter run\'\.
 
+- **`dart.analysisExcludedFolders`**: `array`
+
+  Default: `{}`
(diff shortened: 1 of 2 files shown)


justinmk:
Seems reasonable. We should not have a giant unmanicured list of random things in this repo, main purpose is to serve as starting examples and where possible point users to the current best practice/choice.

NicholasAsimov:
@lithammer I think the command should be `cmd = {"terraform-ls", "serve"};` for the Hashicorp server.

lithammer:
> @lithammer I think the command should be `cmd = {"terraform-ls", "serve"};` for the Hashicorp server.

Ah yes, you're correct. See https://github.com/neovim/nvim-lspconfig/pull/351 for a fix.

Issue #334, opened 2020-08-30
remove make_installer and all auto-install logic


justinmk:
We need to get out of the business of auto-installing things. It's too much scope. 

in the configs where we have  `make_installer` , 70-80% of the code ends up dealing with install logic. That isn't really the point of this repo , it's a distraction.

Note:

- **Documention** of common install steps **is valuable and appropriate for this repo**. E.g.:
  -  URL to download the server installer
  -  URL to upstream documentation explaining how to install the server 
  - brief instructions that user can copy/paste into their shell to install via common package managers such as apt-get/homebrew
- It could be valuable for `:checkhealth` to be the entrypoint that users can reach for, to detect common problems with LSP server installation. For that purpose some of the existing installer logic might be useful.

tjdevries:
I've discussed this a few times in gitter, but I think a better strategy may be to make an org "neovim-lsp" or "neovim-lsp-configurations" and split the different servers into different repos. We can give people access to be maintainers of those repos. Each repo can decide if auto install is possible or not for that server.

I do think the ability to auto install is very helpful, particularly to new comers. So it would be a shame to lose it completely or have people make many many worse versions rather than coming together to make one good one.

justinmk:
> split the different servers into different repos

Many repos adds a lot cognitive and mechanical overhead. The entire purpose of this repo was to give users a convenient way to install "all the configs" (which in terms of data is small). 

> We can give people access to be maintainers of those repos

What's wrong with PRs to one repo? Many repos means we need at least one maintainer for each repo, and if they go missing then Neovim org has to manage all of those repos...

> Each repo can decide if auto install is possible or not for that server.

Contributors can create separate repos already. Putting them under the Neovim org implies some level of commitment, so it doesn't avoid the motivation for removing the install logic in this project.

The "microservice" approach is a way of side-stepping responsibility which is useful for giant organizations, but in OSS that approach is implicit and ever-present. 

> I do think the ability to auto install is very helpful, particularly to new comers. So it would be a shame to lose it completely or have people make many many worse versions rather than coming together to make one good one.

The `doc` element can/should specify useful install steps (links to upstream docs, basic CLI instructions), and solves 95% of the problem without involving the very large scope of trying to automate the steps for the myriad environments users operate in (aka "ad-hoc package manager").

colinxs:
[...]

PR #348, opened 2020-09-06
Rename nvim_lsp to lspconfig

### LINKED ISSUE #97: rename to nvim-lspconfig
Probably should have anticipated this, but people are confused about nvim-lsp (this repo, not Nvim-builtin) vs vim.lsp (Nvim builtin). Although the nvim-lsp README.md makes it clear, it still confuses people in discussions ([example](https://www.reddit.com/r/neovim/comments/eq48cy/anyone_using_nvimlsp/)), and people who are just getting started.

So we should

- rename this repo to ~~nvim-lspconf~~ `nvim-lsp.configs`
- it also makes sense to rename the Lua module: it is confusing that `require'nvim_lsp'` is distinct from `vim.lsp`, since their names are semantically equivalent
    - see https://github.com/neovim/nvim-lsp/issues/97#issuecomment-583717327

I will do so in the next ~week if there are no objections.

cc @norcalli

h-michael:
resolve #97
related #334

This PR includes breaking changes as it involves changing the name of the module.
There is also a discussion in #334 about whether to remove the installer or stop using monorepo.
I think it's better to make these changes at the same time because multiple breaking changes in a short period of time can be confusing to users.


justinmk:
LGTM

> This PR includes breaking changes as it involves changing the name of the module.

Let's define a `nvim_lsp` module that returns an error and/or `:echo` a message like `nvim_lsp` module was renamed to `lspconfig`.

tjdevries:
I don't remember if we've talked about this before, but what about vim.lsp.config?

justinmk:
> I don't remember if we've talked about this before, but what about vim.lsp.config?

But it isn't part of the `lsp` module. It's a separate extension.

Lots of variants were suggested in https://github.com/neovim/nvim-lspconfig/issues/97

teto:
if we target a christmas release, may be good to proceed as early as possible, even if just for the rename (nvim_lsp to lspconfig)

h-michael:
@tjdevries @teto 
Sorry for toooooo late replying. I've added a deprecated message.

teto (approved):

PR #360, opened 2020-09-21
Disable rust-analyzer checkOnSave by default

unrealhoang:
As recently, rust-analyzer has just make checkOnSave as default behavior https://github.com/rust-analyzer/rust-analyzer/commit/24805d1d80af251a124f7e1085bd59dfc93b764f on server side, this behavior is not good for nvim-lsp as it will block save. I think the better default would be off for nvim.


h-michael:
@unrealhoang 
It seems that checking on saving is running on the background thread.
Does it still block?

https://github.com/rust-analyzer/rust-analyzer/blob/4f386afb16297f588484ff24e2b4693218893a94/crates/flycheck/src/lib.rs#L1-L3

mjlbach:
This doesn't seem to be an issue anymore (for me), feel free to reopen if this is not the case.

PR #365, opened 2020-09-23
Update README.md

craigmac:
Mention that vimls provides an install() fn as well for the :LspInstall command


mjlbach:
This should be updated in the vimls lua file, not in README which is autogenerated by docgen

PR #370, opened 2020-09-30
JDT LS: general improvements

HiPhish:
This patch makes the root pattern detection more general and adds a callback for status updates from the server.

### DIFF
--- a/lua/lspconfig/jdtls.lua
+++ b/lua/lspconfig/jdtls.lua
@@ -32,13 +32,30 @@ local cmd = {
 -- could add more build systems, such as Make, but let's keep things simple for
 -- now.
 local root_files = {
-  'build.xml',             -- Ant
-  'pom.xml',               -- Maven
-  'build.gradle',          -- Gradle
-  'build.gradle.kts',      -- Gradle
-  'settings.gradle',       -- Gradle
-  'settings.gradle.kts',   -- Gradle
+  -- Single-module projects
+  {
+    'build.xml',             -- Ant
+    'pom.xml',               -- Maven
+    'settings.gradle',       -- Gradle
+    'settings.gradle.kts',   -- Gradle
+  },
+  -- Multi-module projects
+  {'build.gradle', 'build.gradle.kts'},
 }
+
+
+--- Callback function for the `language/status` notification.
+--
+-- The server sends a non-standard notification when the status of the language
+-- server changes. This can be used to display progress as the server is
+-- starting up.
+local function on_language_status(_, _, result)
+	local command = vim.api.nvim_command
+    command('echohl ModeMsg')
+    command(string.format('echo "%s"', result.message))
+    command('echohl None')
+end
+
 
 configs[server_name] = {
   default_config = {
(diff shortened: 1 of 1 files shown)


lua/nvim_lsp/jdtls.lua
```diff
@@ -92,10 +92,43 @@ end
...
   default_config = {
     filetypes = { "java" };
-    root_dir = util.root_pattern('.git');
+    root_dir = function(fname)
+	return util.find_git_ancestor(fname)
+		or util.root_pattern(unpack(root_files))(fname)
+		or vim.call('getcwd')
+    end;
     init_options = {
       workspace = path.join { vim.loop.os_homedir(), "workspace" };
```
yashLadha, line 133: By default there is no cmd being executed. The cmd is attached after the `on_new_config` method. On the first run it won't be present so probably we can also fix this stuff in the PR as well.

HiPhish: We want to get rid of the installer anyway, so I factor the command out. The issue however is that I need the path to where the server is installed, and there is no standard directory for that. If someone can tell me how I can let the user specify a `server_location` setting and generate the command on the fly it would easy.

Also, what does `on_new_config` do? I see that it is a callback, but for what?

yashLadha: Agree the installer doesn't install the correct jar for java 8 and probably works best for java 11 or greater. So it is best to remove the wrong code and create a correct fix. As the config is broken in it's own.

HiPhish: > Agree the installer doesn't install the correct jar for java 8 and probably works best for java 11 or greater.

I don't know if the current release even works with Java 8 anymore. They used to have instructions for Java 8 and Java 9+ in the README, but now there is only one set of instructions.

> So it is best to remove the wrong code and create a correct fix.

Again, how can I let the user specify the location of the server and generate the command on demand? Something like this:

```lua
default_config = {
	server_path = '/path/to/server',
	cmd = 'java ' .. server_path .. ' -jar server'
}
```

Obviously this won't work, but the idea is that the user should be able to call

```lua
lsp.jdts.setup {
	server_path = '/my/own/path'
}
```

and things would just work.

lua/nvim_lsp/jdtls.lua
```diff
@@ -92,10 +92,43 @@ end
...
+
+--- The presence of one of these files indicates a root directory.
+--
+-- We search for configuration files of the most common Java build systems. We
+-- could add more build systems, such as Make, but let's keep things simple for
+-- now.
+local root_files = {
+  'build.xml',             -- Ant
+  'pom.xml',               -- Maven
+  'build.gradle',          -- Gradle
```
mfussenegger, line 104: Adding `build.gradle` here as well could be problematic for multi-modules setups where you could have a structure like this:

```
/build.gradle
/module1/build.gradle
/module2/build.gradle
```

HiPhish: I have been using those settings for months at work on a multi-module project. But maybe I was just lucky that none of the symbols were in any of the other modules. Also, wouldn't the other files have this same issue?

I think there needs to be a slightly more complicated version of `root_pattern` which looks for one of the target files in the current directory, then peeks up one directory higher, and only if it finds nothing above does it return the current directory. Opinions?

HiPhish:
*bump* I changed the order of root pattern files. This sould take care of Gradle project with sub-projects.

mjlbach:
@mfussenegger Has the author resolved your concerns, and is it ready for merge?

HiPhish:
@mjlbach No, changing the order around does not solve the issue. I have opened #455 for discussion.

mjlbach:
My bad, misread your last comment. Thanks for following up on this!

mjlbach:
@mfussenegger I think this new behavior is good (as of latest update)? It matches what I did with kotlin.

@yashLadha Do you have any further comments?

mfussenegger:
@mjlbach yes looks good to me

PR #377, opened 2020-10-08
Fix root_pattern for wildcards and file in wildignore

tkonolige:
This PR fixes `root_pattern` to find directories that are included in wildignore. I think it is pretty common that people have `.git` in their wildignore, but would like to use it as a root directory marker.

I've also fixed `root_pattern` to work correctly with globs so things like `build_*` will work.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -535,6 +535,12 @@ This server accepts configuration via the `settings` key.
   
   Whether to include symbols that have not been imported in the code completion list and automatically insert the required import when selecting them \(requires restart\)\.
 
+- **`dart.automaticCommentSlashes`**: `enum { "none", "tripleSlash", "all" }`
+
+  Default: `"tripleSlash"`
+  
+  null
+
 - **`dart.buildRunnerAdditionalArgs`**: `array`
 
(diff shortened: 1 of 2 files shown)


stellarhoof:
It took me a couple of hours of debugging to notice this. Nice work!

lithammer:
Looks reasonable to me. @tkonolige can you rebase and fix the merge conflict?

mjlbach:
I don't really have/understand this issue so if you're comfortable merging this @lithammer it's fine to merge

lithammer:
> I don't really have/understand this issue so if you're comfortable merging this @lithammer it's fine to merge

The problem (as I understand it) is this:

```
:echo glob('.git')
.git
:set wildignore+=.git
:echo glob('.git')

:echo empty(glob('.git'))
1
:echo glob('.git', v:true, v:true)
['.git']
```

Basically, if you happened to have something your `wildignore` that matches something that our `root_pattern` would also match. It's ignored. So something like this would never find a `root_dir` given `wildignore=.git`:

```lua
root_dir = util.root_pattern('.git'),
```

PR #386, opened 2020-10-25
Add racket-langserver support

Lazerbeak12345:


Lazerbeak12345:
I've no idea why the mac test failed, but it looks like something with brew... Not something I would know what to do about.

Lazerbeak12345:
Also, is calling it "racket" the best idea? I can change it to `racketls`, `racket-language-server` or something else like that, as this does use [racket-language-server](https://github.com/theia-ide/racket-language-server)

lua/nvim_lsp/racket.lua
```diff
@@ -0,0 +1,23 @@
+local configs = require 'nvim_lsp/configs'
+local util = require 'nvim_lsp/util'
+
+configs.racket = {
+  default_config = {
+    cmd = {"racket", "-l", "racket-langserver"};
+    filetypes = {"racket"};
```
ckipp01, line 7: Btw, this should probably be `scheme`. Be default `.rkt` files will register as that.

mjlbach: You can have both

mjlbach:
@Lazerbeak12345 Please see my notes on https://github.com/neovim/nvim-lspconfig/pull/679 which is a duplicate

mjlbach:
Also, as amusing as they are, please squash your commits :P

mjlbach:
I do also prefer racket_langserver

Lazerbeak12345:
Thanks for the feedback! I’ll look into those things when I have time.

Lazerbeak12345:
I took a look, and I do think @ckipp01 's pr is better.

PR #399, opened 2020-11-13
Use handlers instead of callbacks

Iron-E:
The recent changes in Neovim head cause loading this file to create an error. These changes have resolved that on my machine.

Note: https://github.com/neovim/neovim/pull/12655

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -34,7 +34,7 @@ function configs.__newindex(t, config_name, config_def)
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
   local function add_handlers(config)
-    assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
+    assert(not config.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
     config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -37,8 +37,8 @@ function configs.__newindex(t, config_name, config_def)
     config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
-        assert(lsp.handlers, "Update to Nvim HEAD. This is an incompatible interface.")
-        assert(lsp.handlers["window/logMessage"], "Callback for window/logMessage notification is not defined")
+        assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/nvim-lua/diagnostic-nvim/issues/73")
+        assert(lsp.handlers["window/logMessage"], "Handler for window/logMessage notification is not defined")
         lsp.handlers["window/logMessage"](err, method, params, client_id)
       end
     end
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -34,7 +34,7 @@ function configs.__newindex(t, config_name, config_def)
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
   local function add_handlers(config)
-    assert(config.callbacks, "lsp.callbacks has been obsoleted. See here for more: https://github.com/neovim/neovim/pull/12655")
+    assert(not config.callbacks, "lsp.callbacks has been obsoleted. See here for more: https://github.com/neovim/neovim/pull/12655")
     config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -30,6 +30,9 @@ function configs.__newindex(t, config_name, config_def)
   -- Force this part.
   default_config.name = config_name
 
+  -- Ensure that there are no callbacks registered.
+  assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
+
   -- The config here is the one which will be instantiated for the new server,
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
@@ -37,7 +40,6 @@ function configs.__newindex(t, config_name, config_def)
     config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
-        assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/nvim-lua/diagnostic-nvim/issues/73")
         assert(lsp.handlers["window/logMessage"], "Handler for window/logMessage notification is not defined")
         lsp.handlers["window/logMessage"](err, method, params, client_id)
       end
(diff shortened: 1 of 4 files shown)


lithammer:
I suspect you have to fix the obsolete `callbacks` references in `lua/nvim_lsp/util.lua` and `lua/nvim_lsp/configs.lua` as well...? It might work without obvious errors, but I have a feeling some functionality is just going into a void since they're never called anymore.

Iron-E:
You're right! I missed those. Gonna change them and commit.

Iron-E:
Changed the rest of them.

teto:
@tjdevries

lua/nvim_lsp/configs.lua
```diff
@@ -33,26 +33,26 @@ function configs.__newindex(t, config_name, config_def)
...
-  local function add_callbacks(config)
-    config.callbacks["window/logMessage"] = function(err, method, params, client_id)
+  local function add_handlers(config)
+    config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
-        assert(lsp.callbacks, "Update to Nvim HEAD. This is an incompatible interface.")
-        assert(lsp.callbacks["window/logMessage"], "Callback for window/logMessage notification is not defined")
-        lsp.callbacks["window/logMessage"](err, method, params, client_id)
+        assert(lsp.handlers, "Update to Nvim HEAD. This is an incompatible interface.")
```
tjdevries, line 40: I think we should `assert(not lsp.callbacks, ...` with an informative message and a help section to look at .

Once we do that, we can merge. ("Update to Nvim HEAD" doesn't tell people what / how they need to fix anything)

tjdevries (changes requested):

lua/nvim_lsp/configs.lua
Iron-E: Did the most recent commit address this well?

lua/nvim_lsp/configs.lua
```diff
@@ -33,26 +33,26 @@ function configs.__newindex(t, config_name, config_def)
...
-  local function add_callbacks(config)
-    config.callbacks["window/logMessage"] = function(err, method, params, client_id)
+  local function add_handlers(config)
+    config.handlers["window/logMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.log_level then
         -- TODO(ashkan) remove this after things have settled.
-        assert(lsp.callbacks, "Update to Nvim HEAD. This is an incompatible interface.")
-        assert(lsp.callbacks["window/logMessage"], "Callback for window/logMessage notification is not defined")
-        lsp.callbacks["window/logMessage"](err, method, params, client_id)
+        assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/nvim-lua/diagnostic-nvim/issues/73")
```
tjdevries, line 40: I think this should be moved out of the inner function. This should be on line 36 I think (where we add the handlers).

We don't need to check this every time.

tjdevries: Also, it should reference this PR: https://github.com/neovim/neovim/pull/12655 Which is the one with the most information and most context.

Iron-E: I moved the line out, although now every time I initiate an LSP it causes an error. The repo doesn't have any more mentions of callbacks, so is this working right?

lua/nvim_lsp/configs.lua
```diff
@@ -33,26 +33,26 @@ function configs.__newindex(t, config_name, config_def)
   -- The config here is the one which will be instantiated for the new server,
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
-  local function add_callbacks(config)
-    config.callbacks["window/logMessage"] = function(err, method, params, client_id)
+  local function add_handlers(config)
+    assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
```
tjdevries, line 37: Oh, we should also chdck that config.callbacks is nil.

Sorry to be so pedantic about this.

lua/nvim_lsp/configs.lua
Iron-E: Maybe what we should do is check this each time a new config is registered (to make sure no deprecated settings are being added).

I'll look into that more tonight, since it looks like `callbacks` is getting populated still somehow.

lua/nvim_lsp/configs.lua
Iron-E: It's no problem! We want to make sure it's right before it goes someplace people use it everyday.

I'll add it in a little bit and push, just leaving this here to say I've seen it.

tjdevries: Cool, ping me on Monday if I haven't looked again. Idk what my schedule is tomorrow.

lua/nvim_lsp/configs.lua
```diff
@@ -34,7 +34,7 @@ function configs.__newindex(t, config_name, config_def)
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
   local function add_handlers(config)
-    assert(not lsp.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
+    assert(not config.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
```
Iron-E, line 37: I tested JDT LS using the provided config on master and the proposed config in this pull. Using this `assert` it fails on `master` and works for `patch-1` :relieved: 

As for `lsp.callbacks`, I tried the following:

```sh
nvim -u NORC -c 'lua assert(not vim.lsp.callbacks, "Failed!")'
```

and it failed. So I replaced the `lsp` reference with `config`.

tjdevries: ah, yeah, that's what I mean to say. I wanted you to assert about the config (which is the thing the user can control).

lua/nvim_lsp/configs.lua
```diff
@@ -33,26 +33,26 @@ function configs.__newindex(t, config_name, config_def)
...
     end
 
-    config.callbacks["window/showMessage"] = function(err, method, params, client_id)
+    config.handlers["window/showMessage"] = function(err, method, params, client_id)
       if params and params.type <= config.message_level then
         -- TODO(ashkan) remove this after things have settled.
-        assert(lsp.callbacks and lsp.callbacks[method], "Update to Nvim HEAD. This is an incompatible interface.")
-        assert(lsp.callbacks["window/showMessage"], "Callback for window/showMessage notification is not defined")
-        lsp.callbacks["window/showMessage"](err, method, params, client_id)
+        assert(lsp.handlers and lsp.handlers[method], "Update to Nvim HEAD. This is an incompatible interface.")
```
tjdevries, line 49: I don't think is needed anymore.

Iron-E: You're right. If we're aiming to assert on addition, we don't need to ship back a function which makes assertions about that again.

h-michael:
@Iron-E Sorry I've merged this PR. Could you rebase?
https://github.com/neovim/nvim-lspconfig/pull/348

Iron-E:
No problem. Good time to get that one through

lua/lspconfig/configs.lua
```diff
@@ -33,26 +33,25 @@ function configs.__newindex(t, config_name, config_def)
   -- The config here is the one which will be instantiated for the new server,
   -- which is why this is a function, so that it can refer to the settings
   -- object on the server.
-  local function add_callbacks(config)
-    config.callbacks["window/logMessage"] = function(err, method, params, client_id)
+  local function add_handlers(config)
+    assert(not config.callbacks, "lsp.callbacks has been deprecated. See here for more: https://github.com/neovim/neovim/pull/12655")
```
h-michael, line 37:
```suggestion
    assert(config.callbacks, "lsp.callbacks has been obsoleted. See here for more: https://github.com/neovim/neovim/pull/12655")
```

In some programming languages, "deprecated" may still be used, but it can mean that it will be removed in the future.
I think "absolute" is clearer than "deprecated".

h-michael (approved):
@tjdevries  Is there anything else?

tjdevries:
LGTM I think

h-michael:
@Iron-E Thank you for your work :)

Iron-E:
Happy we could do it!

PR #404, opened 2020-11-16
Move pyright.lua to correct location under lspconfig

mjlbach:
Pyright.lua was not moved to the correct subfolder after the refactoring, this was my mistake.


mjlbach:
@h-michael The recently merged PR had the pyright.lua file under the wrong directory, this fixes it. My bad for not catching the change earlier.

h-michael (approved):

PR #405, opened 2020-11-16
Specify unzip dependency

### LINKED ISSUE #389: LspInstall pyls_ms fails to report missing unzip command
- What language server(If the problem is related to a specific language server):
    - pyls_ms
- Operating system/version:
    Ubuntu 20.04

On a system without the `unzip` command installed, run `:LspInstall pyls_ms` from nvim.

### Actual behaviour

The LspInstall command fails silently, without output. Inspecting the nvim cache directory reveals that a `pyls.nupkg` file has been placed, but no further processing occurred.

### Expected behaviour

The process should have given a clear error that the `unzip` command is not available. From examining the source, it appears this is already done for a few other requirements, but unzip has been skipped.

mjlbach:
Closes #389

I know we're not updating installers anymore, but this closes an open issue


dutchgecko:
This fixes #389 for me.

h-michael (approved):

PR #408, opened 2020-11-16
Added command to organise imports in Pyright

saadparwaiz1:
I thought it might be a bit easier for users if the organise import command was implemented in the config.


mjlbach:
I saw this PR, and while I think it is useful advice I think that ultimately the configurations should be fairly "lean." Ultimately, code_action support should improve in core and eventually would subsume per server specific functions.

saadparwaiz1:
makes sense! closed the pull request!

PR #429, opened 2020-12-01
bugfix for code action not working sometimes

jerrywang1981:
tested on my local.

### DIFF
--- a/lua/lspconfig/jdtls.lua
+++ b/lua/lspconfig/jdtls.lua
@@ -105,13 +105,21 @@ configs[server_name] = {
       -- Due to an invalid protocol implementation in the jdtls we have to
       -- conform these to be spec compliant.
       -- https://github.com/eclipse/eclipse.jdt.ls/issues/376
+      -- Command in org.eclipse.lsp5j -> https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/Command.java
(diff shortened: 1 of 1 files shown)


Matsuuu:
Tried in my Java 11 environment with nvim lsp and jdtls and this fixes up the code action / import problem

lua/lspconfig/jdtls.lua
```diff
@@ -110,6 +110,8 @@ configs[server_name] = {
           -- TODO: (steelsojka) Handle more than one edit?
           if action.command == 'java.apply.workspaceEdit' then
             action.edit = action.arguments[1]
+          elseif type(action.command) == 'table' and action.command.command == 'java.apply.workspaceEdit' then
+            action.edit = action.command.arguments[1]
```
h-michael, line 114: @jerrywang1981 
Why are there two cases of "action.command" and "action.command.command"?

mfussenegger: My guess is that it is because the response can either be an array of commands or an array of code actions:

```
    -- actions is (Command | CodeAction)[] | null
    -- CodeAction
    --      title: String
    --      kind?: CodeActionKind
    --      diagnostics?: Diagnostic[]
    --      isPreferred?: boolean
    --      edit?: WorkspaceEdit
    --      command?: Command
    --
    -- Command
    --      title: String
    --      command: String
    --      arguments?: any[]
```

If it is a codeaction, then the command text is within command.command.

h-michael (approved):

PR #431, opened 2020-12-01
Return all settings when configuration has no section

zegervdv:
For `workspace/configuration` requests that have no `section` all settings data should be returned.


mjlbach:
Is it compliant with the spec to have a workspace/configuration requests without a section? Also, this function is being moved to neovim core in a PR there, so I'm closing this, but you can bring it up on the PR linked here https://github.com/neovim/neovim/pull/13659

Issue #430, opened 2020-12-01
pyright.lua minor setting issue


nialov:
- `nvim --version`: NVIM v0.5.0-ea3aaa0
- nvim-lsp version(commit hash): 09fc9449e88b8a4f97d14e87af3685a48df2137f
- What language server(If the problem is related to a specific language server): pyright

### Description of the problem

Current [pyright.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/pyright.lua) has settings as such:

~~~lua
settings = {
      analysis = { autoSearchPaths= true; };
      pyright = { useLibraryCodeForTypes = true; };
~~~

*analysis* should be wrapped in another parent table *python* as described here:

* https://github.com/microsoft/pyright/blob/master/docs/settings.md

E.g. in my personal init.lua

~~~lua
settings = {python = {analysis = {autoImportCompletions = false}}}
~~~

I could not get autoImportCompletions settings to work using the default way but configuring like above works for me.

lithammer:
Fix: https://github.com/neovim/nvim-lspconfig/pull/432

PR #432, opened 2020-12-01
Wrap pyright settings correctly with python.analysis

lithammer:
Fixes GH-430


mjlbach:
Thanks! Sorry I missed this.

lithammer:
Regarding the failing checks:

> GitHub Actions has encountered an internal error when running your job.

I suspect they just needs to be restarted (I can't do it though).

PR #437, opened 2020-12-03
Update tsserver root pattern, add jsconfig

creativenull:
Since tsserver also supports configuration for a JS project from a [jsconfig file](https://code.visualstudio.com/docs/languages/jsconfig). This will update the root pattern list for tsserver, adding `jsconfig.json` to the list.


mhartington:
Can confirm, tsserver does support jsconfig as well.

PR #439, opened 2020-12-04
Add extension to pyright command name on Windows

benawas:
Same as  #406, but for pyright

### DIFF
--- a/lua/lspconfig/pyright.lua
+++ b/lua/lspconfig/pyright.lua
@@ -2,16 +2,20 @@ local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
 local server_name = "pyright"
+local bin_name = "pyright-langserver"
+if vim.fn.has('win32') == 1 then
+  bin_name = bin_name..".cmd"
+end
 
 local installer = util.npm_installer {
   server_name = server_name;
   packages = {server_name};
-  binaries = {server_name};
+  binaries = {bin_name};
 }
 
 configs[server_name] = {
   default_config = {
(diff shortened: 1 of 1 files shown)


h-michael:
@benawas 
This change not only changed the ".cmd" for Windows, but also changed the value of "binaries" on all OS.

igrep:
@h-michael Watch the patch again! It replaces the extension only if `vim.fn.has('win32') == 1`, like https://github.com/neovim/nvim-lspconfig/pull/298/files, and my https://github.com/neovim/nvim-lspconfig/pull/406/files.

h-michael:
@igrep No. This commit also changes binaries value from `pyright` to `pyright-langserver`.

lithammer:
> @igrep No. This commit also changes binaries value from `pyright` to `pyright-langserver`.

To me it actually looks more correct now since the binary is called `pyright-langserver`, not `pyright`.

h-michael:
I understood the difference between pyright and pytight-langserver.
So pyright is a CLI and pytight-langserver is a language server.
https://github.com/microsoft/pyright/blob/a0620921898dfc5c0ecc704d53dbea3f6b2857e6/packages/pyright-internal/src/server.ts
https://github.com/microsoft/pyright/blob/f8db95da805d1944c605494193691ab062c3ee7b/packages/pyright-internal/src/pyright.ts

h-michael (approved):

lithammer:
pyright is also what the npm package is called, you can't _just_ install pyright-langserver.

PR #440, opened 2020-12-04
Pyright: Add workaround for client/registerCapability for node15

mjlbach:
See: https://github.com/neovim/neovim/issues/13448

closes:   #441

### DIFF
--- a/lua/lspconfig/pyright.lua
+++ b/lua/lspconfig/pyright.lua
@@ -14,6 +14,15 @@ configs[server_name] = {
     cmd = {"pyright-langserver", "--stdio"};
     filetypes = {"python"};
     root_dir = util.root_pattern(".git", "setup.py",  "setup.cfg", "pyproject.toml", "requirements.txt");
+    handlers = {
+      -- pyright ignores dynamicRegistration settings
+      ['client/registerCapability'] = function(_, _, _, _)
+        return {
+          result = nil;
+          error = nil;
+        }
+      end
+    };
     settings = {
       python = {
         analysis = {


mjlbach:
@h-michael Do you have thoughts on this? Would it be better to add a generic catch-all callback instead of per-server overrides?

h-michael:
@mjlbach We send dynamicRegistration as false.
But nevertheless, some language servers will send `client/registerCapability`, right?

The specification said that "Server must not register the same capability both statically through the initialize result and dynamically for the same document selector. If a server wants to support both static and dynamic registration it needs to check the client capability in the initialize request and only register the capability statically if the client doesn’t support dynamic registration for that capability."

I think it's against the specs for a server to make this request in the first place as far as I can see, what do you think?

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#client_registerCapability

mjlbach:
@h-michael Yes, in the linked tracking issue I filed against neovim core I noted that several language servers seem to be ignoring this. https://github.com/neovim/neovim/issues/13448

I agree that this should be filed and fixed with the respective defective language servers, however, users are reporting that language servers "aren't working" so I think merging this and #451 until fixes can be pushed upstream would be a good idea. I'll personally make the PR to remove the mitigation.

h-michael:
@mjlbach If we add the ad-hook implementation, I think that it is better to add this implementation to https://github.com/neovim/nvim-lspconfig/blob/f57212075222c18f56a2118ded20632228294249/lua/lspconfig/configs.lua#L36.

mjlbach:
I was going to add it there, but then we don't have a record of the "bad" language servers that need upstream fixes. i'm trying to fix this in pyright now... I don't know how willing microsoft is to merge PRs for non vscode clients (or at least, help debug them)

mjlbach:
@h-michael Filed a bug upstream for pyright https://github.com/microsoft/pyright/issues/1302 I'll add the ad-hook implementation

mjlbach:
See: https://github.com/neovim/nvim-lspconfig/pull/471

mjlbach:
Closing in favor of #471

Issue #446, opened 2020-12-08
sumneko_lua takes 35-40 seconds to load


sidmadala:
- `nvim --version`: NVIM v0.5.0-dev+4383c0f
- nvim-lsp version(commit hash): 76f11625481eddb7bcbf86b3e4f592a1aef35b37
- `:checkhealth` result: All green
- What language server (If the problem is related to a specific language server): sumneko_lua
- Operating system/version: MacOS Catalina 10.15.4
- Terminal name/version: Alacritty 0.6.0
- `$TERM`: xterm-256color

### How to reproduce the problem from neovim startup

I installed lspconfig using packer.nvim with nvim-completions as a dependency. I installed sumneko_lua using `:LspInstall sumneko_lua` within Neovim. I have pasted my configuration for packer.nvim and nvim-lspconfig below.  I don't know if this is a problem only related to myself.

<details>
<summary>plugins.lua</summary>

```lua
--[[
Packer plugin configuration
--]]

-- Only required if you have packer in your `opt` pack
vim.cmd [[packadd packer.nvim]]

vim.cmd [[ autocmd BufWritePost plugins.lua PackerCompile ]]

require("packer").startup(function()

	-- Essential
	use {"wbthomason/packer.nvim", opt = true}		-- Packer can manage itself as an optional plugin
	use {"svermeulen/vimpeccable"}

	-- Aesthetic
	use {"jnurmine/zenburn", config = require("plugins.colorscheme")}
	use {"itchyny/lightline.vim", config = require("plugins.lightline")}
	use {"Yggdroot/indentLine", config = require("plugins.indentline")}
	use {"junegunn/goyo.vim", config = require("plugins.goyo")}
	use {"kyazdani42/nvim-tree.lua", requires = "kyazdani42/nvim-web-devicons", config = require("plugins.tree")}

	-- LSP
	use {"neovim/nvim-lspconfig", requires = "nvim-lua/completion-nvim", config = require("plugins.lsp_config")}
	use {"nvim-treesitter/nvim-treesitter", config = require("plugins.treesitter")}
	use {"liuchengxu/vista.vim", config = require("plugins.vista")}
	use {"rust-lang/rust.vim"}

	-- Utilities
	use {"tpope/vim-commentary"}
	use {"jiangmiao/auto-pairs"}
	use {"junegunn/fzf.vim", requires = "junegunn/fzf", config = require("plugins.fzf")}
	use {"junegunn/vim-easy-align"}
[...]

PR #450, opened 2020-12-09
sumneko_lua: Use bufdir instead of homedir when resolving root_dir

lithammer:
The problem is that if you open a Lua file outside of a git repository, it will use your home directory as root directory. This causes Sumneko to scan your entire home directory for Lua files. Which is bad from both a security and performance point of view.

Fixes GH-446

### DIFF
--- a/lua/lspconfig/sumneko_lua.lua
+++ b/lua/lspconfig/sumneko_lua.lua
@@ -95,7 +95,7 @@ configs[name] = {
   default_config = {
     filetypes = {'lua'};
     root_dir = function(fname)
-      return util.find_git_ancestor(fname) or vim.loop.os_homedir()
+      return util.find_git_ancestor(fname) or util.path.dirname(fname)
     end;
     log_level = vim.lsp.protocol.MessageType.Warning;
   };
@@ -123,7 +123,7 @@ require'lspconfig'.sumneko_lua.setup{
 If you install via our installer, if you execute `:LspInstallInfo sumneko_lua`, you can know `cmd` value.
 ]];
     default_config = {
-      root_dir = [[root_pattern(".git") or os_homedir]];
+      root_dir = [[root_pattern(".git") or bufdir]];
     };
   };
 }


lua/lspconfig/sumneko_lua.lua
```diff
@@ -123,7 +123,7 @@ require'lspconfig'.sumneko_lua.setup{
 If you install via our installer, if you execute `:LspInstallInfo sumneko_lua`, you can know `cmd` value.
 ]];
     default_config = {
-      root_dir = [[root_pattern(".git") or os_homedir]];
+      root_dir = [[root_pattern(".git") or bufdir]];
```
lithammer, line 126: Not sure what's reasonable to write here.

beauwilliams:
Currently I have

`lsp.sumneko_lua.setup {on_attach = custom_attach,
			root_dir = vim.loop.cwd --Sets root to cwd
}`

The lua language server loads almost instantly. Perhaps this might be a simpler fix?

beauwilliams:
I have noticed that diagnostics does not recognise the `vim` global var. This might be an issue with my cwd config however it should not be..

Not sure if you have same issue

mjlbach:
@beauwilliams https://github.com/nanotee/nvim-lua-guide#sumnekolua-language-server

I think lithammer's implementation is fine/in-line with the other lsp's, and if no one objects I think we should go ahead and merge this as-is

beauwilliams:
Thanks, sorry reference to my above comment regarding vim global var I have fixed that. My bad.

I am still having issues with root pattern setting git and then fallback to cwd but I now believe it is related to the `or`. Likely an issue with how it is being evaluated

`local cwd = vim.loop.cwd
lsp.sumneko_lua.setup {
  on_attach = custom_attach,
  root_dir = vim.loop.cwd,
  --root_dir = util.root_pattern('.git') or cwd, --Not working
    settings = {
        Lua = {
            runtime = {
                -- Tell the language server which version of Lua you're using (LuaJIT in the case of Neovim)
                version = 'LuaJIT',
                -- Setup your lua path
                path = vim.split(package.path, ';'),
            },
            diagnostics = {
                -- Get the language server to recognize the `vim` global
                globals = {'vim'},
            },
            workspace = {
                -- Make the server aware of Neovim runtime files
                library = {
                },
            },
        },
    },
}`

beauwilliams:
Update: Fixed my earlier issue. Again just a silly mistake forgot to wrap the `or` in a function

Here is my fully working lua config, boots instantly for me right now

```lsp.sumneko_lua.setup {
  on_attach = custom_attach,
  root_dir = function(fname)
    return util.find_git_ancestor(fname) or
      util.path.dirname(fname)
  end,
  --root_dir = vim.loop.cwd,
  --root_dir = util.root_pattern('.git') or cwd, --Not working
    settings = {
        Lua = {
            runtime = {
                -- Tell the language server which version of Lua you're using (LuaJIT in the case of Neovim)
                version = 'LuaJIT',
                -- Setup your lua path
                path = vim.split(package.path, ';'),
            },
            diagnostics = {
                -- Get the language server to recognize the `vim` global
                globals = {'vim'},
            },
            workspace = {
                -- Make the server aware of Neovim runtime files
                library = {
                },
            },
        },
    },
}

PR #451, opened 2020-12-10
Add workaround to make yamlls work with node v15.x

david0:
See https://github.com/neovim/neovim/issues/13448 and https://github.com/redhat-developer/yaml-language-server/issues/370.

I don't see a good way how yamlls could find out that registerCapability is not working for workspaceFolder changes, so this workaround would solve that for the moment.

### DIFF
--- a/lua/lspconfig/yamlls.lua
+++ b/lua/lspconfig/yamlls.lua
@@ -14,6 +14,15 @@ configs[server_name] = {
   default_config = {
     cmd = {bin_name, "--stdio"};
     filetypes = {"yaml"};
+    handlers = {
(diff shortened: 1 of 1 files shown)


mjlbach:
I'm contemplating making a PR to just make all servers have this callback by default... I'm not sure how widespread it is yet.

mjlbach:
See: https://github.com/neovim/nvim-lspconfig/pull/471

mjlbach:
@david0 I think since #471 it took care of this

david0:
471 works for me. I'll close this PR.

PR #453, opened 2020-12-11
groovyls: nvim_lsp -> lspconfig

tormodatt:
I think @wuilliam321  had an older version of nvim when he had it running using https://github.com/neovim/nvim-lspconfig/pull/355.

I also just tested this successfully on macOS. Thanks @wuilliam321 !

PR #454, opened 2020-12-11
groovyls: update README as well

tormodatt:
forgot to update README in https://github.com/neovim/nvim-lspconfig/pull/453

I also run this successfully on Java 8, and the LS has 1.8 set in build.gradle as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -2068,18 +2068,18 @@ require'lspconfig'.gopls.setup{}
 https://github.com/prominic/groovy-language-server.git
 
 Requirements:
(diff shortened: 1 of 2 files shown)


tormodatt:
maybe @h-michael  can merge this?

tormodatt:
ahh, then we can close this, as I've already added it in the lua-file

h-michael:
@tormodatt 
Thanks. I fix docgen at https://github.com/neovim/nvim-lspconfig/pull/470 .

PR #462, opened 2020-12-18
Add support for Lua language server alloyed_lua

ojroques:
Add a new language server for Lua: [here's the repo](https://github.com/Alloyed/lua-lsp). I find `sumneko_lua` way too complicated to install so I figured I'd use this one instead.

### DIFF
--- a/lua/lspconfig/alloyed_lua.lua
+++ b/lua/lspconfig/alloyed_lua.lua
@@ -1,0 +1,33 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.alloyed_lua = {
+  default_config = {
+    cmd = {'lua-lsp'};
+    filetypes = {'lua'};
+    root_dir = function(fname)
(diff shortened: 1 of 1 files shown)


saadparwaiz1:
shouldn't the root_dir function look for .luacheckrc and .luacompleterc?

ojroques:
Yes good point. I've updated `root_dir` accordingly.

tssm:
Hey folks, I would like to know if this is going to be merged? I was working on the same without realizing someone already opened a PR

mjlbach:
The main reason I haven't merged this, is because currently alloyed is not a full featured language server, and we already offer sumneko. Symbol/signature/references aren't implemented, it's a bit slower, and the completion results are IMO worse. It also hasn't been updated in 3+ months.

I guess if there's a big demand for this we can merge it, but generally we tend to only have language servers that are either 

1. The only available language server for the respective language
2. Mostly full implementation of the lsp spec/actively developed

mjlbach:
I'm going to close since there hasn't been any activity on the alloyed-lsp repo, but feel free to reopen if the project picks back up again :)

PR #463, opened 2020-12-18
Remove ghcide, fix docgen

mjlbach:
ghcide shouldn't be used directly anymore, the vscode plugin was deprecated, and the failure to fetch the (now missing) package.json was causing docgen to fail to update.

https://github.com/haskell/ghcide/issues/935


mjlbach:
superceeded by #469

PR #464, opened 2020-12-18
Separate autogenerated readme from readme.md

### LINKED ISSUE #333: docgen: generate to separate doc, link to it from README.md
this will solve some common problems:

- contributors often don't know or forget that they need to update README_template.md instead of README.md directly. 
- README.md might be getting too big so it makes sense to have the generated stuff in a separate file

mjlbach:
Closes #333

@justinmk Do you have an opinion on organization/naming?


scripts/docgen.lua
```diff
@@ -224,7 +224,7 @@ local function generate_readme(template_file, params)
   local input_template = readfile(template_file)
   local readme_data = template(input_template, params)
 
-  local writer = io.open("README.md", "w")
+  local writer = io.open("GENERATED_README.md", "w")
```
h-michael, line 227: It doesn't matter to the user whether it is GENERATED or not, so why not just use "CONFIG.md"?

mjlbach: So, I agree with you. The main reason I put GENERATED was because people keep filing PRs in which they've edited README.md, I'll update the PR.

h-michael:
@mjlbach nit: It may be `CONFIG.md` instead of `CONFIG.MD`.

mjlbach:
Whoops, I don't know why I did that. Shouldn't have mattered since the casing is write in docgen but I pushed the fix

h-michael (approved):

PR #469, opened 2020-12-20
Add deprecation message to ghcide

h-michael:
ghcide isn't for end users now.
So I add deprecation message.


lua/lspconfig/ghcide.lua
```diff
@@ -9,7 +11,6 @@ configs.ghcide = {
   };
 
   docs = {
-    package_json = "https://raw.githubusercontent.com/digital-asset/ghcide/master/extension/package.json";
```
h-michael, line 12: This file no longer exists.

PR #474, opened 2020-12-21
New: Provide Verilog/SystemVerilog svls support

dpretet:
Hello there,

here is a pull request for Verilog and SystemVerilog support. I used svls language server and Nvim v0.5.0+950-g5ce328df4. I was able to start without complains by setting the following config in my init.vim:

```lua
require 'lspconfig'.svls.config{}
```

Cheers,
Damien

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,7 @@ that config.
 - [sourcekit](#sourcekit)
 - [sqlls](#sqlls)
 - [sumneko_lua](#sumneko_lua)
+- [svls](#SystemVerilog)
 - [terraformls](#terraformls)
 - [texlab](#texlab)
 - [tsserver](#tsserver)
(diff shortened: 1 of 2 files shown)


mjlbach:
Thanks for the PR! Note that docgen overrites readme.md, so you should add all documentation in the lua file.

dpretet:
Hello @mjlbach ,

Thanks for the feedback @mjlbach :)

1. Does it mean I shouldn't modify the README?
2. The macOS check failed because because of a dependency install issue. Don't know how to handle that.

mjlbach:
After https://github.com/neovim/nvim-lspconfig/pull/464 you might want to check the divide between README.md and CONFIG.md. README.md is not autogenerated and can be modified (but there is nothing in this PR which *should* modify it). CONFIG.md is autogenerated from the docs in each language servers respective lua module, so even if you did modify it, it would be overwritten by CI.

The macos thing is a known issue, not sure how to fix it.

PR #486, opened 2020-12-24
Added perl language server

pretentious7:
Added perl language server, tried to keep things in line with other language server files. Not very familiar with lua so haven't used variables or anything, might improve it in the future.


lua/lspconfig/perlls.lua
```diff
@@ -0,0 +1,37 @@
...
+      perl = {
+        perlCmd = 'perl';
+        perlInc = ' ';
+        fileFilter = {".pm",".pl"};
+        ignoreDirs =  '.git';
+      };
+    };
+    filetypes = {"perl"};
+    root_dir = function(fname)
+      return util.root_pattern(".git")(fname) or vim.loop.os_homedir()
```
mjlbach, line 20: Do you mean to set the root_dir to home here as opposed to the directory containing the file?

lua/lspconfig/perlls.lua
```diff
@@ -0,0 +1,37 @@
...
+    end;
+  };
+  docs = {
+    package_json = "https://github.com/richterger/Perl-LanguageServer/blob/master/clients/vscode/perl/package.json";
+    description = [[
+    https://github.com/richterger/Perl-LanguageServer/tree/master/clients/vscode/perl
+
+    `Perl-LanguageServer`, a language server for Perl.
+
+    To make this work, ensure that you have Perl::LanguageServer installed and the perl command in your $PATH
```
mjlbach, line 30: I would say `To use the language server, ensure that you have Perl-LanguageServer installed and perl command is on your path.`

lua/lspconfig/perlls.lua
pretentious7: aiya, you're right, changed to vim.fin.getcwd()

lua/lspconfig/perlls.lua
pretentious7: Yep, done.

PR #489, opened 2020-12-24
Add pyright.organizeimports command

tzx:
`pyright` allows you to organize imports through `workspace/executeCommand`, so I added a command that does this.

I tried adding a command to generate typestubs but that seems to rely on `window/showMessage`, which is not yet supported (https://github.com/neovim/neovim/issues/11710)


mjlbach:
I'm fine with this, as there are only 2 optional commands for pyright, if it gets anymore, we should break off into its own extension.

PR #491, opened 2020-12-25
Add deno language server

mjlbach:
DRAFT: Add deno language server.

This seems to work on JS/TS. I haven't done any deno work, so not sure what root directories we should also trigger.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -13,6 +13,7 @@ that config.
 - [codeqlls](#codeqlls)
 - [cssls](#cssls)
 - [dartls](#dartls)
+- [denols](#denols)
 - [diagnosticls](#diagnosticls)
 - [dockerls](#dockerls)
 - [efm](#efm)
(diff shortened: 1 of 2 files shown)


mjlbach:
@David-Else, @kuator, @bfulop have any of you tested this? It worked for me on a trivial deno project but I could use feedbak before merging (sorry for the callout)

kuator:
> @David-Else, @kuator, @bfulop have any of you tested this? It worked for me on a trivial deno project but I could use feedbak before merging (sorry for the callout)

No problem, I tried it and it runs, but some features like go to definition and signature help don't work. I'm pretty sure it's due to limitation of the language server itself as it's pretty new.

David-Else:
I am afraid I don't actually use Neovim much, so don't have time to learn how to test this config/related extension. I just chanced upon the thread and shared some info on Deno in case it helped.

The LSP is under heavy development, so missing some features right now. This thread will show what features of the LSP are available and will be updated as they are added: https://github.com/denoland/deno/issues/8643

I am sure it will be fully featured very soon! Thanks for this great config you are adding, I hope to use it in the future when Neovim 0.5 has an official release and I have the time to set it all up and try it out against VS Code.

PR #495, opened 2020-12-26
Gopls quickfix so it doesn't break if no go.mod or .git folder

pretentious7:
See title, in line with other lsp configs.

### DIFF
--- a/lua/lspconfig/gopls.lua
+++ b/lua/lspconfig/gopls.lua
@@ -5,7 +5,9 @@ configs.gopls = {
   default_config = {
     cmd = {"gopls"};
     filetypes = {"go", "gomod"};
-    root_dir = util.root_pattern("go.mod", ".git");
+    root_dir = function(fname)
+      return util.root_pattern(".git","go.mod")(fname) or vim.fn.getcwd()
+    end;
   };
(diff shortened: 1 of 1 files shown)


pretentious7:
ah sorry about it sort of combining with the perlls pr, I can remove those commits if needed, but it'll probably take me a while to figure out and I did the stuff in the review, so is it ok?

mjlbach:
I fixed the branch

pretentious7:
thank you!

btw just for my reference, how did you do it? like what git commands?

mjlbach:
`git rebase -i master`, then I deleted all non-pertinent commits (all but the last one) and force pushed once I checked the diff with your branch and master (I made some educated assumptions about your intentions with this PR :) )

pretentious7:
I see. i need to learnt to use rebase and diff better. thanks!

mjlbach:
Why did you close the PR?

pretentious7:
...I don't understand how PR's work

mjlbach:
Yes, when the PR is merged, the PR will close and the changes will be applied to the target branch (in this case master).

mjlbach:
cwd is not a good fallback, sorry, cleaning up old PRs now

PR #497, opened 2020-12-28
Update tsserver.lua

pretentious7:
peer dependency, useful to make sure, would've saved me atleast an hour or so of fruitless debugging (also seems to be somewhat standard to do for npm docs elsewhere).

(also, I could remove all the install stuff and make the file a lot shorter, if that's welcome? the thread in issues is inconclusive)


mjlbach:
Closing because the documentation changed, and I'll make a PR and merge covering this. Thanks for bringing it up!

Issue #467, opened 2020-12-19
LspInstall for pyright and intelephense fails when node packages are installed with the same name as the plugin


rverton:
- `nvim --version`: v0.5.0-dev+950-g5ce328df4
- nvim-lsp version(commit hash): 1e4a963ba4d00d559247b5fca827142958d4dbdf
- `:checkhealth` result: https://dpaste.com/9PFATQ2MA
- What language server(If the problem is related to a specific language server): pyright, intelephense
- Operating system/version: macOS Big Sur
- Terminal name/version: iterm2
- `$TERM`: xterm-256color
- NPM version: 6.14.9

### How to reproduce the problem from neovim startup

Execute `:LspInstall pyright` or `:LspInstall intelephense`.

### Actual behaviour

The following error is printed (reformatted because its broken): 
```
18 verbose stack Error: Refusing to install package with name "pyright" under a package
18 verbose stack also called "pyright". Did you name your project the same
18 verbose stack as the dependency you're installing?
```

The same happens when installing `intelephense`. Looking in the `/Users/robin/.cache/nvim/lspconfig/pyright` folder (which is created by the LspInstall command), there is a package.json with the following content:
```json
{
  "name": "pyright",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

Running a `npm install pyright` here fails because the package name can't have the same name as the dependency I think. Executing `:LspInstall tsserver` works here, because the actual package which is installed is called `typescript-language-server` and therefore has a different name than the folder.

mjlbach:
I'm not sure I understand this bug. If pyright is *already* installed, and you call :LspInstall pyright it fails?

rverton:
Pyright/intelephense are not installed. If they would be installed, LspInstall would tell it and refuse to go on. 

The LspInstall command creates a folder called “pyright”, puts a package.json in there with the name “pyright” and tries to install the module “pyright” in this I guess. Npm then refuses to do this.

mjlbach:
Does `npm i -g pyright` not work for you?

rverton:
When I install pyright manually via npm, then LSP does not work here. This is what `LspInstallInfo` gives me for pyright:

```
  pyright = {
    bin_dir = "/Users/robin/.cache/nvim/lspconfig/pyright/node_modules/.bin",
[...]

Issue #421, opened 2020-11-22
Does sumneko_lua really need clang to compile?


dlukes:
The `sumneko_lua` install recipe requires `clang`:

https://github.com/neovim/nvim-lspconfig/blob/4f19567fb76438d39ed196ae5b4e06b749e75e16/lua/lspconfig/sumneko_lua.lua#L42-L43

However, I noticed during one of my recent installs that GCC was actually being used for compilation, and when I tried on another machine without `clang` and just commented out the check above to bypass it, everything seems to be working fine (as in, the same).

mjlbach:
Closed by #498

Issue #361, opened 2020-09-21
Can't load tsserver and jdtls


blackCauldron7:
- `nvim --version`:
```
NVIM v0.5.0-704-g905c2eb35
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-f2a9ebb1-a963-4862-8393-c5d9d3de43b3

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
```
- nvim-lsp version(commit hash): `commit 60133c47e0fd82556d7ca092546ebfa8d047466e (grafted, HEAD -> master, origin/master, origin/HEAD)`
- `:checkhealth` result
```
health#nvim_lsp#check
========================================================================
## Checking language server protocol configuration
  - INFO: html: configuration checked.
  - INFO: jdtls: configuration checked.
  - INFO: sumneko_lua: configuration checked.
  - INFO: clangd: configuration checked.
  - INFO: pyls: configuration checked.
  - INFO: cssls: configuration checked.
  - INFO: vimls: configuration checked.
  - INFO: tsserver: configuration checked.
  - INFO: bashls: configuration checked.
```
- What language server(If the problem is related to a specific language server): `tsserver, jdtls`
[...]

Issue #359, opened 2020-09-21
LspInstall not working Windows


RefusesNames:
- `nvim --version`: NVIM v0.5.0-705-g9f704c88a
- nvim-lsp version(commit hash):
    How can I see that? nvim-lspconfig has the commit `60133c47e0fd82556d7ca092546ebfa8d047466e`
- `:checkhealth` result:
    configuration ok, performance ok, remote plugins ok, clipboard ok, python2 and 3 ok, virtualenv ok, node.js ok, perl and ruby provider missing
- What language server(If the problem is related to a specific language server): All of them (`vimls` as an example)
- Operating system/version: Windows 10 Pro (Version 2004)
- Terminal name/version: PowerShell core (7.1.0-preview.7)
- `$TERM`: xterm-256color

### How to reproduce the problem from neovim startup
```
:LspInstall vimls
```

### Actual behaviour
Message: "Installation requires "sh", "npm", "mkdir"

### Expected behaviour
Installation of the language server

tarasyarema:
I guess the first think to do is look here: https://github.com/neovim/nvim-lspconfig/blob/60133c47e0fd82556d7ca092546ebfa8d047466e/lua/nvim_lsp/util.lua#L378.

It seems like the only system implemented are the ones that have `sh`, `npm` and `mkdir`. I think that this can be easly changed to something like this:

```lua
local cmd = io.popen("powershell -command -", "w")
    local install_script = ([[
    mkdir -p "{{install_dir}}"
    cd "{{install_dir}}"
    && npm init -y
    npm install {{packages}} --no-package-lock --no-save --production
    {{post_install_script}}
    ]]):gsub("{{(%S+)}}", install_params)
    cmd:write(install_script)
    cmd:close()
```

lithammer:
I think the idea is to completely drop the installers. See https://github.com/neovim/nvim-lspconfig/issues/334.

tarasyarema:
I was just reading that issue jaja

mjlbach:
Closed by #498

PR #498, opened 2020-12-29
Remove all installers and install logic

### LINKED ISSUE #334: remove make_installer and all auto-install logic

mjlbach:
Closes #334 #26 #92 #289 #359 #361 #421 #467

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -10,6 +10,35 @@ function M.available_servers()
   return vim.tbl_keys(configs)
 end
 
+function M.installable_servers()
+  print("deprecated, see https://github.com/neovim/neovim/wiki/Following-HEAD")
+end
+
+-- Called from plugin/lspconfig.vim because it requires knowing that the last
+-- script in scriptnames to be executed is lspconfig.
+function M._root._setup()
+  M._root.commands = {
+    LspInstall = {
+      function()
+        print("deprecated, see https://github.com/neovim/neovim/wiki/Following-HEAD")
+      end;
+      "-nargs=?";
+      "-complete=custom,v:lua.lsp_complete_installable_servers";
+      description = '`:LspInstall {name}` installs a server under stdpath("cache")/lspconfig/{name}';
+    };
+    LspInstallInfo = {
+      function()
+        print("deprecated, see https://github.com/neovim/neovim/wiki/Following-HEAD")
+      end;
+      "-nargs=?";
+      "-complete=custom,v:lua.lsp_complete_servers";
+      description = 'Print installation info for {name} if one is specified, or all installable servers.';
+    };
+  };
+
+  M.util.create_module_commands("_root", M._root.commands)
+end
+
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -4,71 +4,10 @@ local M = {
   util = require 'lspconfig/util';
 }
 
+M._root = {}
+
 function M.available_servers()
   return vim.tbl_keys(configs)
-end
-
-function M.installable_servers()
-  local res = {}
-  for k, v in pairs(configs) do
-    if v.install then table.insert(res, k) end
-  end
-  return res
-end
-
-M._root = {}
--- Called from plugin/lspconfig.vim because it requires knowing that the last
--- script in scriptnames to be executed is lspconfig.
-function M._root._setup()
-  M._root.commands = {
-    LspInstall = {
-      function(name)
-        if configs[name] == nil then
-          pcall(require('lspconfig/'..name))
-        end
-        local config = configs[name]
-        if not config then
-          return print("Invalid server name:", name)
-        end
-        if not config.install then
-          return print(name, "can't be automatically installed (yet)")
-        end
-        if config.install_info().is_installed then
-          return print(name, "is already installed")
-        end
-        config.install()
-      end;
-      "-nargs=1";
-      "-complete=custom,v:lua.lsp_complete_installable_servers";
-      description = '`:LspInstall {name}` installs a server under stdpath("cache")/lspconfig/{name}';
-    };
-    LspInstallInfo = {
-      function(name)
-        if name == nil then
-          local res = {}
-          for k, v in pairs(configs) do
-            if v.install_info then
-              res[k] = v.install_info()
-            end
-          end
-          return print(vim.inspect(res))
-        end
-        if configs[name] == nil then
-          pcall(require('lspconfig/'..name))
-        end
-        local config = configs[name]
-        if not config then
-          return print("Invalid server name:", name)
-        end
-        return print(vim.inspect(config.install_info()))
-      end;
-      "-nargs=?";
-      "-complete=custom,v:lua.lsp_complete_servers";
-      description = 'Print installation info for {name} if one is specified, or all installable servers.';
-    };
-  };
-
-  M.util.create_module_commands("_root", M._root.commands)
 end
 
 local mt = {}
--- a/lua/lspconfig/als.lua
+++ b/lua/lspconfig/als.lua
@@ -7,62 +7,6 @@ if vim.fn.has('win32') == 1 then
   bin_name = 'ada_language_server.exe'
 end
 
-local function make_installer()
-  local install_dir = util.path.join{util.base_install_dir, server_name}
-
-  local url = 'https://dl.bintray.com/reznikmm/ada-language-server/linux-latest.tar.gz'
-  local download_target = util.path.join{install_dir, "als.tar.gz"}
-  local extracted_dir = "linux"
-  local extract_cmd = string.format("tar -xzf '%s' --one-top-level='%s'", download_target, install_dir)
-
-  if vim.fn.has('win32') == 1 then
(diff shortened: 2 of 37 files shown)


mjlbach:
@teto This is going to take fairly exhaustive testing, and we'll probably have to update the documentation. It's mostly ready, and I think I removed all dead documentation. To be honest, I don't know how to install a lot of these language servers the "normal" way, since all of mine are packaged using nix.

pretentious7:
:clap: :clap: :clap: 

I've been using basically that with typescript for a couple days, so can confirm that works.

lua/lspconfig/jsonls.lua
```diff
@@ -35,7 +18,7 @@ https://github.com/vscode-langservers/vscode-json-languageserver
 
 vscode-json-languageserver, a language server for JSON and JSON schema
 
-`vscode-json-languageserver` can be installed via `:LspInstall jsonls` or by yourself with `npm`:
+`vscode-json-languageserver` can be install via `npm`:
```
teto, line 21: installed ?

lua/lspconfig/metals.lua
```diff
@@ -94,13 +29,11 @@ let g:metals_server_version = '0.8.4+106-5f2b9350-SNAPSHOT'
 ```
 
 Scala language server with rich IDE features.
-`metals` can be installed via `:LspInstall metals`.
+
+See project page for installation instructions
```
teto, line 33: maybe reference that page ?

teto:
thanks for tackling this, it needs to be done before the 0.5 release so that users dont start relying on these primitives.
Once this is merged, we need to reference it in https://github.com/neovim/neovim/wiki/Following-HEAD .

lua/lspconfig/metals.lua
mjlbach: Done, and I added instruction. It's absurdly difficult to find documentation for how to manually install their language server, I had to look under the emacs instructions for a link, I copied the installation steps just to be safe).

mjlbach:
If there are any remaining dicey documentation issues let me know. I think we should also fix whatever is going on with the print warning that cmd is undefined before merging (and hopefully more people chime in that the instructions are great/bad/insufficient. 

This will also break things for people, as cmd will need to be defined. So I imagine an influx of issues. I don't think there's a good intermediate measure.

teto:
I think we should default `cmd` to the name of the binary and let the user put it in PATH. The good thing about this is that instead of a (lua) error (?), we can advise to run `:checkhealth nvim_lsp` and it will tell the user that the binary is not in PATH ? Ideally we need a much more helpful checkhealth for lsp https://github.com/neovim/neovim/pull/13161

mjlbach:
> I think we should default cmd to the name of the binary and let the user put it in PATH.

I initially tried doing this, but many "commands" are simply the path to some random output of an npm build process. My strategy was mostly to keep the names of commands that point to actual binaries one would reasonably put on their path and remove those that were some random build output file. Also there's a slight issue in that windows binaries often will have a different extension. Right now without the error handling I put in, you'll get a nasty red error message if the language client fails to start the server.

lua/lspconfig/metals.lua
```diff
@@ -79,28 +17,36 @@ configs[server_name] = {
...
+chmod +x coursier
+./coursier bootstrap \
+  --java-opt -Xss4m \
+  --java-opt -Xms100m \
+  --java-opt -Dmetals.client=emacs \
+  org.scalameta:metals_2.12:0.9.8 \
+  -r bintray:scalacenter/releases \
+  -r sonatype:snapshots \
+  -o /usr/local/bin/metals -f
+```
```
ckipp01, line 46: I would change this as bootstrapping metals with `-Dmetals.client=emacs` sets some defaults that will cause Metals to not really work as expected. There actually is a very small section on using other vim LSP clients on the vim page here https://scalameta.org/metals/docs/editors/vim.html#using-an-alternative-lsp-client. However, that will be updated soon with nvim LSP specific stuff as well, since we're actually putting effort into having a nvim LSP focused extension for Metals users: https://github.com/scalameta/nvim-metals. This is mainly due to a  lot of defaults not working unless some extra things are offered, and we make heavy use of LSP extensions. Also an alternative to what you have here can I suggest the following?

```suggestion
See full instructions in the Metals documentation:

https://scalameta.org/metals/docs/editors/vim.html#using-an-alternative-lsp-client

Note: that if you're using [nvim-metals](https://github.com/scalameta/nvim-metals), that plugin fully handles the setup and installation of Metals, and you shouldn't set up Metals both with it and this plugin.

To install Metals, make sure to have [coursier](https://get-coursier.io/docs/cli-installation) installed, and once you do you can install the latest Metals with `cs install metals`. You can also manually bootstrap Metals with the following command.

```bash
cs bootstrap \
  --java-opt -Xss4m \
  --java-opt -Xms100m \
  org.scalameta:metals_2.12:<enter-version-here> \
  -r bintray:scalacenter/releases \
  -r sonatype:snapshots \
  -o /usr/local/bin/metals -f
```
```

**Comment by mjlbach** [comment]
@ckipp01 Thanks for the useful feedback! I've merged your changes. I don't use metals, and had a bit of a hard time finding the page you referenced, so I'm glad that is being worked on. I'm happy to nudge people to the nvim-metals plugin in the documentation; I don't think we can ever match the deep integration that server specific plugins have.

Let me know if there are any specific features you want that would make the plugin easier. I've recently filed several PRs with core improving 3.16 compliance.

**Comment by mjlbach** [comment]
@teto Unless I'm mistaken, I think the only bin_name we're missing now is elixir

**Comment by mjlbach** [comment]
@mortepau Do you have suggestions for improving the documentation regarding elixir install and default cmd without LspInstall?

**Comment by saadparwaiz1** [comment]
@mjlbach I'm not sure if this is helpful but I have been using the elixir language server with emacs's lsp-mode and the installation instruction are as simple picking up the zip file from github release and using language_server.sh on unix and language_server.bat on windows. These can be found [here](https://github.com/elixir-lsp/elixir-ls#detailed-installation-instructions)

**Comment by mjlbach** [comment]
@teto I think this is good for another look. I added the missing elixir/sumneko lua documentation, and added some documentation that had always been missing for OmniSharp.

**Comment by teto** [comment]
Is it possible to fix the CONFIG.md conflict. I will let @h-michael merge this when he feels like it, he is the most knowledgeable about this repo.

**Comment by mjlbach** [comment]
Fixed

File: lua/lspconfig.lua [hunk 4]
```diff
@@ -4,71 +4,10 @@ local M = {
...
-  end
-  return res
-end
-
 M._root = {}
--- Called from plugin/lspconfig.vim because it requires knowing that the last
--- script in scriptnames to be executed is lspconfig.
-function M._root._setup()
-  M._root.commands = {
-    LspInstall = {
```
teto, line 24: neovim 0.5 hasn't been released yet so in theory we could just get rid of this but aving the commands print a message like `print("deprecated, see https://github.com/neovim/neovim/wiki/Following-HEAD")  would avoid some pain for everyone (users and contributors on gitter). And we would remove it just before 0.5.

mjlbach: Added in a separate commit to make it easy to remove 🔪

lua/lspconfig/als.lua
```diff
@@ -7,78 +7,19 @@ if vim.fn.has('win32') == 1 then
...
-  on_new_config = function(config)
-    installer.configure(config)
-  end;
   docs = {
     package_json = "https://raw.githubusercontent.com/AdaCore/ada_language_server/master/integration/vscode/ada/package.json";
     description = [[
 https://github.com/AdaCore/ada_language_server
 
-Ada language server. Use `LspInstall als` to install it.
+See project page for installation instructions.
```
teto, line 22: https://github.com/AdaCore/ada_language_server#Install

teto (approved):
I've grepped the source for LspInstall and there is one mention left in health.lua

mjlbach:
I noticed that one, but I'm not sure what it means...

teto:
git blame shows it was due to https://github.com/neovim/nvim-lspconfig/issues/248 . Seems like make_config  is not available prior to te setup call ? maybe keep the `if`  and just remove the comment about LspInstall for now. I think we can merge this today,

mjlbach:
Done, I reworded it to reflect what I (think?) is the purpose.

teto:
cool, can you mention the change on https://github.com/neovim/neovim/wiki/Following-HEAD ? feel free to merge when it's done

mjlbach:
Done, and merging.

Borwe:
Man I miss the auto LspInstall... was so easy.

teto:
I've seen https://github.com/anott03/nvim-lspinstall popup in my github notifs. Dont know how good it is but may be a solution.

PR #509, opened 2020-12-31
Remove deprecated handlers

mjlbach:
See: https://github.com/neovim/neovim/pull/13649


mjlbach:
@tjdevries This removes almost all the handlers, except one workaround that I mentioned on the gitter. If people have nightly builds older than week, it will cause issues for them.

PR #513, opened 2020-12-31
clangd: add support for file status extension and fix capabilities

jdrouhard:
- Add support for file-status extension utilizing new progress support
- Extend capabilities instead of replacing default to add editsNearCursor

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -622,11 +622,13 @@ require'lspconfig'.clangd.setup{}
   - ClangdSwitchSourceHeader: Switch between source/header
   
   Default Values:
-    capabilities = default capabilities, with offsetEncoding utf-8
     cmd = { "clangd", "--background-index" }
     filetypes = { "c", "cpp", "objc", "objcpp" }
-    on_init = function to handle changing offsetEncoding
     root_dir = root_pattern("compile_commands.json", "compile_flags.txt", ".git") or dirname
+    on_init = function to handle changing offsetEncoding
+    capabilities = default capabilities, with offsetEncoding utf-8
+    handlers = handler for textDocument/clangd.fileStatus extension
+    init_options = { clangdFileStatus = true }
 ```
 
 ## clojure_lsp
--- a/lua/lspconfig/clangd.lua
+++ b/lua/lspconfig/clangd.lua
@@ -16,31 +16,37 @@ end
 local function file_status_update(_, _, message, client_id)
   local client = vim.lsp.get_client_by_id(client_id)
   if not client then
-    vim.api.nvim_err_writeln("LSP[" .. client_id .. "] client has shut down after sending the message")
+    vim.api.nvim_err_writeln("LSP[id=" .. client_id .. "] client has shut down after sending the message")
     vim.api.nvim_command("redraw")
   end
   client.messages.status = { uri = message.uri, content = message.state }
-  vim.api.nvim_command('doautocmd <nomodeline> User LspProgressUpdate')
+  vim.api.nvim_command('doautocmd <nomodeline> User LspStatusUpdate')
 end
 
 local root_pattern = util.root_pattern("compile_commands.json", "compile_flags.txt", ".git")
 
-local default_capabilities = vim.tbl_deep_extend('keep', require'vim.lsp.protocol'.make_client_capabilities(), {
+local default_capabilities = vim.tbl_deep_extend('force', vim.lsp.protocol.make_client_capabilities(), {
   textDocument = {
     completion = {
       editsNearCursor = true
     }
-  }
+  },
+  offsetEncoding = {"utf-8", "utf-16"}
 })
 
 configs.clangd = {
-  default_config = util.utf8_config {
+  default_config = {
     cmd = {"clangd", "--background-index"};
     filetypes = {"c", "cpp", "objc", "objcpp"};
     root_dir = function(fname)
       local filename = util.path.is_absolute(fname) and fname
         or util.path.join(vim.loop.cwd(), fname)
       return root_pattern(filename) or util.path.dirname(filename)
+    end;
+    on_init = function(client, result)
+      if result.offsetEncoding then
+        client.offset_encoding = result.offsetEncoding
+      end
     end;
     capabilities = default_capabilities;
     handlers = {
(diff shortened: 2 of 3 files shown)


lua/lspconfig/clangd.lua
```diff
@@ -12,7 +12,27 @@ local function switch_source_header(bufnr)
...
+    vim.api.nvim_err_writeln("LSP[" .. client_id .. "] client has shut down after sending the message")
+    vim.api.nvim_command("redraw")
+  end
+  client.messages.status = { uri = message.uri, content = message.state }
+  vim.api.nvim_command('doautocmd <nomodeline> User LspProgressUpdate')
+end
+
 local root_pattern = util.root_pattern("compile_commands.json", "compile_flags.txt", ".git")
+
+local default_capabilities = vim.tbl_deep_extend('keep', require'vim.lsp.protocol'.make_client_capabilities(), {
```
mjlbach, line 28: Why is this portion necessary? I don't see any modifications to the capabilities from before?

lua/lspconfig/clangd.lua
```diff
@@ -12,7 +12,27 @@ local function switch_source_header(bufnr)
   end)
 end
 
+-- https://clangd.llvm.org/extensions.html#file-status
+local function file_status_update(_, _, message, client_id)
+  local client = vim.lsp.get_client_by_id(client_id)
+  if not client then
+    vim.api.nvim_err_writeln("LSP[" .. client_id .. "] client has shut down after sending the message")
+    vim.api.nvim_command("redraw")
```
mjlbach, line 20: Why is there a redraw here?

mjlbach (commented):
Capabilities are already extended, not replaced. I'm hesitant to add this since the progress stuff isn't stabilized, and it seems like clangd is using a non-standard way of reporting progress.

jdrouhard:
Capabilities are not extended if there's already a capabilities defined in the default  when the `utf8_config` helper is used. See util.lua, line 431. Maybe the proper fix is to extend there too so we don't need this workaround.

And, normally the idea of these specific language server configs are specifically to implement extensions and extra features the server provides. The clangd config already implements the nonstandard switchSourceHeader function, this PR simply adds support for the fileStatus extension.

I believe you ported the lsp-status from the plugin to core, right? I saw you added the "status" member to the `messages` dict but never used it. In the original plugin, that `status` was only used in the extensions like clangd's fileStatus. Maybe a different autocmd would be in order?

lua/lspconfig/clangd.lua
jdrouhard: See other comment. It was a bug with only clangd.

lua/lspconfig/clangd.lua
jdrouhard: This is how `err_message` is defined in handlers.lua in core.

mjlbach:
Teto implemented initial status reporting, I just fixed added support for the workDoneProgress/create tokens. I'm cool with adding the status reporting.

Thank you for pointing that out, I think we should fix the issue with utf-8 configs. I almost lean towards getting rid of the wrapper as it's currently only used for clangd, and just directly add the necessary 
```
  on_init(client, result)
    if result.offsetEncoding then
      client.offset_encoding = result.offsetEncoding
    end
```

jdrouhard:
That sounds good to me. I went ahead and removed the `utf8_config` helper and moved its functionality into clangd.

I also changed the auto cmd for status updates, since it's not really a *progress* update, but I'm open to suggestions on whether we leave it as `LspProgressUpdate` or make a new `LspStatusUpdate`.

CONFIG.md
```diff
@@ -622,11 +622,13 @@ require'lspconfig'.clangd.setup{}
   - ClangdSwitchSourceHeader: Switch between source/header
   
   Default Values:
-    capabilities = default capabilities, with offsetEncoding utf-8
     cmd = { "clangd", "--background-index" }
     filetypes = { "c", "cpp", "objc", "objcpp" }
-    on_init = function to handle changing offsetEncoding
     root_dir = root_pattern("compile_commands.json", "compile_flags.txt", ".git") or dirname
+    on_init = function to handle changing offsetEncoding
```
mjlbach, line 628: Just a reminder not to edit CONFIG.md, because it's autogenerated so it just adds a commit that will be wiped out.

jdrouhard: I was wondering about that...thanks.

jdrouhard:
@mjlbach @teto 

I've been using this a bit more and I'd like to make a change to core to better support the file status extension. Currently, `lsp.util.get_progress_messages()` has the following:

```lua
if next(data.status) ~= nil then
  table.insert(new_messages, {
    name = data.name,
    content = data.status.content,
    uri = data.status.uri,
    status = true
  })
end
```
However, coc/vscode both keep a *map* of URIs to the "content" since we don't tell language servers when we're switching buffers. We actually need to keep a map of URI to the most recent status that clangd gave us, and then return them all in `get_progress_messages()`, or maybe optionally pass a bufnr or URI to that function and return the appropriate status message.

What do you guys think about changing core a bit to better facilitate this?

mjlbach:
I think I'm pro-this, however, I need to dig in more to have opinions on implementation. We could store status by message by URI, or by attached server, or by workspace. Some types of status are per server (indexing) which would be handled by server or workspace indexed statuses, whereas you're saying some may be per buffer?

teto:
does this clangd extension depends on host features ? like kernel config ?

As for the progress report, since there are people using several servers for a single filetype, I can imagine people wanting per-buffer progress reports. I find this overkill but best to make it possible before 0.5 release. We should do as for the diagnostics, aka merge all of them I think (and provide utilities to have them per-server).

jdrouhard:
> 
> does this clangd extension depends on host features ? like kernel config ?
> 
I don't believe so. It just hooks into clangd's per-file worker process and causes the server to send a message when the status changes for that file. It's extremely useful to know when the AST is done being generated for the buffer you have open (especially for translation units that take a bit longer to process).

> As for the progress report, since there are people using several servers for a single filetype, I can imagine people wanting per-buffer progress reports. I find this overkill but best to make it possible before 0.5 release. We should do as for the diagnostics, aka merge all of them I think (and provide utilities to have them per-server).

I agree. Currently, progress messaging is stored on the client object. I think we can also keep per-client/per-buffer status messages, which would mean a status map of URI->status kept on the client object. I will work on a PR for core there, and then update this accordingly. Would that work?

Using my local edits of core and this PR, I've managed to get a status line that gives all the features the coc-clangd extension for coc.nvim gives, which is really great.

jdrouhard:
Closing for now, until we have a more generic approach to handling per-server "status" messaging. Core will need some work. I'm going to reopen a new PR that just fixes the capabilities bug.

mjlbach:
Sounds good! Looking forward to reviewing :)

PR #519, opened 2021-01-03
Remove rust analyzer workaround

mjlbach:
As of the latest nightly, this seems no longer necessary

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -2962,7 +2962,7 @@ require'lspconfig'.omnisharp.setup{}
   Commands:
   
   Default Values:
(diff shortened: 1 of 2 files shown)


leiserfg:
Please if this is working fine with the current nightly, merge as the current state of master misbehaves because of the `prepareSupportDefaultBehavior = 1` cause `prepareSupportDefaultBehavior` is a boolean.

mjlbach:
Want to check and confirm? 

Also, sorry about that. I merged the fix from CoC but didn't check the spec first.

leiserfg:
I used your branch and it's working fine.

PR #520, opened 2021-01-03
Update sumneko lua docs

mjlbach:
cc: @clason

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -4704,9 +4704,23 @@ Lua language server.
 **By default, lua-language-server doesn't have a `cmd` set.** This is because nvim-lspconfig does not make assumptions about your path. You must add the following to your init.vim or init.lua to set `cmd` to the absolute path ($HOME and ~ are not expanded) of you unzipped and compiled lua-language-server.
 
 ```lua
-local sumneko_root_path = "/path/to/lua-language-server"
+local system_name
+if vim.fn.has("mac") == 1 then
+  system_name = "macOS"
+elseif vim.fn.has("unix") == 1 then
+  system_name = "Linux"
+elseif vim.fn.has('win32') == 1 then
+  system_name = "Windows"
+else
+  print("Unsupported system for sumneko")
+end
+
+-- set the path to the sumneko installation; if you previously installed via the now deprecated :LspInstall, use
+local sumneko_root_path = vim.fn.stdpath('cache')..'/lspconfig/sumneko_lua/lua-language-server'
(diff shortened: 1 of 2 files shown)


clason:
Thanks! Maybe we could add for now a notice that if sumneko was previously installed via `:LspInstall`, you should set `sumneko_root_path = vim.fn.stdpath('cache')..'/lspconfig/sumneko_lua/lua-language-server'`? 

(make it as frictionless as possible to migrate for people whose cheese has just been moved ;))

mjlbach:
Added, let me know if this looks good to you and I'll merge.

clason:
Close :) You're using `sumneko_root_path` without defining it first (and the comment about it is after the use -- and uses the wrong variable ;)).

How about 
```lua
 -- set the path to the sumneko installation; if you previously installed via the now deprecated :LspInstall, use
local sumneko_root_path = vim.fn.stdpath('cache')..'/lspconfig/sumneko_lua/lua-language-server'
local sumneko_binary ="/path/to/lua-language-server"..sumneko_root_path.."/bin/"..system_name.."/lua-language-server"
 ```

**Comment by mjlbach** [comment]
Just to clarify, this isn't the sumneko_root_path, this is the binary
`vim.fn.stdpath('cache')..'/lspconfig/sumneko_lua/lua-language-server'`

This is an alternative path to the binary if you install somewhere else
`"/path/to/lua-language-server"..sumneko_root_path.."/bin/"..system_name.."/lua-language-server"`

File: CONFIG.md [hunk 1]
```diff
@@ -4704,9 +4704,24 @@ Lua language server.
...
+  system_name = "macOS"
+elseif vim.fn.has("unix") == 1 then
+  system_name = "Linux"
+elseif vim.fn.has('win32') == 1 then
+  system_name = "Windows"
+else
+  print("Unsupported system for sumneko")
+end
+
+-- If you previously installed via the now deprecated LspInstall, you can set
```
clason, line 4718: obsolete comment line?

mjlbach: Yep, still pushing

clason: will shut up then :)

mjlbach: No please don't! Your commentary is valuable :)

mjlbach: Should be up to date now

clason: Looks good now! (I have no idea if the Windows check is sufficient -- there's also `win64` and `wsl` -- and whether sumneko even supports BSD, so let's wait for complaints there ;))

mjlbach: The win32 should cover win64, as for the bsds... they can handle modifying a couple lines of lua :P

PR #523, opened 2021-01-03
Fix minor typo in sumneko_lua instructions

akinsho:
Just a small typo fix

PR #527, opened 2021-01-04
fix: typo

yamatsum:


mjlbach:
This was already fixed in https://github.com/neovim/nvim-lspconfig/pull/523 but docgen failed. Thanks for noticing! It will be fixed next time a PR is merged and docgen regenerates the config (we don't edit this file manually for future reference).

PR #529, opened 2021-01-04
Fix: typo in lua config

cjjlaidler:
Fix for minor typo in lua config


mjlbach:
See: https://github.com/neovim/nvim-lspconfig/pull/523

I just merged before docgen was finished. We don't ever edit CONFIG.md, it's auto-generated.

PR #534, opened 2021-01-04
Fix sumneko_lua docs

yujinyuz:


mjlbach:
CONFIG.md is autogenerated, I had *thought* I merged the fix in #523, but I overlooked that they did the same thing as you and editing CONFIG.md directly. I just retriggered docgen and it's fixed on master.

yujinyuz:
Thanks! I hadn't realized that it was auto generated

PR #538, opened 2021-01-04
Added some advice to easily install lsp's

Norlock:
Might be useful for users.


mjlbach:
This seems like an unnecessary layer of indirection to me, as you're using neovim, to call lua, to call shell commands. You could either use lua to directly call that file, or (better yet) call the shell commands directly as we recommend in the documentation. This idea is good though, and I think belongs in it's own plugin.

PR #541, opened 2021-01-05
Update CONFIG.md

zhuangya:


mjlbach:
Thanks, but this is autogenerated, you'll need to modify the documentation in the respective luafile, and docgen will update config.md after you file the PR.

PR #547, opened 2021-01-07
Update julia cmd and documentation

### LINKED ISSUE #530: julials: cmd invalid because util.base_install_dir = nil
Latest nightly nvim, latest nvim-lspconfig, did something change in `util.base_install_dir`? When I
```
:lua print(require'lspconfig'.util.base_install_dir)
```
I get `nil`. It was not nil before... Where should `LanguageServer` go? This has made my `julia-ls` not functioning, for some reason, I get
```
[ ERROR ] 2021-01-04T09:10:29+0000 ] /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:455 ]	"rpc"	"julia"	"stderr"	'ArgumentError: Package LanguageServer not found in current path:\n- Run `import Pkg; Pkg.add("LanguageServer")` to install the LanguageServer package.\n\n'
```
The result of `health#lspconfig#check` is
```
  - INFO: julials: configuration checked.
  - INFO: r_language_server: configuration checked.
  - INFO: texlab: configuration checked.
  - INFO: sumneko_lua: configuration checked.
  - INFO: jsonls: configuration checked.
```
The other lsp work, it is just julia that is problematic, but then again it is only julia I had to do an extra step for. The other lsp are installed globally in my system.

mjlbach:
Closes #530 

Julia fans (@coxackie  @martinzh @quantum-booty  @kdheepak ).

Please let me know if this documentation change is ok. It fixes julia "just working" on master, at the cost of a *tiny* bit of overhead on first launch of a julia buffer (100ms on my fairly old laptop)

I had to remove the following from the config to get julia to provide completions for the couple environments I tried it out on (one with flux, one with plots, there was no cross contamination of packages)

```julia
          empty!(LOAD_PATH)
          push!(LOAD_PATH, "@")
```

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -1820,14 +1820,49 @@ require'lspconfig'.jsonls.setup{}
 ## julials
 
 https://github.com/julia-vscode/julia-vscode
+
 `LanguageServer.jl` can be installed with `julia` and `Pkg`:
 ```sh
-julia --project=julials -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
+julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
 ```
-If you want to install the LanguageServer manually, you will have to ensure that the Julia environment is stored in this location:
-```vim
-:lua print(require'lspconfig'.util.path.join(require'lspconfig'.util.base_install_dir, "julials"))
+The default config lazily evaluates the location of the julia language server from the your global julia packages.
+This adds a small overhead on first opening of a julia file. To avoid this overhead, replace server_path in on_new_config with
+a hard-coded path to the server.
+
+```lua
+require'lspconfig'.julials.setup{
+    on_new_config = function(new_config,new_root_dir)
+      server_path = "/path/to/directory/containing/LanguageServer.jl/src"
+      cmd = {
+        "julia",
+        "--project="..server_path,
+        "--startup-file=no",
+        "--history-file=no",
+        "-e", [[
+          using Pkg;
+          Pkg.instantiate()
+          using LanguageServer; using SymbolServer;
+          depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
+          project_path = dirname(something(Base.current_project(pwd()), Base.load_path_expand(LOAD_PATH[2])))
+          # Make sure that we only load packages from this environment specifically.
+          @info "Running language server" env=Base.load_path()[1] pwd() project_path depot_path
+          server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path);
+          server.runlinter = true;
+          run(server);
+        \]\]
+    };
+      new_config.cmd = cmd
+    end
+}
 ```
+You can find the path to the globally installed LanguageServer.jl package with the following command:
+
+```bash
+julia -e 'print(Base.find_package("LanguageServer"))'
+```
+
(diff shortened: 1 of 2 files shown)


mjlbach:
I'm going to go ahead and merge this, since julials is currently broken for users, but feel free to open an issue if this causes any problems.

kdheepak:
Thanks for the timely fix! I didn't know this was a problem. I'll try this updated configuration on my machine and will open an issue if anything comes up!

mjlbach:
So the           

```
empty!(LOAD_PATH)
push!(LOAD_PATH, "@")
```

Wasn't the main problem (although it didn't work for me before I removed this, since it was literally pushing the "@" and not the expansion), it was more than I forgot to update the build_directory which was pointed at the now excised local install feature. I tested this fairly extensively so felt comfortable enough merging, but let me know how it goes :)

martinzh:
Thanks a lot @mjlbach !! It's now working great for me. We the julia fans really appreciate it.

CONFIG.md
```diff
@@ -1820,14 +1820,49 @@ require'lspconfig'.jsonls.setup{}
 ## julials
 
 https://github.com/julia-vscode/julia-vscode
+
 `LanguageServer.jl` can be installed with `julia` and `Pkg`:
 ```sh
-julia --project=julials -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
+julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
```
kdheepak, line 1826: If you do this, then you can remove the `"--project=" .. server_path`. Since the above line installed `LanguageServer` and `SymbolServer` in the global environment, Julia will be able to run `using LanguageServer` fine without starting Julia with the `--project` flag.

When the custom server_path is used, you'll have to use a `--project` flag. However it should not point to `src`, but instead it should point to a folder with a Project.toml that loads `LanguageServer.jl`. And doesn't have to be the `LanguageServer.jl` repository. The way you've set it up, when you point it to `src`, it loads `LanguageServer.jl` and `SymbolServer.jl` from the global environment, and ignores the `--project=path` flag.

Any folder with a Project.toml (and optionally a Manifest.toml) file is a Julia environment. Previously, the way I set it up, I used a new custom folder for the Julia environment where we install `LanguageServer.jl` and `SymbolServer.jl` so that `LanguageServer.jl` doesn't have to installed in the global environment. Your `server_path` example is great for this, so I'll use that going forward. The `--project` flag must point to a folder with a `Project.toml` though, for this to work as the user intends.

I'm testing it locally, and I'm running into other issues with this set up. Once I figure out what is going on I'll write it up in more detail here. I think the issue I'm having right now is unrelated to this configuration (I'm using Julia nightly and I'm not sure is `LanguageServer.jl` works with that), but I'm trying to figure that out at this point.

mjlbach: So, my thoughts on this were to give an explicit example that used whatever the currently active julia environment is, such that if someone . I realize it's global, and did this mostly to align with the other servers that use npm/cargo because it's simpler for people debug. Totally open to changing this back! I think more people would complain if we defaulted to per project with "why isn't my language server installed/launching". It's easy enough to override, but let me know what you think. I'll merge whatever PR you submit with what you believe the most amenable configuration is for JLS users.

Looking into this, I think the issues are on JLS's side regarding the crash (we're not doing anything special on save with newlines as far as I can tell. It's definitely not something with the language client.

PR #517, opened 2021-01-02
Issue template improvements

mjlbach:
@teto What are your thoughts on including a minimal init.lua?

Issue #493, opened 2020-12-26
Using the suggested mappings overrides `c-]` which breaks navigation in the help system


yatesco:
The `:help lsp` suggests `nnoremap <silent> <c-]> <cmd>lua vim.lsp.buf.definition()<CR>`, which overrides the default key binding for navigating across a tag. 

This means that <c-]> has no effect when viewing help files because the help system doesn't use an LSP server.

What's the idiomatic way to solve this (this is a little out of my depth)? I thought of a function which checks to see if an LSP is active and if not calls the "native" `c-t`, and binding that function to `c-t` globally, but I might be talking nonsense. 

I could also only apply the mapping to non-help files but I can't figure out how to do that - I certainly don't want to add a specific instruction in each `ftplugin` file.

Help :-)

mjlbach:
To be honest, I prefer gd  `vim.api.nvim_buf_set_keymap(bufnr, 'n', 'gd', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)`

If you want to *only* bind LSP keybindings when servers are available for the filetype, you can see my config.

https://github.com/mjlbach/nix-dotfiles/blob/49a6c89b455a79669c17f17d471b9922fd193b65/nixpkgs/configs/neovim/init.vim#L311-L359

It would be possible to fallback to c-[ if the language server is not running. If you still want that I can come up with a snippet.

yatesco:
Your `init.vim` was most instructive, but a snippet to check whether a language server is running might be useful too - thanks!

mjlbach:
You can use `vim.lsp.buf_get_clients()`. I think the way I have it in my configuration is the best way to selectively bind keybinds only if a language server is set up for that buffer (that I know of)

lithammer:
You can also do conditional mapping by checking the `client.resolved_capabilities` table in the `on_attach` callback.

```lua
local lspconfig = require("lspconfig")

local function on_attach(client, bufnr)
  local buf_set_keymap = function(...)
    vim.api.nvim_buf_set_keymap(bufnr, ...)
  end

  local opts = {noremap = true, silent = true}

  if client.resolved_capabilities.document_formatting then
    buf_set_keymap(
      "n", "<Leader>f", "<cmd>lua vim.lsp.buf.formatting()<CR>", opts
    )
  end
end

lspconfig.foo.setup({
  on_attach = on_attach
})
```

mjlbach:
[...]

PR #475, opened 2020-12-21
lspconfig: allow accessing nil config with pcall

mjlbach:
Closes #191

The following example from the readme should now "just work"
```lua
local lspconfig = require'lspconfig'
local configs = require'lspconfig/configs'
-- Check if it's already defined for when I reload this file.
if not lspconfig.foo_lsp then
  configs.foo_lsp = {
    default_config = {
      cmd = {'/home/ashkan/works/3rd/lua-language-server/run.sh'};
      filetypes = {'lua'};
      root_dir = function(fname)
        return lspconfig.util.find_git_ancestor(fname) or vim.loop.os_homedir()
      end;
      settings = {};
    };
  }
end
lspconfig.foo_lsp.setup{}
```

PR #549, opened 2021-01-07
Update README.md

mjlbach:
I keep getting feedback that it is hard to use neovim's built-in language server (and nvim-lspconfig). I hope that these documentation changes make it more clear.

Warning, typos and dead links ahead. Please help me find them! And let me know where there are regressions. I did cut some things, and reorganized the repo a bit. I'll add a table of contents too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,76 +1,52 @@
 # nvim-lspconfig
 
-Collection of common configurations for the [Nvim LSP client](https://neovim.io/doc/user/lsp.html).
+Collection of common configurations for Neovim's built-in [language server client](https://neovim.io/doc/user/lsp.html).
+The configurations are supported on a best-effort basis, and rely on contributions 
+from regular users to stay up-to-date.
 
-## Best-effort, unsupported
-
-The configurations here are strictly **best effort and unsupported**.
-
-This repo is (1) a place for Nvim LSP users to collaboratively provide starting
-examples for the many LSP backends out there, and (2) a reference for the
-current best practice (or most popular) regarding choice of server, setup, etc.
-
-## Requires Nvim 0.5 HEAD
-
-While Nvim LSP undergoes development, the configs in this repo assume that you
-are using the latest [Nvim HEAD/nightly build](https://github.com/neovim/neovim/releases/tag/nightly).
-
-Update Nvim and nvim-lspconfig before reporting an issue.
-
-## Contributions
-
-It's up to you to send improvements so that these configs align with current
-best practices for a given language.
-
-1. Read [CONTRIBUTING.md](CONTRIBUTING.md).
-   Ask questions in [Neovim Gitter](https://gitter.im/neovim/neovim).
-2. Choose a language from [the coc.nvim wiki](https://github.com/neoclide/coc.nvim/wiki/Language-servers) or
-  [emacs-lsp](https://github.com/emacs-lsp/lsp-mode#supported-languages).
-3. Create a new file at `lua/lspconfig/SERVER_NAME.lua`.
-   - Copy an [existing config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/)
-     to get started. Most configs are simple. For an extensive example see
-     [texlab.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/texlab.lua).
+This repo handles automatically launching, initializing, and configuring language servers
+that are installed on your system.
 
 ## Install
 
-- Requires [Nvim HEAD/nightly](https://github.com/neovim/neovim/releases/tag/nightly) (v0.5 prerelease).
-- nvim-lspconfig is just a plugin. Install it like any other Vim plugin, e.g. with [vim-plug](https://github.com/junegunn/vim-plug):
+* Requires [Neovim HEAD/nightly](https://github.com/neovim/neovim/releases/tag/nightly) (v0.5 prerelease). The configs in this repo
+  assume that you are using the latest [Neovim HEAD/nightly build](https://github.com/neovim/neovim/releases/tag/nightly). 
+  Update Neovim and nvim-lspconfig before reporting an issue.
+
+* nvim-lspconfig is just a plugin. Install it like any other Vim plugin, e.g. with [vim-plug](https://github.com/junegunn/vim-plug):
   ```
   :Plug 'neovim/nvim-lspconfig'
   ```
-- Call `:packadd nvim-lspconfig` in your vimrc if you installed nvim-lspconfig to `'packpath'` or if you use a package manager such as minpac.
+## Quickstart
+1. Install a language server, e.g. [pyright](CONFIG.md#pyright) via npm -i -g pyright" `npm i -g pyright`
+2. Install `nvim-lspconfig` via your plugin manager
+3. Add the language server setup to your init.vim. The server name must match those found in the table of contents in [CONFIG.md](CONFIG.md)
+```vim
+lua << EOF 
+require'lspconfig'.pyright.setup{}
+EOF 
+```
+4. Open a file that is placed in a directory recognized by the server 
+(see server configuration in [CONFIG.md](CONFIG.md); e.g., for [pyright](CONFIG.md#pyright), this is 
+any directory containing ".git", "setup.py",  "setup.cfg", "pyproject.toml",
+or "requirements.txt")
+5. See [Keybindings and completion](#Keybindings-and-completion) for mapping useful functions and enabling
+omnifunc completion
 
 ## Usage
 
-Each config provides a `setup()` function, to initialize the server with
-reasonable defaults and some server-specific things like commands or different
-diagnostics.
+**All provided examples are in Lua,** see `:help :lua-heredoc` to use Lua from your init.vim,
+or the quickstart above for an example of a lua heredoc.
 
+Each config provides a `setup()` function to initialize the server with reasonable default
+initialization options and settings, as well as some server-specific commands. This is 
+invoked in the following form, where `<server>` corresponds to the language server name
+in [CONFIG.md](CONFIG.md).
+
+`setup()` takes optional arguments <config>, each of which overrides the respective 
+part of the default configuration. The allowed arguments are detailed [below](#setup-function).
 ```lua
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
-require'lspconfig'.<config>.setup{name=…, settings = {…}, …}
-```
-
-If you want to add this to your vimrc, you will need to enclose it in a `lua` block.
-
-```vim
-lua <<EOF
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
-require'lspconfig'.<config>.setup{name=…, settings = {…}, …}
-EOF
-```
-
-Find the [config](#configurations) for your language, then paste the example
-given there to your `init.vim`. **All examples are given in Lua,** see `:help
-:lua-heredoc` to use Lua from your init.vim.
-
-Some configs may define additional server-specific functions, e.g. the `texlab`
-config provides `lspconfig.texlab.buf_build({bufnr})`.
-
-If you want to see the location of log file, you can run this in neovim:
-
-```
-:lua print(vim.lsp.get_log_path())
+require'lspconfig'.<server>.setup{<config>}
 ```
 
 ### Example: using the defaults
@@ -79,7 +55,6 @@ To use the defaults, just call `setup()` with an empty `config` parameter.
 For the `gopls` config, that would be:
 
 ```lua
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
 require'lspconfig'.gopls.setup{}
 ```
 
@@ -95,8 +70,9 @@ lspconfig.gopls.setup{
 }
 ```
 
-The [documentation](#configurations) for each config lists default values and
-additional optional properties.
+The [documentation](CONFIG.md) for each config lists default values and
+additional optional properties. For a more complicated example overriding 
+the `name`, `log_level`, `message_level`, and `settings` of texlab:
 
 ```lua
 local lspconfig = require'lspconfig'
@@ -116,32 +92,28 @@ lspconfig.texlab.setup{
 
 ### Example: custom config
 
-To configure a custom/private server, just require `lspconfig/configs` and do
-the same as we do if we were adding it to the repository itself.
+To configure a custom/private server, just 
 
-1. Define the config: `configs.foo_lsp = { … }`
-2. Call `setup()`: `require'lspconfig'.foo_lsp.setup{}`
+1. load the lspconfig module: `local lspconfig = require('lspconfig')`,
+2. Define the config: `lspconfig.configs.foo_lsp = { … }`
+3. Call `setup()`: `lspconfig.foo_lsp.setup{}`
 
 ```lua
 local lspconfig = require'lspconfig'
-local configs = require'lspconfig/configs'
--- Check if it's already defined for when I reload this file.
-if not lspconfig.foo_lsp then
-  configs.foo_lsp = {
-    default_config = {
-      cmd = {'/home/ashkan/works/3rd/lua-language-server/run.sh'};
-      filetypes = {'lua'};
-      root_dir = function(fname)
-        return lspconfig.util.find_git_ancestor(fname) or vim.loop.os_homedir()
-      end;
-      settings = {};
-    };
-  }
-end
+lspconfig.configs.foo_lsp = {
+  default_config = {
+    cmd = {'/home/ashkan/works/3rd/lua-language-server/run.sh'};
+    filetypes = {'lua'};
+    root_dir = function(fname)
+      return lspconfig.util.find_git_ancestor(fname) or vim.loop.os_homedir()
+    end;
+    settings = {};
+  };
+}
 lspconfig.foo_lsp.setup{}
 ```
 
-### Example: override default config
+### Example: override default config for all servers
 
 If you want to change default configs for all servers, you can override default_config like this.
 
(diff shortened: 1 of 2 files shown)


mjlbach:
@elianiva You spotted a lot of typos on my issue_template so if you have the time :)

README.md
```diff
@@ -1,81 +1,69 @@
...
-current best practice (or most popular) regarding choice of server, setup, etc.
-
-## Requires Nvim 0.5 HEAD
-
-While Nvim LSP undergoes development, the configs in this repo assume that you
-are using the latest [Nvim HEAD/nightly build](https://github.com/neovim/neovim/releases/tag/nightly).
+This repo is :
 
-Update Nvim and nvim-lspconfig before reporting an issue.
+* a place for Nvim LSP users to collaboratively provide starting
```
elianiva, line 8: I think `a place for Nvim LSP client users` sounds better, or maybe `a place for Nvim builtin LSP client users` ?

mjlbach: This is always something we've been unclear about. We should make it clear that neovim implements a language server client, in accordance with language server protocol, via which it communicates with language servers.

README.md
```diff
@@ -1,81 +1,69 @@
...
-```lua
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
-require'lspconfig'.<config>.setup{name=…, settings = {…}, …}
-```
 
-If you want to add this to your vimrc, you will need to enclose it in a `lua` block.
+Note: if you installed nvim-lspconfig to `'packpath'` you need to call 
+`:packadd nvim-lspconfig` in your vimrc.  This is the case for only 
+some package managers, such as minpac.  You do not need to do this 
+step with vim-plug or packer.
```
elianiva, line 30: You also need to do `packadd nvim-lspconfig` if you specify `opt = true` for packer.nvim.

mjlbach: That's a really good point, will fix

README.md
```diff
@@ -1,81 +1,69 @@
...
+or the source file for each server to check the root trigger.
 
-Find the [config](#configurations) for your language, then paste the example
-given there to your `init.vim`. **All examples are given in Lua,** see `:help
-:lua-heredoc` to use Lua from your init.vim.
+## Usage
 
-Some configs may define additional server-specific functions, e.g. the `texlab`
-config provides `lspconfig.texlab.buf_build({bufnr})`.
+**All examples are given in Lua,** see `:help :lua-heredoc` to use Lua from your init.vim,
```
elianiva, line 48: I think this should be `All examples given are in Lua`.

mjlbach: 'All provided examples`?

mjlbach:
Also, maybe we should add a "keymaps" section, to cover mapping.

mjlbach:
And a brief note about omnifunc completion

README.md
elianiva: That sounds better to me.

elianiva:
> Also, maybe we should add a "keymaps" section, to cover mapping.

Agree, that would be a nice section to have in my opinion

mjlbach:
Ah great, I accidentally just pushed over some old changes. I'll try to touch this up and add what's missing.

README.md
```diff
@@ -154,9 +140,52 @@ lspconfig.util.default_config = vim.tbl_extend(
...
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>wr', '<cmd>lua vim.lsp.buf.remove_workspace_folder()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>wl', '<cmd>lua print(vim.inspect(vim.lsp.buf.list_workspace_folders()))<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>D', '<cmd>lua vim.lsp.buf.type_definition()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>rn', '<cmd>lua vim.lsp.buf.rename()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', 'gr', '<cmd>lua vim.lsp.buf.references()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>e', '<cmd>lua vim.lsp.diagnostic.show_line_diagnostics()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '[d', '<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', ']d', '<cmd>lua vim.lsp.diagnostic.goto_next()<CR>', opts)
+  vim.api.nvim_buf_set_keymap(bufnr, 'n', '<leader>q', '<cmd>lua vim.lsp.diagnostic.set_loclist()<CR>', opts)
+end
```
lithammer, line 178: Same comment here ^

https://github.com/neovim/nvim-lspconfig/pull/517#discussion_r553183906

mjlbach: Yeah, I pushed before I saw yours. Do you think this is a good idea to put in the readme? The biggest issue people are saying is that it's hard to configure, so I imagine having some sort of "sane defaults" mentioned somewhere would be good, instead of being buried in issues.

lithammer: I don't see how sane defaults and making the examples terser by reducing some redundancy are related? I think we should provide nice examples since this is going to be copied _a lot_. But then again, "nice" is in the eye of the beholder.

lithammer: I'm fine with skipping this though to keep the example free of too many abstractions/indirections.

mjlbach: Oh no no, that's not what I meant haha. I like your change. I meant what *I* included, independently of using the helper function.

mjlbach: Sorry for the miscommunication.

lithammer: > Oh no no, that's not what I meant haha. I like your change. I meant what _I_ included, independently of using the helper function.

Oooh! I thought you were referring to my helper function. Not the examples themselves. Anyway, I think it's a good idea to include some sample configuration including key mappings for the user. I assume coc.nvim and vim-lsp added it to their readme for similar reasons.

> Sorry for the miscommunication.

No worries! 🙂

README.md
```diff
@@ -154,9 +140,55 @@ lspconfig.util.default_config = vim.tbl_extend(
...
+  buf_set_keymap('n', '<space>wa', '<cmd>lua vim.lsp.buf.add_workspace_folder()<CR>', opts)
+  buf_set_keymap('n', '<space>wr', '<cmd>lua vim.lsp.buf.remove_workspace_folder()<CR>', opts)
+  buf_set_keymap('n', '<space>wl', '<cmd>lua print(vim.inspect(vim.lsp.buf.list_workspace_folders()))<CR>', opts)
+  buf_set_keymap('n', '<space>D', '<cmd>lua vim.lsp.buf.type_definition()<CR>', opts)
+  buf_set_keymap('n', '<space>rn', '<cmd>lua vim.lsp.buf.rename()<CR>', opts)
+  buf_set_keymap('n', 'gr', '<cmd>lua vim.lsp.buf.references()<CR>', opts)
+  buf_set_keymap('n', '<space>e', '<cmd>lua vim.lsp.diagnostic.show_line_diagnostics()<CR>', opts)
+  buf_set_keymap('n', '[d', '<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>', opts)
+  buf_set_keymap('n', ']d', '<cmd>lua vim.lsp.diagnostic.goto_next()<CR>', opts)
+  buf_set_keymap('n', '<space>q', '<cmd>lua vim.lsp.diagnostic.set_loclist()<CR>', opts)
```
lithammer, line 181: Maybe it would be a good idea to add an example of how to make a conditional keymapping depending on the resolved capabilities (ref GH-493)?

```lua
if client.resolved_capabilities.document_formatting then
  buf_set_keymap("n", "<Leader>f", "<cmd>lua vim.lsp.buf.formatting()<CR>", opts)
end
```

And document highlighting:

```lua
if client.resolved_capabilities.document_highlight then
  util.nvim_multiline_command [[
    augroup lsp_document_highlight
      autocmd!
      autocmd CursorHold <buffer> lua vim.lsp.buf.document_highlight()
      autocmd CursorMoved <buffer> lua vim.lsp.buf.clear_references()
    augroup END
  ]]
end
```

mjlbach: omg why is nvim_multiline_command not in core.

elianiva: what? I didn't know we have `nvim_multiline_command`, that would be so useful 😆

mjlbach: It's only in `lspconfig`, yet another reason to prefer `lspconfig` over CoC (just kidding)

mjlbach: Also this reminds me to port the function signature of this version of on_attach to core.

lithammer: > omg why is nvim_multiline_command not in core.

Good question. Writing an `augroup` using `vim.cmd` isn't super pretty.

mjlbach: What servers actually support document highlight? I can't test it.

lithammer: > What servers actually support document highlight? I can't test it.

Pyright and bash-language-server at least. Note that `CursorHold` can be pretty slow depending on your configuration since the default for `'updatetime'` is `4000`.

mjlbach: Ah, then we may want to add notes about these (mine doesn't seem to, onedark)

clason: https://github.com/neovim/neovim/pull/13614

mjlbach: Thanks! When merged I can add a reference

mjlbach:
Do people feel this is more readable than the old readme? Strikes a good balance in terms of length? I had several non vim friends try to follow this, and I addressed the major stumbling blogs, with the exception that some of them had issues with vim-plug (apart from the plugin).

README.md
```diff
@@ -1,76 +1,61 @@
...
-diagnostics.
-
-```lua
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
-require'lspconfig'.<config>.setup{name=…, settings = {…}, …}
-```
 
-If you want to add this to your vimrc, you will need to enclose it in a `lua` block.
+Note: if you installed nvim-lspconfig to `'packpath'` you need to call 
+`:packadd nvim-lspconfig` in your vimrc.  This is the case for only
```
clason, line 25: This is actually no longer true (unless you put it in `pack/*/opt`, where this is intended).

README.md
```diff
@@ -1,76 +1,61 @@
 # nvim-lspconfig
 
-Collection of common configurations for the [Nvim LSP client](https://neovim.io/doc/user/lsp.html).
-
-## Best-effort, unsupported
-
+Collection of common configurations for Neovim's built-in [language server client](https://neovim.io/doc/user/lsp.html).
 The configurations here are strictly **best effort and unsupported**.
```
clason, line 4: "unsupported" by whom? Clearly they are supported by the contributors to this repository ;)

Maybe what is meant is "are supported on a best-effort basis", meaning that *server* configuration changes may not be picked up unless someone notices, reports, and suggest fixes for them?

README.md
```diff
@@ -1,76 +1,61 @@
...
-2. Choose a language from [the coc.nvim wiki](https://github.com/neoclide/coc.nvim/wiki/Language-servers) or
-  [emacs-lsp](https://github.com/emacs-lsp/lsp-mode#supported-languages).
-3. Create a new file at `lua/lspconfig/SERVER_NAME.lua`.
-   - Copy an [existing config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/)
-     to get started. Most configs are simple. For an extensive example see
-     [texlab.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/texlab.lua).
+* a place for users of Neovim's language server client to collaboratively provide starting 
+  examples for the many LSP backends out there
+* a reference for the current best practice (or most popular) regarding 
+  choice of server, setup, etc.
```
clason, line 11: I'd recommend staying away from "best choice" recommendations; rather focus on documenting configuration options for servers and hints for getting them installed and running (where not trivial).

README.md
```diff
@@ -1,76 +1,61 @@
...
-```
 
-If you want to add this to your vimrc, you will need to enclose it in a `lua` block.
+Note: if you installed nvim-lspconfig to `'packpath'` you need to call 
+`:packadd nvim-lspconfig` in your vimrc.  This is the case for only 
+some package managers, such as minpac and packer (in the case where `opt = true`).  You do not need to do this 
+step with vim-plug.
 
+## Quickstart
+1. Install a language server `npm i -g pyright`
```
clason, line 30: "...server, e.g. [pyright](link to pyright) via `npm -i -g pyright`"

README.md
```diff
@@ -1,76 +1,61 @@
...
 ```vim
-lua <<EOF
-vim.cmd('packadd nvim-lspconfig')  -- If installed as a Vim "package".
-require'lspconfig'.<config>.setup{name=…, settings = {…}, …}
-EOF
+lua << EOF 
+require'lspconfig'.pyright.setup{}
+EOF 
 ```
+4. Open a file that in a directory with a file that triggers the language server's root detection.
```
clason, line 38: "that is placed in a directory recognized by the server (see server configuration; e.g., for pyright, this is any directory...)

README.md
```diff
@@ -116,11 +102,11 @@ lspconfig.texlab.setup{
...
 ### Example: custom config
 
-To configure a custom/private server, just require `lspconfig/configs` and do
-the same as we do if we were adding it to the repository itself.
+To configure a custom/private server, just 
 
-1. Define the config: `configs.foo_lsp = { … }`
-2. Call `setup()`: `require'lspconfig'.foo_lsp.setup{}`
+1. require `lspconfig/configs`,
+2. Define the config: `configs.foo_lsp = { … }`
```
clason, line 108: Is that correct? Don't you need to have `local configs = require'lspconfig.configs` for this to work?

README.md
```diff
@@ -154,9 +140,75 @@ lspconfig.util.default_config = vim.tbl_extend(
 )
 ```
 
+## Individual server settings and initialization options
+
+See [CONFIG.md](CONFIG.md) for documentation and configuration of individual language servers.
+This document contains installation instructions for each language server, and is 
+auto-generated from the documentation in the lua source. Do not submit PRs 
+modifying CONFIG.md directly; CONFIG.md will be overwrriten by docgen
```
clason, line 148: "overwritten"

mjlbach: Fixed. Thanks!

README.md
mjlbach: I hadn't touched a lot of the original language, so this was good feedback. I've tweaked the places it occurs, but let me know if the changes didn't address the issue.

README.md
mjlbach: Changed, also streamlined the intro. The configurations are reasonable well supported now anyways.

README.md
mjlbach: I've never manually managed pack path. Are we safe to just delete this line?

README.md
```diff
@@ -231,6 +280,38 @@ lspconfig.SERVER.setup{config}
...
+to enable logging 
+
+```vim
+lua << EOF
+vim.lsp.set_log_level("debug")
+EOF
+```
+Attempt to run the language server, and view the log present at:
+```
+:lua print(vim.lsp.get_log_path())
```
clason, line 304: I can't find the magic invocation to open that file directly using the `get_log_path` function, but if anyone knows that, this might be better :)

README.md
```diff
@@ -231,6 +280,38 @@ lspconfig.SERVER.setup{config}
...
+```
+Most of the time, the reason for failure is present in the logs.
+
+## Contributions
+
```
clason, line 309: Maybe add explicitly "If you are missing a language server on the list [CONFIG.md], contributing a new configuration for it would be appreciated. You can follow these steps:"

README.md
clason: I think there are three possibilities here:
1. People use vim-plug or a similar plugin manager that works by managing rtp.
2. People use min-pac or packer that works with vim native packages.
3. People install plugins/packages by hand.

In case 1 and 2, the respective plugin/package manager should have ample documentation for adding plugins/packages. In case 3, the user knows what they are doing (hopefully) :)

The only issue would be if lspconfig did not behave like a normal plugin/package (which used to be the case for packages until a short while ago because Lua handling in packages had hiccups, but this has been fixed).

So, yes, I'd say it's safe to delete the note (especially since you -- sensibly -- give an example for vim-plug).

clason (commented):
Sorry, somehow I always end up in review mode and forget to submit pending comments 😠

README.md
mjlbach: Addressed, thanks

README.md
mjlbach: `v:lua.vim.lsp.get_log_path()` returns the log path, but I don't know how to open the returned string rather than a buffer of the function name itself 😆

README.md
mjlbach: Done!

README.md
mjlbach: Agreed. It's unlikely this is someone's *first* plugin, and I had 2 of the 4 non-vim users I had test out the readme get confused by this.

README.md
clason: I got it: `:lua vim.cmd('e'..vim.lsp.get_log_path())`

README.md
mjlbach: This whole functions is broken, see https://github.com/neovim/nvim-lspconfig/issues/191

I "fixed this" by silently ignoring the require error, but teto had some concerns. https://github.com/neovim/nvim-lspconfig/pull/475

README.md
mjlbach: Should we open in a split by default? vsp?

clason: I'd keep it simple...

(Alternatively, you could spell it out: `~/.local/share/nvim/lsp.log` under Linux/BSD/macOS, `~\AppData\Local\nvim-data\lsp.log` under Windows?)

README.md
clason: Yeah, I have the strong opinion that any code snippet in a README should be working when copied and pasted (up to necessary customization)...

README.md
mjlbach: We could wait until https://github.com/neovim/neovim/pull/13707 is merged I guess

mjlbach: Went ahead with`:lua vim.cmd('e'..vim.lsp.get_log_path())`

README.md
clason: And I don't think this should error.
```lua
local configs = require 'lspconfig.configs'
configs.texlab = {}
```
works fine for me?

mjlbach: That's true, the main thing that errors is testing for membership, because the metatable defines new index's to 'require' the module (which doesn't exist when defining your own server)

clason: (Meaning, you don't need the `if not ...` guard. If the config already exists, it is just overwritten.)

teto:
If you want I can enable the wiki too so that people can share their configs/tricks etc. It's also less of a headache to edit.
People will always complain, maybe some  tried nvim-lspconfig sometime ago and it was indeed hard. It's now settling down, it's better out of the box (not perfect by any means but still). Also it relies on *new* techs (neovim's lua,  lsp servers are faulty too, the spec keeps evolving). Once released there will be more doc on the web and it will eventually be easier for everyone.

mjlbach:
I was wondering if discussions might also be a good fit, we could pin a thread for each language server.

clason:
1. Isn't CONFIG.md (partly) generated automatically (the supported and default options)? That'd be hard to translate to a Wiki. If we do that (and there are good reasons for it), we'd have to have a clear split: config etc. in the documentation, installation instructions and "here's how to trick the project root detection" in the Wiki? But then you'd spread the information over more places, which should be avoided, all things being equal. I guess my opinion would be not to do that yet; wait until maintaining the documentation (via drive-by PRs if possible) becomes a hassle?
2. Here, too, is the question where you'd draw the line between issues and discussions -- which questions would you route to which? And how to deal with the inevitable wrongly-placed question?

mjlbach:
I deleted my earlier comment, but one idea is enabling GH discussions, but locking the number of topics so there is one per server, and disabling the notifications to org members by default.

clason:
> I deleted my earlier comment, but one idea is enabling GH discussions, but locking the number of topics so there is one per server, and disabling the notifications to org members by default.

And keep the config still in the documentation? So basically instead of Gitter, you'd point people with questions about installing the server and "it's not working when I do `nvim test.py`" to GH Discussions? That would indeed make sense and cut back on having to answer the same questions repeatedly.

mjlbach:
> And keep the config still in the documentation? 
Yes, I'm opposed to a wiki, because I think it would be a duplicate of CONFIG.md, and we already have README.md and CONFIG.md. 

Side note: I'm working on a automated vimdoc thing to make a`:help` for each language server, but that's going to take awhile.

> So basically instead of Gitter, you'd point people with questions about installing the server and "it's not working when I do nvim test.py" to GH Discussions? That would indeed make sense and cut back on having to answer the same questions repeatedly.

I think it's better for the gitter channel, and probably better for the user, as people are more easily going to read through a discussion thread than find the right combination of search terms in gitter to get to the old messages.

README.md
mjlbach: I just went ahead and removed it for now, I'll add it back in #191

README.md
```diff
@@ -154,9 +126,75 @@ lspconfig.util.default_config = vim.tbl_extend(
...
+  if client.resolved_capabilities.document_highlight then
+    require('lspconfig').util.nvim_multiline_command [[
+      :hi LspReferenceRead cterm=bold ctermbg=red guibg=LightYellow
+      :hi LspReferenceText cterm=bold ctermbg=red guibg=LightYellow
+      :hi LspReferenceWrite cterm=bold ctermbg=red guibg=LightYellow
+      augroup lsp_document_highlight
+        autocmd!
+        autocmd CursorHold <buffer> lua vim.lsp.buf.document_highlight()
+        autocmd CursorMoved <buffer> lua vim.lsp.buf.clear_references()
+      augroup END
```
lithammer, line 186: I noticed that using a `augroup` here probably isn't the best idea. It kind of works to begin with, but if you open a new buffer that starts a new server (i.e. runs `on_attach` again), it will stop working for the old buffer/server. At least that was what it looked like to me.

I suspect the "problem" is `<buffer>` (which we want to have).

David-Else:
@mjlbach Could you add an explanation of how to install without a plugin manager using `packadd`? It seems strange to not use the official built-in package manager over 3rd party solutions.

nvim treesitter says: https://github.com/nvim-treesitter/nvim-treesitter#installation : 
> You can install nvim-treesitter with your favorite package manager (or using the native package feature of vim, see :h packages).

and adds https://github.com/nvim-treesitter/nvim-treesitter#i-run-into-errors-like-module-nvim-treesitterconfigs-not-found-at-startup:
> I run into errors like module 'nvim-treesitter.configs' not found at startup

> This is because of rtp management in nvim, adding packadd nvim-treesitter should fix the issue.

clason:
@David-Else Just put `nvim-lspconfig` into `stdpath('data')/site/pack/*/start`, nothing else required; see `:h packages`. (The "rtp management issues" have been solved in the mean time.)

It would be equally strange to not recommend a working, well-tested, and maintained solution to a wide audience over manual steps that require at least some understanding of relatively delicate vim internals (startup and packages). :)

David-Else:
I am really new to NeoVIm and have no idea how to set up a plugin manager, or zero interest either if there is one built-in! Thanks for the info, it would be great to add that sentence to the instructions for other newbies :)

clason:
_Especially_ as a newbie, I would strongly recommend looking into a plugin manager like vim-plug (in my opinion the most robust and easiest to use one), as they also help with installing, updating, and otherwise managing plugins. (If you pick a plugin manager that uses packages internally, it's README will do a much better job of explaining the ins and outs and caveats of packages than a single sentence here does. No info is better than misleading info!)

mjlbach:
> @mjlbach Could you add an explanation of how to install without a plugin manager using packadd? It seems strange to not use the official built-in package manager over 3rd party solutions.

It's really funny for me to read this, because when I started using vim there wasn't a built-in plugin manager! Pathogen and later vim-plug were the defacto standards.

>and adds https://github.com/nvim-treesitter/nvim-treesitter#i-run-into-errors-like-module-nvim-treesitterconfigs-not-found-at-startup:

    I run into errors like module 'nvim-treesitter.configs' not found at startup

    This is because of rtp management in nvim, adding packadd nvim-treesitter should fix the issue.

We actually had something similair, although less aptly worded in our readme, and I tried to do some A/B testing with friends (although n~=5), most of whom were first time vim users. Almost everyone got confused by packpath being included in all of the examples.

I think if someone is managing packpath manually, they'll generally know the issue is with the built-in plugin manager and really at the extreme case we'd have to add a mini-tutorial on package management in vim with all the most common package managers. Since the readme was getting a bit long, since I added a bunch of very explicit examples, including example keybindings and such, we just completely trimmed out all but the most common package manager.

Anyways, that's just my 2 cents, if this ends up being the wrong call, I'll add a little section on the native package manager, but you really should check out vim-plug or packer, it makes the upgrade and pinning experience much less painful

PR #561, opened 2021-01-08
README.md: auto keybindings for all servers

pierreprinetti:
Set keybindings on all servers by default in the suggested keybindings
snippet.

Please be aware that this is literally my first line of LUA ever.


mjlbach:
This will not work. See discussion on our other thread as to why, and also why if it did it would not be a good idea.

PR #651, opened 2021-01-09
Commands: add LspInfo

### LINKED ISSUE #254: per buffer/filetype checkhealth
I work with several LSP servers, some of them capricious.
I would love to see a command similar to `NeomakeInfo` or `CocInfo` (IIRC) e.g. to get more detailed information/diagnostics about the lsp servers:
- show init settings
- show current server status

This could be integrated as part of a per-buffer `checkhealth`.

mjlbach:
Closes #254 (?)
This is my first attempt at playing with floating windows. I wanted to get some feedback and input on the direction, and maybe from the floating window UI experts.

`:LspInfo` opens the window, clicking anywhere not on the window in mouse mode, or hitting escape closes.

Things that might be nice:
- [x] Format the text better (is there some common way we do this with lua string literals?)
- [x] More info? Maybe list the servers that *could* have launched per filetype
- [x] Padding on left.
- [x] ~~Fix up some of the lsp specific preview window functions in core so we can reuse them for UI components~~ Why is plenary.windows not in core yet, oh god, it's so much nicer. Still need TJ's ok.
- [x] Not just dump a bunch of messy code in a function and call it a day (make everything look nice and clean and document)
- [x] Add a splash of color. Don't know how to do this yet
- [ ] ~Dynamically resize the floating window on size change. I don't know how to do this, winresize autocmd doesn't seem to work. Low priority though, since you can Just reopen the window.~ I'm going to do this in core when we make ui helpers there.

I have 0 artistic talent/experience with UI, so suggestions on how to implement any of the above stylistic things welcome.

Screen Shot 2021-01-11 at 2 17 01 PM


mjlbach:
Now this is a super draft because I stole some stuff from plenary and need to talk to @tjdevries about how to use it. I don't think we should make plenary a dependency of this plugin, and I also hope that someday some of these UI helpers make their way into core.

cloggier:
@mjlbach How about making window creation configurable same way [Packer does](https://github.com/wbthomason/packer.nvim#custom-initialization) with its `display.open_fn`? You could by default just use `enew` for example and leave implementation of more fancy windows to users. I would gladly set it to use function I have for Packer.

mjlbach:
Or... I can just ~~steal~~ receive divine inspiration from others. 

In all honesty, I was going to just dump everything to messages like we usually do, so I'd rather this *not* become a whole UI thing and rather have a nice compartmentalized floating window for LspInfo.

lua/lspconfig.lua
```diff
@@ -18,6 +19,60 @@ end
 -- script in scriptnames to be executed is lspconfig.
 function M._root._setup()
   M._root.commands = {
+      LspInfo = {
```
elianiva, line 22: Not that it matters or anything but I couldn't help myself, shouldn't this `LspInfo` line gets outdented 2 space to the left? :laughing:

elianiva:
Would it be possible to show the current handlers/capabilities in a better format? I would love to see a *formatted* output from `vim.lsp.buf_get_clients` where it shows the handlers/capabilites in a list instead of raw lua table format.

mjlbach:
> Would it be possible to show the current handlers/capabilities in a better format?

That's essentially what I'm doing here, could you explain more about what I'm not doing? The first "section" is from buf_get_clients, the second from get_active_clients, the third from parsing the enabled configs, and matching the filetype against the current buffers filetype, then attempting to resolve the default root directory.

The issue with capabilties is the list is insanely long, and it would make this much less readable. I thought about enabling only a couple, like code actions, which get asked about frequently.

elianiva:
Yes, I think we should only include the common handlers like hover, code action, diagnostic, etc. We don't need *all* of them.

Also you're right, I changed my mind, I don't think I want capabilities stuff anymore. I didn't realise that the capabilities output is stupidly long and it would make LspInfo too cluttered.

mjlbach:
> Also you're right, I changed my mind, I don't think I want capabilities stuff anymore. I didn't realise that the capabilities output is stupidly long and it would make LspInfo too cluttered.

If it makes you feel any better I internally debated about this for awhile 😆 

> Yes, I think we should only include the common handlers like hover, code action, diagnostic, etc. 

The issue I have with this, is that the list will all be the same (for the handlers at least), since neovim core doesn't enable/disable handlers based on server capabilities. Are you talking about user defined handlers?

elianiva:
I'm thinking of showing it only if the user override it so they'll know what's the handler after they override it (maybe all fields that they override?) I have this issue a while ago when I'm trying to override a resolved capabilities, I need to check it from the output given by `buf_get_clients`, it's kinda annoying that I have to scroll through the _very long_ output to find which one that I've overriden.

mjlbach:
It will take a bit of refactoring, since we don't *currently* store what you have overridden, but I can see this being valuable so I'll try to add it, and if I don't someone can extend this PR.

I want to get this in soon, because I think it addresses a huge usability issue. My hope is people use it and complain about it be too/not enough verbose, and we can do a second pass at it.

mjlbach:
Ok, here's what a "bad" config vs a "good" config look like with the new UI.
Screen Shot 2021-01-09 at 4 51 37 PM
Screen Shot 2021-01-09 at 4 51 57 PM

mjlbach:
Note to self: this line is bad https://github.com/mjlbach/nvim-lspconfig/blob/58ee73802962495900fcfd80fb5f2ecc05840969/lua/lspconfig.lua#L77

It's why the command takes a few hundred ms, we could cache this when we start the servers.

lua/lspconfig.lua
```diff
@@ -18,6 +19,98 @@ end
...
+          for _, filetype_match in ipairs(config_table.filetypes) do
+            if buffer_filetype == filetype_match then
+              local matching_config_info = {
+                "",
+                "Config: "..config.name,
+                "\tcmd: "..cmd,
+                "\tcmd is executable: ".. cmd_is_executable,
+                "\tidentified root: "..(config_table.root_dir or "None"),
+                "\tcustom handlers: "..table.concat(vim.tbl_keys(config_table.handlers), ", "),
+              }
```
tjdevries, line 101: you may be able to re-use some of the `health#*` APIs from checkhealth to do a few of these things if you want. idk if that's better or not.

tjdevries:
This should not be very configurable. It should be pretty dumb and simple (especially to start). It doesn't even _have_ to have a floating window. You can just make a new tab / buffer / split / etc. if you don't want to make it floaty.

mjlbach:
Added a screenshot of the final version, this could probably be improved, but I think the returns are marginal at this point. I added a bit of color, but if someone with a stronger artistic sense wants to figure out how to make this prettier than I will gladly accept the PR :) (provided it does not increase code complexity, I agree with TJ that this is a convenience and not meant to be customizable/extensible).

mjlbach:
Oh, not pictured: Error highlighting
Screen Shot 2021-01-11 at 2 25 48 PM
Screen Shot 2021-01-11 at 2 26 16 PM

csmoe:
@mjlbach lsp_info display is a bit mess on my nvim.
version:
```
NVIM v0.5.0-dev+944-g82100a6bd
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: /Applications/Xcode_12.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/Users/runner/work/neovim/neovim/build/config -I/Users/runner/work/neovim/neovim/src -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Applications/Xcode_12.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk/usr/include -I/Library/Frameworks/Mono.framework/Headers -I/Users/runner/work/neovim/neovim/build/src/nvim/auto -I/Users/runner/work/neovim/neovim/build/include
Compiled by runner@Mac-1607923147408.local

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
```
image

mjlbach:
Hmm, maybe I should disable transparency. What's your term emulator/compositor?

csmoe:
> Hmm, maybe I should disable transparency. What's your term emulator/compositor?

apple terminator Version 2.11 (440)

andrewferrier:
If it's possible, it would be really interesting to see if it would be possible to make this command show the codeActions which can be executed.

andrewferrier:
Personally, I would also suggest that transparency by default is maybe not the best choice for a wider audience. I know transparency is very faishonable (!), but it's not very friendly to those with visibility issues. (I could go on a long rant about transparency, but I guess I'm just trying to make a general point that I think it should be optional).

csmoe:
@andrewferrier yep, transparency was disabled here https://github.com/neovim/nvim-lspconfig/pull/659

mjlbach:
@csmoe you beat me to it! My witty response is lost to the ether.

Re: Code actions, sure. If you could file an issue I'll take a look at it.

mjlbach:
Also, FWIW, the transparency was inherited from the plenary default function I used. Normally I'm not a transparency person myself 😆

PR #652, opened 2021-01-09
utils: add depth first and breadth first distinction in root pattern

mjlbach:
Some have brought up issues with root_patterns behaving unintuitively. I've added a breadth/depth first distinction and refactored all modules to use the new helper functions.

NOTE: I did not optimize the new modules to use the simplest root function possible for each one. Comments welcome, I'm not going to merge this unless there is a strong signaling that this clarifies things.

@ahmedelgabri @lithammer I think you've both commented on root_patterns/finders before, let me know if I should bring anyone else in.


mjlbach:
Also @hiphish

lithammer (commented):
I think I like it. I guess a question is; when is "depth first" wanted? Usually it can be solved by just chaining calls (e.g. GH-648). But maybe it's better to explicit about it to prevent someone from making a refactoring like this:

```diff
-root_dor = function(fname)
-  return util.root_pattern("Makefile")(fname) or
-    util.root_pattern(".git")(fname)
-end,
+root_dir = util.root_pattern("Makefile", ".git"),
```

An alternative API could be to keep the `root_pattern` function and make it breadth first (since it seems to be the most common use-case). Then add a helper function to make it nicer to convert to depth first (instead of having to use an anonymous function). This can probably be useful if you for some reason want to combine both breadth first with depth first as well.

```lua
root_dir = util.root_patterns(  -- note the "s" in patterns
  util.root_pattern("Makefile"),
  util.root_pattern(".git"),
),
```

```lua
local function root_patterns(...)
  return function(fname)
    for _, func in ipairs(unpack(...)) do
      local path = func(fname)
      if path then
        return path
      end
    end
  end
end
```

mjlbach:
I chain calls all the time, which is why I marked this as a draft 😆 People seemed unhappy with there not being a depth/breadth option, or at least, confused because it was not clear from looking at the code (I guess) that it was breadth first. I don't think this is a strictly necessary PR, but I figured it would be easier to see if there is utility in adding this, and we can discuss it here instead of having many open issues about figuring out cleaner root pattern strategies.

HiPhish:
I think `breadth_first_root_pattern` is a bit verbose. Then again, you only use it like once per server, so it's not really much of an issue. But the `breadth_first` part of the name bothers me. I think it was me who started calling it that simply because i wanted to assign a name to it, but it's not really a breadth/depth search.

Breadth of width first is when you search through a tree starting at the root and digging _down_. Here however we are moving _up_ the tree, the difference is in whether we try all the patterns at the same time during one traversal, or whether we have a separate traversal for each pattern. I don't know what to call that type of search.

mjlbach:
It's still (in the formal sense) a breadth first search over a tree, it's just in this case, the root node of the tree is the starting working directory, and the children belong to the parent directory.

HiPhish:
> it's just in this case, the root node of the tree is the starting working directory

Is it? I was under the impression that the search starts at the directory containing the file and then moves up the file tree from there. I might have misunderstood the algorithm then.

mjlbach:
Yes, the algorithm is not specific to file tree, but is a general means of searching trees (so here, we're just defining the tree to be starting at the deepest level of the file system hierarchy)

HiPhish:
OK, but the algorithm only moves straight up, right? It doesn't traverse sibling directories. In that case the "tree" being traversed is just a straight line up, and if there are no turns to be taken, then there is no difference between breadth-first and depth-first. Take the following graph:

```
                A     
                |     
      +---------+---------+
      |                   |
      B                   C
      |                   |
+-----+-----+     +-------+------+
|           |     |              |
D           E     F              G
            |
            H
```

Starting at `A`, the breadth-first traversal order is `A, B, C, D, E, F, G, H` and the dept-first order is `A, B, D, E, H, C, F, G` because when we descend we have up to two choices. However, if we start at `H` the order is always `H, E, B, A` because there always exactly one choice on how to ascend.

mjlbach:
You're right in the sense that's it's technically not a tree, but a DAG since children have only one child, but can have multiple parents. The breadth and depth first still will ultimately visit the same nodes, so I think the terminology is fine.

HiPhish:
You be the judge then. My concern is that it might confuse new users because they expect "breadth-first" to mean something else, and the only way of finding out what it actually means is to read the source code (or documentation if it existed), which makes the name meaningless. If everyone else is fine with the name though, then so be it, I don't have any better names to suggest.

PR #654, opened 2021-01-10
julials: add note about crash without binary mode

mjlbach:
@kdheepak 

I don't like this, but if they aren't going to fix it upstream, this is the only fix. No other language server behaves this way, and it's out of compliance with the spec to crash.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -1858,6 +1858,18 @@ https://github.com/julia-vscode/julia-vscode
 ```sh
 julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
 ```
+
+LanguageServer.jl implemented a unique hard check that the contents of a file match the content sent
+to the language server, which requires a workaround, as neovim (and vim) and new lines to files
+by default on save. Add the following to your init.
+
+require('lspconfig').util.nvim_multiline_command [[
+  augroup SpecialJuliaLanguageServerWorkarounds
+    autocmd!
+    au FileType julia set binary noeol
+  augroup end
+\]\]
+
 The default config lazily evaluates the location of the julia language server from the your global julia packages.
 This adds a small overhead on first opening of a julia file. To avoid this overhead, replace server_path in on_new_config with
 a hard-coded path to the server.
--- a/lua/lspconfig/julials.lua
+++ b/lua/lspconfig/julials.lua
@@ -39,6 +39,18 @@ https://github.com/julia-vscode/julia-vscode
 ```sh
 julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
 ```
+
+LanguageServer.jl implemented a unique hard check that the contents of a file match the content sent
+to the language server, which requires a workaround, as neovim (and vim) and new lines to files
+by default on save. Add the following to your init.
+
+require('lspconfig').util.nvim_multiline_command [[
+  augroup SpecialJuliaLanguageServerWorkarounds
+    autocmd!
+    au FileType julia set binary noeol
+  augroup end
+\]\]
+
 The default config lazily evaluates the location of the julia language server from the your global julia packages.
 This adds a small overhead on first opening of a julia file. To avoid this overhead, replace server_path in on_new_config with
 a hard-coded path to the server.


mjlbach:
sorry @quantum-booty * I misplaced your hyphen.

kdheepak:
Thanks for doing all this debugging and figuring out what is going on!

mjlbach:
Did this work for you? It seems to work for me. I am a little frustrated about this situation tbh, but 🤷‍♀️

mjlbach:
I also didn't see any special handling with coc's julia plugin/vim-lsp/languageclient-neovim, so I assume they will all start showing this behavior once people bump languageserver.jl

kdheepak:
I haven't tested it yet. I'll do so in a few hours.

The `LanguageServer.jl` devs tend to follow the spec very strictly and are very strict about their implementation. This has caused a few issues when the LSP implementation differs between VSCode and other clients.

I read through this section in the spec: https://microsoft.github.io/language-server-protocol/specification#textDocument_didSave

It says "The document save notification is sent from the client to the server when the document was saved in the client.".

It doesn't say anything about the text matching, but it does say that the buffer text can be optionally sent. 

One thing I guess what I'm trying to figure out is why the text that is sent by LSP here:

https://github.com/neovim/neovim/blob/a1ec36ff31ff1fd545dd135faa47dc782b41a687/runtime/lua/vim/lsp.lua#L870-L891

doesn't match the file contents? Do you know why this is the case? It is triggered as a `BufWritePost` so I would have expected the change to be in the text that is sent along with the notification. I read `:help fixeol`, but I don't understand why there's a disparity here in the first place.

https://github.com/neovim/neovim/blob/a1ec36ff31ff1fd545dd135faa47dc782b41a687/runtime/lua/vim/lsp.lua#L913

You mentioned that `set nofixeol` seems to be broken. Can you clarify? 

Could we not just fix the LSP implementation to send the correct included text? If can check the value of `fixeol` [here](https://github.com/neovim/neovim/blob/a1ec36ff31ff1fd545dd135faa47dc782b41a687/runtime/lua/vim/lsp.lua#L881) and append the new line to the text sent accordingly.

Alternatively, we can disable sending the text on save for Julia if it going to be an issue.

mjlbach:
> It says "The document save notification is sent from the client to the server when the document was saved in the client. It doesn't say anything about the text matching, but it does say that the buffer text can be optionally sent.

I think that's my main point, the whole text matching thing is outside of the spec and an implementation detail.

>doesn't match the file contents? Do you know why this is the case? It is triggered as a BufWritePost so I would have expected the change to be in the text that is sent along with the notification.

I don't think it's an issue with our language client implementation so much as a feature of vim. The newline is actually appending during [buf_write](https://github.com/mjlbach/neovim/blob/b535575acdb037c35a9b688bc2d8adc2f3dece8d/src/nvim/fileio.c#L2132-L2144) right [here](https://github.com/mjlbach/neovim/blob/b535575acdb037c35a9b688bc2d8adc2f3dece8d/src/nvim/fileio.c#L3179-L3249). We send updates on BufWritePost, but by then it's too late because Julia has already read the changes from the disk and decided to crash. We'd have to do something during BufWritePre/BufWrite (they're the same thing) to pre-fix eol and send the update to the server *before* vim internally applied the change during the save process (so vim wouldn't make any changes within buf_write).


> I read :help fixeol, but I don't understand why there's a disparity here in the first place. You mentioned that set nofixeol seems to be broken. Can you clarify?

I was wrong, it's not broken, you need 'noeol' and ('bin' or 'nofixeol') for it to not append a new line, so set nofixeol is insufficient by itself. See [here](https://github.com/mjlbach/neovim/blob/b535575acdb037c35a9b688bc2d8adc2f3dece8d/src/nvim/memline.c#L4115-L4121)

> Could we not just fix the LSP implementation to send the correct included text? If can check the value of fixeol here and append the new line to the text sent accordingly.

There's nothing *wrong* with the LSP implementation. We could hack around LanguageServer.jl's requirement, but it's a lot more complicated than just appending a newline since we need to support all operating systems and match vim's internal handling of this in buf_write, because if vim changes anything then the crash will occur, and again this feels like a huge hack.

mjlbach:
Also just generally JLS's check on this is just very fragile, in the case of unreliable filesystems, IO error under high system load, network communciation to remote language server, basically any time the two become decoupled instead of handling the error it will just crash.

quantum-booty:
Setting binary mode doesn't help either X_X, at this point I just give up and commented out the textdocument checking in JuliaLanguageServer source code, and it worked.

mjlbach:
@quantum-booty you set binary *and* nofixeol? You need both. I tested this on like 20 different julia files.

quantum-booty:
Yeah it doesn't work. And to be honest I'd rather comment out the language server checking rather than using the binary option.

```
au BufNewFile,BufRead *.jl set filetype=julia

set binary
set noeol
set nofixeol

call plug#begin('~/.config/nvim/plugged')
Plug 'nvim-lua/completion-nvim'
Plug 'neovim/nvim-lspconfig'
call plug#end()

set updatetime=50
set completeopt+=menuone,noselect,noinsert
set completeopt-=preview
set shortmess+=c

autocmd BufEnter * lua require'completion'.on_attach()

lua <<EOF
require'lspconfig'.julials.setup{}
require('lspconfig').util.nvim_multiline_command [[ augroup SpecialJuliaLanguageServerWorkarounds autocmd! au FileType julia set binary noeol augroup end ]]
EOF

lua vim.lsp.set_log_level('error')
" log_path: ~/.local/share/nvim/lsp.log
```

mjlbach:
Sad. Ok, maybe I'll just include a short sed command to remove the line from LanguageServer.jl in the readme instead

mjlbach:
I'm just going to close this while we wait for my upstream fix (to nvim-core) to be merged.

mjlbach:
neovim/neovim#13792 was merged, this should be fixed

PR #655, opened 2021-01-10
Remove package_json link from Metals docs.

ckipp01:
Many of the settings that this link shows are reliant on
specific things like code lenses working. I think it's a better
idea to not show these as there are a handful of them that if a
user would use them, they won't work at all, and in the worst
case they actual may cause errors due to unsupported handlers
or LSP extensions.

### DIFF
--- a/lua/lspconfig/metals.lua
+++ b/lua/lspconfig/metals.lua
@@ -18,7 +18,6 @@ configs[server_name] = {
       };
   };
   docs = {
-    package_json = "https://raw.githubusercontent.com/scalameta/metals-vscode/master/package.json";
     description  = [[
 https://scalameta.org/metals/
 


mjlbach:
Sounds good, I've been trying to do this for all of the plugins that were causing confusion as well, but I missed this one. Thanks for noticing. I'm not sure if you saw, but I listed nvim-metals under our recommended plugins, and gave you all shoutout on reddit :)

mjlbach:
FYI I did merge but the GHA for docgen is broken, so this will likely not show up until the issue is fixed on GH's side and we pushed another change to the repo

ckipp01:
>  gave you all shoutout on reddit :)

I saw that, thanks! :)

> FYI I did merge but the GHA for docgen is broken, so this will likely not show up until the issue is fixed on GH's side and we pushed another change to the repo

No worries at all 👍🏼

PR #657, opened 2021-01-11
rust_analyzer: read project root from cargo

mjlbach:
@lithammer @elianiva 

I left the current behavior as a fallback. Note, this does increase launch time when first opening a file in a project. This is the only "correct" way to do it, but we could also just read from each cargo.toml and look for the workspace key. I think it would be slightly faster, but maybe less robust.


elianiva:
LGTM! I think it's slow because of `vim.fn.json_decode`? Though in my case the delay isn't that much, it's barely noticeable.

lua/lspconfig/rust_analyzer.lua
```diff
@@ -6,7 +6,15 @@ configs.rust_analyzer = {
...
+      local cargo_root
+      if vim.v.shell_handler == 0 then
+        cargo_root = vim.fn.json_decode(cargo_metadata)["workspace_root"]
+      else
+        cargo_root = nil
+      end
+      return cargo_root or
+      util.find_git_ancestor(fname) or
       util.root_pattern("rust-project.json")(fname) or
       util.root_pattern("Cargo.toml")(fname)
```
lithammer, line 19:
```suggestion
      return cargo_root or
        util.find_git_ancestor(fname) or
        util.root_pattern("rust-project.json")(fname) or
        util.root_pattern("Cargo.toml")(fname)
```

lua/lspconfig/rust_analyzer.lua
```diff
@@ -6,7 +6,15 @@ configs.rust_analyzer = {
     cmd = {"rust-analyzer"};
     filetypes = {"rust"};
     root_dir = function(fname)
-      return util.find_git_ancestor(fname) or
+      local cargo_metadata = vim.fn.system("cargo metadata --format-version 1")
+      local cargo_root
+      if vim.v.shell_handler == 0 then
+        cargo_root = vim.fn.json_decode(cargo_metadata)["workspace_root"]
+      else
+        cargo_root = nil
```
lithammer, line 14: This `else` looks redundant.

mjlbach:
@lithammer fixed. Are you two happy with this?

lithammer:
> Note, this does increase launch time when first opening a file in a project.

Speaking of this. I would if it would be possible to somehow run the LSP stuff in a background job or whatever. Basically to prevent slow stuff in the `root_dir` callback from impacting launch time. No one will notice if the servers takes 1-2s seconds to start in the background, but they sure notice Vim itself being slow to start.

> This is the only "correct" way to do it, but we could also just read from each cargo.toml and look for the workspace key. I think it would be slightly faster, but maybe less robust.

Definitely think we should try to stick with "supported" solutions as long as possible.

lithammer:
> @lithammer fixed. Are you two happy with this?

I am 🙂

lithammer (approved):

mjlbach:
> Speaking of this. I would if it would be possible to somehow run the LSP stuff in a background job or whatever. Basically to prevent slow stuff in the root_dir callback from impacting launch time. No one will notice if the servers takes 1-2s seconds to start in the background, but they sure notice Vim itself being slow to start.

So, yes and no. I was playing around what was and wasn't possible with luv. I would like to make some things async (like loading configs, since right now it takes 1 ms per config and some people enable them all), but right now I don't think we can call bindings to vim functions within the luv context, which would mean a *lot* of our current code not work.

Example:

```lua
local uv = vim.loop

local x_func = function(package_path) --work,in threadpool
    package.path = package_path
    result = vim.fn.decode_json("{\"test\".json}")
    return result
end

local ctx = uv.new_work(
    x_func,
    vim.schedule_wrap(function(n) print(n) end)    --after work, in loop thread
)
uv.queue_work(ctx, package.path)
uv.queue_work(ctx, package.path)
uv.queue_work(ctx, package.path)

uv.run('default')
```

mjlbach:
The other thing to keep in mind, is that although from the Language Server's perspective, all it needs is a list of commands. Initialization is currently fairly stateful client side. Each of the different handlers (on_init, on_attach, etc) can read and modify vim global state, so those would always block/have to be run sequentially.

PR #661, opened 2021-01-12
LspInfo: fix display when client cmd contains newlines

### LINKED ISSUE #660: LspInfo fails when cmd has newline character
The new `LspInfo` functionality is awesome, and works for most cases flawlessly. I noticed a hick-up: I just run LspInfo in a buffer with a Julia project, and got:
Screenshot from 2021-01-12 17-42-20
Pressing enter makes the floating window appear empty, and one has to :q to get back to the buffer.

mjlbach:
Closes #660

@coxackie 

 I realize it's not pretty (yet), I'll add new line trimming, I just want to make sure this fixes the issue in all cases and doesn't add weird corner problems (it shouldn't)

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -2132,6 +2132,7 @@ require'lspconfig'.julials.setup{}
   Commands:
   
   Default Values:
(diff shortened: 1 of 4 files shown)


mjlbach:
It works, but I may just rewrite some things to make it keep the new lines, as it currently looks pretty ugly.

Screen Shot 2021-01-12 at 4 17 35 PM

teto:
I dont think it's ugly but focusing the floating window may not be evident for everyone, maybe creating a new tab as checkhealth does is better.

mjlbach:
The newline problem will still be an issue with displaying in a tab, so this PR will still be necessary and the "ugliness" will remain (which I kinda fixed). I can get rid of the floating window. I personally find floating windows less disruptive, since it's easily dismissed with esc by clicking.

mjlbach:
Fixed.

Screen Shot 2021-01-14 at 10 11 46 PM

Now that this command is solidifying, I should probably move it out of lspconfig.lua into it's own module and clean things up a bit, but I'll do that in it's own PR. @coxackie, do you want to test?

coxackie:
Sure thing. Thanks.

mjlbach:
No, this PR is updated.

coxackie:
works pretty well - no errors. many thanks.

 For what it's worth, I also support the floating window option. This is just some temporary info that one should be able to discard with a simple press of ESC.

PR #668, opened 2021-01-14
use new global bin command

mhartington:
This is a placeholder until Angular releases a non-rc version of their update `@angular/language-server` package, which includes a `bin` entry for calling directly.


mjlbach:
Nice! Just ping me when this releases and we will merge. One less headache for users :)

mhartington:
@mjlbach this should be good to go!

mjlbach:
Wonderful! Merged.

PR #669, opened 2021-01-14
Fix LspInfo when filetypes not defined

### LINKED ISSUE #663: LspInfo fails un-gracefully when filetypes aren't set
Certain diagnostic servers like efm and diagnosticls don't have default filetypes. You have to set these in order for the language server to work. LspInfo should warn you when these aren't set.

Reported on gitter by @cloggier

mjlbach:
Closes #663

CC: @cloggier

Will now warn you if filetypes not defined for a server.

Screen Shot 2021-01-14 at 2 01 39 PM

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -87,25 +87,36 @@ function M._root._setup()
             cmd = "cmd not defined"
             cmd_is_executable = cmd
           end
-
-          for _, filetype_match in ipairs(config.filetypes) do
-            if buffer_filetype == filetype_match then
+          if config.filetypes then
+            for _, filetype_match in ipairs(config.filetypes) do
+              if buffer_filetype == filetype_match then
+                local matching_config_info = {
+                  "",
+                  "Config: "..config.name,
+                  "\tcmd: "..cmd,
+                  "\tcmd is executable: ".. cmd_is_executable,
+                  "\tidentified root: "..(config.get_root_dir(buffer_dir) or "None"),
+                  "\tcustom handlers: "..table.concat(vim.tbl_keys(config.handlers), ", "),
+                }
+               vim.list_extend(buf_lines, matching_config_info)
+              end
+            end
+          else
               local matching_config_info = {
                 "",
                 "Config: "..config.name,
-                "\tcmd: "..cmd,
-                "\tcmd is executable: ".. cmd_is_executable,
-                "\tidentified root: "..(config.get_root_dir(buffer_dir) or "None"),
-                "\tcustom handlers: "..table.concat(vim.tbl_keys(config.handlers), ", "),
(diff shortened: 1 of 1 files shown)


weilbith:
I think this fix is not complete. I'm locally on latest master (at time of writing) and still get this error for line `66`. Changing this locally solves it. So it appears to be the last occurrence.

mjlbach:
How do you have an attached/active client without filetypes defined?

weilbith:
This is the case for my `efm-langserver` setup. The following is a snippets that hides some parts of my setup, but eventually you can find some information. But for me it works...

```vim
lua require('lspconfig').efm.setup({
    \   init_options = {
    \     documentFormatting = true,
    \     hover = false,
    \     documentSymbol = false,
    \     codeAction = false,
    \     completion = false,
    \   },
    \   settings = {
    \     rootMarkers = { "package.json", "pyproject.toml", "Cargo.toml", ".git/" },
    \     languages = {
    \       sh = {
    \         require('efm_config.linting').shellcheck,
    \         require('efm_config.formatting').shfmt,
    \	    }
    \     },
    \   }
    \ })
```

mjlbach:
I can't reproduce this. EFM never starts for me if I don't define filetypes, and if I don't define filetypes, it won't be an active or attached client, so it won't be triggered by that help function.

Screen Shot 2021-01-15 at 9 45 16 AM

weilbith:
Okay. Idk. For me it works. If I comment out the specified line, the `LspInfo` command successfully lists the EFM server. 

@lukas-reineke sorry to pull you in here. But I got inspired for EFM by your Reddit post and also took your dotfile as template how to setup. Moreover I remember that you implemented to the server the feature to support the config update capability. Maybe you know why this works even it apparently is unexpected.

mjlbach:
This made me realize LspInfo won't work when you manually launch a client (which doesn't have filetypes set), so I fixed it. try the PR and lmk if it works for you

lukas-reineke:
I'm not 100% sure I understand the problem, but

> How do you have an attached/active client without filetypes defined?

you don't need to specify a filetype, the client defaults to start with `BufReadPost`
https://github.com/neovim/nvim-lspconfig/blob/87900ffcccd47bb6a09483ccf0d6840dc1d0d0d6/lua/lspconfig/configs.lua#L72-L76

Then the client can decide if it can handle that filetype or not. efm works like this, so it can also do wildcard matching, and you don't have to specify the filetypes twice.

mjlbach:
I realized this about the FT fallback after posting (and merged the PR fixing concatenating nil filetypes), not sure why I never ran into an issue with filetypes being unlisted. IMO this is undesirable behavior and we should have a wildcard attribute under filetypes. Otherwise if someone sets up a server and forgets the filetype attribute, the language server will unduly launch.

PR #671, opened 2021-01-15
kotlin: update roots to fallback to build.gradle

### LINKED ISSUE #455: Root pattern for Gradle projects?
[Gradle](https://gradle.org/) is a build system used by JVM languages such as Java or Kotlin. A Gradle project can bei either a plain project or contain sub-project. The question is, what should the root pattern be?

A plain project will have one file `build.gradle` in its root directory. A project with sub-projects on the other hand will have multiple sub-directories, each with their own `build.gradle`, and one global `settings.gradle` file. In order to find the root directory we have to first search for a `settings.gradle` file, and only then fall back to searching for `build.gradle`. Otherwise the sub-project will be considered the root of the entire project.

My question is, what should the `root_dir` configuration be? At first I tried the following:

```lua
util.root_pattern('settings.gradle', 'build.gradle')
```

However, after looking at the implementation of `root_pattern` I can see that this is wrong. Neovim will for each directory (starting with the one containing the current file) look for all the root files first before moving up a directory. What I want is the opposite though: start with the first file, search upwards, and when that fails search for the second file.

Here is another attempt:

```lua
function(startpath)
    return util.root_pattern('settings.gradle')(startpath) or util.root_pattern('build.gradle')(startpath)
end
```

What do you think? Should we have a depth-first alternative for `util.root_path`? Or do we leave it to server configuration s to use the latter?

mjlbach:
Closes #455 

@ahmedelgabri @HiPhish does this work/seem right for you?

### DIFF
--- a/lua/lspconfig/kotlin_language_server.lua
+++ b/lua/lspconfig/kotlin_language_server.lua
@@ -21,10 +21,13 @@ end
 --  Kotlin. I am not sure whether the language server supports Ant projects,
 --  but I'm keeping it here as well since Ant does support Kotlin.
 local root_files = {
+  'settings.gradle',       -- Gradle (multi-project)
+  'settings.gradle.kts',   -- Gradle (multi-project)
   'build.xml',             -- Ant
   'pom.xml',               -- Maven
-  'settings.gradle',       -- Gradle (mutli-project)
-  'settings.gradle.kts',   -- Gradle (mutli-project)
+}
+
+local fallback_root_files = {
   'build.gradle',          -- Gradle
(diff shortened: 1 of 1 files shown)


HiPhish (commented):
LGTM

mjlbach:
if you want to update your  JDTLS PR to have the same fallback behavior I'll merge it, which should close your issue

HiPhish:
> if you want to update your JDTLS PR to have the same fallback behavior I'll merge it, which should close your issue

I have done it now.

PR #676, opened 2021-01-18
orgnize the autocmd into an augroup

skt041959:
Thus we can use `doautoall LSP_<name> FileType` after lazy loading nvim-lspconfig


mjlbach:
You can now lazy load nvim_lspconfig by disabling autostart, see the readme.

PR #685, opened 2021-01-20
bashls: Set root_dir to nil when editing file in ~

lithammer:
A problem with bash-language-server is that it has a bad habit of trying to recursively scan the entire home folder if you for example tried to edit a shell script directly in `~` (e.g. `~/foo.sh`).

This causes a bunch of weird behaviour and causes the server to take forever to start.

To prevent this we set `root_dir` to `nil` when the resolved `root_dir` happened to be `~`. This solution is kind of hinted towards if you look at their own [README](https://github.com/bash-lsp/bash-language-server/blob/master/README.md) for how you should configure it with coc.nvim (note the `ignoredRootPaths`):

    "languageserver": {
      "bash": {
        "command": "bash-language-server",
        "args": ["start"],
        "filetypes": ["sh"],
        "ignoredRootPaths": ["~"]
      }
    }

### DIFF
--- a/lua/lspconfig/bashls.lua
+++ b/lua/lspconfig/bashls.lua
@@ -1,3 +1,4 @@
+local log = require('vim/lsp/log')
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
@@ -7,7 +8,17 @@ configs[server_name] = {
   default_config = {
     cmd = {"bash-language-server", "start"};
     filetypes = {"sh"};
-    root_dir = util.path.dirname;
+    root_dir = function(fname)
+      local dirname = util.path.dirname(fname)
+
+      -- Prevent the bash-language-server from scanning the entire home folder.
+      if dirname == vim.loop.os_homedir() then
+        log.info('Not setting root_dir to prevent scanning everything in $HOME')
+        return nil
+      end
+
+      return dirname
+    end
   };
   docs = {
     description = [[
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -2118,6 +2118,12 @@ This server accepts configuration via the `settings` key.
   
   Display plots within VS Code\. Might require a restart of the Julia process\.
 
+- **`julia.useProgressFrontend`**: `boolean`
+
+  Default: `true`
+  
+  null
+
 - **`julia.useRevise`**: `boolean`
 
   Default: `true`
@@ -2326,6 +2332,12 @@ This server accepts configuration via the `settings` key.
   Default: `vim.empty_dict()`
   
   Array items: `{description = "Unicode character to translate to",type = "string"}`
+  
+  null
+
+- **`lean.input.eagerReplacementEnabled`**: `boolean`
+
+  Default: `true`
   
   null
 


lua/lspconfig/bashls.lua
```diff
@@ -7,7 +8,17 @@ configs[server_name] = {
   default_config = {
     cmd = {"bash-language-server", "start"};
     filetypes = {"sh"};
-    root_dir = util.path.dirname;
+    root_dir = function(fname)
+      local dirname = util.path.dirname(fname)
+
+      -- Prevent the bash-language-server from scanning the entire home folder.
+      if dirname == vim.loop.os_homedir() then
+        log.info('Not setting root_dir to prevent scanning everything in $HOME')
```
lithammer, line 16: Not sure how to phrase this in a good way... or if we should bother.

mjlbach: This is probably fine. You tested the log right? 

Maybe change the info to:
`Detected root directory is $HOME will will hang bashls. Move your file to to a subdirectory and reopen.`

mjlbach:
100000x this. Good idea.

lua/lspconfig/bashls.lua
mjlbach: And maybe note this in the comment
```
  -- Prevent the bash-language-server from hanging while scanning the entire home folder.
```

mjlbach:
FYI pretty sure coc.nvim does this too:
https://github.com/neoclide/coc.nvim/wiki/Language-servers#bash

lua/lspconfig/bashls.lua
lithammer: > This is probably fine. You tested the log right?

Yeah, except the default log level is warn(ing), so most people will likely miss it. I could change it to `log.warn` though.

But maybe the default log level should be changed instead? https://github.com/neovim/neovim/blob/1785ac3e3787e84846d753ceb73a239f5575a691/runtime/lua/vim/lsp/log.lua#L22

mjlbach: I would just change it to warn right now, this is kinda a "warning" in that the language server won't start. Has this issue been filed upstream? I tried searching the bashls issue tracker and didn't see anything. There should be an option not to index home directory given some people are going to try to edit bashrc/profile etc.

lithammer:
An annoying thing I noticed is that the server never actually initializes if it receives `nil`. So nothing is working which will be a frustrating user experience as well. Not sure how to handle that 🤔

mjlbach:
I would check what coc does, IMO it's fine not to start, but then maybe print to :messages why instead of logging to the log.

lithammer:
> I would check what coc does, IMO it's fine not to start, but then maybe print to :messages why instead of logging to the log.

Seems like it mostly doesn't start either. Just provides some snippet completion included in [coc-sh](https://github.com/josa42/coc-sh).

So I did some digging, and it turns out the glob pattern they use to find shell scripts is configurable via the `GLOB_PATTERN` environment variable (https://github.com/bash-lsp/bash-language-server/pull/187). It defaults to `**/*@(.sh|.inc|.bash|.command)`.

I wonder if not a better solution would be to override it to only look into the current directory. It's gonna be good enough in most cases, maybe even better (how often do you actually want completion from shell scripts somewhere else in a repo). Something like this:

```lua
    cmd_env = {
      GLOB_PATTERN = vim.env.GLOB_PATTERN or "*@(.sh|.inc|.bash|.command)"
    },
```

lithammer:
Ideally I would like to override `GLOB_PATTERN` conditionally based on the resolved `root_dir`. But since it's an environment variable I don't think it's possible.

Would've been better if it was a regular `settings` or `init_options` 🤷‍♂️

lithammer:
Alternative solution using `GLOB_PATTERN`: GH-686

mjlbach:
Closing since I merged GH-686

PR #686, opened 2021-01-20
bashls: Set GLOB_PATTERN to prevent recursive scanning

lithammer:
Alternative solution to GH-685. Personally I prefer this approach since it means editing files in `~` will work just fine.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -146,6 +146,9 @@ require'lspconfig'.bashls.setup{}
   
   Default Values:
     cmd = { "bash-language-server", "start" }
+    cmd_env = {
+      GLOB_PATTERN = "*@(.sh|.inc|.bash|.command)"
+    }
     filetypes = { "sh" }
     root_dir = vim's starting directory
 ```
--- a/lua/lspconfig/bashls.lua
+++ b/lua/lspconfig/bashls.lua
@@ -6,6 +6,13 @@ local server_name = "bashls"
 configs[server_name] = {
   default_config = {
     cmd = {"bash-language-server", "start"};
+    cmd_env = {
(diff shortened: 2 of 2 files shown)


mjlbach:
Oh really? That's nice! Any disadvantages? I'll try it out, but I think you've read up more on this than me.

lithammer:
> Any disadvantages?

Well things like go-to-definition will work in fewer cases. On the other hand, you'll likely also get fewer unrelated workspace symbols because it won't be scraping an entire `node_modules/` directory three levels down or some such.

The nature of shell scripting means that the current directory is usually more important than some nested directory as well. At least from my experience. I suspect a large majority of shell scripts are stand-alone anyway (because you want them portable and easy to run).

mjlbach:
I haven't forgot about this! I think there is a way to have this env-var differ based on if we are in $HOME or not, but I haven't worked on it yet.

lithammer:
> I haven't forgot about this!

No worries 🙂

> I think there is a way to have this env-var differ based on if we are in $HOME or not, but I haven't worked on it yet.

It's a classic chicken-and-egg problem 🤔 

An option could be to restart the server with a custom `GLOB_PATTERN` if we detect we're in `~`. But I don't know how to restart the server in a graceful way. Feels a bit dirty too.

Ultimately, the best course of action is perhaps to bring this upstream...? Convincing them that recursive scanning when the workspace is `~` probably isn't what you want the majority of the time. Or at the very least turn `GLOB_PATTERN` into a real setting instead of an environment variable.

mjlbach:
I'm merging this because I think it's a good idea, do you want to file the upstream issue or should I?

lithammer:
I filed https://github.com/bash-lsp/bash-language-server/issues/277

PR #690, opened 2021-01-23
chore: make LspInfo output more compact

teto:
to save screen estate. I've added some indentation too to better identify the blocks. Let me know what you think @mjlbach

### DIFF
--- a/lua/lspconfig/lspinfo.lua
+++ b/lua/lspconfig/lspinfo.lua
@@ -13,11 +13,16 @@ return function ()
   local bufnr, win_id = win_info.bufnr, win_info.win_id
 
   local buf_lines = {}
+
+  local buf_client_names = {}
+  for _, client in ipairs(buf_clients) do
+    table.insert(buf_client_names, client.name)
+  end
+
   local header = {
-    "Available servers:",
-    "\t"..table.concat(vim.tbl_keys(configs), ', '),
+    "Configured servers: "..table.concat(vim.tbl_keys(configs), ', '),
     "",
-    "Clients attached to this buffer: "..tostring(#buf_clients)
+    tostring(#buf_clients).." client(s) attached to this buffer: "..table.concat(buf_client_names, ', '),
   }
   vim.list_extend(buf_lines, header)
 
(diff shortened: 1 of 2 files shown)


mjlbach:
I think I like it? 

Before:
Screen Shot 2021-01-23 at 8 50 09 AM
After:
Screen Shot 2021-01-23 at 9 00 43 AM

mjlbach:
Two things. One, I think the first tab/space looks a bit weird next to  `Configured servers:`, two, maybe we should add highlighting of the filetypes name, as `python` looks a bit blended in with the line.

I also noticed that `clients attached to this buffer` is not correct (my b), but that's outside the scope of this PR (Unless you just want to throw it in)

mjlbach:
Also, does current_id make sense? we never let the client change ids

mjlbach:
Also maybe add the tab alignment to the last section to match the others

teto:
I use indents to make it easier to differentiate between sections but that may not be what you want.



LspInfo is a good way to debug LSP server for the current buffer but the list of active clients could be moved to neovim's checkhealth  while LspInfo would cater to buffer specific debug. Anyway that's out of scope for this PR.

mjlbach:
I think it looks great now! Are you happy with it?

mjlbach:
The reason I like having active clients, is because I want to know if a client is active but not attached to the buffer. This isn't immediately apparent now, but it's super useful for multi-root workspaces (which kinda but not really works right now)

teto:
I am happy with it but more importantly I want you tobe happy with it :p

mjlbach:
Oh, did you want to add the tabs to the subheadings of  Config:  too?

teto:
tab should be fixed for the last section, please check

lua/lspconfig/lspinfo.lua
```diff
@@ -38,14 +43,14 @@ return function ()
...
+  local indent = "  "
   local function make_client_info(client)
     return {
       "",
-      "Client: "..tostring(client.id),
-      "\tname: "..client.name,
-      "\troot: "..client.workspaceFolders[1].name,
-      "\tfiletypes: "..table.concat(client.config.filetypes or {}, ', '),
-      "\tcmd: "..remove_newlines(client.config.cmd),
+      indent.."Client: "..client.name.." (id "..tostring(client.id)..")",
```
mjlbach, line 50: I mean that you add these empty spaces after client/root/filetype/cmd to align them.

lua/lspconfig/lspinfo.lua
```diff
@@ -86,28 +91,29 @@ return function ()
...
           }
          vim.list_extend(buf_lines, matching_config_info)
         end
       end
     else
         local matching_config_info = {
           "",
-          "Config: "..config.name,
-          "\tfiletype: ".."No filetypes defined, please define filetypes in setup().",
+          indent.."Config: "..config.name,
```
mjlbach, line 107: And here there aren't those same number of spaces

lua/lspconfig/lspinfo.lua
```diff
@@ -86,28 +91,29 @@ return function ()
...
         if buffer_filetype == filetype_match then
           local matching_config_info = {
-            "",
-            "Config: "..config.name,
-            "\tcmd: "..cmd,
-            "\tcmd is executable: ".. cmd_is_executable,
-            "\tidentified root: "..(config.get_root_dir(buffer_dir) or "None"),
-            "\tcustom handlers: "..table.concat(vim.tbl_keys(config.handlers), ", "),
+            indent.."",
+            indent.."Config: "..config.name,
```
mjlbach, line 95: Or here

PR #710, opened 2021-02-03
rust_analyzer: Fix the root_dir shell command

stevearc:
The shell function was added in #657. Searching google and [the neovim repository](https://github.com/neovim/neovim/search?q=shell_handler&type=commits) yields no reference to a `v:shell_handler` special variable. I believe that `v:shell_error` was intended to be used. [The docs for v:shell_error](https://neovim.io/doc/user/eval.html#v:shell_error) certainly sounds like what we want, and testing it with this change produces the correct results.


mjlbach:
Thanks! Don't know what I was thinking.

PR #715, opened 2021-02-04
Automatically reattach to buffers after reloading

### LINKED ISSUE #272: Resourcing $MYVIMRC breaks nvim-lsp
- `nvim --version`: `v0.5.0-60c581b`
- nvim-lsp version(commit hash): `94cce7448f34f172492d98ff6efe10986f73eb9e`
- `:checkhealth` result: Everything passes
- Operating system/version: macOS 10.15.5
- Terminal name/version: Terminal.app
- `$TERM`: xterm-256color

If I source my $MYVIMRC again lsp stops working. i.e omnifunc doesn't nothing and no diagnostics pop up.

```viml
  if !exists("g:nhooyr_lsp")
    "let g:nhooyr_lsp = 1
    lua << EOF
    local lsp_loaded, lsp = pcall(require, 'nvim_lsp')
    if not lsp_loaded then
      return
    end

    lsp.gopls.setup{}
    lsp.tsserver.setup{}
    lsp.vimls.setup{}
    lsp.clangd.setup{}
EOF
  endif
```

So I added an if guard around it to load only once.

mjlbach:
Closes #272 

Also depends on https://github.com/neovim/neovim/pull/13877/files

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -87,10 +87,12 @@ function configs.__newindex(t, config_name, config_def)
     M.cmd = config.cmd
 
     -- In the case of a reload, close existing things.
+    local reload = false
     if M.manager then
       for _, client in ipairs(M.manager.clients()) do
         client.stop(true)
       end
+      reload = true
       M.manager = nil
     end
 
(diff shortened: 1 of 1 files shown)


mjlbach:
@nhoory can you test both the neovim PR and this PR together to see if it addresses your issue? It should

mjlbach:
@nhooyr can you test both the neovim PR and this PR together to see if it addresses your issue? It should

mjlbach:
@teto This affects you too. If autostart is disabled, this PR will *not* restart the language servers on resourcing init.lua, you'll have to start them manually with the autostart() call. If autostart is enabled, then the language client will try to start/attach for all active buffers. Does that make sense?

teto:
as long as resourcing doesn't kill the active ones it's fine. Since this can get tricky, it it worth documenting somewhere.

mjlbach:
Resourcing currently (and has always) killed active servers, all this PR added was automatically re-attaching buffers, which it wouldn't have done before for either autostart true/false

PR #721, opened 2021-02-05
Add option to disable autostart

### LINKED ISSUE #691: ability to disable autostart
I would like to decouple the configuration of the language server and its autostart so that I can configure my LSPs globally but decide on a per project basis whether to launch the server or not (via .nvimrc for instance)

mjlbach:
Closes #691 

@teto The idea is to make an autocommand based on the root directory which invokes a wrapper around try_add that subsequently checks filetype.

1. Disable autostart when you configure your language server
```lua
local sumneko_cmd
if vim.fn.executable("lua-language-server") == 1 then
  sumneko_cmd = {"lua-language-server"}
else
  local sumneko_root_path = vim.fn.getenv("HOME").."/.local/bin/sumneko_lua"
  sumneko_cmd = {sumneko_root_path.."/bin/macOS/lua-language-server", "-E", sumneko_root_path.."/main.lua" }
end

nvim_lsp.sumneko_lua.setup {
  cmd = sumneko_cmd;
  autostart = false; -- This is the important new option
  on_attach = on_attach,
  capabilities = capabilities;
  settings = {
      Lua = {
          runtime = {
              -- Tell the language server which version of Lua you're using (LuaJIT in the case of Neovim)
              version = 'LuaJIT',
              -- Setup your lua path
              path = vim.split(package.path, ';'),
          },
          diagnostics = {
              -- Get the language server to recognize the `vim` global
              globals = {'vim'},
          },
          workspace = {
              -- Make the server aware of Neovim runtime files
              library = {
              },
          },
      },
  },
}
```
2. `:lua require('lspconfig').sumneko_lua.launch()` To activate the server (and subsequent autolaunch for buffers opened within this root) for the given root directory

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -55,13 +55,30 @@ function configs.__newindex(t, config_name, config_def)
     else
       trigger = "BufReadPost *"
     end
-    api.nvim_command(string.format(
-        "autocmd %s lua require'lspconfig'[%q].manager.try_add()"
-        , trigger
-        , config.name
-        ))
+    if not (config.autostart == false) then
+      api.nvim_command(string.format(
+          "autocmd %s lua require'lspconfig'[%q].manager.try_add()"
+          , trigger
+          , config.name
+          ))
+    end
 
     local get_root_dir = config.root_dir
+
+    function M.launch()
+      local root_dir = get_root_dir(api.nvim_buf_get_name(0), api.nvim_get_current_buf())
+      api.nvim_command(string.format(
+          "autocmd %s lua require'lspconfig'[%q].manager.try_add_wrapper()"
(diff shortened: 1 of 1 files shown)


mjlbach:
I guess it would be nice to have every buffer "just work" without triggering the autocmd, but I'm not sure the best thing to do for this.

teto:
thanks I will give it a spin. the addition of a launch command is appreciated.

teto:
thanks that works. I am a bit wary of `launch` that could be confusing with `start` ?

mjlbach:
I'm not sure what else to call it, it does a bit more than start since it watches for additional buffers within the workspace. Name suggestions welcome for sure

dbalatero:
@mjlbach I tested it and it worked for my restart case fyi!

PR #728, opened 2021-02-12
readme: Fix lsp_document_highlight augroup to work for buffers

lithammer:
Without this the `augroup` will be removed from old buffers when a new language server is attached. Found the solution here: https://stackoverflow.com/a/31822230/1862923


lithammer:
Wow, faster than Lucky Luke 😛

mjlbach:
I was just clicking refresh until you fixed this :P

PR #744, opened 2021-02-19
Implement some denols functions

kitagry:
Thank you for the nice plugin.

I added some implementations for denols. https://github.com/denoland/deno/tree/master/cli/lsp#custom-requests

- deno/cache
- deno/virtualTextDocument

`deno/cache`: This command will instruct Deno to attempt to cache a module and all of its dependencies. 

`deno/virtualTextDocument`: Requests a virtual text document from the LSP, which is a read only document that can be displayed in the client. This allows clients to access documents in the Deno cache, like remote modules and TypeScript library files built into Deno. The Deno language server will encode all internal files under the custom schema `deno:`, so clients should route all requests for the `deno:` schema back to the `deno/virtualTextDocument` API.

Furthermore, I added `DenolsDefinition` command. This command wraps `vim.lsp.buf.definition` command, and
 resolve `deno:` schema of the definition result.

Thank you.

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -1,7 +1,39 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
+local lsp = vim.lsp
 
 local server_name = "denols"
+
+local function buf_cache(bufnr)
+  local params = {}
+  params["referrer"] = { textDocument = { uri = vim.uri_from_bufnr(bufnr) } }
+  params["uris"] = {}
+  lsp.buf_request(bufnr, "deno/cache", params)
+end
+
+local function virtual_text_document_handler(uri, result)
+  if not result or #result ~= 1 then return nil end
+
+  local lines = vim.split(result[1].result, "\n")
+  local bufnr = vim.uri_to_bufnr(uri)
+
+  local current_buf = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+  if #current_buf ~= 0 then return nil end
+
+  vim.api.nvim_buf_set_lines(bufnr, 0, -1, nil, lines)
+  vim.api.nvim_buf_set_option(bufnr, "readonly", true)
+  vim.api.nvim_buf_set_option(bufnr, "modified", false)
+end
+
+local function virtual_text_document(uri)
+  local params = {
+    textDocument = {
+      uri = uri,
+    },
+  }
+  local result = lsp.buf_request_sync(0, "deno/virtualTextDocument", params)
+  virtual_text_document_handler(uri, result)
+end
 
 configs[server_name] = {
   default_config = {
@@ -12,6 +44,27 @@ configs[server_name] = {
       enable = true;
       lint = false;
       unstable = false;
+    };
+    handlers = {
+      ["textDocument/definition"] = function(err, method, result)
+        if not result or vim.tbl_isempty(result) then return nil end
+
+        for _, res in pairs(result) do
+          if string.sub(res.targetUri, 1, 7) == "deno://" then
+            virtual_text_document(res.targetUri)
+          end
+        end
+
+        lsp.handlers["textDocument/definition"](err, method, result)
+      end;
+    };
+  };
+  commands = {
+    DenolsCache = {
+      function()
+        buf_cache(0)
+      end;
+      description = "Cache a module and all of its dependencies."
     };
   };
   docs = {
@@ -26,4 +79,6 @@ Deno's built-in language server
   };
 }
 
+configs.denols.buf_cache = buf_cache
+configs.denols.virtual_text_document = virtual_text_document
 -- vim:et ts=2 sw=2


lua/lspconfig/denols.lua
```diff
@@ -1,8 +1,76 @@
...
 local util = require 'lspconfig/util'
+local lsp = vim.lsp
 
 local server_name = "denols"
 
+local function buf_cache(bufnr)
+  local params = {}
+  params['referrer'] = { textDocument = { uri = vim.uri_from_bufnr(bufnr) } }
+  params['uris'] = {}
+  lsp.buf_request(0, 'deno/cache', params)
```
lithammer, line 11:
```suggestion
  lsp.buf_request(bufnr, 'deno/cache', params)
```

lua/lspconfig/denols.lua
```diff
@@ -1,8 +1,76 @@
...
+  local lines = split(result[1].result, '\n')
+  local bufnr = vim.uri_to_bufnr(uri)
+
+  local current_buf = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+  if #current_buf ~= 0 then return nil end
+
+  vim.api.nvim_buf_set_lines(bufnr, 0, -1, nil, lines)
+  vim.api.nvim_buf_call(bufnr, function()
+    vim.cmd('set readonly')
+    vim.cmd('set nomodified')
```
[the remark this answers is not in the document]
lithammer: You could also use the `vim.bo` api.

```suggestion
    vim.bo.readonly = true
    vim.bo.modified = false
```

lua/lspconfig/denols.lua
```diff
@@ -1,8 +1,76 @@
...
+  default_definition_handler(_, method, result)
+end
+
+local function deno_definition()
+  lsp.handlers['textDocument/definition'] = function(m, method, result)
+    deno_definition_handler(m, method, result)
+    lsp.handlers['textDocument/definition'] = default_definition_handler
+  end
+  lsp.buf.definition()
+end
```
lithammer, line 72: Seems a bit fragile/dangerous/dirty to juggle the handlers like this 🤔

kitagry: Hmm. Would you like the followings?

```lua
configs[server_name] = {
  default_config = {
    ...
    handlers = {
      ['textDocument/definition'] = function(err, method, result)
        if not result or vim.tbl_isempty(result) then return nil end

        for _, res in pairs(result) do
          if string.sub(res.targetUri, 1, 7) == 'deno://' then
            virtual_text_document(res.targetUri)
          end
        end

        vim.lsp.handlers['textDocument/definition'](err, method, result)
      end;
    };
  };
```

lithammer: Well, it's not up to me. It was mostly an observation. But _I_ definitely prefer overriding server-specific handler rather than modifying the global one (like your example above).

Another alternative is to leave the `textDocument/definition` handler alone and implement everything in the `deno_definition` function. Something like this:

```lua
local util = require('vim.lsp.util')

local function deno_definition()
  local params = util.make_position_params()

  vim.lsp.buf_request(0, 'textDocument/definition', params, function(...)
    local result = select(3, ...) or {}

    for _, res in pairs(result) do
      if string.sub(res.targetUri, 1, 7) == 'deno://' then
        virtual_text_document(res.targetUri)
      end
    end

    vim.lsp.handlers['textDocument/definition'](...)
  end)
end
```

mjlbach:
This is big enough that I'm leaning towards "This would make a great deno.nvim plugin"

David-Else:
@mjlbach Did you check https://github.com/denoland/deno/tree/main/cli/lsp#custom-requests

> A client should implement these in order to have a fully functioning client that integrates well with Deno:

It would be great to just have denols work 100% out the box without needing yet another plugin.

mjlbach:
The point is that as servers implement more and more non-lsp specification functionality the maintenance burden will increase, especially since they have no obligation to maintain compatibility across versions to their custom extensions to the lsp specification. If we were, one day in the future, to bundle default language server configurations as we do filetypes (hypothetically, this isn't on the roadmap), this would never fly. 

I'm not 100% set on this, but in CoC-land each server gets it's own CoC-x plugin which handles iterating on the specification in step with upstream server changes , and that's not been happening with nvim. My hypothesis is that the defaults in nvim-lspconfig are usually "good enough" that the annoyances/unimplemented features don't sufficiently motivate people to create a X.nvim plugin.

Anyways, that's my piece, if people feel strongly that adding 1-2 commands is necessary for Deno, that's probably fine, but if we start getting where each server has 10+ custom commands, then these likely need to be broken out into their own plugins.

cloggier:
https://github.com/nvim-lua/lsp_extensions.nvim seems to be a good place for stuff like that perhaps?

kitagry:
I agree with you.
But, `deno/virtualTextDocument` is a necessary element for `textDocument/definition` in denols.
So, I think it is an essential protocol.

lua/lspconfig/denols.lua
```diff
@@ -1,8 +1,76 @@
...
 local server_name = "denols"
 
+local function buf_cache(bufnr)
+  local params = {}
+  params['referrer'] = { textDocument = { uri = vim.uri_from_bufnr(bufnr) } }
+  params['uris'] = {}
+  lsp.buf_request(bufnr, 'deno/cache', params)
+end
+
+local function split(str, delimiter)
```
mjlbach, line 14: Are you aware of  `:help vim.gsplit()`

lua/lspconfig/denols.lua
```diff
@@ -14,6 +82,20 @@ configs[server_name] = {
...
     };
   };
+  commands = {
+    DenolsCache = {
+      function()
+        buf_cache(0)
+      end;
+      description = "cache a module and all of its dependencies"
+    };
+    DenolsDefinition = {
```
mjlbach, line 92: Why is this a command? Shouldn't this just transparently override the definition handler?

lua/lspconfig/denols.lua
kitagry: I was not aware of it. Thank you!

lua/lspconfig/denols.lua
kitagry: It's ok. I'll fix it. Thank you.

mjlbach: Just curious. I added the initial deno support, but I don't actively use it, so you're the expert :)

lua/lspconfig/denols.lua
```diff
@@ -26,4 +81,6 @@ Deno's built-in language server
   };
 }
 
+configs.denols.buf_cache = buf_cache
```
mjlbach, line 84: What is the purpose of these two lines?

lua/lspconfig/denols.lua
```diff
@@ -13,6 +47,27 @@ configs[server_name] = {
...
+        lsp.handlers["textDocument/definition"](err, method, result)
+      end;
+    };
+  };
+  commands = {
+    DenolsCache = {
+      function()
+        buf_cache(0)
+      end;
+      description = "cache a module and all of its dependencies"
```
mjlbach, line 69: Maybe slightly more descriptive?

lua/lspconfig/denols.lua
```diff
@@ -1,8 +1,42 @@
...
+
+  local lines = vim.split(result[1].result, "\n")
+  local bufnr = vim.uri_to_bufnr(uri)
+
+  local current_buf = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+  if #current_buf ~= 0 then return nil end
+
+  vim.api.nvim_buf_set_lines(bufnr, 0, -1, nil, lines)
+  vim.api.nvim_buf_call(bufnr, function()
+    vim.bo.readonly = true
```
kitagry, line 25: Hmm, luacheck was failed. Is there the way to resolve it?

```
lua/lspconfig/denols.lua:25:5: setting read-only field bo.readonly of global vim
```

mjlbach:
Left a couple minor comments, I'll leave it to the deno crowd to tell me when this is ready.

lua/lspconfig/denols.lua
kitagry: I followed the method of texlab.

https://github.com/neovim/nvim-lspconfig/blob/8f642586e4213d257099704c3dacceb7f4d64627/lua/lspconfig/texlab.lua#L107-L108

lua/lspconfig/denols.lua
mjlbach: I don't know actually, I believe this  should work. @tjdevries do you have any ideas about this?

To work around you can try `:help nvim_buf_set_option`

lua/lspconfig/denols.lua
mjlbach: This actually predates my maintainership so I don't know how we use these 😆 I can take a look

mjlbach: I think it's fine

lua/lspconfig/denols.lua
kitagry: `nvim_buf_set_option` works well. Thank you!

mjlbach: I think this is probably the more "correct" way to do than buf_call, but it's weird luacheck gave you issues

mjlbach:
@lithammer @cloggier @David-Else Do you have any more comments?

Otherwise, @kitagry if you could squash your commits I'll merge tomorrow.

lua/lspconfig/denols.lua
lithammer: https://github.com/neovim/nvim-lspconfig/pull/746

mjlbach: Thank you, I saw. I think we should probably prefer nvim_buf_set_option generally since it's more specific, but I'd be in favor of merging #746

David-Else:
@kitagry Is it possible to add some very brief documentation to https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#dartls so users know the functions are available and how to access them?

mjlbach:
That should be fine semi automatically by docgen. Just to clarify, please don't edit config.md directly

kitagry:
Thank you for the review!

David-Else:
@kitagry If I use `:DenolsCache` when the denols server is running I get:
```
E5108: Error executing lua ....mount_nvimIFO6Oe/usr/share/nvim/runtime/lua/vim/lsp.lua:704: not found: "deno/cache" request handler for client "denols".
```
Am I using it in the right way, or is this a bug? I am using the Linux appimage.

Please advise how to test the other features too, cheers.
```
 nvim -v
NVIM v0.5.0-dev+1102-g595f6e4d6
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az207-238

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "
/home/runner/work/neovim/neovim/build/nvim.AppDir/usr/share/nvim"
```

kitagry:
@David-Else 
Reproduced. It's my mistake. Would you try #756 ?

David-Else:
@kitagry I tried to test your version by using:
```
    Plug 'kitagry/nvim-lspconfig', { 'branch': 'fix-deno-cache' }
```
but can't install it.

```
  x fix-deno-cache:
      Cloning into '/home/david/.config/nvim/plugged/fix-deno-cache'...
      fatal: repository 'https://github.com/kitagry/nvim-lspconfig/tree/fix-deno-cache.git/' not found
```
I am new to vimplug, please advise.

David-Else:
@kitagry I have tested it and can confirm it works :)

David-Else:
@kitagry How do I display `deno:/status.md`?

kitagry:
This is a bit difficult. Because neovim can't deal with `deno:/status.md` as uri.

```vim
lua vim.uri_to_buf("deno:/status.md")
" E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/uri.lua:116: URI must contain a scheme: deno:/status.md
```

And, deno lsp changed the behavior.

1. v1.7.4 and earlier returns deno:/// with three slashes
2. v1.7.5 returns deno:/ with only one slash

So, both of these behaviors need to be resolved. I'll open issue, and investigate how this is resolved in vim-lsp and coc.nvim

David-Else:
@kitagry Nice one! I had a look online and this thread seems very relevant:

https://github.com/denoland/deno/issues/9514

There are links to relevant coc issues in there too.

PR #746, opened 2021-02-20
luacheck: Allow setting vim.{g,b,w,t,v,env,o,bo,wo}

lithammer:


mjlbach:
Luacheck seems to still be very unhappy haha

lithammer:
> Luacheck seems to still be very unhappy haha

Yeah ended up having to add a bunch of stuff. Not sure if this is what we want? We would need to update `.luacheckrc` every time a new function or table is used. But maybe that doesn't happen very often...?

lithammer:
Passing now at leaast 🤷‍♂️

mjlbach:
I am mildly lukewarm on this, as I think the scenarios in which people will fail luacheck are also scenarios in which there exist non-syntactical sugar nvim_api calls which should be preferred anyways

mjlbach:
Closing because these setters are going to be replace by vim.opt, and we should encourage people to directly use the API anyways as it's more explicit

PR #751, opened 2021-02-22
Add configuration for Erlang Language Server

88maomao:


lua/lspconfig/erlangls.lua
```diff
@@ -0,0 +1,31 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+
+configs.erlangls = {
+  default_config = {
+    cmd = { "erlang_ls" };
+    filetypes = {"erlang"};
+    root_dir = function(fname)
+      return util.root_pattern("rebar.config", "erlang.mk", ".git")(fname) or vim.loop.os_homedir()
```
mjlbach, line 10: why homedir?

lua/lspconfig/erlangls.lua
```diff
@@ -0,0 +1,31 @@
...
+    end;
+    settings = {};
```
mjlbach, line 12: Why empty settings?

lua/lspconfig/erlangls.lua
```diff
@@ -0,0 +1,31 @@
...
+  };
+  docs = {
+    description = [[
+https://erlang-ls.github.io
+
+Language server for the Erlang.
```
mjlbach, line 18: delete *the* erlang

lua/lspconfig/erlangls.lua
```diff
@@ -0,0 +1,31 @@
...
+
+```lua
```
mjlbach, line 20: You don't need this since it's on your path

lua/lspconfig/erlangls.lua
88maomao: because of https://github.com/neovim/nvim-lspconfig/blob/80290ccc3a624b515541d1bc4e26257abf4951de/lua/lspconfig/elixirls.lua#L9

lua/lspconfig/erlangls.lua
88maomao: because of https://github.com/neovim/nvim-lspconfig/blob/80290ccc3a624b515541d1bc4e26257abf4951de/lua/lspconfig/sqls.lua#L11

lua/lspconfig/erlangls.lua
lithammer: > because of
> 
> https://github.com/neovim/nvim-lspconfig/blob/80290ccc3a624b515541d1bc4e26257abf4951de/lua/lspconfig/elixirls.lua#L9

I suspect it's incorrectly using `os_homedir`. It's basically never what you want and it's often preferred to use `util.path.dirname(fname)`.

lua/lspconfig/erlangls.lua
88maomao: What do you mean?

lua/lspconfig/erlangls.lua
88maomao: So what would be a better fallback in this case?

lithammer:
```lua
return util.root_pattern("rebar.config", "erlang.mk", ".git")(fname) or
  util.path.dirname(fname)
```

88maomao: thank you

mjlbach:
Can you add installation instruction to the docs? See rust-analyzer etc

mjlbach:
@88maomao ping me when you've added the installation docs and I'll merge

88maomao:
> Can you add installation instruction to the docs? See rust-analyzer etc

@mjlbach It is already there on line 15

mjlbach:
Ah, that's not really what I had in mind. More like "Erlang-ls must be built from source. Instructions are available on the main [repo](https://github.com/erlang-ls/erlang_ls). After installing the dependencies, Erlang OTP 21+ and rebar3 3.9.1+, and run make in the erlang-ls root directory."

Rust-analyzer was a bad example to point you to, my bad.

lua/lspconfig/erlangls.lua
```diff
@@ -0,0 +1,31 @@
...
+
+Installation instruction can be found [here](https://github.com/erlang-ls/erlang_ls).
+
+    Install requirements:
+      - [Erlang OTP 21+](https://github.com/erlang/otp)
+      - [rebar3 3.9.1+](https://github.com/erlang/rebar3)
+    Clone [erlang_ls](https://github.com/erlang-ls/erlang_ls)
+    Compile the project with `make` and copy resulting binaries somewhere in your $PATH eg. `cp _build/*/bin/* ~/local/bin`
+
+Language Server for Erlang.
```
mjlbach, line 25: Nit: Add this above around line 17 like before to match the styling.

Issue #754, opened 2021-02-23
reload workspace on rust project


etrnal70:
Newly added crates aren't detected unless reopening nvim or restarting server. I think restarting server isn't a good idea especially for large project since rust-analyzer always run cargo check the whole project.

I think rust-analyzer have cargo autoreload option in the settings, but it's not working. Other option is reloading workspace by triggering rust-analyzer/reloadWorkspace (coc have this).

I can have a PR though for the second one, it's adapted from texlab command and working so far, but i'm unsure if it's the desired choice

mjlbach:
I would go ahead and file the PR and we can discuss. What settings did you send to the server for autoreload? Is there a case where this is undesirable?

If it's not working (and we're doing things right on our end), we should file a bug upstream.

etrnal70:
Maybe you can test it ?

mjlbach:
Closed by #755

ibraheemdev:
How do I use the `CargoReload` command? `:CargoReload` gives me `Not an editor command: CargoReload`

ibraheemdev:
@etrnal70 
```lua
lsp.rust_analyzer.setup({
    on_attach = on_attach;
    settings = {
        ["rust-analyzer"] = {
            server = {
                path = "~/.cargo/bin/rust-analyzer"
            };
            cargo = {
                allFeatures = true
            }
        }
    }
})
```

etrnal70:
It's working for me here with `:CargoReload` :/

PR #755, opened 2021-02-23
Add :CargoReload to reload cargo workspace

etrnal70:
Solve #754

### DIFF
--- a/lua/lspconfig/rust_analyzer.lua
+++ b/lua/lspconfig/rust_analyzer.lua
@@ -4,7 +4,7 @@ local lsp = vim.lsp
 
 local function reload_workspace(bufnr)
   bufnr = util.validate_bufnr(bufnr)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/rust_analyzer.lua
```diff
@@ -1,5 +1,15 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
+local lsp = vim.lsp
+
+local function reload_workspace(bufnr)
+  bufnr = util.validate_bufnr(bufnr)
+  lsp.buf_request(bufnr, 'rust-analyzer/reloadWorkspace', _,
+      function(err, _, result, _)
+        if err then error(tostring(err)) end
+        print("Workspace reloaded")
```
lithammer, line 10: I think maybe this should be `vim.notify` rather than `print`.

jack0son:
How do you enable this command? Running `:LspCargoReload` gives me `not an editor command`.

PR #757, opened 2021-02-24
Set terraformls to also operate on `hcl` files

WhyNotHugo:
The file format that terraformls checks is HCL (Hashicorp Configuration
Language). Tools like `vim-hcl` actually set this `filetype` correctly.

Make the LSP pick up these files correctly too.

### DIFF
--- a/lua/lspconfig/terraformls.lua
+++ b/lua/lspconfig/terraformls.lua
@@ -4,7 +4,7 @@ local util = require 'lspconfig/util'
 configs.terraformls = {
   default_config = {
     cmd = {"terraform-ls", "serve"};
-    filetypes = {"terraform"};
+    filetypes = {"terraform", "hcl"};
     root_dir = util.root_pattern(".terraform", ".git");
   };
   docs = {


lithammer:
I don't think this is correct, `tf` is something else and has nothing to do with Terraform. You need https://github.com/hashivim/vim-terraform or https://github.com/sheerun/vim-polyglot for Neovim to correctly set the filetype.

WhyNotHugo:
I'm not using `vim-terraform` (don't want any of the features it offers really), so `vim` picks these files up as `tf`.

Does this PR actually sound wrong though? The file extension is typically `.tf`, and the file format is `hcl`. It seems wrong that the LSP **only** works on a `filetype=terraform`, but not `tf` (the extension) nor `hcl` (the file format).

mjlbach:
We literally use what's in filetype as a vim `:help filetype`, we're not matching on the extension.

lithammer:
The PR in its current form wrong for the same reason that using something like `ft=python` is wrong. `ft=tf` actually refers to [TinyFugue](http://tinyfugue.sourceforge.net). So _your_ Vim thinks you're coding a TinyFugue file and not HCL/Terraform. Therefore starting `terraform-ls` is wrong. You wouldn't start [python-language-server](https://github.com/palantir/python-language-server) when editing a Ruby file.

Relying on the file extension can be tricky since as you've noticed, it's ambiguous. We can't make a blanket decision that _all_ `.tf` files are Terraform files, because evidently, they're not. Even if it might be true most of the time.

Since the `tf` syntax file is of no use when editing Terraform files (probably even does more harm than good). You can solve it by setting the [modeline](https://vimhelp.org/options.txt.html#modeline), e.g. `vi: ft=terraform`. Alternatively create an `autocmd`, something like:

```vim
autocmd BufRead *.tf setlocal=terraform
```

Starting the server for `ft=hcl` has some merit though I guess. Especially if you happen to have a syntax file  for`ft=hcl` (e.g. [vim-hcl](https://github.com/b4b4r07/vim-hcl)), but not for `ft=terraform`.

lithammer:
And since this plugin is meant to be easy to extend/configure. You can also override `filetypes` to suit your use-case:

```lua
lspconfig.terraformls.setup({
  filetypes = {"terraform", "tf"},
})
```

WhyNotHugo:
I was unaware of TinyFugue. Out-of-the-box, nvim pick up `something.tf` as `ft=tf`, hence my confusion (the highlighting seemed very odd, but somehow worked for me :sweat_smile: ).

I've updated the PR to set make `terraformls` run on `hcl` files too. This plays nicely with `vim-hcl` too.

WhyNotHugo:
BTW, thanks for all the clear feedback here!

lithammer:
> I was unaware of TinyFugue.

You and me both 😉 Never heard of it until today. Last time I played MUD was like 20 years ago.

lithammer:
> I've updated the PR to set make `terraformls` run on `hcl` files too. 

Please update the PR title and description to reflect this.

lithammer:
I should perhaps add that I'm not actually a maintainer. So take what I say with a grain of salt 😛

WhyNotHugo:
Updated.

> I should perhaps add that I'm not actually a maintainer.

No worries, your points are still valid.

mjlbach:
> I should perhaps add that I'm not actually a maintainer. So take what I say with a grain of salt 😛 

@lithammer you're basically a maintainer in my view 😃

mjlbach:
@lithammer (now it's not just in my view ;) )

mjlbach:
@lithammer Was the consensus we should merge this? I have never used terraform

lithammer:
Hmm, it's not clear-cut since not all HCL files are Terraform files. But the downsides are few I guess. So I'm in favour of merging. We can always revert if we start receiving issues.

Issue #25, opened 2019-11-17
Add support for settings schema from VSCode package.json


norcalli:
https://github.com/rust-lang/rls-vscode/blob/master/package.json
https://github.com/lampepfl/dotty/blob/master/vscode-dotty/package.json
https://github.com/mauve/vscode-terraform/blob/master/package.json

There is a very simple part of the existing VSCode extensions that we can automatically adopt by parsing `contributes.configuration`. There are other sections that we can get as well.

I'm considering adding this to `docs` just for auto generating documentation about what the default settings are, but we could use the same schema to provide client side validation of the settings as well.

So currently, I think this will be sufficient:

```lua
docs = {
  package_json = "https://github.com/rust-lang/rls-vscode/blob/master/package.json";
}
```

and then we can pull it from docgen (and cache it in case it hasn't changed). 

Also, we can put these in the

<details>
<summary>Default server settings:</summary>
<pre>
nvim_lsp[server].setup {
  settings = {
    ...
  }
}
</pre>
</details>

since these settings will often be long.

norcalli:
Implemented in #29

norcalli:
Follow up in #49. Adds the ability to scrape the vspackage itself from microsoft directly, which is more accurate. We should prefer that in the future.

justinmk:
@norcalli  What's remaining to be done here?

saadparwaiz1:
Hi @mjlbach. If it's fine with you I want to try and attempt something for this issue? I currently have a partially complete attempt at this [here](https://github.com/saadparwaiz1/validate-nvim-lsp-settings) if you like this approach let me know and I can work on this further and make a PR.

The approach is pretty simple use the package.json such as one for [pyls](https://github.com/saadparwaiz1/validate-nvim-lsp-settings/blob/main/json/pyls.json)  and convert it to an equivalent [lua](https://github.com/saadparwaiz1/validate-nvim-lsp-settings/blob/main/lua/server_settings/pyls.lua) table. Then validation should be as simple as iterating over the keys in user settings to the default server settings.

I am pretty new to using lua, so the codebase isn't that great. but I can try and improve things

mjlbach:
Sounds terrific to me. I lean towards it being a helper function which isn't loaded by default that people can call with their current configs table to check if all settings are valid. No runtime penalty for those who already have working configs and the like (people could wrap/call the helper in their config if they want).

saadparwaiz1:
[...]

PR #758, opened 2021-02-24
Attempt at server setting validation

saadparwaiz1:
The PR will add validate_server_settings to utils, letting users validate server specific settings. 

- [x] Create Lua Table For Default Settings From Package.json
- [x] Create Basic Implementation for validate_server_settings
- [x]  Improve validate_server_settings to check nested tables

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -250,15 +250,40 @@ local function read_package_json(package_json)
   return settings
 end
 
+local function deep_equal(tbl_a, tbl_b)
+  for k,v in pairs(tbl_b) do
+    if tbl_a[k] == nil then
+      return false
+    elseif type(v) == "table"  and tbl_a[k]['type'] == nil then
+      if not deep_equal(tbl_a[k], v) then
+        return false
+      end
+    else
+      local v_type = type(v)
+      local tbl_v_type = type(tbl_a[k]['default'])
+
+      if tbl_a[k]['type'] ~= nil and tbl_v_type == "nil" then
+        tbl_v_type = tbl_a[k]['type']
+        tbl_v_type = string.gsub(tbl_v_type, "object", "table")
+        tbl_v_type = string.gsub(tbl_v_type, "array", "table")
+        tbl_v_type = string.gsub(tbl_v_type, "integer", "number")
+      end
+
+      if v_type ~= tbl_v_type then
+        return false
+      end
+    end
+  end
+  return true
+end
+
 function M.validate_server_settings(server_settings, package_json)
   -- TODO (saad parwaiz) improve validation
   local default_server_settings = read_package_json(package_json)
-  for k,_ in pairs(server_settings) do
-    if default_server_settings[k] == nil then
-      return {}
-    end
+  if deep_equal(default_server_settings, server_settings) then
+    return server_settings
   end
-  return server_settings
+  return {}
 end
 
 -- Returns a function(root_dir), which, when called with a root_dir it hasn't
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -108,6 +108,7 @@ function M.has_bins(...)
   return true
 end
 
+
 -- Some path utilities
 M.path = (function()
   local function exists(filename)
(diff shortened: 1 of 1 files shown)


saadparwaiz1:
@mjlbach here's the attempt for #25. I still need to implement validation_server_settings properly, so it checks for nested tables but this is the basic idea.

Here's how I'm currently using it in my config with sumnko_lua:

require('lspconfig').sumneko_lua.setup{
	cmd = {"lua-langserver"},
	on_attach = on_attach,
	capabilities = capabilities,
	settings = require('lspconfig.util').validate_server_settings({
		Lua = {
			runtime = {
				version = 'LuaJIT',
				path = vim.split(package.path, ';')
			},
			workspace = {
				library = {
				},
			},
			diagnostics = {
				globals = { 'vim' }
			}
		}
	}, '/Users/saadparwaiz/package.json')
}
``

saadparwaiz1:
Think it more or less works now. if there are any changes you want to make or if you want to test it out please let me know

mjlbach:
Can you add a helper function that reads package.json's url from the server config?

mjlbach:
Also we should add a helper that iterates over all servers and checks them using their attached package.json (if it is provided)

saadparwaiz1:
for the helper method that reads the url should it also download the package.json to the stdpath('cache')?

mjlbach:
I would say yes, or mimc the method that the lua docgen uses

mjlbach:
Can't we just directly stream json from the url?

saadparwaiz1:
yeah, we probably can! sounds good I'll get started on it

lua/lspconfig/util.lua
```diff
@@ -223,6 +223,68 @@ M.path = (function()
...
+  file:close()
+
+  local settings = {}
+  for k, v in pairs(content) do
+      set_value(settings, k, v)
+  end
+  return settings
+end
+
+local function deep_equal(tbl_a, tbl_b)
```
mjlbach, line 253: Is there a reason you can't use vim.deep_equal here?

saadparwaiz1: I didn't know it existed! I'll have a look. Sorry for not working on this, been busy with a few other things recently :(

mjlbach: Lol it's ok. I have ICCV coming up so I haven't been active as well.

tamago324:
Hello, I'm working on a plugin called [nlsp-settings.nvim](https://github.com/tamago324/nlsp-settings.nvim).
This plugin provides a mechanism to allow completion like coc-settings.json.
Would it be possible to integrate this into nvim-lspconfig? Or would it be better to continue the project as [nlsp-settings.nvim](https://github.com/tamago324/nlsp-settings.nvim)?

Also, I thought the idea of loading the config from the URL package.json was great!

Thank you!

mjlbach:
This looks terrific! You should definitely add it to the wiki.

I would say there are probably parts that should stay separate (anything to do with autocompletion, comp-e, snippets, etc). If you want to move functions for reading/validating settings (similar to the spirit of this PR) into nvim-lspconfig, I think that's ok? 

We try to keep nvim-lspconfig as "stock" as possible and dependency free, so as to not get in the way of plugin authors like yourself :)

tamago324:
Thank you! I'll add it to the wiki later!

The function for reading settings might be better to be in nvim-lspconfig. If I can successfully implement it, I'll send a pull request or comment here.

I like the idea of eliminating dependencies and keeping it small. I agree with that idea.

kabouzeid:
I agree that a utility function to read the settings from json would fit quite well into nvim-lspconfig. Managing and autocompletion for the JSON configs are better dealt with outside of this project as it is now. nlsp-settings looks great, thanks @tamago324.

mjlbach:
@tamago324 @saadparwaiz1 Any updates? I'm aiming for 0 open issues before the 0.5 release 😆

saadparwaiz1:
> @tamago324 @saadparwaiz1 Any updates? I'm aiming for 0 open issues before the 0.5 release 😆

I'm currently in the middle of exam season :( I'd love it if someone can take this over

mjlbach:
No worries at all, just wanted to check before I took things over :)

tamago324:
@mjlbach I have created a [script](https://github.com/tamago324/nlsp-settings.nvim/blob/main/scripts/gen_schemas.lua) to generate a JSON file from package.json for use with JSON schema.
I hope you find it useful.

Simply out of curiosity, if you use JSON Schema to validate, are you assuming that you will use a library like [jsonschema](https://github.com/api7/jsonschema)?

mjlbach:
I'm not sure this feature warrants a whole dependency on an external parsing library. I think what saadparwaiz1 already did is basically what norcalli originally had in mind, it just needs a bit of cleanup and maybe an additional helper function to validate the configs of all .setup language servers.

tamago324:
@mjlbach Thank you very much for your answer.
I understand now.

folke:
Maybe this could be integrated in `:checkhealth`?

PR #760, opened 2021-02-24
lspinfo: add neovim log path in header

teto:
I am usually against this type of overengineering but LSP configuration is so complex that it may be a good idea to parse the server commands and look for their logfile as well so we can display it when we see it.
For instance hls has:
```
	settings = {
			haskell = {
			completionSnippetsOn = false,
			formattingProvider = "stylish-haskell",
				logFile = "/tmp/nvim-hls.log",
				hlintOn = false
		}
	}
```
Or even more generic, a per server debug function.

### DIFF
--- a/lua/lspconfig/lspinfo.lua
+++ b/lua/lspconfig/lspinfo.lua
@@ -21,6 +21,7 @@ return function ()
 
   local header = {
     "Configured servers: "..table.concat(vim.tbl_keys(configs), ', '),
+    "Neovim logs at: "..(vim.lsp.get_log_path()),
     "",
     tostring(#buf_clients).." client(s) attached to this buffer: "..table.concat(buf_client_names, ', '),
   }


mjlbach:
I've also thought about logging separate clients to their own log files as well (this would make it much more readable), maybe as an option. Overall I agree with this PR though (I think someone brought this up on reddit).

teto:
That person must be quite clever :)
once we have nvim_log in the API, everything will go to the same file anyway so not sure about a per-server neovim log.
Just to make sure, what I meant was haskell-language-server has its own logfile configured via lsp settings` logFile = "/tmp/nvim-hls.log",` and maybe we could print it along with haskell-language-server configuration.

mjlbach:
Doxed! 

That makes sense, I think we could also implement (in a plugin ofc) some nice filtering of the lsp.log file to only show a given session/server's log files. Maybe using telescope or the like.

PR #762, opened 2021-02-27
Add hls specific lspinfo

teto:
first commit removes deprecated suttf, second one is an experiment to add server specific lspinfo.


mjlbach:
LGTM, we might want to document writing a "good" configuration at some point to note the features we've added, like server specific info, custom commands, custom handlers, etc.

PR #765, opened 2021-03-03
Fix denols textDocument/definition in deno 1.8

### LINKED ISSUE #761: denols: cannot jump to definition in deno v1.7.5
related https://github.com/neovim/nvim-lspconfig/pull/744#issuecomment-785751001

- `nvim --version`: v0.5.0-dev
- nvim-lsp version(commit hash): 4d5dbea4f
- What language server (If the problem is related to a specific language server): denols
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): no
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): no
- Is the problem isolated to a particular language server: yes
- Operating system/version: windows10 wsl ubuntu20.04

```
$ deno --version
deno 1.7.5 
```

main.ts

```ts
import { createApp } from "https://servestjs.org/@v1.1.9/mod.ts";
const app = createApp();
```

And, move cursor to `createApp` in line 2. And then, run `:lua vim.lsp.buf.definition()`.

### Actual behaviour

returns error.

```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/uri.lua:116: URI must contain a scheme: deno:/https/servestjs.org/%40v1.1.9/app.ts
```

### details

This error occurs because denols change the behavior from v1.7.5.

1. v1.7.4 and earlier returns `deno:///` with three slashes
2. v1.7.5 returns `deno:/` with only one slash

This issue is related to https://github.com/denoland/deno/issues/9514.
I'll investigate how this is resolved in vim-lsp and coc.nvim.
Thank you.

kitagry:
fix #761 

`lua vim.lsp.buf.definition()` works in normal buffer, but doesn't work in `deno://` buffer.
`lua require('lspconfig').denols.definition()` or `:DenolsDefinition` works in `deno://` buffer.

And, I'd fixed `vim.lsp.buf.references`, too.

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -5,8 +5,7 @@ local lsp = vim.lsp
 local server_name = "denols"
 
 local function deno_uri_to_uri(uri)
-  -- 1.7.4 LS returns deno:///https/deno.land/std@0.85.0/http/server.ts
-  -- 1.7.5 LS returns deno:/https/deno.land/std%400.85.0/http/server.ts
+  -- denols returns deno:/https/deno.land/std%400.85.0/http/server.ts
   -- nvim-lsp only handles deno://
   if string.sub(uri, 1, 6) == "deno:/" and string.sub(uri, 7, 7) ~= "/" then
     return string.gsub(uri, "^deno:/", "deno://", 1)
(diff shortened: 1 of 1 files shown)


David-Else:
@kitagry Thanks for all this deno work!

I don't fully understand the code, but looking at the comments, are you adding backward compatibility for `1.7.4`? If so, I don't think it is needed. It might be a good idea to add a minimum requirements `deno 1.8` version to the docs?

Could you test your code with the new version of Deno `1.8` https://github.com/denoland/deno/releases/tag/v1.8.0 ?

kitagry:
I'm sorry for the late reply.

Yes, you are correct. I'll fix the comment.

I'd tested this with Deno1.8, and it works well:tada:

PR #766, opened 2021-03-05
Check if lsp cmd is a string

Sh3Rm4n:
If `config.cmd[1]` is not a string type, vim.fn.executable() will throw an error, when LspInfo is called.

### DIFF
--- a/lua/lspconfig/lspinfo.lua
+++ b/lua/lspconfig/lspinfo.lua
@@ -83,7 +83,7 @@ return function ()
   vim.list_extend(buf_lines, matching_config_header)
   for _, config in pairs(configs) do
     local cmd_is_executable, cmd
-    if config.cmd then
+    if config.cmd and type(config.cmd[1]) == "string" then
       cmd = remove_newlines(config.cmd)
       if vim.fn.executable(config.cmd[1]) == 1 then
         cmd_is_executable = "True"


mjlbach:
I'm a little confused why we would do this, as this short-circuits one of the debugging feedback we do.

Sh3Rm4n:
Well to be honest, I have not given it much thought. 
I had the problem, that `vim.fn.executable()` gets a non-string and will throw a vim string error `E4somethingsomething`, when I ran `:LspInfo` and this is the "fix" I came up with, to suppress this error. 

I haven't really methodically debugged this, so I can't say for sure, why a non-string was passed. 
I'm fine, if this solution is not the right way to go, but me as a user expected at a more insightful error. 

What do you think would be the better approach?

mjlbach:
There is no situation in which the first entry to `cmd` should be an int, cmd should be table of strings, where the first string is either the absolute path or the name of an executable on your path.

Sh3Rm4n:
> cmd should be table of strings, where the first string is either the absolute path or the name of an executable on your path.

That is correct, but I had a situation, where this clearly wasn't the case. And it resulted into an error when I executed `:LspInfo`.

I do understand, that when the `LspInfo` code is executed, it has to assume, that `cmd` is a valid list of strings, a.k.a. executable with arguments, but as `:LspInfo` is kind of a debug tool for the lsp configuration, it would be helpful if it displayed a useful error. 

My approach is not good either, because I just suppress the error, but in case of configuration error like these, I would see some information printed into the floating window. But now, for me with my specific configuration (might be buggy!) at least, it throws the vim error (E928 string required to be exact) and nothing is printed in the floating window.

PR #767, opened 2021-03-05
add filetype in LspInfo floatwindow

glepnir:
@mjlbach  add filetype in lspinfo float window. it will be useful for some indent plugins


mjlbach:
Seems good to me :)

glepnir:
morning? evening here:)

mjlbach:
Just woke up actually haha, lspsaga beautiful lspinfo replacement when??

glepnir:
just use for my indent-guides plugin. maybe can do  lspinfo window more beautiful in lspsaga .try to implement it later

mjlbach:
I would wait because I can refactor lspinfo so the logic portion is decoupled from the UI portion

PR #769, opened 2021-03-06
add phpactor

andersondanilo:


dantleech:
I've just tried this with nvim nightly - phpactor starts and shows indexing message, but get this with `:LspInfo`:

```
E5108: Error executing lua ...e/daniel/.vim/plugged/nvim-lsp/lua/lspconfig/lspinfo.lua:56: attempt to index field 'workspaceFolders' (a nil value)
```

And no messages seem to get sent for other operations.

Phpactor doesn't support workspace folders at the moment - but it's not a requirement for LSP.

mjlbach:
I added the workspace folder thing to both LspInfo and core, it should be agnostic to whether the server supports workspace folders so not sure why that is happening. We just use it to read the root directory.

jesseleite:
Seems to work great! I can jump to definitions, and I'm seeing phpactor completions with nvim-compe 🤘

I do get this error when running `:LspInfo`, but I get the same error when running with intelephense, so I'm not suire it's anything phpactor specific...

CleanShot 2021-03-20 at 17 53 38

mjlbach:
That's because the PR needs to be rebased

jesseleite:
@andersondanilo Should phpactor be showing inline diagnostics like this though? These show for me with intelephense, but not with phpactor 🤔

CleanShot 2021-03-20 at 18 50 05

andersondanilo:
@jesseleite, It's supported: https://phpactor.readthedocs.io/en/develop/lsp/support.html

However, i tested right now and i'ts not displayed, i did not notice before since i use another plugin for syntax and linter.

> @andersondanilo Should phpactor be showing inline diagnostics like this though? These show for me with intelephense, but not with phpactor 
> 
> CleanShot 2021-03-20 at 18 50 05

mjlbach:
Seems like you need PHPStan installed based on the table you linked:

 https://github.com/phpactor/language-server-phpstan-extension

I don't do PHP, so I won't be able to debug this. FYI inline diagnostics are just rendered diagnostics, so this likely means the server is not sending diagnostics (which supports my phpstan theory)

jesseleite:
@andersondanilo I just installed phpstan as a phpactor extension via:

```php
phpactor extension:install "phpactor/language-server-phpstan-extension"
```

That alone doesn't render the diagnostics. Is there something missing in the nvim config for this? 🤔

Issue #729, opened 2021-02-13
help wanted for powershell server


jmwatte:
Not so good at Lua, I guess, but when I put [[....]] and everything between it  

`
cmd={[["C:/Program Files/PowerShell/7/pwsh.exe" -NoLogo -NoProfile -ExecutionPolicy RemoteSigned -File "C:/users/jmwinters/.vscode/extensions/ms-vscode.powershell-2020.6.0/modules/PowerShellEditorServices/Start-EditorServices.ps1" -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -LogPath "C:/users/jmwinters/Desktop/pses.log.txt" -LogLevel Diagnostic -BundledModulesPath "C:/users/jmwinters/.vscode/extensions/ms-vscode.powershell-2020.6.0/modules" -Stdio -SessionDetailsPath "C:/users/jmwinters/Desktop/.pses_session"]]}
`

in an online lua coder and I paste the output straight into a cmd prompt in windows 10, it starts up the service.  But when I use it like above to configure the server it , I get this checkhealth message ...

`
## Checking language server protocol configuration
  - INFO: vimls: configuration checked.
  - ERROR: powershell_lsp: The given command "C:/Program Files/PowerShell/7/pwsh.exe -NoLogo -NoProfile -ExecutionPolicy RemoteSigned -File \"C:/users/jmwinters/.vscode/extensions/ms-vscode.powershell-2020.6.0/modules/PowerShellEditorServices/Start-EditorServices.ps1\" -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -LogPath \"C:/users/jmwinters/Desktop/pses.log.txt\" -LogLevel Diagnostic -BundledModulesPath \"C:/users/jmwinters/.vscode/extensions/ms-vscode.powershell-2020.6.0/modules\" -Stdio -SessionDetailsPath \"C:/users/jmwinters/Desktop/.pses_session\"" is not executable.

`
Not executable... So.... something gets lost in translation, I guess. Checked the shell and it is cmd.exe on windows 10. Any help would be appreciated.

jmwatte:
Didn't want the bug label there. Help label, maybe...

mjlbach:
This belongs in the discussion forum. Command needs to be a table, where each entry in the table is either a binary path (the first) or argument to the binary path. You have spaces, which is why the executable check fails.

sakhnik:
Any plans to include PS support out of the box?

[...]

PR #779, opened 2021-03-14
Add configuration for PowerShellEditorServices

sakhnik:
Follow up #729
Please let me know what to elaborate on, I can keep improving and testing it.

### DIFF
--- a/lua/lspconfig/powershell_es.lua
+++ b/lua/lspconfig/powershell_es.lua
@@ -1,0 +1,61 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = "powershell_es"
+local temp_path = vim.fn.stdpath('cache')
+
+local function make_cmd(bundle_path)
+  if bundle_path ~= nil then
+    local command_fmt = [[%s/PowerShellEditorServices/Start-EditorServices.ps1 -BundledModulesPath %s -LogPath %s/powershell_es.log -SessionDetailsPath %s/powershell_es.session.json -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
+    local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
+    return {"pwsh", "-NoLogo", "-NoProfile", "-Command", command}
+  end
+end
+
+configs[server_name] = {
+  default_config = {
+    on_new_config = function(new_config, _)
+      local bundle_path = new_config.bundle_path
+      new_config.cmd = make_cmd(bundle_path)
+    end,
+    filetypes = {"ps1"},
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname) or vim.fn.getcwd()
+    end;
+  };
+  docs = {
+    description = [[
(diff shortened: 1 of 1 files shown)


mjlbach:
CONFIG.md is autogenerated, so your changes there will get overwritten.

lua/lspconfig/pses.lua
```diff
@@ -0,0 +1,45 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = "pses"
+local temp_path = vim.fn.stdpath('cache')
+
+local function make_cmd(bundle_path)
+  if bundle_path ~= nil then
+    local command = bundle_path .. [[/PowerShellEditorServices/Start-EditorServices.ps1 -BundledModulesPath ]] .. bundle_path .. [[ -LogPath ]] .. temp_path .. [[/pses.log -SessionDetailsPath ]] .. temp_path .. [[/pses.session.json -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
```
mjlbach, line 9: This is a little messy, string.format might be a better choice with this many substitutions.

mjlbach:
Otherwise I think this is probably fine :)

sakhnik:
> CONFIG.md is autogenerated, so your changes there will get overwritten.

I used the `scripts/docgen.sh` by the way. But I read this as I should leave it intact.

lua/lspconfig/pses.lua
sakhnik: I can fix this

mjlbach:
Oh that's fine then, I would remove it from the commit as it keeps the history cleaner, but it's otherwise fine.

lua/lspconfig/pses.lua
```diff
@@ -0,0 +1,46 @@
...
+      return util.find_git_ancestor(fname) or vim.fn.getcwd()
+    end;
+  };
+  docs = {
+    description = [[
+https://github.com/PowerShell/PowerShellEditorServices
+
+Language server for PowerShell.
+
+The default configuration doesn't set cmd unless `bundle_path` is specified.
```
mjlbach, line 32: I'd make this a bit more clear, mention how to install the server (if it's simple) and then mention briefly what bundle path is.

lua/lspconfig/pses.lua
```diff
@@ -0,0 +1,46 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = "pses"
```
mjlbach, line 4: Debating whether it should be powershell_es, because this will be the entry in table of contents so it might be more descriptive

sakhnik: Agree

lua/lspconfig/pses.lua
sakhnik: Done

sakhnik:
Thanks!

Liquidmantis:
Huge thanks to @sakhnik for this contribution!  Powershell pays the bills, and I recently went all-in on the native Neovim LSP, gutted CoC and therefore coc-powershell from my config, and had to start a new Powershell project today.  I found this while looking for documentation on configuring the Powershell ES language server and it makes setup trivial.  Thanks!

PR #783, opened 2021-03-16
Beancount Support

polarmutex:
I am the maintainer of the beancount language server and would like to add the config for the lang server to this repo


mjlbach:
So you are, and so it shall be.

PR #789, opened 2021-03-22
refactor(vala): refine vala project root detection

ttys3:
take `https://github.com/phw/peek` or `https://github.com/benwaffle/vala-language-server` for example,

```shell
git clone https://github.com/benwaffle/vala-language-server
cd vala-language-server
nvim
```
we opened `src/server.vala` then the lsp will run `meson build` under the wrong dir `src/`

because the `src` dir has a file named `meson.build` which actully is not a real `meson.build`,
it is included by the root `meson.build`, and the lsp server will complain the "ERROR: First statement must be a call to project"

this PR:

1. detect the valid `meson.build` file by detection first line include `project(` detective or not
~~2. prefer cwd if root is a descendant, otherwise fallback to git repo~~

### DIFF
--- a/lua/lspconfig/vala_ls.lua
+++ b/lua/lspconfig/vala_ls.lua
@@ -1,13 +1,34 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
+local meson_matcher = function (path)
+  local pattern = "meson.build"
+  local f = vim.fn.glob(util.path.join(path, pattern))
+  if f == '' then
+    return nil
+  end
+  for line in io.lines(f) do
+    -- skip meson comments
+    if not line:match('^%s*#.*') then
+      local str = line:gsub('%s+', '')
+      if str ~= '' then
+        if str:match('^project%(') then
+          return path
+        else
+          break
+        end
+      end
+    end
+  end
+end
+
 configs.vala_ls = {
   default_config = {
     cmd = {'vala-language-server'},
     filetypes = {'vala', 'genie'},
-    root_dir = function(fname)
-      return util.root_pattern("meson.build")(fname)
-        or util.find_git_ancestor(fname)
+    root_dir = function (fname)
+      local root = util.search_ancestors(fname, meson_matcher)
+      return root or util.find_git_ancestor(fname)
     end,
   },
   docs = {


lua/lspconfig/vala_ls.lua
```diff
@@ -5,9 +5,25 @@ configs.vala_ls = {
...
+            if str ~= '' and str:match('^project%s*%(')  then
+              return path
+            else
+              break
+            end
+          end
+        end
+      end)
+      -- prefer cwd if root is a descendant, fallback to git repo
+      return (util.path.is_descendant(cwd, root) and cwd or root) or
```
lithammer, line 25: Why prefer cwd over `util.path.dirname(fname)`?

ttys3: Removed and force pushed, please check again

lua/lspconfig/vala_ls.lua
```diff
@@ -5,9 +5,25 @@ configs.vala_ls = {
...
     cmd = {'vala-language-server'},
     filetypes = {'vala', 'genie'},
-    root_dir = function(fname)
-      return util.root_pattern("meson.build")(fname)
-        or util.find_git_ancestor(fname)
+    root_dir = function (fname)
+      local root = util.search_ancestors(fname, function (path)
+        local pattern = "meson.build"
+        local f = vim.fn.glob(util.path.join(path, pattern))
+        if not util.path.exists(f) then
```
lithammer, line 12: You can skip a syscall by just checking if `f` is an empty string since `glob` has already verified that the path exists.

```suggestion
        if f ~= '' then
```

lua/lspconfig/vala_ls.lua
```diff
@@ -5,9 +5,25 @@ configs.vala_ls = {
...
+          return nil
+        end
+        for line in io.lines(f) do
+          local str = line:gsub('%s+', '')
+          if str ~= '' then
+            if str:match('^project%s*%(') then
```
lithammer, line 18: I'm not at all familiar with Vala. But is there no CLI command that can be used instead? For example, for Rust you can do `cargo metadata` to get the workspace/project root.

ttys3:
I'm afraid there's no tool to use to report the workspace root.

according to <https://wiki.gnome.org/Projects/Vala/Tools>

> Meson - a front end to the Ninja build system with Vala support (The main build system for Vala)

and ref to <https://mesonbuild.com/Syntax.html#comments>

> A comment starts with the `#` character and extends until the end of the line.

and also the `meson.build` file "First statement must be a call to project"

ttys3:
split the logic part into `meson_matcher()`

lithammer (approved):

lithammer:
Thanks!

PR #792, opened 2021-03-22
Added the dot language server to the mix

DoktorMike:
- Following recommendations and removing default conf
- Following existing language servers with default root_dir


DoktorMike:
This is a new PR as I messed up the squashing of the old commits.. 😧

mjlbach:
Oh nice, for future reference I just do `git rebase -i master`, use block mode to change all of the commits (except the first one) from pick to squash, and then `:wqa` and `git push -f`

PR #793, opened 2021-03-24
sumneko: do not send telemetry data by default

mhinz:
By default, sumneko sends telemetry data containing a [randomized but unique identifier](https://github.com/sumneko/lua-language-server/blob/master/script/service/telemetry.lua#L12) to a Chinese IP. The token can be found in `lua-language-server/log/token`.

The string sent looks like this in my case (I X'ed the identifier):

```
$...pulse.XXXXXXXXXXXXXXXX.Neovim 0.5.0.J...platform.XXXXXXXXXXXXXXXX.macOS 64.libc++ libc++ 10000.clang Clang 12.0.0.
```

In case of errors, that log is sent, too. You can visit the IP to get a summary of the collected data: http://119.45.194.183

Telemetry data should always be opt-in, IMHO.

### DIFF
--- a/lua/lspconfig/sumneko_lua.lua
+++ b/lua/lspconfig/sumneko_lua.lua
@@ -10,6 +10,7 @@ configs[name] = {
       return util.find_git_ancestor(fname) or util.path.dirname(fname)
     end;
     log_level = vim.lsp.protocol.MessageType.Warning;
+    settings = { Lua = { telemetry = { enable = false }}};
   };
   docs = {
     package_json = "https://raw.githubusercontent.com/sumneko/vscode-lua/master/package.json";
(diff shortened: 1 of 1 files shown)


clason:
Arguably, we should even disable it by default by adding it to `settings` instead of / in addition to `docs`.

clason:
By the way, do we need the additional `root_dir` in the `docs` key here? And if so, why are the two different?

@mjlbach

mjlbach:
@clason I added recently a feature to docgen to read the text of the function during docgen, so we don't need to write out the root_dir anymore, but I didn't go through and remove it from the configs.

mjlbach:
See: https://github.com/neovim/nvim-lspconfig/commit/1c3ac1319cc2f7c4523cf3f072460cbba2825c4d

PR #795, opened 2021-03-26
lspinfo: avoid overeager highlighting

mhinz:
This avoids the `matchadd()` shotgun highlighting.

Before:

After:

I also changed the highlight group for the filetype from `Title` to `Identifier`, but that's totally up to debate. It just felt right to use two different groups for different things.


mjlbach:
Thanks for cleaning up my mess 😆

Issue #796, opened 2021-03-26
When root_dir is not found, cmake-language-server does not work 


LuoZhongYao:
- `nvim --version`: NVIM v0.5.0-dev+1176-g9481310c2
- nvim-lsp version(commit hash): 83d458a0150b3a883454efc0a9ebc4a34162512c
- What language server (If the problem is related to a specific language server): cmake-language-server
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): no
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): no
- Is the problem isolated to a particular language server: yes
- Operating system/version: Linux localhost.localdomain 5.10.8-100.fc32.x86_64 #1 SMP Sun Jan 17 19:52:43 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux Fedora 32

### How to reproduce the problem from neovim startup
1. Create a directory so that it and its ancestor directory do not have the following files and directories. git,compile_commands.json,build 
2. nvim CmakeLists.txt

### Actual behaviour
No completion 

### Expected behaviour
completion 

### Minimal init.vim or init.lua and code sample
```vim
runtime! debian.vim

call plug#begin(stdpath('config') . '/plugged')
Plug 'Shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }
Plug 'Shougo/deoplete-lsp'
Plug 'lifepillar/vim-solarized8'
Plug 'godlygeek/tabular'
Plug 'neovim/nvim-lspconfig'
call plug#end()

let g:deoplete#enable_at_startup = 1
lua << EOF
local nvim_lsp = require('lspconfig')
local on_attach = function(client, bufnr)
  local function buf_set_keymap(...) vim.api.nvim_buf_set_keymap(bufnr, ...) end
  local function buf_set_option(...) vim.api.nvim_buf_set_option(bufnr, ...) end

  buf_set_option('omnifunc', 'v:lua.vim.lsp.omnifunc')

  -- Mappings.
  local opts = { noremap=true, silent=true }
  buf_set_keymap('n', 'gD', '<Cmd>lua vim.lsp.buf.declaration()<CR>', opts)
  buf_set_keymap('n', 'gd', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
  buf_set_keymap('n', 'K', '<Cmd>lua vim.lsp.buf.hover()<CR>', opts)
[...]

PR #797, opened 2021-03-26
Use the file path as root_dir when cmake does not find root_dir

LuoZhongYao:


lua/lspconfig/cmake.lua
```diff
@@ -1,11 +1,16 @@
...
+root_pattern = util.root_pattern(".git", "compile_commands.json", "build");
 configs.cmake = {
   default_config = {
     cmd = {"cmake-language-server"};
     filetypes = {"cmake"};
-    root_dir = util.root_pattern(".git", "compile_commands.json", "build");
+    root_dir = function(fname)
+      local filename = util.path.is_absolute(fname) and fname
+        or util.path.join(vim.loop.cwd(), fname)
+      return root_pattern(filename) or util.path.dirname(filename)
```
mjlbach, line 12: This should be able to be just  `util.path.dirname(fname)`, no? The above two lines can be deleted.

mjlbach: Also, I would keep root_pattern in the return instead of moving it to line 4 so our automated docgen can copy the function text into the docs.

LuoZhongYao: I don't know lua, I copy the configuration of clangd

mjlbach: No worries :)  It's correct now

mjlbach:
Closes #796

PR #806, opened 2021-03-28
pyright: add package_json path

tamago324:
I think we can use the [package.json](https://raw.githubusercontent.com/microsoft/pyright/master/packages/vscode-pyright/package.json) in vscode-pyright for the package.json of pyright, what do you think?

I'd be happy to hear your opinion on this.


mjlbach:
Seems fine :)

tamago324:
Thank you :)

PR #808, opened 2021-03-28
yamlls: Fix file-not-found on windows

### LINKED ISSUE #807: yamlls default config causes file not found on windows
- `nvim --version`: NVIM v0.5.0-dev+1176-g9481310c2
- nvim-lsp version(commit hash): afc2648
- What language server: yamlls
- Is the problem isolated to a particular language server: Yes
- Operating system/version: Windows 10 Build 19042

Open a yaml file on windows with neovim and lsp_config configured to use yamlls

### Actual behaviour
You get a file not found error as a result of the language server having a different name on windows.
On windows the filename of yamlls as installed by npm is going to be `yaml-language-server.cmd`. not `yaml-language-server` like on unix systems.

I don't know if this is always true, hence why this is an issue and not a PR.

### Expected behaviour
Working ls without having to manually configure the binaries name.

### Minimal init.lua and code sample

Offending config:
```lua
require'lsp_config'.yamlls.setup {}
```

Working config:
```lua
local yamlls_bin
if vim.fn.has("win32") == 1 then
	yamlls_bin = "yaml-language-server.cmd"
elseif vim.fn.has("unix") == 1 then
	yamlls_bin = "yaml-language-server"
end

require'lsp_config'.yamlls.setup {
	cmd = { yamlls_bin, "--stdio" },
}
```

TurboCartPig:
You currently get a file not found error on windows as a result of yamlls having a different name than on other systems.
Yamlls as installed by npm is going to be called `yaml-language-server.cmd`, not `yaml-language-server` like on unix systems.

Closes #807

### DIFF
--- a/lua/lspconfig/yamlls.lua
+++ b/lua/lspconfig/yamlls.lua
@@ -2,7 +2,13 @@ local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
 local server_name = "yamlls"
-local bin_name = "yaml-language-server"
+
+local bin_name
+if vim.fn.has("win32") == 1 then
+  bin_name = "yaml-language-server.cmd"
+else
+  bin_name = "yaml-language-server"
+end
 
 configs[server_name] = {
   default_config = {


mjlbach:
Hmm, maybe we should just document to do this instead:

https://github.com/neovim/neovim/issues/13911#issuecomment-796564539

TurboCartPig:
If someone is working on a systemic fix for this then that is certainly better, but I personally don't see the harm in keeping this until such a fix is implemented.

Making this a wiki/readme entry is also not a bad choice, up to you.

mjlbach:
If you add this in your init.lua, does it solve the issue? 

```lua
vim.loop.spawn = (function ()
  local spawn = vim.loop.spawn
  return function(path, options, on_exit)
    local full_path = vim.fn.exepath(path)
    return spawn(full_path, options, on_exit)
  end
end)()
```

TurboCartPig:
That seems to work.

mjlbach:
I'm just going to add a note in the readme then, it should get upstreamed soon-ish

TurboCartPig:
Great!

I'll just close these issues then.

PR #810, opened 2021-03-29
Show Capabilities in :LspInfo

kabouzeid:
Bildschirmfoto 2021-03-29 um 2 36 13 PM
Adds the "capabilities" line.

### DIFF
--- a/lua/lspconfig/lspinfo.lua
+++ b/lua/lspconfig/lspinfo.lua
@@ -44,6 +44,33 @@ return function ()
     return cmd
   end
 
+  local function available_capabilities(resolved_capabilities)
+    -- these are the capabilities that might be interesting to the user
+    local display_keys = {
+      "call_hierarchy",
+      "code_action",
+      "code_lens",
+      "completion",
+      "declaration",
+      "document_formatting",
+      "document_highlight",
+      "document_range_formatting",
+      "document_symbol",
+      "execute_command",
+      "find_references",
+      "goto_definition",
+      "hover",
+      "implementation",
+      "rename",
+      "signature_help",
+      "type_definition",
+    }
+    return vim.tbl_filter(function(key)
+      -- keep only the capabilities that are interesting & available
+      return vim.tbl_contains(display_keys, key) and resolved_capabilities[key] == true
+    end, vim.tbl_keys(resolved_capabilities))
+  end
+
   local indent = "  "
   local function make_client_info(client)
     local server_specific_info = ""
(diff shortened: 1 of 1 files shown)


mjlbach:
Sorry for the delay. I'm a little mixed on this, as `:LspInfo` is already pretty crowded, and we'd have to manually add a list of capabilities (as you've done). Open for discussion about it though (and thanks for the PR :) )

kabouzeid:
I understand :)

I usually only use `:LspInfo` to figure out why something isn't working; with the capabilities info I can quickly check whether the language server even supports what I'm trying to do (e.g. range formatting).

Also it's a nice way to "discover" the functionality of running language servers.

mjlbach:
Some thoughts:

1. Can we automatically read these capabilities from the language server client in core
2. Can we use a fold in the floating window to hide some things (thinking forward to if people want to add anything else to lspinfo).
3. We should probably do some type of indentation on multiline things (like filetypes and capabilities in your screenshot). This could be in a separate PR.

kabouzeid:
Interesting ideas, it got me thinking. What do you think about setting `nowrap` in the `:LspInfo` buffer?

Bildschirmfoto 2021-04-09 um 11 57 18 PM

mjlbach:
I wouldn't want to disable wrapping until we have reliable solution for manually splitting the lines, otherwise I guarantee someone will file an issue not realizing you can navigate in the float the same way you can a normal buffer.

pidgeon777:
Showing capabilities of the currently available LSP servers would be great, indeed.

kabouzeid:
feel free to copy this from my config https://github.com/kabouzeid/dotfiles/blob/453c794d1b7ffdecf5bc085383a6e25d99dd0292/nvim/lua/lsp-capabilities.lua

you can then define a command `:LspCapabilities` for this as follows:
```lua
vim.api.nvim_command("command! LspCapabilities lua require'lsp-capabilities'()")
```

ranjithshegde:
> feel free to copy this from my config https://github.com/kabouzeid/dotfiles/blob/453c794d1b7ffdecf5bc085383a6e25d99dd0292/nvim/lua/lsp-capabilities.lua
> 
> you can then define a command `:LspCapabilities` for this as follows:
> 
> ```lua
> vim.api.nvim_command("command! LspCapabilities lua require'lsp-capabilities'()")
> ```

How about instead of it populating `:LspInfo` it could be a standalone command `:LspCapabalities`? It seems like a really useful thing to have these capabilities displayed on call

pidgeon777:
> How about instead of it populating :LspInfo it could be a standalone command :LspCapabalities? It seems like a really useful thing to have these capabilities displayed on call

I also vote for this feature to be added, a way to display all of the current capabilities of a given server in a readable way would be very appreciated, and also would help with debugging.

mjlbach:
It's not possible to display "all" of the features of the language server as we can only account for what is within the protocol that we can detect. I'm going to close this PR as it's out of date, and also I'm still mixed as for the most part it will be redundant, and the number of language servers that don't support any given feature is decreasing. I'd rather this just be handled at the documentation level to keep lspinfo information dense.

kylo252:
@mjlbach, how about the idea of `:LspCapabalities` with auto-complete and/or a prompt to choose the client? or maybe  `:LspClientInfo`?

mjlbach:
Soft pass (I could be convinced, but unlikely), it would take a lot to convince me that this wouldn't just be confusing to users, as it would only cover the on-spec stuff and would be redundant 90% of the time. Also we show a warning if you try to do something that the server doesn't support, so I don't see how this extra information is relevant.

PR #811, opened 2021-03-29
Search ancestors in symlinks

Badhi:
The build environment I use has the whole repo structure duplicated through symlinks. On such occasions, it would be great to have an option to avoid resolving the symlinks while walking down the hierarchy

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -185,8 +185,10 @@ M.path = (function()
   end
 
   -- Iterate the path until we find the rootdir.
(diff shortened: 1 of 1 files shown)


mjlbach:
Can you give me a test-case (a docker container perhaps) where the current lspconfig filesystem utils fail? Sorry for the delay, I'm going through old PRs today and cleaning things up.

Badhi:
I'm sorry I'm unable to recreate the issue right now which lead me to think of this change. Maybe I will reopen a PR if the issue comes up again

PR #812, opened 2021-03-30
Allows settings to be modifiable from the outside

tamago324:
Currently, there is no way to change the settings after calling `lspconfig.xxx.setup()`.
That is, we cannot change the values set in client.config.settings that are used when we start a new server.
Since the value in `client.config.settings` is used by the default handler in `workspace/configuration`, we thought it would be nice to be able to change it.

This will allow you to change the value by doing something like `lspconfig[server_name].settings = new_settings`.

Thank you!

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -96,6 +96,10 @@ function configs.__newindex(t, config_name, config_def)
       M.manager = nil
     end
 
+    -- Can be changed from outside
+    local config_settings = vim.tbl_deep_extend("keep", vim.empty_dict(), config.settings)
+    M.settings = vim.tbl_deep_extend('keep', config_settings, default_config.settings)
+
     local make_config = function(_root_dir)
       local new_config = vim.tbl_deep_extend("keep", vim.empty_dict(), config)
       new_config = vim.tbl_deep_extend('keep', new_config, default_config)
@@ -112,6 +116,10 @@ function configs.__newindex(t, config_name, config_def)
       if config.on_new_config then
         pcall(config.on_new_config, new_config, _root_dir)
       end
+
+      -- Reset values that have been changed from outside
+      -- Used by 'workspace/configuration' handler
+      new_config.settings = M.settings
 
       new_config.on_init = util.add_hook_after(new_config.on_init, function(client, _result)
         function client.workspace_did_change_configuration(settings)


tamago324:
By using `on_new_config`, I found that I do not need to implement this pull request. like https://github.com/tamago324/nlsp-settings.nvim/pull/3

However, I found it a bit annoying to set `on_new_config` in `lspconfig.xxx.setup()`.

Which way do you think is cooler?
Please let me know what you think.

mjlbach:
Don't disagree with the change, just wondering In what scenario are you trying to change the settings of a running client?

tamago324:
Thank you for saying you are not disagreed with the change.

> Don't disagree with the change, just wondering In what scenario are you trying to change the settings after invoking setup({settings={...}}) but not running client?

I envision a scenario like this.

1. In init.lua, lspconfig.xxx.setup() will be executed.

```
-- in init.lua
lspconfig.sumneko_lua.setup({
  -- ...

  -- setup() 実行時には、1 の時点の settings が設定される
  settings = nlspsettings.sumneko_lua.get(),
})
```

2. Open the file for `sumneko_lua` with `:NlspConfig sumneko_lua` and change the `settings` in JSON. (If there are any clients connected to the `sumneko_lua` server, it will notify the server of the change in `workspace/didChangeConfiguration` and update the value in `client.config.settings`).

3. Open the project in the new `root_dir`.

4. `make_config()` will be executed and the `settings` passed in `setup()` in step 1 will be set to the client.

We believe that the settings used in make_config() in step 4 should be passed the settings to which the changes in step 2 have been applied.
Therefore, the implementation of this pull request will allow settings to be easily changed from the outside.


> If you want to change the settings of a running client, you can access it via client.config.settings. If you want to change the settings that are sent after on_init via the initial workspace/didChangeConfiguration, you can set the settings in setup({}).

Yes. I am aware that it can be changed that way.
I also found out that I can use on_new_config to set the config during initialization. As I said earlier, if we use on_new_config, we don't need this pull request. But it is a bit tedious.

> Aren't you controlling the invocation of setup({}) manually with nlsp-settings anyways?

In nlsp-settings, `setup({})` is not executed manually.
However, it manually controls the `settings` passed to `setup({})`.
Therefore, there will be a difference from the `settings` when `setup()` is executed, and the implementation of this pull request will allow us to apply the difference from outside.

lua/lspconfig/configs.lua
```diff
@@ -113,6 +117,10 @@ function configs.__newindex(t, config_name, config_def)
         pcall(config.on_new_config, new_config, _root_dir)
       end
 
+      -- Reset values that have been changed from outside
+      -- Used by 'workspace/configuration' handler
+      new_config.settings = M.settings
```
mjlbach, line 122: Won't this overwrite any settings that are set in on_new_config? This is the one thing I'm a bit concerned about

mjlbach: Commented on wrong line, but I think you know what I meant lol

tamago324: Thanks for the review.
Oh, that's right, the value set in on_new_config will be overwritten... I just realized that.
Now I realize that it is better to update settings externally by doing something like `setup({on_new_config = nlspsettings.on_new_config})`.

tamago324: I think this pull request is incorrect, so can I close it?

tamago324:
If I need help, I may come back to ask questions.
Thanks!

mjlbach:
Thanks for the PR! I am planning on moving the settings management into core (I have an open PR for this) which I can keep in mind for addressing your needs.

tamago324:
Is that an https://github.com/neovim/neovim/pull/13659 thing? I didn't know this! I'm curious!

PR #817, opened 2021-04-02
replace `vim.fn` in sumneko config

FelipeLema:
`vim.fn` was deprecated. This uses `vim.env` which seems to be working in my machine


mjlbach:
CONFIG.md is auto-generated, so you'll need to edit sumeko.lua. 

Also, vim.fn was *not* deprecated, just vim.fn calls which have an equivalent vim.api call.

FelipeLema:
oh, I see... I guess it's Ok to keep it, then

Thanks for clearing that out

PR #818, opened 2021-04-02
Doc: Remove the util prefix to avoid ambiguity and also to unify

ya-luotao:
The others do not have the `util` prefix

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -107,7 +107,7 @@ require'lspconfig'.als.setup{}
   Default Values:
     cmd = { "ada_language_server" }
     filetypes = { "ada" }
-    root_dir = util.root_pattern("Makefile", ".git")
+    root_dir = root_pattern("Makefile", ".git")
 ```
 
 ## angularls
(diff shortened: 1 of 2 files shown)


mjlbach:
CONFIG.md is an autogenerated file. Updating this will do nothing, and it also accurately reflects [the code](https://github.com/neovim/nvim-lspconfig/blob/f9785053a4ef4aaa2d0aac958bc09a1a289d2fbf/lua/lspconfig/als.lua#L15)

ya-luotao:
> CONFIG.md is an autogenerated file. Updating this will do nothing, and it also accurately reflects [the code](https://github.com/neovim/nvim-lspconfig/blob/f9785053a4ef4aaa2d0aac958bc09a1a289d2fbf/lua/lspconfig/als.lua#L15)

Thanks for the reply, but I guess the CONFIG.md reflects the L38 instead of L15?

ya-luotao:
I submitted this PR because this section seems inconsistent with the others, and it made me a little confused at first.

mjlbach:
If line 38 were not present, then the docgen would automatically extract the code from line 15, yielding the same result as is present. There is no ambiguity, it's just some people bring more functions into scope than others when writing the configs.

ya-luotao:
Okay, that makes sense. Let me close this PR.

PR #820, opened 2021-04-02
Add tflint configuration

shihanng:
tflint (https://github.com/terraform-linters/tflint) is a linter for Terraform which can also act as a language server. Would it be okay to include its configuration to the project? Thank you.

### DIFF
--- a/lua/lspconfig/tflint.lua
+++ b/lua/lspconfig/tflint.lua
@@ -1,0 +1,22 @@
+local configs = require "lspconfig/configs"
+local util = require "lspconfig/util"
+
+configs.tflint = {
+  default_config = {
+    cmd = {"tflint", "--langserver"},
+    filetypes = {"terraform"},
+    root_dir = util.root_pattern(".terraform", ".git", ".tflint.hcl")
+  },
+  docs = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/tflint.lua
```diff
@@ -0,0 +1,22 @@
+local configs = require "lspconfig/configs"
+local util = require "lspconfig/util"
+
+configs.tflint = {
+  default_config = {
+    cmd = {"tflint", "--langserver"},
+    filetypes = {"terraform"},
+    root_dir = util.root_pattern(".terraform", ".git", ".tflint.hcl")
```
lithammer, line 8: Why is `.git` preferred over `.tflint.hcl`?

shihanng: I didn't know there is a order of precedence in `util.root_pattern`. If there is I think it should be `".tflint.hcl", ".terraform", ".git"`. What do you think @lithammer ?

lithammer: Sorry I forgot that the search is breadth first (and not depth first). So the order is irrelevant.

PR #822, opened 2021-04-03
minimal config: make bootstrap on first start

mjlbach:


mjlbach:
@lithammer pros and cons

pro: less idiomatic starter config

con: I'm pretty sure there is no way to mess this up now

lithammer:
lgtm, but what's the problem we're trying to solve?

mjlbach:
The main issue with the current setup, is that if a user doesn't open the minimal_init.lua, the autocmd to install via packer won't fire. Even if we changed this, the user would still have to restart neovim after the packages installed for lspconfig to work. This changes it to block until all packges are installed, then to reload the packages (making it all work on the first startup).

lithammer (approved):

PR #830, opened 2021-04-09
Fix conditional keymap in README

csaez:
The recommended configuration in the `README.md` doesn't setup the keymap for range formatting correctly: this needs to be done independent from the full document formatting and making sure the keymap is set in visual mode as it's meant to work on the selection.


mjlbach:
In this PR you're overridding the same keymap (leader f), twice. Before it was either-or

csaez:
I believe the key map is set in different modes, the first if statement set a key map in normal mode (formatting the whole document), while the second set it in visual mode (formatting the selection only), making both available.

mjlbach:
Ah, that's fine then. Thanks!

PR #834, opened 2021-04-12
Make intelephense work on Windows

Julpikar:
This commit is just to add `.cmd` to the binary name `intelephense` so that the binary name becomes `intelephense.cmd`. LSP Neovim on Windows need to execute command 'intelephense.cmd' for LSP server to work.


mjlbach:
Did you see the section on windows in the readme? Adding that snippet to your config will make every server work on windows without modifying cmd.

Julpikar:
Oh okay thanx.

PR #838, opened 2021-04-14
Add cssls description

nabezokodaikon:
Added supplement to cssls description

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -418,6 +418,7 @@ npm install -g vscode-css-languageserver-bin
 
 
 ```lua
+--Enable (broadcasting) snippet capability for completion
 local capabilities = vim.lsp.protocol.make_client_capabilities()
 capabilities.textDocument.completion.completionItem.snippetSupport = true
 
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -418,13 +418,7 @@ npm install -g vscode-css-languageserver-bin
 
 
 ```lua
---Enable (broadcasting) snippet capability for completion
-local capabilities = vim.lsp.protocol.make_client_capabilities()
-capabilities.textDocument.completion.completionItem.snippetSupport = true
-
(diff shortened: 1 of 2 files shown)


mjlbach:
You should add a note on why to enable snippet support (is there any particular reason as to why this differs from what is already in the wiki?) and not include the commit to CONFIG.md

nabezokodaikon:
Current description of cssls This is because cssls does not work with `require'lspconfig'.cssls.setup{}`.

mjlbach:
These are the lines you should copy: https://github.com/neovim/nvim-lspconfig/blob/f9785053a4ef4aaa2d0aac958bc09a1a289d2fbf/lua/lspconfig/html.lua#L32-L34


Please add the above, squash, and remove the commited changes to CONFIG.md (that file is auto-generated, and your changes in css lua file will be enough).

lua/lspconfig/cssls.lua
```diff
@@ -27,6 +27,19 @@ https://github.com/vscode-langservers/vscode-css-languageserver-bin
 ```sh
 npm install -g vscode-css-languageserver-bin
 ```
+
+Neovim does not currently include built-in snippets. `vscode-html-languageserver` only provides completions when snippet support is enabled.
```
mjlbach, line 31: This need to be changed from html to vscode-css-languageserver

CONFIG.md
```diff
@@ -418,7 +418,7 @@ npm install -g vscode-css-languageserver-bin
 
 
 ```lua
-require'lspconfig'.cssls.setup{}
+require'lspconfig'.cssls.setup {}
```
mjlbach, line 421: remove this change and then squash

nabezokodaikon:
It's complicated, so start over.

mjlbach:
for future reference, `git rebase -i master`, change every commit but the first one from `pick` to `squash`

PR #839, opened 2021-04-14
Added description to introduce cssls

nabezokodaikon:


lua/lspconfig/cssls.lua
```diff
@@ -27,6 +27,19 @@ https://github.com/vscode-langservers/vscode-css-languageserver-bin
 ```sh
 npm install -g vscode-css-languageserver-bin
 ```
+
+Neovim does not currently include built-in snippets. `vscode-html-languageserver` only provides completions when snippet support is enabled.
```
mjlbach, line 31: Still need to fix this type :) vscode-html-languageserver -> vscode-css-languageserver

mjlbach:
Looks good, now you just need to squash.

cloggier:
The same applies to other lsps so maybe extract it into separate section and link to from relevant servers?

mjlbach:
We can't do that, I'd have to rewrite the docgen (and I won't)

nabezokodaikon:
Is my PR method correct?
I feel bad. I am not used to PR.

mjlbach:
It's great :) Thanks for your contribution.

PR #841, opened 2021-04-14
Use the file path as root_dir when ccls does not find root_dir

Adnios:
Problem: when root_dir is not found, ccls does not work.

This pull request is using the file path as root_dir when ccls does not find root_dir


lithammer:
Thanks!

PR #855, opened 2021-04-20
Add importMap to denols

tzx:
I was just trying out deno and seems like the current lsp config doesn't resolve imports from importMaps because it doesn't have that capability. I added it with a sane (i think?) default. 

Before:
image

After:
image

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -100,6 +100,7 @@ configs[server_name] = {
       enable = true;
       lint = false;
       unstable = false;
+      importMap = "./import_map.json";
     };
     handlers = {
       ["textDocument/definition"] = denols_handler;


mjlbach:
@David-Else I don't know enough about deno to know whether this is a common pattern/it will break common usecases. Can you advice? Maybe @williamboman knows as well.

williamboman:
https://www.npmjs.com/package/@jsenv/importmap-node-module uses `project.importmap` as the filename for the import map. I'm not particularly familiar with import maps but my experience so far has been that it's still highly experimental.

It's worth noting that simply adding this setting will not fix things - one would still have to actually create and manage the `import_map.json` file. I also don't know how well the denols server handles import_map filenames that doesn't exist, will it introduce undefined behavior or crashes?

I feel like it'd be best to leave this to user's to configure themselves, but again I don't use deno.

mjlbach:
Thanks! Let's leave this to the users then.

Issue #856, opened 2021-04-20
bufnr in on_attach lsp config callback is nil


mnowotnik:
I think the autocmd at [this line](https://github.com/neovim/nvim-lspconfig/blob/5c005ce93367ad85933eff80887228bca2a7efee/lua/lspconfig/configs.lua#L142) should be
```lua
          api.nvim_command(string.format(
              "autocmd BufEnter <buffer=%d> ++once lua require'lspconfig'[%q]._setup_buffer(%d,%d)"
              , bufnr
              , config_name
              , client.id
              , bufnr
              ))
```
Currently, bufnr is being passed to user on_attach callback as a nil.

mjlbach:
Closed by #857

PR #857, opened 2021-04-20
bug: bufnr not passed to on_attach

mnowotnik:
#856

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -139,10 +139,11 @@ function configs.__newindex(t, config_name, config_def)
           M._setup_buffer(client.id, bufnr)
         else
           api.nvim_command(string.format(
-              "autocmd BufEnter <buffer=%d> ++once lua require'lspconfig'[%q]._setup_buffer(%d)"
+              "autocmd BufEnter <buffer=%d> ++once lua require'lspconfig'[%q]._setup_buffer(%d,%d)"
               , bufnr
               , config_name
               , client.id
+              , bufnr
               ))
         end
       end)


mnowotnik:
Looks like it failed on downloading some spec:
`Error: Error fetching file: Failed downloading https://luarocks.org/busted-2.0.0-1.rockspec - busted-2.0.0-1.rockspec`

lithammer:
Re-triggered the jobs and now it's green.

lua/lspconfig/configs.lua
```diff
@@ -139,10 +139,11 @@ function configs.__newindex(t, config_name, config_def)
           M._setup_buffer(client.id, bufnr)
         else
           api.nvim_command(string.format(
-              "autocmd BufEnter <buffer=%d> ++once lua require'lspconfig'[%q]._setup_buffer(%d)"
+              "autocmd BufEnter <buffer=%d> ++once lua require'lspconfig'[%q]._setup_buffer(%d,%d)"
               , bufnr
               , config_name
               , client.id
+              , bufnr
               ))
```
lithammer, line 147: Hmm, I must admit that I don't really get why want this `else` block at all. What's the difference between just calling `M._setup_buffer(client.id, bufnr)` directly? Do you know @mjlbach?

mjlbach: Pretty sure it's because schedule_wrap doesn't guarantee the function executes while the current buffer is in focus, so if it's not it execute an autocommand on the next BufEnter (which wouldn't trigger if it was already the active buffer)

lithammer: Why do we care about buffer focus at this point though? Especially since we have `bufnr` which would explicitly target the correct buffer regardless what focused buffer is.

mjlbach: I didn't write it, so I don't know the full motivation. One advantage is it differs attaching the client, and running on_attach, until you actually enter the buffer, so if a user adds buffer local keymaps without specifying the buffer (so using the current buffer), it "just works"

lithammer: That's a valid point. Especially if you have a habit of doing something like this:

```console
$ nvim src/**.py
```

mnowotnik: True, and somebody probably uses that [behavior](https://www.hyrumslaw.com/).

Issue #239, opened 2020-05-18
Lazy require language server configurations to avoid increasing neovim startup time


ms-jpq:
Hi there,

I noticed that my vim startup time doubled, when I added in  a bunch of lsp configs (I added many).

So I lazy loaded all of mine `setup(...)` in `FileType` `autocmd`.

I had to dig around a bit and add in `lsp.<client>.manager.try_add()` for each client, or else the first buffer of `FileType` would not be loaded.

Of course this isn't too difficult, but I think there are some room for improvement.

We could perhaps add in a param for `require("nvim_lsp").<client>.setup(...)` such that it does this automatically.

This would not be a breaking change by any means, and should speed up things for our users.

Edit:

I don't know why this is labled as a bug automatically, how do I change it?

tjdevries:
Can you show what your final config looked like?

ms-jpq:
Sure thing it goes something like this:

`bash.lua`

```lua
local ftp = require "libs/ftp"

local ft = {"python"}
local lang = function ()

  if not fn.executable("dotnet") then
    return
  end

  local lsp = require "nvim_lsp"
  lsp.pyls_ms.setup{}
  lsp.pyls_ms.manager.try_add()

end
ftp.defer(ft, lang)
```


`libs/ftp.lua`

```lua
local set = require "libs/set"
local registry = require "libs/registry"

local _ftp = set.new()

local defer = function (ft, ftplugin)
  local ftp = function ()
    if set.contains(_ftp, ftplugin) then
      return
    end
    set.add(_ftp, ftplugin)
    ftplugin()
    print("-- 加载: " .. table.concat(ft, ",") .. " --")
  end
  registry.auto("FileType", ftp, ft) -- registry basically does the autocmd thing
end
```

`libs/registry.lua`

```lua
local _callbacks = {}
local inc = std.count()

local auto = function (events, func, filter)

  local evnts = std.wrap(events)
  local events = table.concat(evnts, ",")
  local filter = table.concat(std.wrap(filter or "*"), ",")
  local idx = inc()
  local group = "augroup " .. idx
  local cls = "autocmd!"
  local cmd = "autocmd " .. events .. " " .. filter .. " lua require('" .. _registry .. "').call(" .. idx .. ")"
  local done = "augroup END"

  for event in ipairs(evnts) do
    assert(fn.exists("##" .. event))
  end

  api.nvim_command(group)
  api.nvim_command(cls)
  api.nvim_command(cmd)
[...]

PR #861, opened 2021-04-22
Use dofile instead of require for faster loading

mjlbach:


lua/lspconfig.lua
```diff
@@ -84,7 +86,7 @@ end
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
-    pcall(require, 'lspconfig/'..k)
+    dofile(script_path .. 'lspconfig/' .. k .. ".lua")
```
justinmk, line 89: https://stackoverflow.com/a/31149198/152142 seems like there are some risks with dofile(), are we certain they are avoided here? Perhaps worth mentioning in a brief comment above this line.

justinmk:
For the commit message:

- closes https://github.com/neovim/nvim-lspconfig/issues/239 ?
- The stats in https://github.com/neovim/nvim-lspconfig/pull/860 were useful to note.

lua/lspconfig.lua
mjlbach: Yes, I'm not quite sure what "gotchas" there are with `dofile`, I believe the reasons it is faster are the drawbacks, it doesn't have to search package.path (thus requiring the debug workaround to get the path to the lspconfig.lua file), it doesn't cache lookups (which doesn't matter for us since we are only loading these files once to add to the global config table), and someone mentioned on IRC that it may avoid JIT compilation, which is fine because these are just config files.

mjlbach:
I'll update the commit, to copy the stats here, this improves sourcing *all* current configs to < 1 ms, requiring lspconfig itself takes 4 ms

For all 76 servers, with this "fix", < 1 ms: 

```
139 lspconfig: start
144 lspconfig: end
144 lspconfig server setup: start
144 lspconfig server setup: end
```

Without this fix (123 ms):

```
225 lspconfig: start
232 lspconfig: end
232 lspconfig server setup: start
335 lspconfig server setup: end
```

nanotee:
It does seem to be faster on my machine (saves around 3-4 ms)

One thing that broke for me is the ability to put configurations in `vim.fn.stdpath('config') .. '/lua/lspconfig/custom-server.lua'`. I know there are [other ways](https://github.com/neovim/nvim-lspconfig#example-custom-config) to write custom configs so I don't know if it's that important

mjlbach:
Hmm, yeah, I think it's ok to "deprecate" that for now, the performance improvement is mostly for large numbers of servers. If you have < 5 you won't notice much of a difference. I also wanted to make `lspconfig` ludicrously because I've seen on reddit claims it slows down starting neovim, and now at ~6 ms I don't think anyone will say anything 😆

mjlbach:
I'm going to merge this, if it breaks anything for anyone, please file an issue.

PR #862, opened 2021-04-22
feat: try to run lsp client at startup

mnowotnik:
This change makes it so that an lsp client gets attached to a directly opened file, i.e.

```
nvim /path/to/file.lua
```

Currently, when a file gets opened this way, the lsp client is started, but the file buffer itself is not getting attached via `manager.try_add()`. A user has to work around this problem, by postponing the opening of a file. For example:

```
nvim -c ':e /path/to/file.lua'
```

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -56,6 +56,11 @@ function configs.__newindex(t, config_name, config_def)
       trigger = "BufReadPost *"
     end
     if not (config.autostart == false) then
+      if not config.filetypes or vim.tbl_contains(config.filetypes, vim.bo.ft) then
+        vim.schedule(function()
+          M.manager.try_add()
+        end)
+      end
       api.nvim_command(string.format(
           "autocmd %s lua require'lspconfig'[%q].manager.try_add()"
           , trigger


mjlbach:
I'm confused, as what you're describing does not happen for me. The only issue is when you create a new file with certain clients via neovim, you have to save the file to disk and force reattach.

mjlbach:
See, automatically attached, no intervention from me:
https://user-images.githubusercontent.com/13316262/115742387-78d18500-a345-11eb-8ceb-e00b45f2fe43.mov

mnowotnik:
Probably because in my config the lsp setup is delayed (for example I run sumneko_lua in a vim.schedule callback in order to gather rtp paths).

I guess the PR description is incorrect then. 

What I want is the ability to run `require('lspconfig').<some_config>.setup({...})` any time after neovim startup and have the existing opened buffers properly attached. My other use case is a vim command that runs lsp setup on demand. It doesn't work without this change. `vim.lsp` supports this use case (runtime/lua/vim/lsp.lua, line 851):

```lua
      -- If we had been registered before we start, then send didOpen This can
      -- happen if we attach to buffers before initialize finishes or if
      -- someone restarts a client.
      for bufnr, client_ids in pairs(all_buffer_active_clients) do
        if client_ids[client_id] then
          client._on_attach(bufnr)
        end
      end
```

The above lines are not executed by nvim-lspconfig's `setup` since `start_client` is only launched on `FileType` or `BufReadPost` events, both of which have already been triggered by then in the scenarios above.

mjlbach:
Why not just use LspStart then?

mnowotnik:
Ok, I could do that. Thanks for the tip!

Issue #858, opened 2021-04-21
How to intercept ["workspace/executeCommand"] to use a custom lua function instead of sending to LS?


lbiaggi:
Hi, I’m trying to contribute / use this [ltex-ls](https://github.com/valentjn/ltex-ls) with neovim built-in client, so far, this is what I could achive alone [ltex.lua](https://gist.github.com/lbiaggi/a3eb761ac2fdbff774b29c88844355b8).

This LS has a nonstandard way to implement 3 actions at the client side with involves intercepting these commands from workspace/executeCommand
1. add word to a custom dict ``` (_ltex.addToDictionary) ```
2. add ignore rule.  ``` (_ltex.disableRules) ```
3. add false positive. ```(_ltex.hideFalsePositives) ```

These actions are generated by the default handler of code action, what I want is when they happen instead of trying to communicate with LS, execute my custom function.

Regarding coding in lua / neovim lsp client config, what I can improve, there is a better solution for anything already done?

Regards,
Lucas

mjlbach:
I don't really understand what you're asking. This question belongs on discourse, but I'll try to ask some clarifying questions here.

What do you mean "intercepting a command"? We send the request for executeCommand [here](https://github.com/mjlbach/neovim/blob/e343437bb6b82a48b1001d98a08fb5c63dccda30/runtime/lua/vim/lsp/buf.lua#L371-L377) and process the handler [here](https://github.com/mjlbach/neovim/blob/e343437bb6b82a48b1001d98a08fb5c63dccda30/runtime/lua/vim/lsp/handlers.lua#L20-L25)

You can override either of these.

> Regarding coding in lua / neovim lsp client config, what I can improve, there is a better solution for anything already done?

This is a broad question? Are you asking with regards to your snippet or the language client in general?

lbiaggi:
Default handler give this code action menu 

https://user-images.githubusercontent.com/5401584/115696459-17a3b480-a339-11eb-9d1d-fe056171310d.mp4

if I accept any of the first three, the debug log will be similar to these lines

```
[...]

PR #863, opened 2021-04-23
[WIP] Support ltex-ls

lbiaggi:
Initial working code to support [ltex-ls](https://github.com/valentjn/ltex-ls) with suggestion to address #858 
thanks, @mjlbach, @mfussenegger

Looking for help/suggestion about these things:
- How to generalize `dictionary, disabledRules, hiddenFalsePositives`, they currently need to be hardcoded using a valid table entry with an existing file previously created.
- Discover a better way to expose tables with file entries.

### DIFF
--- a/lua/lspconfig/ltex.lua
+++ b/lua/lspconfig/ltex.lua
@@ -1,0 +1,37 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.ltex = {
+  default_config = {
+    cmd = { 'ltex-ls' },
+    filetypes = { 'tex', 'bib', 'markdown' },
+    root_dir = util.find_git_ancestor,
+    settings = {
+      ltex = {
+        enabled = { 'latex', 'tex', 'bib', 'markdown' },
+        checkFrequency = 'edit',
+        language = 'en',
+        diagnosticSeverity = 'information',
+        setenceCacheSize = 2000,
+        additionalRules = {
+          enablePickyRules = true,
+          motherTongue = 'en',
+        },
+        dictionary = {},
+        disabledRules = {},
+        hiddenFalsePositives = {},
+      },
+    },
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/valentjn/vscode-ltex/develop/package.json',
+    description = [[
+https://github.com/valentjn/ltex-ls
+
+LTeX Language Server: LSP language server for LanguageTool 🔍✔️ with support for LaTeX 🎓, Markdown 📝, and others
+
+To install, download the latest [release](https://github.com/valentjn/ltex-ls/releases) and ensure `ltex-ls` is on your path.
+
+]],
+  },
+}


mjlbach:
I started reviewing, but I would highly recommend looking at another configuration example (rust-analyzer is a good one) to see the form these should take, you shouldn't have a setup{} call in the config for example, you shouldn't check for configs.ltex-ls etc.

lbiaggi:
> I started reviewing, but I would highly recommend looking at another configuration example (rust-analyzer is a good one) to see the form these should take, you shouldn't have a setup{} call in the config for example, you shouldn't check for configs.ltex-ls etc.

Ok

mjlbach:
Ok, so I looked more into this when I wanted to use it. I am considering reconsidering allowing custom commands to be registered per language server. This https://github.com/neovim/nvim-lspconfig/pull/863/files#diff-b6698738921d6c7c76e0bb67c33623ba1d213719c995c8353468bcdb3411049bR156-R186 is extremely messy, and I'd like to avoid it somehow. I'm not sure what the appropriate way to do this is @mfussenegger

For now, I would recommend instead of overriding execute_command, I would override the handler for `textDocument/codeAction`, which can be passed as a key to setup. This will be cleaner and avoid mutating the global function which is used by other language servers.

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,186 @@
...
+    default_config = {
+        cmd = {"ltex-ls"};
+        filetypes = {'tex', 'bib', 'md'};
+        root_dir = function(filename)
+            return util.path.dirname(filename)
+        end;
+        dictionary_files = { ["en"] = {vim.fn.getcwd() .. "dictionary.ltex"} };
+        disabledrules_files = { ["en"] = {vim.fn.getcwd() .. "disable.ltex"} };
+        falsepositive_files = { ["en"] = {vim.fn.getcwd() .. "false.ltex"}};
+        settings = {
```
mjlbach, line 127: I'm pretty sure everything under settings is either a default, or too opinionated to be worth using.

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,186 @@
...
+                    motherTongue= "en",
+                };
+                dictionary = {};
+                disabledRules = {};
+                hiddenFalsePositives = {};
+            },
+        };
+        on_attach = function(client, bufnr)
+                    -- local lang = client.config.settings.ltex.language
+            for lang,_ in ipairs(client.config.dictionary_files) do       --
```
mjlbach, line 145: settings are automatically sent on_attach, I don't thin you need to update this.

mjlbach:
With the size of this my module, my recommendation is to create ltex-ls.nvim and add a smaller minimal config in lspconfig to start. The following "works":

```lua
local lspconfig = require'lspconfig'
local configs = require'lspconfig/configs'

local server_name = "ltex-ls"

configs[server_name] = {
  default_config = {
    cmd = {"ltex-ls"};
    filetypes = {'tex', 'bib', 'markdown'};
    root_dir = function(filename)
      return util.path.dirname(filename)
    end;
  };
  docs = {
    package_json = 'https://raw.githubusercontent.com/valentjn/vscode-ltex/develop/package.json',
    description = [[
https://github.com/valentjn/ltex-ls

LTeX Language Server: LSP language server for LanguageTool 🔍✔️ with support for LaTeX 🎓, Markdown 📝, and others

To install, download the latest [release](https://github.com/valentjn/ltex-ls/releases) and ensure `ltex-ls` is on your path.

]];
  };
};
```

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,186 @@
...
+            return nil
+        end
+    end
+    return addToFile(filetype, lang, file, value)
+end
+
+configs.ltex = {
+    default_config = {
+        cmd = {"ltex-ls"};
+        filetypes = {'tex', 'bib', 'md'};
```
mjlbach, line 120: should be `markdown`

mfussenegger:
> Ok, so I looked more into this when I wanted to use it. I am considering reconsidering allowing custom commands to be registered per language server. This https://github.com/neovim/nvim-lspconfig/pull/863/files#diff-b6698738921d6c7c76e0bb67c33623ba1d213719c995c8353468bcdb3411049bR156-R186 is extremely messy, and I'd like to avoid it somehow. I'm not sure what the appropriate way to do this is @mfussenegger
> 
> For now, I would recommend instead of overriding execute_command, I would override the handler for `textDocument/codeAction`, which can be passed as a key to setup. This will be cleaner and avoid mutating the global function which is used by other language servers.

https://github.com/neovim/neovim/pull/14115 would help :) 

In nvim-jdtls I provide a dedicated `code_action` function:

https://github.com/mfussenegger/nvim-jdtls/blob/8bd4eac08c637961b10ac01c124fb5091484c265/lua/jdtls.lua#L638-L702

But I only do that because otherwise I can't access the original code_action_params which are required for the java specific client commands. (the VSCode code action API gives access to them, so language servers naturally ended up depending on them)

mjlbach:
> neovim/neovim#14115 would help :)

I totally forgot about that :) , we should stop punting on the decision about breaking symmetry in the args. Maybe we can convince @tjdevries to review next Friday.

> Otherwise I'd probably also override the handler. But the problem with that is that different plugins can become incompatible with each other, so the handler should only be overriden in the scope of the language specific client and not globally.

For sure, in this case I'd advocate just adding a default handler in the config that overrides it just for ltex-ls, as I believe the current PR overrides a global function which IMO is not good. We can always make this cleaner once neovim/neovim#14115 in one form or another.

lua/lspconfig/ltex.lua
lbiaggi: Ok

lua/lspconfig/ltex.lua
lbiaggi: This came from my use case, where I use two languages in the same document with different dict files.

Do you have a better suggestion how to do this with multiple languages?

lua/lspconfig/ltex.lua
lbiaggi: > With the size of this my module, my recommendation is to create ltex-ls.nvim and add a smaller minimal config in lspconfig to start. The following "works". We can gradually add back in the functionality while making any necessary changes in neovim core to support more "off-spec" language servers like this one:
> 
> ```lua
> local configs = require'lspconfig/configs'
> local util = require 'lspconfig/util'
> 
> local server_name = "ltex-ls"
> 
> configs[server_name] = {
>   default_config = {
>     cmd = {"ltex-ls"};
>     filetypes = {'tex', 'bib', 'markdown'};
>     root_dir = function(filename)
>       return util.path.dirname(filename)
>     end;
>     settings = {
>       ltex = {
>         checkFrequency="edit",
>       };
>     };
>   };
>   docs = {
>     package_json = 'https://raw.githubusercontent.com/valentjn/vscode-ltex/develop/package.json',
>     description = [[
> https://github.com/valentjn/ltex-ls
> 
> LTeX Language Server: LSP language server for LanguageTool 🔍✔️ with support for LaTeX 🎓, Markdown 📝, and others
> 
> To install, download the latest [release](https://github.com/valentjn/ltex-ls/releases) and ensure `ltex-ls` is on your path.
> 
> ]];
>   };
> };
> ```
> 
> edit: Actually we need to send at least one setting for the server to trigger... seems like a bug.

What exactly didn't work? At least in my test, except to the "add to ..." and pickyrules which is disabled by default.

mjlbach: The server doesn't provide any diagnostics if not at least one `settings` item is sent, I'm advocating for the first round of inclusion of lspconfig, using the simpler config you quoted while we clean up this PR.

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,183 @@
...
+    return dict
+end
+
+local function findLtexLang()
+    local buf_clients = vim.lsp.buf_get_clients()
+    for _, client in ipairs(buf_clients) do
+        if client.name == "ltex" then
+            return client.config.settings.ltex.language
+        end
+    end
```
mjlbach, line 21: use pairs here, ipairs will stop at the first break in the table sequence

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,183 @@
...
+end
+
+local function findLtexFiles(filetype, value)
+    local buf_clients = vim.lsp.buf_get_clients()
+    for _, client in ipairs(buf_clients) do
```
mjlbach, line 26: pairs

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,183 @@
...
+                return nil
+            end
+        end
+    end
+end
+
+local function updateConfig(lang, configtype)
+    local buf_clients = vim.lsp.buf_get_clients()
+    local client = nil
+    for _, lsp in ipairs(buf_clients) do
```
mjlbach, line 49: pairs

lbiaggi: done

abalmos:
I would love to see this merged. Is there a hold up we could help with? 

In the meantime, this is an option: https://github.com/brymer-meneses/grammar-guard.nvim

mjlbach:
The hacks around buffer_execute command can be removed with the new handler signature, and this: https://github.com/neovim/neovim/pull/14115 once merged will make adding client side commands easier.

I also have an unresolved comment that hasn't been addressed.

David-Else:
@lbiaggi Neovim 0.5.1 is out, I think https://github.com/neovim/neovim/pull/15504 might be the relevant change in the handler signature @mjlbach mentioned. Really looking forward to this getting merged :)

mjlbach:
The client side command registry wasn't backported so this would still need to be gated behind a 0.6 'has' https://github.com/neovim/neovim/pull/14115

lbiaggi:
> The client side command registry wasn't backported so this would still need to be gated behind a 0.6 'has' [neovim/neovim#14115](https://github.com/neovim/neovim/pull/14115)
> 
> Also, given the amount of client-side code I think this is probably better as a separate plugin, we'll see.

Is there a better way to check nvim version (from lua) than this?
```lua
vim.fn.has('nvim-0.6')
```

mjlbach:
That's what I would recommend.

williamboman:
> I would love to see this merged. Is there a hold up we could help with?

I've also added this in https://github.com/williamboman/nvim-lsp-installer/pull/134 for the time being. Hopefully should be fairly straight forward to eject the custom code from that plugin, and only keep the installer, once things are figured out.

David-Else:
@oblitum Thanks for the info! Unfortunately I am not using coc, so will have to wait until this PR lands.

abalmos:
@mjlbach This can get merged now I think? It makes `ltex` work and there is an issue on grammar guard to implement the local command handlers for false positives, etc. (https://github.com/brymer-meneses/grammar-guard.nvim/issues/4) I plan to attempt that PR if someone else doesn't do it first, but that is likely at least 4-5 weeks in the future.

lua/lspconfig/ltex.lua
```diff
@@ -0,0 +1,33 @@
+configs.ltex = {
+  default_config = {
+    cmd = { "ltex-ls" },
+    filetypes = { "tex", "bib", "markdown" },
+    settings = {
```
williamboman, line 5: `root_dir` seems to be missing. It used to be like this, but not sure if it's a good one for this LSP (I don't use it)

```suggestion
     root_dir = function(filename)
       return util.path.dirname(filename)
     end,
    settings = {
```

mjlbach: it should just be a git fallback until I add scratch server support

oblitum: > but not sure if it's a good one for this LSP (I don't use it)

@williamboman this applies to all lsp4j based ones (ltex-ls, jdtls, etc):

- https://github.com/eclipse/lsp4j/blob/40a1fe44015626f8ea63dd7ed36ebbd5a1ce84f8/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageClient.java#L116-L118

mjlbach:
I'm merging this by popular demand (lspconfig is ofc community contributed), a couple notes:

* It's really hard to have a universal default for grammar/language (english is hardcoded here)
* ltex-ls requires many custom client-side/off-spec commands
* ltex-ls especially is one of those servers that users probably want to use in arbitrary places, so scratch server support would be nice to have, but is currently blocked by the lspconfig rewrite

For those reasons, I highly encourage people to either contribute to grammar-guard, or start their own custom plugin which provides additional features (which we can then reference in the wiki as with the other growing number of language specific plugins that integrate or bypass lspconfig to use the client directly). Lspconfig is always going to try to be as "on-spec" and minimal as possible (for servers that are on-spec/may not necessitate their own plugin since it would be about 10 LOC, are unpopular, etc.). I updated CONTRIBUTING to make this more clear the other day.

David-Else:
@mjlbach (on Rocky Linux) I am getting:
```                                                                                                                                                                                                                                                             
  - get app, recor   cmd:               ltex-ls                                                                                                                             
                     cmd is executable: Unable to find executable. Please check your path and ensure the server is installed
```
It is on my $PATH:
```
$ ltex-ls --version
{
  "ltex-ls": "14.0.0",
  "java": "11.0.13"
}
```
It  worked fine using the setup you supplied with `local bin_path = '/home/david/bin/ltex-ls-14.0.0/bin/ltex-ls`, but I remember it would not work with `~/bin/ltex-ls-14.0.0/bin/ltex-ls` for some strange reason.

I add the executables to my $PATH like this:
`.bashrc`
```
alias ltex-ls="/home/david/bin/ltex-ls-14.0.0/bin/ltex-ls"
alias ltex-cli="/home/david/bin/ltex-ls-14.0.0/bin/ltex-cli"
```
Any idea what could be happening?

disrupted:
> I add the executables to my $PATH like this: `.bashrc`
> 
> ```
> alias ltex-ls="/home/david/bin/ltex-ls-14.0.0/bin/ltex-ls"
> alias ltex-cli="/home/david/bin/ltex-ls-14.0.0/bin/ltex-cli"
> ```
> 
> Any idea what could be happening?

if you're just aliasing the commands, the directory is not in your `$PATH` and the commands are only known to your shell. What you'd want is `export PATH="$PATH:~/bin/ltex-ls-14.0.0/bin"` or you could move it to another directory within your `PATH`.

oblitum:
> english is hardcoded

The thread previously referred by me [here](https://github.com/neovim/nvim-lspconfig/pull/863#issuecomment-946283956) is originally about language auto detection, which ltex/language-tool does support (with the given limitations as per thread).

David-Else:
@disrupted Can't believe I made such a dumb error! Here is the install script I just made, in case in saves anyone some time. No need to add anything to the path if you install to the proper directory and add a symlink :)
```
#!/bin/bash
set -euo pipefail

BIN_INSTALL_DIR=/usr/local/bin

LTEXLS_LOCATION=valentjn/ltex-ls/releases/download/15.0.0/
LTEXLS_FILENAME=ltex-ls-15.0.0.tar.gz
LTEXLS_SHA=76bed42397cd52a05415e107cd1163c05e13f00fa2e8d4e6c841f7f7f0fd4a67d99474dd6f86aa4ebef5d46e4cd227310badc52bc1b2c055c5aa390a9ac706a4

if [ "$(id -u)" != 0 ]; then
    echo "You're not root! Run script with sudo" && exit 1
fi

# Call with arguments (${1} location,${2} filename,${3} sha)
download_verify() {
    curl -LOf "https://github.com/${1}${2}"
    echo "${3} ./${2}" | sha512sum --check
}

download_verify "$LTEXLS_LOCATION" "$LTEXLS_FILENAME" "$LTEXLS_SHA"
tar --no-same-owner -C $BIN_INSTALL_DIR/ -xf $LTEXLS_FILENAME --no-anchored 'bin' --strip=1
tar --no-same-owner -C $BIN_INSTALL_DIR/ -xf $LTEXLS_FILENAME --no-anchored 'lib' --strip=1
ln -s $BIN_INSTALL_DIR/bin/ltex-ls $BIN_INSTALL_DIR/ltex-ls
rm $LTEXLS_FILENAME
```

David-Else:
Also, if people want to use the ngram data this is now working great:
```
-- optionally install ngram data into home directory (~/en) and uncomment languageModel = '~/'
-- https://dev.languagetool.org/finding-errors-using-n-gram-data

require('lspconfig').ltex.setup {
  settings = {
    ltex = {
      additionalRules = {
        enablePickyRules = true,
        motherTongue = 'en',
        -- languageModel = '~/',
      },
    },
  },
}
```
You get a lot more corrections, brings it closer to the online grammarly results while keeping everything on your own computer.

oblitum:
I'd advice to not extract directly at `~` and have its own ngram directory with `en`, etc. underneath. So to avoid potential problems in case you happen to put some specially (for `ltex`) named directory under `~` without realizing.

David-Else:
If the user doesn't add `language` then they get no spell checking, only grammar checking. Maybe adding it should be default, or maybe this is intentional as users are expected to be using the built in Neovim spell check?
```
require('lspconfig').ltex.setup {
  settings = {
    ltex = {
      language = 'en-US',
    },
  },
}
```
I understand that adding words to the dictionary is something that needs a plugin.

valentjn:
@David-Else The reason for the missing spelling corrections is that the default config in this PR uses the language `en`. [As the docs say](https://valentjn.github.io/ltex/settings.html#ltexlanguage), if you want spelling corrections, you have to use a specific variant like LT<sub>E</sub>X's default value `en-US`. I don't know if this is on purpose or not (if Neovim already has a spell checker that's activated by default, then it makes sense).

On a related note, generally, I would advise not setting defaults explicitly and let LT<sub>E</sub>X take care of those, if possible—except for good client-specific reasons. This makes changes of settings structure/default values easier and it enables a more consistent experience across clients. Plus, there are some thoughts behind most default values. Picky rules for instance, which this PR activates, are disabled by default as they generate quite some false positives, which can confuse users (e.g., valentjn/vscode-ltex#419).

mjlbach:
@valentjn Thank you so much for stopping by, it's extremely appreciated when upstream authors provide direct input.

> @David-Else The reason for the missing spelling corrections is that the default config in this PR uses the language en. As the docs say, if you want spelling corrections, you have to use a specific variant like LTEX's default value en-US. I don't know if this is on purpose or not (if Neovim already has a spell checker that's activated by default, then it makes sense).

`setspell` is not on by default, and IMO ltex-ls should be preferred over it.

> On a related note, generally, I would advise not setting defaults explicitly and let LTEX take care of those, if possible—except for good client-specific reasons.

Is there *any* setting here we need to be sending? I'm fine just wiping this, I believe I had asked about this before during review from the original PR author.

```
 ltex = {
        enabled = { 'latex', 'tex', 'bib', 'markdown' },
        checkFrequency = 'edit',
        language = 'en',
        diagnosticSeverity = 'information',
        setenceCacheSize = 2000,
        additionalRules = {
          enablePickyRules = true,
          motherTongue = 'en',
        },
        dictionary = {},
        disabledRules = {},
        hiddenFalsePositives = {},
      },
    },
```

I'm fine doing a quick follow-up PR based on your feedback, although it would be great if a neovim user who actively uses ltex-ls can take ownership over maintaining these settings from time-to time (perhaps keeping it in sync with grammar guard).

valentjn:
LT<sub>E</sub>X itself doesn't need any setting to be set. Missing settings are automatically set to their [default values](https://valentjn.github.io/ltex/settings.html).

The only setting that's actually necessary in this case should be [`ltex.enabled`](https://valentjn.github.io/ltex/settings.html#ltexenabled), because I think [Neovim uses the language ID `tex` for L<sup>A</sup>T<sub>E</sub>X documents](https://neovim.io/doc/user/filetype.html). LT<sub>E</sub>X follows the LSP spec, which recommends `latex` (`tex` is preferred for plain T<sub>E</sub>X or undetermined T<sub>E</sub>X dialects). That means there won't be L<sup>A</sup>T<sub>E</sub>X checking unless `tex` is added to `ltex.enabled`.

We had that problem before for the coc.nvim version in valentjn/vscode-ltex#425, where the problem was fixed by changing the mapping of filetypes to LSP language IDs in coc.nvim. Not sure if something like this would be possible here.

If not, then I'd suggest setting `ltex.enabled` to the default plus `'tex'`:

```
'bibtex', 'html', 'latex', 'markdown', 'org', 'restructuredtext', 'rsweave', 'tex'
```

The other settings can be removed. I haven't tried it, though.

mjlbach:
Yes, we can pass `get_language_id` to start client to do the remapping, similar to the solution in coc.nvim. I'll file the PR, thanks for the input!

mjlbach:
ref: for those in the thread https://github.com/neovim/nvim-lspconfig/pull/1364

Haven't tested, will check everything works later.

David-Else:
I am not sure grammar guard is happening any time soon, getting the ability to define a file to be the user dictionary and have the code action to add to it work would be pretty amazing at this point :)

mjlbach:
I encourage someone else to start an additiona ltexls.nvim then, it's out of scope of lspconfig to add off-spec handlers/requests. Lspconfig should always be a fallback/backup to a dedicated extension (or the defacto if the client is on-spec I guess).

PR #864, opened 2021-04-23
Add zeta-note

nanotee:
https://github.com/artempyanykh/zeta-note

### DIFF
--- a/lua/lspconfig/zeta_note.lua
+++ b/lua/lspconfig/zeta_note.lua
@@ -1,0 +1,32 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = "zeta_note"
+
+configs[server_name] = {
+  default_config = {
+    filetypes = {"markdown"};
+    root_dir = util.root_pattern(".zeta.toml");
+  };
+  docs = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/zeta_note.lua
```diff
@@ -0,0 +1,32 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = "zeta_note"
+
+configs[server_name] = {
+  default_config = {
+    filetypes = {"markdown"};
+    root_dir = util.root_pattern(".zeta.toml");
```
mjlbach, line 9: You might want to do the current working directory fallback, see pyright, not sure if .zeta.toml is actually required.

nanotee: This is what the vscode extension does: https://github.com/artempyanykh/zeta-note-vscode#installation

> Add `.zeta.toml` to your workspace root folder. The extension is automatically activated only when `.zeta.toml` file is present. This is done to avoid running Zeta Note on random Markdown files, but rather only inside your notes folder.

mjlbach: Ah ok! Great :) disregard

PR #869, opened 2021-04-27
Validate buffer before setup

Shatur95:
I wrote a plugin to switch sessions and I am intermittently encountering the following error:
изображение
[This](https://github.com/Shatur95/neovim-session-manager/blob/74bc3400d8cdb28a2089153614b210098bd6aa1e/lua/session_manager/init.lua#L19) code triggers the issue.
This PR fixes this problem by adding a buffer validation check. I think that this check will not hurt even for such a specific use case.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -188,6 +188,9 @@ function configs.__newindex(t, config_name, config_def)
   end
 
   function M._setup_buffer(client_id, bufnr)
+    if not vim.api.nvim_buf_is_valid(bufnr) then
+      return
+    end
     local client = lsp.get_client_by_id(client_id)
     if client.config._on_attach then
       client.config._on_attach(client, bufnr)


mjlbach:
You shouldn't be calling setup_buffer on invalid buffers. Nothing in nvim-lspconfig does that, and if it does, it's a bug.

Shatur95:
> You shouldn't be calling setup_buffer on invalid buffers. Nothing in nvim-lspconfig does that, and if it does, it's a bug.

This called automatically. My code just stops the LSP, removes all buffers, and does `:source session`.

mjlbach:
If you're not using lspconfig's built-in server management (LspStop for example), you should make sure to clear the autocommand  https://github.com/neovim/nvim-lspconfig/blob/0840c91e25557a47ed559d2281b0b65fe33b271f/lua/lspconfig/configs.lua#L141-L142

Shatur95:
The plugin is very simple, it just stops the LSP clienе using `vim.lsp.stop_client`. Should I be using something different for this?

mjlbach:
I'm not sure why your plugin is having issues, but calling vim.lsp.stop_client should normally be fine (that's what LspStop does). If you can reproduce the issue outside of your plugin, that would be useful.

Shatur95:
> If you can reproduce the issue outside of your plugin, that would be useful.

I will try. I know for sure that these lines are causing the problem, but this bug happen and periodically.

Shatur95:
@mjlbach, I think I figured out what was going on. When the buffer is removed, the `BufEnter` event is also called. the buffer switches to the next when the current buffer is removed. You can check this yourself using the following command:

```
autocmd BufEnter * lua print('called')
```

And then by removing all buffers using the following code:

```
for _, buffer in ipairs(vim.api.nvim_list_bufs()) do
  if vim.api.nvim_buf_is_valid(buffer) then
      vim.api.nvim_buf_delete(buffer, {})
  end
end
```

But you should have several buffers to reproduce.

So, I think that the fix is okay. Additional check will not hurt and will allow to avoid race condition.

mjlbach:
I'd recommend not unloading the buffers like that, you can use a snippet like:

```
vim.lsp.stop_client(vim.lsp.get_active_clients())

local cur_buf = vim.api.nvim_get_current_buf()
for _, buffer in ipairs(vim.api.nvim_list_bufs()) do
  if vim.api.nvim_buf_is_valid(buffer) and buffer ~= cur_buf then
      vim.api.nvim_buf_delete(buffer, {})
  end
end
vim.schedule(function() vim.api.nvim_buf_delete(0, {}) end)
```

I'm not sure what implications there are to silently not attaching buffers.

Shatur95:
You are right, this is the best way to do it. Thank you for the help!

PR #873, opened 2021-05-01
utils: fix script_path on windows

mjlbach:
@rleathart this addresses your issue on Windows?


rleathart:
@mjlbach Yea, this change fixes the issue

PR #875, opened 2021-05-03
Add configuration for java-language-server

zupatisc:
I found this to be an easier setup compared to jdtls so I think it should be included here as an alternative.

### DIFF
--- a/lua/lspconfig/java_language_server.lua
+++ b/lua/lspconfig/java_language_server.lua
@@ -1,0 +1,22 @@
+local lspconfig = require('lspconfig')
+local configs = require('lspconfig/configs')
+
+local name = 'java_language_server'
+
+configs[name] = {
+  default_config = {
+    cmd = {};
+    filetypes = {'java'};
+    root_dir = lspconfig.util.root_pattern('build.gradle', 'pom.xml', '.git');
+    settings = {}
+  };
+  docs = {
+    description = [[
+https://github.com/georgewfraser/java-language-server
+
+Java language server
+
+Point `cmd` to `lang_server_linux.sh` or the equivalent script for macOS/Windows provided by java-language-server
+]]
+  }
+}


lua/lspconfig/java_language_server.lua
```diff
@@ -0,0 +1,20 @@
+local lspconfig = require('lspconfig')
+local configs = require('lspconfig/configs')
+
+local name = 'java_language_server'
+
+configs[name] = {
+	default_config = {
+		cmd = { '/usr/share/java/java-language-server/lang_server_linux.sh'};
```
lithammer, line 8: This `cmd` isn't portable.

zupatisc: I wasn't aware it has to be, besides where the script is installed will vary. I could add a comment about adding the script to ones PATH and change the cmd to reflect that.

mjlbach: If it's not portable, don't include cmd. See jdtls

zupatisc: Alright I've removed the cmd and instead added a short description on what to point it to.

lithammer:
Please change the indentation from tabs to spaces (2).

lua/lspconfig/java_language_server.lua
```diff
@@ -0,0 +1,22 @@
...
+		root_dir = lspconfig.util.root_pattern('build.gradle', 'pom.xml', '.git');
+		settings = {}
+	};
+	docs = {
+		description = [[
+https://github.com/georgewfraser/java-language-server
+
+Java language server
+
+Point `cmd` to `lang_server_linux.sh` or the equivalent script for MACOS/Windows provided by java-language-server
```
lithammer, line 19:
```suggestion
Point `cmd` to `lang_server_linux.sh` or the equivalent script for macOS/Windows provided by java-language-server
```

lithammer:
Looks good. I'll merge after you squash the commits.

zupatisc:
Not sure if you are notified so just a quick tag to let you know I squashed the commits @lithammer

lithammer:
Thanks!

PR #879, opened 2021-05-04
gopls: change root_dir function for files in $GOPATH/pkg/mod

Snyssfx:
Closes an issue #804

### DIFF
--- a/lua/lspconfig/gopls.lua
+++ b/lua/lspconfig/gopls.lua
@@ -1,11 +1,24 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
+local last_root_dir = nil
+local gopath = os.getenv("GOPATH")
+if gopath == nil then gopath = "" end
+
+local function prev_server_for_pkgmod(fname)
+  local fullpath = vim.fn.expand(fname, ":p")
+  if string.find(fullpath, gopath.."/pkg/mod") and last_root_dir ~= nil then
+    return last_root_dir
+  end
+  last_root_dir = util.root_pattern("go.mod", ".git")(fname)
+  return last_root_dir
+end
+
 configs.gopls = {
   default_config = {
     cmd = {"gopls"};
     filetypes = {"go", "gomod"};
-    root_dir = util.root_pattern("go.mod", ".git");
+    root_dir = prev_server_for_pkgmod;
   };
   -- on_new_config = function(new_config) end;
   -- on_attach = function(client, bufnr) end;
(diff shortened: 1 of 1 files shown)


mjlbach:
This needs to be rebased :)

Snyssfx:
@mjlbach done)

lua/lspconfig/gopls.lua
```diff
@@ -1,11 +1,24 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
+local last_root_dir = nil
+local gopath = os.getenv("GOPATH")
+if gopath == nil then gopath = "" end
+
+local function prev_server_for_pkgmod(fname)
+  local fullpath = vim.fn.expand(fname, ":p")
+  if string.find(fullpath, gopath.."/pkg/mod") and last_root_dir ~= nil then
```
rliebz, line 10: There's a couple potential issues with determining the module path this way—`GOPATH` might not be set, in which case the `go` toolchain's default value would be used, or it may be set to more than one value, in which case the first value would be used.

The good news is, Go will resolve its environment variables for us with the defaults the toolchain uses, and there is an variable specifically for this path.

```lua
local gomodcache = vim.fn.trim(vim.fn.system("go env GOMODCACHE"))

-- ...

if string.find(fullpath, gomodcache) and last_root_dir ~= nil then
```

If it's not a safe assumption that the `go` executable is installed on the system here, it might make sense to add error handling.

mjlbach: @rliebz we can continue this elsewhere if you have good ideas, but I want to summarize why I haven't merged this/my thoughts on some revisions to the `setup {}` api.

setup's role is to detect a "workspace" really, and to attach all buffers under that workspace to the language server rooted at "root_dir", rootdir/workspace folders are part of the spec. When you open a system library, in vscode (for example), the language server completely ignores it as it is not a child of the current workspace, nor has the user actively opened a new workspace in the library file. With the upcoming changes in #1066, we're going to add an interactive root picker for determining workspace dir.

I have a couple thoughts on how to handle these out-of-root files. One option, is to do as vscode does. We do not attach to files that do not match our workspace dir, but then utilize something like the interactive root picker to allow the user to opt-in to adding the system libraries to the server workspace.

The other option might be to add a system_libraries entry to setup {} that could auto-root the server in the system libraries (or somewhere in gopath in this case), and only attach buffers in the user navigates into that.

I'd like to keep root_dir as it is, and not add these hacks to attach language servers to out of workspace projects, but formalize a mechanism that can be customized by the user.

Anyways, sorry, that was long-winded. I've been thinking about the best solution for this a lot and have not yet come up with something I find compelling. Getting rid of the util.path.dirname/cwd fallback will at least solve the multiple-language-server-per-project issue.

clason:
If this is still relevant after the recent changes, can you rebase and update (or close otherwise)?

mjlbach:
I'm aware of this PR and it's still relevant, but it's still unclear what to do about it. You can check the discussion in the linked issue.

glepnir:
Looks like this pr is out of date.

PR #880, opened 2021-05-04
Add unified language server support

eruizc-dev:
Adds [Unified Language Server](https://github.com/unifiedjs/unified-language-server) for markdown and text files.

I had to make two entries: `unified_language_server_md` and `unified_language_server_txt` due to it having different arguments depending on what you want to check.

I ran the linters with success but had to manually generate the docs because the script would change way too many things.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -74,6 +74,8 @@ that config.
 - [texlab](#texlab)
 - [tflint](#tflint)
 - [tsserver](#tsserver)
+- [unified_language_server_md](#unified_language_server_md)
+- [unified_language_server_txt](#unified_language_server_txt)
 - [vala_ls](#vala_ls)
 - [vimls](#vimls)
 - [vls](#vls)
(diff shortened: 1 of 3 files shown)


mjlbach:
You dont need to generate the docs with these commits (in fact, don't). You can modify cmd at runtime with on_new_config (or perhaps one of the other hooks), to avoid having duplicate entries like this.

eruizc-dev:
> You dont need to generate the docs with these commits (in fact, don't). You can modify cmd at runtime with on_new_config (or perhaps one of the other hooks), to avoid having duplicate entries like this.

I'll try what you mention and re-activate the pr later!

eruizc-dev:
I'm quite in a struggle here, since I can't find a way to integrate both markdown and text support into a single language server.

I've seen the hooks but they're mostly per project and not per file. `on_new_config` updates when `root_dir` changes, the issue here is I may have a split with a markdown file on the left and a text on the right, each one requiring its own independent language server.

Even with other hooks or settings neither of those seem to allow me to have two instances of the same language server running with different params.

If anyone knows an easy fix please let me know, if not I'll be closing this PR because after using this language server for a while I realized it is not what I expected and provides zero value. Also looks like it's not maintained anymore and I don't feel like wasting time in something that nobody's gonna use.

mjlbach:
I was a bit surprised you wanted to package it based on the fact that it seems inactive. We can close the PR, and if you find yourself using it in the future feel free to reopen.

eruizc-dev:
I write enough documentation to make a spellchecker something useful for me. I've been looking at plenty of spellcheckers and this one was the one I found the most interesting initally. After using it for a few days wanted to make this PR, but after a couple more days I found out it's not as good as I thought it was.

mjlbach:
ltex-lsp seems like a possible alternative, it wraps language tool, not sure if it can be used outside of latex. There's also https://github.com/lewis6991/spellsitter.nvim and https://github.com/inkarkat/vim-SpellCheck

PR #884, opened 2021-05-06
Allow EFM to work on non-git repos

WhyNotHugo:
The current settings for EFM make it so that it will only work within a git repository.

This change adds a fallback, so that if the current file is _not_ inside a git repository
then the current directory is used instead.

This is useful, for example, when opening a random python file, so as to get feedback
from the system flake8.

### DIFF
--- a/lua/lspconfig/efm.lua
+++ b/lua/lspconfig/efm.lua
@@ -8,7 +8,9 @@ local bin_name = "efm-langserver"
 configs[server_name] = {
   default_config = {
     cmd = {bin_name};
-    root_dir = util.root_pattern(".git");
+    root_dir = function(fname)
+      return util.root_pattern(".git")(fname) or util.path.dirname(fname)
+    end;
   };
 
   docs = {
@@ -18,7 +20,7 @@ https://github.com/mattn/efm-langserver
 General purpose Language Server that can use specified error message format generated from specified command.
 ]];
     default_config = {
-      root_dir = [[root_pattern(".git")]];
+      root_dir = [[util.root_pattern(".git")(fname) or util.path.dirname(fname)]];
     };
   };
 }


lua/lspconfig/efm.lua
```diff
@@ -8,7 +8,9 @@ local bin_name = "efm-langserver"
 configs[server_name] = {
   default_config = {
     cmd = {bin_name};
-    root_dir = util.root_pattern(".git");
+    root_dir = function(fname)
```
mjlbach, line 11: You should use util.path.dirname for this

https://github.com/neovim/nvim-lspconfig/blob/f68f7d4c30abb3e4b8ebe6e3a604cc6f9eb924e8/lua/lspconfig/pyright.lua#L30-L33

lua/lspconfig/efm.lua
```diff
@@ -8,7 +8,9 @@ local bin_name = "efm-langserver"
...
+      return util.root_pattern(".git")(fname) or vim.fn.getcwd()
```
lithammer, line 12: I think `util.path.dirname(fname)` might be better than `vim.fn.getcwd()` because it makes more sense if you do something like this:

```console
$ nvim ../../foo/bar/baz.py
```

```suggestion
      return util.root_pattern(".git")(fname) or util.path.dirname(fname)
```

lithammer: Damn, @mjlbach beat me to it 🤷‍♂️

mjlbach: Convergent evolution :P

lua/lspconfig/efm.lua
WhyNotHugo: Ah, makes sense. Done.

lua/lspconfig/efm.lua
```diff
@@ -18,7 +20,7 @@ https://github.com/mattn/efm-langserver
 General purpose Language Server that can use specified error message format generated from specified command.
 ]];
     default_config = {
-      root_dir = [[root_pattern(".git")]];
+      root_dir = [[root_pattern(".git") or "vim's starting directory"]];
```
mjlbach, line 23: You should copy from above, it's not vim's starting directory anymore

lua/lspconfig/efm.lua
WhyNotHugo: Done. (Impressive feedback timing!)

lithammer (approved):

mjlbach:
After you address my comment re: doc, I'd change the commit messsage to be "efm: include current file directory in root detection"

lua/lspconfig/efm.lua
```diff
@@ -18,7 +20,7 @@ https://github.com/mattn/efm-langserver
 General purpose Language Server that can use specified error message format generated from specified command.
 ]];
     default_config = {
-      root_dir = [[root_pattern(".git")]];
+      root_dir = [[root_pattern(".git") or dirname]];
```
mjlbach, line 23: I would make this `util.root_pattern(".git")(fname) or util.path.dirname(fname)`

WhyNotHugo: I followed the same pattern the docs for other have here.

WhyNotHugo:
Changed commit message but slightly abbreviated since that went over the limit.

lua/lspconfig/efm.lua
mjlbach: Yes, they are wrong :)

WhyNotHugo: Why not include the whole block though?

```lua
    root_dir = function(fname)
      return util.root_pattern(".git")(fname) or util.path.dirname(fname)
    end;
```

I don't think the snippet above can be copy-pasted anyway due to the function taking a param, right?

mjlbach: It's just text, in fact, if you leave out root_dir entirely that's exactly what will be pasted by docgen.

mjlbach:
FYI you don't need to request my review, I'm just waiting for the tests to pass

PR #885, opened 2021-05-06
Allow bashls to work with Makefile and Zsh script

rekaerst:
Allow bash-language-server to work with Makefile and Zsh scripts since their commands are similar


mjlbach:
I don't think we should add this as a default, but users can add this locally in their config if they understand that some diagnostics are will be incorrect

rekaerst:
IIRC bashls doesn't have diagnostics.

mjlbach:
According to the readme, it does.

rekaerst:
My apology.
This could be closed I think.

PR #900, opened 2021-05-12
modify jdtls env var example for one that works on all systems

rafamadriz:
Why ?

Not all systems install openjdk in the same path, or even with the same name as the one provided. This replaces the example for one that works across all systems.

### DIFF
--- a/lua/lspconfig/jdtls.lua
+++ b/lua/lspconfig/jdtls.lua
@@ -141,7 +141,7 @@ inferred. Please set the following environmental variables to match your install
 ```bash
 export JAR=/path/to/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/repository/plugins/org.eclipse.equinox.launcher_1.6.0.v20200915-1508.jar
 export GRADLE_HOME=$HOME/gradle
-export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-11.0.9.11-9.fc33.x86_64/
+export JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:/bin/java::")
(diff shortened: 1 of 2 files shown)


mjlbach:
Thanks but config.md is auto-generated (that should be clear in the pull request template), so you'll need to modify jdtls.lua

rafamadriz:
Sorry didn't know that, just fixed it.

lua/lspconfig/jdtls.lua
```diff
@@ -141,7 +141,7 @@ inferred. Please set the following environmental variables to match your install
 ```bash
 export JAR=/path/to/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/repository/plugins/org.eclipse.equinox.launcher_1.6.0.v20200915-1508.jar
 export GRADLE_HOME=$HOME/gradle
-export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-11.0.9.11-9.fc33.x86_64/
+export JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:/bin/java::")
```
lithammer, line 144: Unfortunately this isn's as portable as you think it is. It only works on GNU Linux. The `-f` flag doesn't exist on the BSD variant (e.g. macOS).

Seems like the post portable solution might be `realpath(1)` or Perl/Python one-liner. Though I guess they also have some caveats.

```bash
perl -MCwd -le 'print Cwd::abs_path shift' /usr/bin/java
```
```bash
python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' /usr/bin/java
```
```bash
realpath /usr/bin/java
```

https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac

PR #910, opened 2021-05-16
docs: simplify and reorganize

mjlbach:


README.md
```diff
@@ -1,170 +1,60 @@
...
-          if params and params.type <= vim.lsp.protocol.MessageType.Warning.Error then
-            vim.lsp.handlers["window/showMessage"](err, method, params, client_id)
-          end
-        end;
-    }
-  }
-)
+```vim
+lua << EOF
+require'lspconfig'.rust-analyzer.setup{}
```
elianiva, line 49: should be `rust_analyzer`?

mjlbach: Done, thanks!

mjlbach:
Why are people reacting with confused emoji? Please comment if you have suggestions @kuator @shadmansaleh @andersevenrud

README.md
```diff
@@ -1,170 +1,60 @@
...
-that are installed on your system.
+A collection of common configurations for Neovim's built-in [language server client](https://neovim.io/doc/user/lsp.html).
+This repo handles automatically launching and initializing language servers that are installed on your system.
 
 ## Install
 
-* Requires [Neovim HEAD/nightly](https://github.com/neovim/neovim/releases/tag/nightly) (v0.5 prerelease). The configs in this repo
-  assume that you are using the latest [Neovim HEAD/nightly build](https://github.com/neovim/neovim/releases/tag/nightly). 
-  Update Neovim and nvim-lspconfig before reporting an issue.
+* Requires [Neovim HEAD/nightly](https://github.com/neovim/neovim/releases/tag/nightly) (v0.5 prerelease). Update Neovim and nvim-lspconfig before reporting an issue.
```
levouh, line 8: I think the reduction to the "wall of text" here is awesome, as a lot of people seemingly don't realize that to get going, you can literally just install a server, setup this repo, and write a single line to your init script to have an insanely useful baseline. If that isn't plug-and-play, I'm not really sure what is.

One thing here to continue with the streamlining might be to add a table of contents @mjlbach? The README after this update serves as really useful overview, but people need to understand up front that this is not the end of the capabilities provided by this repository. The interface/baseline is really where the true power _starts_, so showing them how to get started, then providing resources to immediately change/add seems like it'd be useful.

mjlbach: > The interface/baseline is really where the true power starts, so showing them how to get started, then providing resources to immediately change/add seems like it'd be useful.

What do you mean by this? The cut content will go into the wiki, so I was thinking of adding a mini-table of contents (sort of like what is there already) to this missing content.

kuator:
Herd mentality in my case

README.md
levouh: You've noted exactly what I was trying to say; the _README_ is a good "short read", but it is also important to show where people can learn more. My comment was mainly around the idea that people might not want to scroll to the bottom of this already short _README_, to do so, and presenting an early "catering to the users interest level" earlier on might serve better.

andersevenrud:
My real reaction based on emojis would probably be :thinking: , but that's not an option :sweat_smile: 

The reduction of noise is great, but there was removal of some stuff that I found helpful, ex `setup()`  Maybe this is provided elsewhere ?

README.md
levouh: This meaning pretty much a table of contents to the _README_ itself at the very top, one of the sections in that table of contents would maybe be a table to the Wiki to learn more, etc.

levouh: Also, and I know you haven't gotten there yet, but this:

image

is a bit daunting (especially to people who have trouble moving their eyes slightly to the right ;) ). Not sure what other things _Github_ provides to spruce that up, but that might help as well.

mjlbach:
> The reduction of noise is great, but there was removal of some stuff that I found helpful, ex setup() Maybe this is provided elsewhere ?

You mean the explanation of the signature of the setup function? I was planning on moving that to the wiki. I do think it's valuable, but I'm trying to remove anything that seems complicated/requiring custom code from the top level.

README.md
mjlbach: Sadly it doesn't provide anything. I tried deleting the HOME page but it results in a generic list of entries without the nice sidebar. I'm going to turn HOME into a table of contents like the coc wiki after moving content from the readme into the wiki.

README.md
```diff
@@ -192,134 +82,29 @@ local on_attach = function(client, bufnr)
   buf_set_keymap('n', '[d', '<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>', opts)
```
levouh, line 82: What these each do is pretty obvious to someone who has used the LSP client before, but inline comments here might be useful? Obviously all of this can be found in the _Wiki_ as well, but just to entice people a bit more to the client itself (understanding that this is indeed just wrapping config for it, but to most it provides their first experience with the client at all).

andersevenrud:
> You mean the explanation of the signature of the setup function?

Yup.

> I was planning on moving that to the wiki. I do think it's valuable, but I'm trying to remove anything that seems complicated/requiring custom code from the top level.

Hm. Are you open to having these as plain ol' markdown files in the repo ? Gotta say I like to have the option just to read this straight off the repo (especially seeing that it's already installed on my machine).

README.md
```diff
@@ -1,170 +1,60 @@
...
-      root_dir = function(fname)
-        return lspconfig.util.find_git_ancestor(fname) or vim.loop.os_homedir()
-      end;
-      settings = {};
-    };
-  }
-end
-lspconfig.foo_lsp.setup{}
-```
+Enabling most language servers is as easy as:
```
camspiers, line 45: I would say:

Enabling most language servers is as easy as installing that server globally, and configuring its setup:

mjlbach:
> Hm. Are you open to having these as plain ol' markdown files in the repo ? Gotta say I like to have the option just to read this straight off the filesystem (especially seeing that this is always installed).

Totally, but I want to do it in a way that it's somewhat 'hidden'. We really need to shake the narrative that this is complicated to use...

README.md
mjlbach: Would a see `:help vim.lsp.*` suffice? The built-in docs are quite good

README.md
mjlbach: Thanks, I'll make that edit.

README.md
elianiva: > add a table of contents

if you mean README's table of content, I don't think that's necessary as Github is now capable of generating TOC automatically and put it on a menu above the README. People just need to get used to it :p

image

README.md
```diff
@@ -1,170 +1,60 @@
...
-modifying CONFIG.md directly; CONFIG.md will be overwritten by docgen
-
-**You do not need to copy settings or init_options from this configuration into your config**
+For a full list of servers, see [CONFIG.md](CONFIG.md). This document contains installation instructions and additional, optional customization suggestions for each language server. For some servers that are not on your system path (jdtls, elixirls) you will be required to manually add `cmd` as an entry in the table passed to setup.
 
 ## Keybindings and completion
 
-The following maps most of the standard functions to keybindings, and maps omnicomplete to use
-the lsp.omnifunc. See `:help lsp` for more details
+nvim-lspconfig does not include any keybindings or completion enabled by default. The following snippet uses the on-attach function to map keybindings and omnifunc (\<c-x\>\<c-o\>) for completion whenever the language server client activates on a given buffer:
```
camspiers, line 56: I think there is an educational point to be made here, I suspect a lot of users coming from coc for example don't understand the autocompletion features are being provided by coc, and so when they come here, they don't realize either they need to learn how to trigger completion via omnifunc's binding <c-x><c-o> and hook up omnifunc via `buf_set_option('omnifunc', 'v:lua.vim.lsp.omnifunc')` as below, or that they need to install another plugin, e.g. deoplete, or compe etc and then configure that. So while this aids the user in understanding that it doesn't include keybindings and complete by default, it doesn't help them to understand what to expect from lspconfig after pasting the below.

mjlbach: What do you mean? I do include the keybinding in the above for omnifunc snippet for omnifunc, and it is added in the example configurations + keybinding. What would you suggest we add to clarify?

 Totally agree re: the confusion about autocompletion vs completion...

andersevenrud:
> Totally, but I want to do it in a way that it's somewhat 'hidden'. We really need to shake the narrative that this is complicated to use...

Not sure if making it "hidden" is a good idea. If the README file has a table of contents and a section that directs users towards interface documentation and advanced usage, that might meet a nice middle ground.

Btw, I only recently transitioned over to the built-in LSP and all of this Lua stuff, and in hindsight I can kind of see why it's considered "complicated" to some users (especially coming from `coc.nvim` etc), because now you have to add (or, which is more likely the case copy/paste) a bunch of small snippets to get going, while other solutions can be close to zero-config. So this probably  goes a bit deeper than having clean and simple docs :) But I'm sure someone out there makes something to bridge this ease-of-use gap some day (or already has).

Personally, I prefer this method because I have total control over everything.

README.md
levouh: I'd absolutely add that, but I think part of what you've mentioned around people thinking that the builtin LSP client is "hard to use" comes from a lack of understand to how much is already there for you. So here, I think adding comments like "move back and forth between server-reported diagnostics" might help flip that switch by putting things in plain English?

Obviously this isn't necessary by any means, and duplicating what is already in `:h lsp` is useless, but I think it might help bridge the gap in understanding that some new users have.

mjlbach:
> Btw, I only recently transitioned over to the built-in LSP and all of this Lua stuff, and in hindsight I can kind of see why it's considered "complicated" to some users (especially coming from coc.nvim etc), because now you have to add (or, which is more likely the case copy/paste) a bunch of small snippets to get going, while other solutions can be close to zero-config. So this probably goes a bit deeper than having clean and simple docs :) But I'm sure someone out there makes something to bridge this ease-of-use gap some day (or already has).

I want to have a conversation about this because it keeps coming up. If you don't want to map any keybindings, it should be approximately 3 lines to activate every language server in nvim-lspconfig. The *only* overhead comes from:

1. Mapping keybindings to neovim functions
2. Either enabling omnifunc (1 line) or compe (~10 lines)

We can't do 1 by default, everyone has strong opinions about keybinds. I also assume like with CoC, most people want to choose which language servers are enabled.

README.md
camspiers: I think I just mean to emphasize that a user who wants auto-completion (rather than just completion) will need to explore plugins that specialize in adding such features to nvim.

Something like the following, though I acknowledge this is more long winded:

> nvim-lspconfig does not automatically add keybindings (e.g. for go to definition, hover) or configure nvim's inbuilt omnifunc completion by default. If auto-completion beyond nvim's inbuilt completion is desired another plugin is needed. However, the following example snippet demonstrates how to provide an on-attach function in order to map keybindings and nvim's omnifunc (\<c-x\>\<c-o\>) for completion whenever the language server client activates on a given buffer:

camspiers: I also think it's on us here to educate a user about what an on_attach function and even what buffer local mappings are.

Something like, again I understand it is long winded:

Given that we don't want lsp related keybindings added to all buffers, only buffers in which a server is active, we take advantage of the on_attach property when configuring each server. This function, when provided, will be executed when a server auto attaches for an appropriate buffer, meaning you will now have buffer local mappings provided as well as the appropriate configuration of omnifunc.

README.md
mjlbach: Done

README.md
mjlbach: I'll debate what to add notes to, I added a small note to omnifunc and the thing about ':help`

andersevenrud:
> I want to have a conversation about this because it keeps coming up.

In the case of transitioning users, maybe this is because of expectations ? If the documentation here was more explicit about what this actually provides then it could help with the case.

> Maybe if we just make it extremely specific that you only need to do 2 things to get started (keybinds + choose completion method?)?

Maybe. Keeping the README dead simple with the bare minimum instructions on how to get rolling would be nice, and then redirect for further reading material as I previously mentioned.

README.md
mjlbach: Ooh I didn't even know about that, looks nice!

README.md
mjlbach: I think I worked those both in, I tried to reduce the word count a bit but it's in the prelude to the example keybindings + completion

shadmansaleh:
My reaction was mainly because, I think the docs were already very clear . I think main reason people are complaining about nvim-lsp being complicated because they are comparing with coc.nvim.  The fact that in most cases in coc.nvim you can setup lsp by runnung `:CocInstall` once makes this look setting up yourself look complicated . I think people should realize coc.nvim and nvim-lsp aren't the samething and has different goals . @mjlbach like I've told you before you're paying too much attention to reddit.

I have nothing against this change by the way :)

README.md
camspiers: Sounds good

mjlbach:
> You're paying too much attention to reddit

:fire: I don't know what you are talking about :fire: 

> I have nothing against the changes by the way :)

Do you think the changes improve the readability/on-boarding? I'll make sure the "lost" content is in a contributing/advanced_configuration.md.

camspiers:
@shadmansaleh My view is that it is on us to help correct certain common incorrect assumptions folks transition from coc have, e.g. around completion, auto-completion, server install, automated server install, automatic server install on startup.

The costs of addressing this well in the docs seem minimal, but the benefits could be that people at more informed about what the project is (its scope), we receive less reports from users about it missing X or Y, or X or Y not working, and the community has a better sense of what pieces are required in a lsp nvim setup to achieve what something like coc does, e.g. using something like compe for autocompletion, using something like lspinstall for server installation, and eventually finding a nicer solution for automatic server install).

andersevenrud:
@shadmansaleh And another important thing when it comes to comparing nvim LSP features to existing plugins that *incorporates* LSP is that the latter usually also provides features that tightly integrates with the UI. You have to install for example saga, lsp_signature, and other supplements to get the same-ish experience.

camspiers:
@mjlbach Just read over your new changes. They are spot on. Awesome.

mjlbach:
> And another important thing when it comes to comparing nvim LSP features to existing plugins that incorporates LSP is that the latter usually also provides features that tightly integrates with the UI. You have to install for example compe, saga, lsp_signature, and other supplements to get the same-ish experience.

Which features are you missing? @elianiva and I already sherlocked borders from lspsaga, and signature support is built-in with C-k from the keymappings

andersevenrud:
> Which features are you missing? @elianiva and I already sherlocked borders from lspsaga, and signature support is built-in with C-k from the keymappings

Oh, I didn't know that borders was an option!

As for the "missing" stuff, it was mainly thinking of how everything is displayed in floating windows and that you get signature help on typing (also [this](https://github.com/neovim/neovim/issues/14444) which is missing at this moment), also some of the CoC language server clients bridges the gaps when it comes to certain features. The one that comes to mind is with TypeScript where you get snippet support for signature autocompletion (which is not in tsserver).

mjlbach:
> As for the "missing" stuff, it was mainly thinking of how everything is displayed in floating windows (and some other UI sugar)

Hmm everything in core is using floating windows. What is the missing UI sugar?

> that you get signature help on typing 

Yeah as you noted, that can be configured, I personally don't like this 

>  also some of the CoC language server clients bridges the gaps when it comes to certain features. The one that comes to mind is with TypeScript where you get snippet support for signature autocompletion (which is not in tsserver).

Typescript has always been a pain, the typescript team is apparently releasing a language server soonish (at least according to the rumor mill), at which point all of these concerns will be gone.

andersevenrud:
> Hmm everything in core is using floating windows. What is the missing UI sugar?

Ex of prompt and not a floating window: `vim.lsp.buf.code_action()` and `vim.lsp.buf.references()` displays a new split.

Maybe I'm missing something obvious here :blush: 

> Yeah as you noted, that can be configured, I personally don't like this

Maybe make a mention of how to do this for people that this behavior ?

> Typescript has always been a pain, the typescript team is apparently releasing a language server soonish (at least according to the rumor mill), at which point all of these concerns will be gone.

:crossed_fingers: I haven't caught the sniff of this...but my first thought is that this would be a port of the VSCode stuff that the CoC extension used :thinking:

mjlbach:
> Ex of prompt and not a floating window: vim.lsp.buf.code_action() and vim.lsp.buf.references() displays a new split.

Ah yes, we leverage inputlist and quickfix for those IIRC. Yeah different strokes, as you probably know lspsaga does floating windows for those. Not sure how people feel, personally I think quickfix list is more "native" and useful (integration with cdo) and the like

> Maybe make a mention of how to do this for people that prefer this behavior?

Maybe in the wiki page? I don't use it so someone else can contribute it.

andersevenrud:
> Yeah different strokes

Indeed. I'm grown to prefer having all of this stuff popping up under my cursor in my workflow. I usually work with high resolutions and a lot of splits, so getting all this info directly in my line of sight is the most comfortable.

> Maybe in the wiki page

I'm not really sure about placing documentation into a Wiki. I feel like it all should be available via the repository source in some form. Doing it this way you can just generate github pages instead of relying on the Wiki feature.

README.md
```diff
@@ -1,181 +1,78 @@
...
+cd test_python_project
+git init
+touch main.py
 ```
+4. Launch neovim, the language server will now be attached and providing diagnostics (see `:LspInfo`)
+`nvim main.py`
+5. See [Keybindings and completion](#Keybindings-and-completion) for mapping useful functions and enabling omnifunc completion
 
-### Example: using the defaults
+## Automatically launching language servers
```
David-Else, line 38: I don't think this heading is the best choice here. It implies that you need to do something to automatically launch a language server. Maybe change it to ``The project root and the `root_dir` pattern``?

README.md
```diff
@@ -1,181 +1,78 @@
...
-To use the defaults, just call `setup()` with an empty `config` parameter.
-For the `gopls` config, that would be:
+In order to automatically launch a language server, lspconfig searches up the directory tree from your current buffer to find a file matching the `root_dir` pattern defined in each server's configuration.  For [pyright](CONFIG.md#pyright), this is any directory containing ".git", "setup.py",  "setup.cfg", "pyproject.toml", or "requirements.txt").
 
-```lua
-require'lspconfig'.gopls.setup{}
-```
-
-### Example: override some defaults
+Language servers require each project to have a `root` in order to provide completion and search across symbols that may not be defined in your current file, and to avoid having to index your entire filesystem on each startup.
```
David-Else, line 42: If you change the heading then this sentence should be the first, to introduce the key concept of the `root` before going into detail.

David-Else:
Overall this is a big improvement, awesome!

I think what is missing is a general overview of what this plugin does in the context of the entire ecosystem of 'extension' plugins that nearly everyone will be forced to interact with to get a nice working LSP system.

It seems implied that everyone coming to this repo knows exactly what the Neovim LSP is, what it does, and more importantly what it does not do. As 99% of people using the LSP will be installing this extension, AND it is in the official Neovim github I think it should provide an overview section. In it there should be things like:

- Features the LSP offers (code completion, linting, formatting, using general LSPs to use other tools like Prettier etc)
- The fact that you will need an autocompletion plugin, and links to the ones available.
- You will want a plugin that alerts you of code actions
- Comparisons to CoC in a nice table of features (maybe)

Basically it would address the complete beginner, and also act as an advert for people thinking about swapping from another editor or CoC.

Maybe this is totally out of scope? If it is, then this info should be somewhere on the official Neovim website. If it were not for ploughing through random Reddit posts  I would not have been able to use Neovim LSP, and that is not a good situation.

README.md
```diff
@@ -192,134 +89,30 @@ local on_attach = function(client, bufnr)
...
-```
+The `on_attach` hook is used to only activate the bindings after the language server attaches to the current buffer.
 
 ## Debugging
-The two most common points of failure are 
+The two most common reasons a language server does not start or attach are:
 
-1. The language server is not installed. You should be able to run the `cmd` 
-defined in each lua module from the command line.
+1. The language server is not installed. nvim-lspconfig does not install language servers for you. You should be able to run the `cmd` defined in each server's lua module from the command line and see that the language server starts. If the `cmd` is an excutable name, ensure it is on your path.
```
nanotee, line 109: Perhaps the README should ask users to run `:checkhealth` since it reports whether a command is executable?

tjdevries:
Can we just have it be a running list of your best reddit replies?

README.md
mjlbach: My issue with this is it's not attention grabbing. People might breeze past a discussion on "project roots" thinking it's some advanced feature but this is 99% of the issues with language servers

README.md
mjlbach: No worries!

mjlbach:
> Features the LSP offers (code completion, linting, formatting, using general LSPs to use other tools like Prettier etc)

Done

> The fact that you will need an autocompletion plugin, and links to the ones available.

For clarification, you don't need an autocompletion plugin to use completion, but I linked to the wiki in the feature overview

> You will want a plugin that alerts you of code actions

WDYM? There currently is *not* a way in the spec to notify you of where code actions are available without just trying a code action to my knowledge

> Comparisons to CoC in a nice table of features (maybe)

I'd like to avoid direct comparison with CoC, vim-lsp, etc. Don't want to feed into the competition narrative.

I integrated your suggestions, one disadvantage is it now increases the number of lines to the quick-start section. Let me know what you think

mjlbach:
@andersevenrud I added an "ADVANCED_README.md" (name suggestions accepted), with all of the old info.

David-Else:
@mjlbach Love the new LSP overview, thanks.

> WDYM? There currently is not a way in the spec to notify you of where code actions are available without just trying a code action to my knowledge. I guess some people like lsp-lightbulb

Yeah, I was talking about `lsp-lightbulb`. I see it as an essential, I think a lot of new users will want this functionality.

Fair enough to avoid competition narrative with CoC. I do think it is the benchmark that Neovim LSP will be judged, and many Reddit comments for ages have just been 'should I use native LSP or CoC?', I thought  the comparison could answer this very popular question.

mjlbach:
> Yeah, I was talking about lsp-lightbulb

Aren't the diagnostics from this one cached? I think it will still show invalid code actions on a line as it requires caching code actions when you hover over a specific area. I'm hoping they extend the protocol to allow us to query document code actions so we can display something in the sign column. It can go in the wiki but I don't want to put any plugins in the readme.

andersevenrud:
@mjlbach 

> I added an "ADVANCED_README.md" (name suggestions accepted), with all of the old info.

Great!

Also loving these new changes/addition since my last visit :)

mjlbach:
> Can we just have it be a running list of your best reddit replies?

@tjdevries no one wants to see that, especially not reddit 😆

mjlbach:
Just to summarize the remaining commentary:

* @levouh You think there should be a table of contents, but possibly github's auto TOC mitigates this
* @levouh You think we should add a little bit of wording above each mapped function to explain what they do (already did some of this)

* @David-Else You think "Automatically launching language servers" should be renamed to explicitly call out root dir

Is everyone happy with the location/split of README.md and ADANCED_README.md? Seems like we have a lot of markdown at the top level now 😆 

Let me know if I missed anything! I've tried to incorporate all of the other feedback.

andersevenrud:
> You think there should be a table of contents, but possibly github's auto TOC mitigates this

I vote for having the ToC inside the markdown file for visibility. That Github auto ToC feature seems a bit obscure to me... I stumbled across it by accident :sweat_smile:

mjlbach:
I'm going to merge for now and we can do a follow-up with this feedback.

README.md
```diff
@@ -1,181 +1,93 @@
...
+Neovim supports the Language Server Protocol (LSP), which means it acts as a client to language servers and includes a Lua framework `vim.lsp` for building enhanced LSP tools. LSP facilitates features like:
+
+- go-to-definition
+- find-references
+- hover
+- completion
+- rename
+- format 
+- refactor
+
```
teto, line 17: it looks exhaustive, add `...` ? wiith maybe a link to `:h lsp-method` so that one can find the exhaustive list.

README.md
```diff
@@ -1,181 +1,93 @@
...
+Neovim provides an interface for all of these features, and the language server client is designed to be highly extensible to allow plugins to integrate language server features which are not yet present in Neovim core such as [**auto**-completion](https://github.com/neovim/nvim-lspconfig/wiki/Autocompletion) (as opposed to manual completion with omnifunc) and [snippet integration](https://github.com/neovim/nvim-lspconfig/wiki/Snippets).
```
teto, line 18: I would remove the `be highly extensible` that looks like a bit too much marketing :)

README.md
```diff
@@ -1,181 +1,93 @@
...
-
-See [CONFIG.md](CONFIG.md) for documentation and configuration of individual language servers.
-This document contains installation instructions for each language server, and is 
-auto-generated from the documentation in the lua source. Do not submit PRs 
-modifying CONFIG.md directly; CONFIG.md will be overwritten by docgen
-
-**You do not need to copy settings or init_options from this configuration into your config**
+For a full list of servers, see [CONFIG.md](CONFIG.md). This document contains installation instructions and additional, optional customization suggestions for each language server. For some servers that are not on your system path (jdtls, elixirls) you will be required to manually add `cmd` as an entry in the table passed to setup. Most language servers can be installed in less than a minute.
 
 ## Keybindings and completion
```
teto, line 70: for me this part is already advanced. I would lean to move it to another page and reference it here. The longer the root README.md, the less likely people will read I believe.

README.md
```diff
@@ -332,19 +140,27 @@ Attempt to run the language server, and open the log with:
...
+Please see the [wiki](https://github.com/neovim/nvim-lspconfig/wiki) for additional topics, including:
+
+* [Installing language servers automatically](https://github.com/neovim/nvim-lspconfig/wiki/Installing-language-servers-automatically)
+* [Snippets support](https://github.com/neovim/nvim-lspconfig/wiki/Snippets-support)
+* [Project local settings](https://github.com/neovim/nvim-lspconfig/wiki/Project-local-settings)
+* [Recommended plugins for enhanced language server features](https://github.com/neovim/nvim-lspconfig/wiki/Language-specific-plugins)
+
+## Windows
+
+In order for neovim to launch certain executables on Windows, it must append `.cmd` to the command name. To work around this, manually append `.cmd` to the entry `cmd` in a given plugin's setup{} call.
```
teto, line 163: I read the `it must append ...` as if neovim was doing it. Rephrasing it so that it is clearer it is the user's responisbility`You may need to append `.cmd` to the cmd entry on windows otherwise neovim wont be able to launch the executable.` . Maybe this is a check that could be done in checkhealth/LspInfo (hoping a windows user reads this).

README.md
mjlbach: by this you mean the following right? 

`For some servers that are not on your system path (jdtls, elixirls) you will be required to manually add `cmd` as an entry in the table passed to setup. Most language servers can be installed in less than a minute.`

README.md
mjlbach: Done in r3 PR

README.md
mjlbach: Done in r3 PR, but we should fix this somehow... I had a workaround but it breaks packer, there's an upstream issue in core about it.

README.md
mjlbach: Done in r3 PR

README.md
teto: I meant moving the whole  `## Keybindings and completion` section to its own page or on the advanced page. 

As for the sentence you quote `Most language servers can be installed in less than a minute.` sounds unhelpful I would remove it. I would condense the next one with:
`For servers not on your system path (jdtls, elixirls) you must set `cmd` to the full path .` and maybe add a tag t `lspconfig-setup()-cmd-arg ` yo ucould reference here.

David-Else:
Regarding a TOC, great idea, in case you have not encountered the incredible power of https://pandoc.org/, here is my personal script to quickly generate a TOC from a Markdown:
```
#!/usr/bin/env bash

# use pandoc to create a TOC

# The $# variable will tell you the number of input arguments the script was passed
if [ $# -eq 0 ]; then
    echo "No arguments supplied, please supply file name."
fi

#  -s                    --standalone
#  -f FORMAT, -r FORMAT  --from=FORMAT, --read=FORMAT
#  -t FORMAT, -w FORMAT  --to=FORMAT, --write=FORMAT
#  -o FILE               --output=FILE

#  -f markdown-smart       disables the smart extension that changes ''
pandoc -s -f markdown-smart -t gfm --toc -o "$1-with-toc.md" "$1"
```

README.md
mjlbach: Ooh maybe, but the issue is there are people who wanted this to be literally the first thing because if you copy/paste this you are basically done configuring 😆

levouh:
>You think there should be a table of contents, but possibly github's auto TOC mitigates this

@mjlbach the built-i one from GitHub is nice, but I'm not sure how many people are aware it exists. You can add it manually, or do something like what @David-Else noted with `pandoc` or similar. I can't imagine it makes a _huge_ difference, but like I said it just helps cater to interest level more easily.

>You think we should add a little bit of wording above each mapped function to explain what they do

Sort of, my only comment being that `nvim-lspconfig` is often people's first introduction to the LSP client _at all_, so flashing some of the cool features as well as making it apparent where to learn more might help shrink that gap in understanding.

windwp:
I am not sure but we should have a document about using lspconfig with nvim-compe. 
Many people don't know how to use the auto completion and they keep asking about that. 
Better we have link from readme  https://github.com/neovim/nvim-lspconfig#keybindings-and-completion for use nvim-compe

mjlbach:
@windwp It's already in the wiki, and referenced from the new readme. I expanded the page: https://github.com/neovim/nvim-lspconfig/wiki/Autocompletion

Also, we have a full config that uses compe here: https://github.com/neovim/nvim-lspconfig/wiki/Complete-init.lua-example

PR #914, opened 2021-05-17
Update documentation round 3

mjlbach:
I merged the majority of the changes, which I think were already positive. Copying the feedback here.

levouh: You think there should be a table of contents, but possibly github's auto TOC mitigates this

levouh: You think we should add a little bit of wording above each mapped function to explain what they do (already did some of this)

David-Else: You think "Automatically launching language servers" should be renamed to explicitly call out root dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,9 @@ Neovim supports the Language Server Protocol (LSP), which means it acts as a cli
 - format 
 - refactor
 
-Neovim provides an interface for all of these features, and the language server client is designed to be highly extensible to allow plugins to integrate language server features which are not yet present in Neovim core such as [**auto**-completion](https://github.com/neovim/nvim-lspconfig/wiki/Autocompletion) (as opposed to manual completion with omnifunc) and [snippet integration](https://github.com/neovim/nvim-lspconfig/wiki/Snippets).
+See `:h lsp-method` for more.
+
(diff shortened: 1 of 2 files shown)


David-Else:
> David-Else: You think "Automatically launching language servers" should be renamed to explicitly call out root dir

You gave a reason to not do that, so it is up to you. I was reading it from a technical writing point of view rather than with the knowledge you have about issues.

levouh:
Pasting from my smooth-brain reply in the old PR.

---

>You think there should be a table of contents, but possibly github's auto TOC mitigates this

@mjlbach the built-i one from GitHub is nice, but I'm not sure how many people are aware it exists. You can add it manually, or do something like what @David-Else noted with `pandoc` or similar. I can't imagine it makes a _huge_ difference, but like I said it just helps cater to interest level more easily.

>You think we should add a little bit of wording above each mapped function to explain what they do

Sort of, my only comment being that `nvim-lspconfig` is often people's first introduction to the LSP client _at all_, so noting some of the cool features (in plain English) as well as making it apparent where to learn more (like you said, `:help vim.lsp.*`) might help shrink that gap in understanding.

levouh:
There are also a few things like the random `Window` tidbits at the bottom of the _README_ and whatnot that seem a bit out of place.

mjlbach:
> There are also a few things like the random Window tidbits at the bottom of the README and whatnot that seem a bit out of place.

Not sure what to do about this, I get a lot of questions about windows support though. I've been resisting just doing a tree-wide if windows then cmd = cmd .. ".cmd" end type thing

PR #918, opened 2021-05-18
Add doc/tags to gitignore

parmort:
I keep my plugins in submodules in my dotfiles repo, which I load through the builtin plugin manager. When I generate helptags, a file is put in `doc/tags`, which should be ignored by git. This ends up being annoying because it is shown in the git status in my main dotfiles repo as "uncommitted changes".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .luacheckcache
 neovim
+doc/tags


jcmuller:
You can always keep a list of globally ignored files. `tags` is definitely one that should be there.

```sh
$ git config --global core.excludesfile '~/.gitignore'
```

Then add all the files that are typically around in your own development setup but don't necessarily need to exist in every single project's local gitignores.

parmort:
Oh ok, thanks. Completely forgot that file existed. I probably would have gone searching for that, had the other doc/tags not been ignored project-locally, which is why I thought this project should as well.

PR #922, opened 2021-05-21
Require config if dofile cannot find file

tiagovla:
When a plugin (or user's config) tries to extend **lspconfig** with a config file i.e. `lua/lspconfig/my_server.lua`, dofile won't be able to find it because it uses the absolute path.

Now:
```lua
require("lspconfig/my_server") -- it's now required because of dofile
require("lspconfig").my_server.setup {}
```

With this PR, it should work the way it used to:
```lua
require("lspconfig").my_server.setup {}
```

Would it be possible to add the require back in case dofile does not find the file? It shouldn't have any impact in performance because it would only use require in case dofile does not find it.


mjlbach:
I'd prefer not to, we've never recommended (or officially supported) the method you're asking for, and it's not how we support adding servers as specified in ADVANCED_README.md

PR #924, opened 2021-05-21
jdtls: add package_json path

tamago324:
I think we can use the [package.json](https://github.com/redhat-developer/vscode-java/blob/master/package.json) in [vscode-java](https://github.com/redhat-developer/vscode-java) for the `package.json` of `jdtls`, what do you think?

Thank you!


tamago324:
Thanks!

PR #934, opened 2021-05-23
add eclipse/lemminx XML Language Server

tormodatt:


lua/lspconfig/lemminx.lua
```diff
@@ -0,0 +1,51 @@
...
+    ...
+}
+```
+]];
+    default_config = {
+      cmd = {
+        "java", "-jar", bin_name
+      },
+      filetypes = {"xml"};
+      root_dir = "vim's starting directory";
```
mjlbach, line 46: All of the stuff you added under default config should already be copied, I would just leave it out.

lua/lspconfig/lemminx.lua
```diff
@@ -0,0 +1,51 @@
...
+local bin_name = "org.eclipse.lemminx-uber.jar"
+
+configs[name] = {
+  default_config = {
+    cmd = {
+      "java", "-jar", bin_name
+    },
+    filetypes = {"xml"};
+    root_dir = function(filename)
+      return util.path.dirname(filename)
```
mjlbach, line 14: Are there no other preferred project markers?

lua/lspconfig/lemminx.lua
```diff
@@ -0,0 +1,51 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local name = "lemminx"
+local bin_name = "org.eclipse.lemminx-uber.jar"
+
+configs[name] = {
+  default_config = {
+    cmd = {
+      "java", "-jar", bin_name
```
mjlbach, line 10: I would just copy the bin_name in here

lua/lspconfig/lemminx.lua
tormodatt: Personally, I would like this to work for scratch-files as well, as I often create a file with an xml-response and format it using this. I tend to put my scratch-files in the /tmp-directory, so that they get deleted after a while, and are then not a part of a project structure.

This LSP does not support `textDocument/definition` yet (https://github.com/eclipse/lemminx). Correct me if I'm wrong, but I guess that is the main point of setting the project root higher up than the file location?

Or maybe it possible to use the git-root, and if that does not exists use the file location?

mjlbach: You can just add the current file as a fallback, see pyright. I don't use this language server, so I'm not sure if there is a situation where having the project root set correctly matters. It is used for other features/internally by the language server.

tormodatt: In maven projects with eg. multiple modules, there can be a hierarchy of pom.xml files that could reference the ones above, so I put git-root as default, and current file as fallback.

lua/lspconfig/lemminx.lua
tormodatt: done

laher:
It's mostly working for me, yay!

These things all seem to work:
 * completions
 * rename
 * error messages - bad xml
 * error messages - it's telling me when I have schema problems.
 * It tells me when some features are not implemented (such as textDocument/declaration).
 * diagnostics (previous/next)

But, a couple of things don't seem to work and don't seem to be declared as "not implemented":
 * the hover() function doesn't seem to do anything. 
 * Nor does 'go to definition'
In both cases I'd hope to see what's defined in the schema for this element / attribute...

lua/lspconfig/lemminx.lua
laher: I'm confused by that amendment. For me, if I put the jar on my PATH, the result is this:

```
❱ org.eclipse.lemminx-uber.jar    
zsh: exec format error: org.eclipse.lemminx-uber.jar
```

Could it ever work? Maybe with a specifically prepared (hashbanged) jarfile?

lua/lspconfig/lemminx.lua
datho7561: ^ `textDocument/definition` is set up to navigate between opening and closing tags in LemMinX (see https://github.com/redhat-developer/vscode-xml/blob/master/docs/Features.md#jump-to-startend-tag). LemMinX also supports type definition `textDocument/typeDefinition`, which should link from an XML element to the definition of the element in the schema file.

laher: ah. that's working for me. It's different to how intellij interprets "go to definition" but that's all good, I can see why lemminx would do it this way. 

... There seems to be some quirks with resolving xsd catalogs, but that's going to be an upstream lemminx thing

Thanks. Great stuff.

lua/lspconfig/lemminx.lua
datho7561: If you leave the `cmd` as `bin_name`, its easy enough to instead use the binary version of LemMinX that's available here: https://download.jboss.org/jbosstools/vscode/stable/lemminx-binary

laher: Ah, that's great. Maybe update the description field to mention that? Currently it gives instructions for downloading the non-executable jar file. Doesn't need to be full instructions IMO

lua/lspconfig/lemminx.lua
```diff
@@ -0,0 +1,57 @@
...
+
+local name = "lemminx"
+local bin_name = "org.eclipse.lemminx-uber.jar"
+
+configs[name] = {
+  default_config = {
+    cmd = {
+      bin_name
+    },
+    filetypes = {"xml"};
```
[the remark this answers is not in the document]
tormodatt: true! I added those

tormodatt:
> the hover() function doesn't seem to do anything.

it works for me for e.g. maven pom's @laher 

Skjermbilde 2021-06-01 kl  23 27 56

lua/lspconfig/lemminx.lua
tormodatt: yeah, jar's are required to be run with the java-binary as far as I know. so `java -jar org.eclipse.lemminx-uber.jar` in that case.

I changed to the binary-method in stead as suggested. I didn't know that existed

lua/lspconfig/lemminx.lua
tormodatt: good to know! added it to the doc. should be added to the lemminx readme as well I suppose though

datho7561: > good to know! added it to the doc. should be added to the lemminx readme as well I suppose though

see redhat-developer/vscode-xml#491

laher:
> > the hover() function doesn't seem to do anything.
> 
> it works for me for e.g. maven pom's @laher
> 
> Skjermbilde 2021-06-01 kl 23 27 56

This is working for me too, thanks. Sorry for confusion there. 

:clap:

laher (approved):
Approving FWIW (I'm not a member of the project, just a grateful user). 
Thanks for making this.

lua/lspconfig/lemminx.lua
```diff
@@ -0,0 +1,43 @@
...
+    filetypes = {"xml", "xsd", "svg"};
+    root_dir = function(filename)
+      return util.root_pattern(".git")(filename) or util.path.dirname(filename)
+    end;
+  };
+  docs = {
+    description = [[
+https://github.com/eclipse/lemminx
+
+Features:
```
mjlbach, line 21: I would match the documentation style of other plugins, see elixirls for the standard "We don't modify your path blah blah" bit I put in. I also wouldn't recommend listing the capabilities as these may become out of date.

mjlbach:
@Mergifyio rebase

mjlbach:
This is failing style and commit lint, please fix and rebase.

tormodatt:
should be fine now @mjlbach

lua/lspconfig/lemminx.lua
tormodatt: fixed

mjlbach:
@Mergifyio rebase

PR #937, opened 2021-05-25
Fix eof to upper case, (E5107 error)

mrquaketotheworld:


lithammer:
Thanks!

mrquaketotheworld:
@lithammer You are welcome :)

Issue #943, opened 2021-05-27
Temp buffer (floatwindow) with go type cause high CPU usage of gopls


ray-x:
The issue found is when I use Vista + gopls. Vista can preview tag/function in floatwindows. 
When I open preview, Vista will show a pretty floating window with highlight with go syntax (by setting file type to go)

If my gopls rootpath is 
```lua
{
root_dir = function(fname)
      return util.root_pattern("go.mod", ".git")(fname) or util.path.dirname(fname)
    end
}
```
`util.path.dirname(fname)` part give me a problem.
It returns an empty string (not nil)
That empty string will be pass to rootpath and as there is no separator. In the end, it becomes "/", which is a root dir in mac.

That will ask gopls to search all files in "/" and slow down gopls.

I think a fix should address this. I will submit a PR.

mjlbach:
Shouldn’t this be the fallback anyways?

________________________________
From: rayx ***@***.***>
Sent: Thursday, May 27, 2021 3:15:23 PM
To: neovim/nvim-lspconfig ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [neovim/nvim-lspconfig] Temp buffer (floatwindow) with go type cause high CPU usage of gopls (#943)


The issue found is when I use Vista + gopls. Vista can preview tag/function in floatwindows.
When I open preview, Vista will show a pretty floating window with highlight with go syntax (by setting file type to go)

If my gopls rootpath is

{
root_dir = function(fname)
      return util.root_pattern("go.mod", ".git")(fname) or util.path.dirname(fname)
    end
}

util.path.dirname(fname) part give me a problem.
It returns an empty string (not nil)
That empty string will be pass to rootpath and as there is no separator. In the end, it becomes "/", which is a root dir in mac.

That will ask gopls to search all files in "/" and slow down gopls.

I think a fix should address this. I will submit a PR.

ray-x:
I think "/" is the fallback...

ray-x:
go.mod and .git is not essential to a go project folder. 
e.g a folder only contain  hello_world.go, I can still use `go run hello_world.go` to run the code

It is true that by default, the gopls setting works perfectly.

mjlbach:
Yeah but in that case the current fallback *should* work, returning early isn't the correct fix...

ray-x:
for function dirname(filename)
if filename is nil or empty string. both should return directly.
[...]

PR #944, opened 2021-05-27
fix: prevent dirname return '/' if path is empty string

ray-x:
Related to issue #943


mjlbach:
Are you creating a temp go buffer outside a go project?

ray-x:
The plugin of Vista will create a go buffer. I also saw similar behaviour in other plugins with floating window preview features.

PR #945, opened 2021-05-28
Update documentation: correct rome link to latest website

stevelacy:


mjlbach:
Ah whoops, I shouldn't have merged that. Please see the pr template, you shouldn't make changes to Config.me as it's auto generated, please change in the rome lua file itself

stevelacy:
@mjlbach Sorry about that, will note for the future. Editing a file directly on gh does not show the template.

PR #956, opened 2021-06-03
Texlab: update options for 3.1.0

clason:
Unfortunately, one of the new options is called `texlab.latexindent.local`, which Lua doesn't like. Let's see what CI thinks...

### DIFF
--- a/lua/lspconfig/texlab.lua
+++ b/lua/lspconfig/texlab.lua
@@ -61,7 +61,8 @@ configs.texlab = {
         build = {
           executable = "latexmk";
           args = {"-pdf", "-interaction=nonstopmode", "-synctex=1", "%f"};
-          isContinuous = false;
+          onSave = false;
+          forwardSearchAfter = false;
         };
         auxDirectory = '.';
         forwardSearch = {
@@ -73,6 +74,11 @@ configs.texlab = {
           onEdit = false;
         };
         diagnosticsDelay = 300;
+        latexFormatter = 'latexindent';
+        latexindent = {
+          ["local"] = nil; -- local is a reserved keyword
+          modifyLineBreaks = false;
+        };
         bibtexFormatter = 'texlab';
         formatterLineLength = 80;
       };


mjlbach:
You can always do { ["local"] = nil } (or do you want vim.NIL here?)

clason:
It's a "string, default null" -- so I followed the similar root directory option.

If there's a better value here, I'd be happy to change it!

lua/lspconfig/texlab.lua
```diff
@@ -73,6 +73,11 @@ configs.texlab = {
           onEdit = false;
         };
         diagnosticsDelay = 300;
+        latexFormatter = 'latexindent';
+        latexindent = {
+          local = nil; -- sic
```
lithammer, line 78: `local` is a keyword.

clason:
@mjlbach should be good now -- let me know if you want the `nil` replaced by `''` or `{}` here.

clason:
@mjlbach bump?

mjlbach:
Can you squash your commits?

mjlbach:
Also do we need to pass local if it's nill anyways?

clason:
> Also do we need to pass local if it's nill anyways?

We don't, but I always understood this to double as the documentation for the possible options to set. (A non-default option would be a path to a local `file.yaml`.)

mjlbach:
I've generally come of the opinion to not change the defaults unless it's widely expected, as using this as documentation tends to confuse users who think these settings are mandatory to use the language server (I see a lot of people copying/pasting defaults into their own configs). I usually just try to make the link to settings obvious somehow, or we autogenerate it from the package.json.

clason:
Hmm. That would remove maintenance burden if we don't keep the list of options up to date. But now that the documentation is completely autogenerated, where would you get -- and prominently display -- the link to the docs? What if there's no `package.json`?

clason:
@mjlbach squashed and used conventional commit

mjlbach:
I think we should go through at some point and remove the default config options, merging now

PR #964, opened 2021-06-10
[RDY] show autostart status in LspInfo

teto:
autostart can explain why a server is not active so I think it deserves a place in `:LspInfo`. Seems like this PR wont work though because the `autostart` setting is not forwarded to the generated client, there is an `autostart()` function in its stead. 

Maybe we can rename the `autostart()` function ?


teto:
I ended up using saving the autostart client value in _autostart

PR #966, opened 2021-06-11
Import tailwindcss from lspinstall

polyzen:


mjlbach:
This needs some documentation, see other language servers for inspiration.

polyzen:
@mjlbach, should be good to go!

lua/lspconfig/tailwindcss.lua
```diff
@@ -0,0 +1,87 @@
...
+      util.find_node_modules_ancestor(fname) or
+      util.find_git_ancestor(fname)
+    end,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/tailwindlabs/tailwindcss-intellisense/master/packages/tailwindcss-language-server/package.json',
+    description = [[
+https://github.com/tailwindlabs/tailwindcss-intellisense
+
+Tailwind CSS Language Server
```
mjlbach, line 81: Can you add the npm command here? see cssls, htmls, pyright, etc.

polyzen: It's not yet available on npm, afaik.

polyzen: Arch Linux users can install the server using https://aur.archlinux.org/packages/tailwindcss-language-server/

mjlbach: It's pretty easy to install manually, if you need to see how for reference nvim-lspinstall https://github.com/kabouzeid/nvim-lspinstall/blob/878645fffab9b14eec104ec95a64c4165b6138e8/lua/lspinstall/servers/tailwindcss.lua#L4-L12 does basically the same thing I did to test this

polyzen: Comment with intent to ship to npm:
https://github.com/tailwindlabs/tailwindcss-intellisense/issues/301#issuecomment-850542512

Work towards that:
https://github.com/tailwindlabs/tailwindcss-intellisense/commit/bdb398bf80a5557b329e5b1c7f2019c58e49cd00

package.json for standalone server:
https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/tailwindcss-language-server/package.json

mjlbach: That's great news! I'd still like to include the basic instructions to install this when we merge it, as that's usually the first question/issue people have when we add a server without it.

polyzen: Is it preferred to suggest building it themselves or extracting the server from the pre-built VS Code extension?

mjlbach: I'd probably just do like nvim-lspinstall for now and we'll await the npm package/update when it happens

polyzen: Installation instructions added 👍

Issue #965, opened 2021-06-10
Sass (*.scss) files get errors as if CSS files


benfrain:
- `nvim --version`: NVIM v0.5.0-dev+1383-g4ce61742c: 
- nvim-lsp version(commit hash): HEAD-4ce6174_2
- What language server (If the problem is related to a specific language server): 

cssls (npm install -g vscode-css-languageserver-bin)

- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): 

no, works fine in VSCode and Sublime Text

- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): N/A
- Is the problem isolated to a particular language server:
- Operating system/version: macOS  11.4

### How to reproduce the problem from neovim startup
Install `npm install -g vscode-css-languageserver-bin`, use standard `require'lspconfig'.cssls.setup{}` in init file, open a sass (scss file) and try adding something like `@use "./required/_variables" as v;` 

### Actual behaviour
You will see a Warning `Unknown at rule @use` for that line.

### Expected behaviour
You should not see a warning

### Minimal init.vim or init.lua and code sample

Used minimal_lua from link. Test file content:

```scss
@use "./required/_variables" as v;

.post-Bio_Title {
    display: flex;
    padding: 1em 0 0;
    margin: 0;
    border-top: 5px double v.$color-blockquote;
}

```

### Health check
<details>
<summary>Checkhealth result</summary>

health#lspconfig#check
========================================================================
## Checking language server protocol configuration
  - INFO: cssls: configuration checked.

</details>

### LSP log

<details>
<summary>Log file</summary>
[lsp.log](https://github.com/neovim/nvim-lspconfig/files/6634749/lsp.log)
</details>
<img width="1325" alt="image" src="https://user-images.githubusercontent.com/1318466/121603143-3f152480-ca40-11eb-8ca5-79cfad9888a0.png">

neovim-discourse:
This issue has been mentioned on **Neovim Discourse**. There might be relevant details there:

https://neovim.discourse.group/t/lsp-cssls-settings/568/4

rudotriton:
That lang server hasn't been updated for years. Try the css lang server from [vscode-langservers-extracted](https://www.npmjs.com/package/vscode-langservers-extracted) instead and also remember to update the `cmd`.

mjlbach:
[...]

PR #968, opened 2021-06-12
Update to provide updated CSS instructions

benfrain:
Updated language server references based on https://github.com/neovim/nvim-lspconfig/issues/965#issuecomment-860008798
Also added basic rules override example.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -413,14 +413,14 @@ require'lspconfig'.codeqlls.setup{}
 
 ## cssls
 
-https://github.com/vscode-langservers/vscode-css-languageserver-bin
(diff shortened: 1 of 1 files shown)


mjlbach:
Thanks for the PR, but CONFIG.md is auto-generated, so you'll want to modify the (lua) source file directly.

benfrain:
> Thanks for the PR, but CONFIG.md is auto-generated, so you'll want to modify the (lua) source file directly.

Doh! Thanks will do!

benfrain:
Hi @mjlbach I opened a new pull request with the edits in the Lua file. Hope that is more useful.

PR #971, opened 2021-06-12
Update to provide updated CSS instructions

benfrain:
Updated language server references based on #965
Also added basic rules override example.

Apologies if this is still not correct, unfamiliar with Lua!

### DIFF
--- a/lua/lspconfig/cssls.lua
+++ b/lua/lspconfig/cssls.lua
@@ -31,7 +31,7 @@ https://github.com/hrsh7th/vscode-langservers-extracted
 npm i -g vscode-langservers-extracted
 ```
     
-Neovim does not currently include built-in snippets. vscode-css-language-server only provides completions when snippet support is enabled. To enable completion, install a snippet plugin and add the following override to your language client capabilities during setup.
+Neovim does not currently include built-in snippets. `vscode-css-language-server` only provides completions when snippet support is enabled. To enable completion, install a snippet plugin and add the following override to your language client capabilities during setup.
 
 ```lua
 --Enable (broadcasting) snippet capability for completion
@@ -43,27 +43,6 @@ require'lspconfig'.cssls.setup {
 }
 ```
     
-```lua
-require'lspconfig'.cssls.setup{}
-
-  Commands:
-  
-  Default Values:
-    cmd = { "vscode-css-language-server", "--stdio" }
-    filetypes = { "css", "scss", "less" }
-    root_dir = root_pattern("package.json")
-    settings = {
-      css = {
-        validate = true
-      },
-      less = {
-        validate = true
-      },
-      scss = {
-        validate = true
-      }
-    }
-```
 ]];
     default_config = {
       root_dir = [[root_pattern("package.json")]];
--- a/lua/lspconfig/cssls.lua
+++ b/lua/lspconfig/cssls.lua
@@ -21,7 +21,77 @@ configs[server_name] = {
   };
   docs = {
     description = [[
-https://github.com/vscode-langservers/vscode-css-languageserver-bin
+
+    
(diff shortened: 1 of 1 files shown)


lua/lspconfig/cssls.lua
```diff
@@ -21,7 +21,77 @@ configs[server_name] = {
...
+Neovim does not currently include built-in snippets. vscode-css-language-server only provides completions when snippet support is enabled. To enable completion, install a snippet plugin and add the following override to your language client capabilities during setup.
+
+```lua
+--Enable (broadcasting) snippet capability for completion
+local capabilities = vim.lsp.protocol.make_client_capabilities()
+capabilities.textDocument.completion.completionItem.snippetSupport = true
+
+require'lspconfig'.cssls.setup {
+  capabilities = capabilities,
+}
```
mjlbach, line 43: Pretty much everything after this point is unecessary IMO

lua/lspconfig/cssls.lua
```diff
@@ -21,7 +21,77 @@ configs[server_name] = {
   };
   docs = {
     description = [[
-https://github.com/vscode-langservers/vscode-css-languageserver-bin
+
+    
+https://www.npmjs.com/package/vscode-langservers-extracted
```
mjlbach, line 26: I would change this to the github URL of the project

mjlbach:
Do you not also need to update `cmd` with the new binary name? I don't use this language server.

lua/lspconfig/cssls.lua
benfrain: As a new user, I added the example as I found the original section  confusing/insufficient (with the nested defaults). I’m a fan of a cut and paste example but obviously go with what makes sense from your POV.

mjlbach: I think the cut/paste example is overly specific to your usecase.  Also one of the blocks basically duplicates what docgen does

benfrain: Hi, commit https://github.com/neovim/nvim-lspconfig/pull/971/commits/61ba064065b301657cc3ad31e26fd89795044d41 should have it as you desire.

lua/lspconfig/cssls.lua
```diff
@@ -21,15 +21,17 @@ configs[server_name] = {
...
+
 ```sh
-npm install -g vscode-css-languageserver-bin
+npm i -g vscode-langservers-extracted
 ```
-
-Neovim does not currently include built-in snippets. `vscode-css-languageserver` only provides completions when snippet support is enabled.
-To enable completion, install a snippet plugin and add the following override to your language client capabilities during setup.
+    
+Neovim does not currently include built-in snippets. vscode-css-language-server only provides completions when snippet support is enabled. To enable completion, install a snippet plugin and add the following override to your language client capabilities during setup.
```
mjlbach, line 34: Out of curiosity, why did you remove the backticks around the server name?

lua/lspconfig/cssls.lua
mjlbach: Lines 45 to 66 still need to be removed

lua/lspconfig/cssls.lua
benfrain: Sheer stupidity!

lua/lspconfig/cssls.lua
```diff
@@ -40,6 +42,7 @@ require'lspconfig'.cssls.setup {
   capabilities = capabilities,
 }
 ```
+
```
mjlbach, line 45: unnecessary space :)

mjlbach:
@benfrain if you want, you can also do html and json language servers since these are also available from the same repo :) no pressure though. Thanks for the PR!

benfrain:
Sure, should have a better idea what I'm doing now!

PR #981, opened 2021-06-14
update cssls binary

NFrid:


mjlbach:
Please read the pinned issue, the binary name was changed intentionally.

PR #983, opened 2021-06-15
improve(tailwindcss): missing validation settings, provide tab size for hover

kabouzeid:
Sync from: https://github.com/kabouzeid/nvim-lspinstall/blob/main/lua/lspinstall/servers/tailwindcss.lua

### DIFF
--- a/lua/lspconfig/tailwindcss.lua
+++ b/lua/lspconfig/tailwindcss.lua
@@ -65,6 +65,28 @@ configs[server_name] = {
         eruby = 'erb',
       },
     },
+    settings = {
+      tailwindCSS = {
+        validate = true,
+        lint = {
+          cssConflict = "warning",
+          invalidApply = "error",
+          invalidScreen = "error",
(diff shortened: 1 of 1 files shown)


mjlbach:
Out of curiosity, why all the settings? Does tailwind not provide defaults?

kabouzeid:
The defaults are defined in the package.json of the VSCode extension (https://github.com/tailwindlabs/tailwindcss-intellisense/blob/8aa60b122d189f30e418b1bd666fb9f1d759d2d6/packages/vscode-tailwindcss/package.json#L87-L169). The server itself doesn't provide fallback values for all settings, so for now we have to do what the VSCode extension does and send the settings to the server.

They are currently preparing an npm release for the language server, and I'll open an issue after that. We probably won't need this anymore in a few weeks/months.

mjlbach:
Thanks for clarifying, that sounds good to me. Let's go ahead and merge and then revisit once the server is published on npm/we can file issues upstream.

PR #1000, opened 2021-06-27
Increase the default memory and timeout limits in lean3ls

Julian:
Matches our VSCode's default found here:

https://github.com/leanprover/vscode-lean/blob/ca764c60acbfaeaca55a681bc0be65f84b0cae44/package.json#L32-L41


Julian:
Pull request #1000, what do I win?

PR #1009, opened 2021-06-29
terraform-ls is for terraform only

dimbleby:
The HCL format is used by lots of hashicorp tools - and some others - not only terraform: whereas the terraform language server is only meant for terraform and only useful for terraform.

This MR directly contradicts - and undoes - #757, therefore pinging @WhyNotHugo and @lithammer. I see there was some discussion in that issue: I think you simply came to the wrong conclusion.

I am currently considering switching hashivim/vim-terraform to use filetype hcl.terraform for terraform files, it was while investigating the consequences of that change that I spotted this.

hashivim/vim-terraform#168

If we do go that way, I'll likely follow up with an MR to use that filetype as a trigger for tflint and terraform-ls.

(NB this repository doesn't currently try to run tflint for non-terraform HCL files, which I think is right).

### DIFF
--- a/lua/lspconfig/terraformls.lua
+++ b/lua/lspconfig/terraformls.lua
@@ -4,7 +4,7 @@ local util = require 'lspconfig/util'
 configs.terraformls = {
   default_config = {
     cmd = {"terraform-ls", "serve"};
-    filetypes = {"terraform", "hcl"};
+    filetypes = {"terraform"};
     root_dir = util.root_pattern(".terraform", ".git");
   };
   docs = {


mjlbach:
Thanks for the PR! I don't use terraform, but I see you're one of the main contributors to vim-terraform so I'm inclined to defer to you.

WhyNotHugo:
This breaks `terraform-ls` for anyone _not_ using vim-terraform (as is my case).

Terraform files are hcl files, and without `vim-terraform`, they're correctly set to `ft=hcl`.

> The syntax of Terraform configurations is called HashiCorp Configuration Language (HCL).

https://www.terraform.io/docs/configuration-0-11/syntax.html

WhyNotHugo:
OH, I use [vim-hcl](https://github.com/jvirtanen/vim-hcl) for highlighting and stuff.

WhyNotHugo:
Oh, looks like vim-hcl [has updated accordingly](https://github.com/jvirtanen/vim-hcl) too, so never mind.

dimbleby:
Oh, interesting!  Thanks for that pointer, it's great that the two plugins now agree - and makes a pretty strong case _against_ using `hcl.terraform` (or `terraform.hcl`, I'm not sure which!) in hashivim/vim.

PR #1010, opened 2021-06-30
Set java-language-server default cmd

drybalka:
This sets the default `cmd` to `java-language-server`, which will make it work out of the box on machines with an installed java-language-server package (for example from AUR). It is also more consistent with other servers and avoids the cryptic error "E5108: Error executing lua Vim:E928: String required", when trying to get `:LspInfo` with no `cmd` provided, as it is now.

### DIFF
--- a/lua/lspconfig/java_language_server.lua
+++ b/lua/lspconfig/java_language_server.lua
@@ -5,7 +5,7 @@ local name = 'java_language_server'
 
 configs[name] = {
   default_config = {
-    cmd = {};
+    cmd = { "java-language-server" };
     filetypes = {'java'};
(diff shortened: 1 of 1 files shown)


mjlbach:
I think this is only the case for the AUR https://repology.org/project/java-language-server/versions, in which case I don't think we should merge this.

I  would updated the documentation instead, matching elixir for example (which also does not have cmd set) 

https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#elixirls

mfussenegger:
Also keep in mind that eclipse uses a data folder to store project specific data,  and to avoid issues you want to use a different data folder per project.  See https://github.com/mfussenegger/nvim-jdtls#data-directory-configuration

I don't think the java-language-server launcher in the AUR accounts for that

PR #1017, opened 2021-07-02
Fixed a small gotcha where the server_path is set to nonsense if you print stuff in your startup.jl

DoktorMike:
When I start julia I print out bayes formula in `startup.jl` which looks like this

```
          p(x|θ)p(θ)
p(θ|x) = ------------
          ∫p(x,θ)dθ
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.1 (2021-04-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |
```

Which makes the executed julia command for finding the location of LanguageServer get part of bayes formula as a path. This fix makes sure that startup.jl is not executed.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -3704,7 +3704,7 @@ require'lspconfig'.ocamllsp.setup{}
   
   Default Values:
     cmd = { "ocamllsp" }
-    filetypes = { "reason", "menhir", "ocamlinterface", "ocaml", "ocamllex" }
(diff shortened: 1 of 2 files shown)


mjlbach:
Thanks! If you remove and reopen without the docgen step I'll merge.

PR #1020, opened 2021-07-02
Fix infinite loop in root_pattern running on Windows mapped network drives.

nochiel:
ref. https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-psdrive?view=powershell-7.1 e.g.

# Steps

0. Use Windows powershell to map a network drive e.g:
```powershell
New-PSDrive -persist -scope global -Root \\$env:COMPUTER_NAME\Users -Name T -PSProvider FileSystem 
```
1. Edit a file in that drive for whose type you have an lspconfig setup. e.g. `nvim T:\hello.go`

2. nvim-lspconfig goes into an infinite loop.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -157,7 +157,16 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
+    _path = path
     path = uv.fs_realpath(path)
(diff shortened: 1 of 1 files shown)


mjlbach:
This shouldn't be an issue anymore with the changes I made to root directory detection, I believe (I also added a guard against infinite recursion). Feel free to open if not.

PR #1022, opened 2021-07-03
fix: validate cmd is of type table

### LINKED ISSUE #1021: If `cmd` for LSP server is set to a string, `lspconfig` commands, including `LspInfo` break with hard to understand errors
- `nvim --version`: 0.5.0
- nvim-lspconfig version(commit hash): 8435587084210c0779a27ca59d98ce4b9c6af0f6
- What language server (If the problem is related to a specific language server): rust_analyzer
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): not tried
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): not tried
- Is the problem isolated to a particular language server: :shrug:
- Operating system/version: NixOS

Specify

```lua
require('lspconfig').rust_analyzer.setup {
  cmd = '/wherever/bin/rust-analyzer'
}
```

instead of 

```lua
require('lspconfig').rust_analyzer.setup {
  cmd = { '/wherever/bin/rust-analyzer' }
}
```

### Actual behaviour

```
:LspInfo
E5108: Error executing lua ...are/vim-plugins/nvim-lspconfig/lua/lspconfig/lspinfo.lua:32: bad argument #1 to 'pairs' (table expected, got string)
:LspStart
E5108: Error executing lua ...06-30/share/vim-plugins/nvim-lspconfig/lua/lspconfig.lua:29: attempt to index a nil value
```

and similar.

### Expected behaviour

The failure should be much better behaved in this instance I feel. Ideally I think you would receive an error at the time invalid `setup` method is invoked. Or the invalid servers should be ignored and somehow reflected as invalidly configured otherwise.

mjlbach:
Closes #1021

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -32,6 +32,7 @@ function configs.__newindex(t, config_name, config_def)
 
   function M.setup(config)
     validate {
+      cmd = {config.cmd, 't', default_config.cmd ~= nil};
       root_dir = {config.root_dir, 'f', default_config.root_dir ~= nil};
       filetypes = {config.filetype, 't', true};
       on_new_config = {config.on_new_config, 'f', true};


nagisa:
Thanks for a quick fix!

PR #1026, opened 2021-07-03
ci: lint and format with stylua

mjlbach:
Since this repo is community contributed, many of the stylistic choices differ between users (making things look a bit messy). Additionally, some lua syntax (calling functions with table, dropping parens) can be confusing for new contributors. I feel weird nitpicking style (and it's time consuming) the diff LGTM, so I think this will reduce our maintenance burden going forward.


mjlbach:
my bot, I feel so betrayed 😢

Issue #1031, opened 2021-07-04
#1022 breaks `powershell_es`


uyha:
- `nvim --version`: 0.5
- nvim-lspconfig version(commit hash): 5e41f4a7b84a26d57d7eda5c91396bccb1752f57
- What language server (If the problem is related to a specific language server): `powershell_es`
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): not tried
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): not tried
- Is the problem isolated to a particular language server: `powershell_es`
- Operating system/version: Windows 10 Build 19043

After #1022, the check for `cmd` being not `nil` is incorrect for `powershell_es` since it sets `cmd` based on the `bundle_path` variable.

### How to reproduce the problem from neovim startup
Config `powershell_es` with `bundle_path` as indicated in `CONFIG.md`.
### Actual behaviour
Issue error: `E5113: Error while calling lua chunk: ...ck\packer\start\nvim-lspconfig\lua\lspconfig\configs.lua:34: cmd: expected table, got nil`
### Expected behaviour
`powershell_es` is loaded as normally without error

mjlbach:
Should be fixed by #1032. Sorry about that, we should probably change how powershell works to modify a cmd string or something, I would like to implement stricter validation for our configs.

PR #1032, opened 2021-07-04
fix: be less strict about cmd validation

mjlbach:
Some language servers lazily build cmd based on a parameter passed to setup, like powershell_es. We should change this eventually, but for now be less strict.


mjlbach:
closes  #1031

PR #1035, opened 2021-07-04
fix: nil check for config

ChristianChiarulli:


mjlbach:
I would just do an early return, also not sure why this is happening which should probably get addressed (now it will just fail silently)

PR #1038, opened 2021-07-05
Update ADVANCED_README.md

kabouzeid:
Don't see why this would be necessary?


mjlbach:
powershell_es doesn't set cmd, some people set it in on_new_config which has elicited questions previously

kabouzeid:
I don't understand 🤔 Shouldn't this be documented in the powershell config then? It's not necessary to set `cmd` in `on_new_config`, it just needs to be set *somewhere*.

mjlbach:
superseded by #1349

PR #1040, opened 2021-07-05
Revert "Fix starting the LSP server when creating a new file"

mjlbach:
Reverts neovim/nvim-lspconfig#955

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -161,7 +161,7 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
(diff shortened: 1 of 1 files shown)


mivok:
Oh no. Did #955 end up breaking things?

mjlbach:
Yes, mostly around fugitive buffers as the commenter noted. I didn't like matching on the URI name (and I was a bit busy at the time) so I reverted it as a temporary fix.

Do you want to check out:
https://github.com/neovim/nvim-lspconfig/pull/1041
?

I just added a guard and additionally check that the first parent is a real directory (I think this is fine, it won't let people create an arbitrarily nested file but you wouldn't be able to save that without creating the intermediate directories anyways).

PR #1041, opened 2021-07-05
feat: try to resolve root for new files

mjlbach:
There were issues with fugitive and other non filesystem buffer paths before. Still needs testing for edge cases...

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -161,15 +161,16 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
-    path = uv.fs_realpath(path)
     local function it(s, v)
-      if not v then
+      if v and not is_fs_root(v) then
+        v = dirname(v)
+      else
         return
       end
-      if is_fs_root(v) then
+      if not uv.fs_realpath(v) then
         return
       end
-      return dirname(v), path
+      return v, path
     end
     return it, path, path
   end
@@ -263,7 +264,14 @@ function M.search_ancestors(startpath, func)
   if func(startpath) then
     return startpath
   end
+  local guard = 100
   for path in M.path.iterate_parents(startpath) do
+    -- Prevent infinite recursion if our algorithm breaks
+    guard = guard - 1
+    if guard == 0 then
+      return
+    end
+
     if func(path) then
       return path
     end


lua/lspconfig/util.lua
```diff
@@ -161,12 +161,12 @@ M.path = (function()
...
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
-    path = uv.fs_realpath(path)
+    path = uv.fs_realpath(path) or path
     local function it(s, v)
       if not v then
         return
       end
-      if is_fs_root(v) then
+      if is_fs_root(v) or not uv.fs_realpath(path) then
```
rish987, line 169: I don't think this works, at least from the perspective of #1045 -- with [Lua generic for](https://www.lua.org/pil/7.2.html), the iterator is called to get the initial value, so in this case we'd be deferring the `nil` return to here and the loop below still wouldn't enter.

lua/lspconfig/util.lua
```diff
@@ -161,12 +161,12 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
-    path = uv.fs_realpath(path)
+    path = uv.fs_realpath(path) or path
     local function it(s, v)
```
rish987, line 165:
```suggestion
    local function it(s, v)
    v = dirname(v)
```

lua/lspconfig/util.lua
rish987:
```suggestion
      if is_fs_root(v) or not uv.fs_realpath(v) then
```

lua/lspconfig/util.lua
```diff
@@ -161,12 +161,12 @@ M.path = (function()
...
       if not v then
         return
       end
-      if is_fs_root(v) then
+      if is_fs_root(v) or not uv.fs_realpath(path) then
         return
       end
       return dirname(v), path
```
rish987, line 172:
```suggestion
      return v, path
```

rish987 (commented):
_I believe_ this is what needs to be done to ensure that we still iterate over all of the parents, even if the given file doesn't exist yet.

lua/lspconfig/util.lua
```diff
@@ -161,12 +161,12 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
-    path = uv.fs_realpath(path)
+    path = uv.fs_realpath(path) or path
     local function it(s, v)
       if not v then
```
rish987, line 166: Also with these changes I'm pretty sure this block can now be removed.

mjlbach:
You can try the latest version @rish987

lua/lspconfig/util.lua
```diff
@@ -161,15 +161,16 @@ M.path = (function()
 
   -- Iterate the path until we find the rootdir.
   local function iterate_parents(path)
-    path = uv.fs_realpath(path)
     local function it(s, v)
-      if not v then
+      if not is_fs_root(v) then
```
rish987, line 165:
```suggestion
      if v and not is_fs_root(v) then
```
(to make it consistent w/ previous behavior when `path == nil`)

lua/lspconfig/util.lua
```diff
@@ -263,7 +264,15 @@ function M.search_ancestors(startpath, func)
   if func(startpath) then
     return startpath
   end
+  local guard = 100
   for path in M.path.iterate_parents(startpath) do
+    print(path)
```
rish987, line 269:
```suggestion
```
(I think you left this here by accident)

rish987:
Awesome, just tried it and this is working for us now! LGTM

mjlbach:
Great! Thanks for interactively edge-case testing :)

PR #1044, opened 2021-07-06
Fixed vim-plug installation method

paniash:
Removed the unnecessary colon in the vim-plug installation section in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ These features are not implemented in this repo, but in Neovim core. See `:help 
 
 * Install nvim-lspconfig like any other Vim plugin, e.g. with [vim-plug](https://github.com/junegunn/vim-plug):
 ```vim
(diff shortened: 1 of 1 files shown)


justinmk:
how does this "fix" anything? The colon is a signal that it's a vim command, rather than shell etc.

clason:
I think @paniash read this as "put this vimscript code in your `init.vim` (which is how `Plug` is most often used). This could be made explicit rather than just removing the colon, which would be a more useful improvement to the docs.

mjlbach:
@justinmk I had some people ask me about it awhile back who thought they had to run `:Plug 'neovim/nvim-lspconfig'`

I also checked what the nvim-treesitter had in their docs and figured we should align with them. We can add more explicit instructions, this is often new users first plan

https://github.com/nvim-treesitter/nvim-treesitter#installation

PR #1045, opened 2021-07-06
fix(leanls,lean3ls): Fix Lean 3/4 root directory search.

rish987:
Fix for the case where a new file is created within a valid project structure.


mjlbach:
Can you try https://github.com/neovim/nvim-lspconfig/pull/1041 instead?

PR #1050, opened 2021-07-09
Add formatting true flag to init_options for solargraph

CKolkey:
Per this issue: https://github.com/castwide/solargraph/issues/455

Solargraph needs to be told it supports formatting, otherwise it won't return document_formatting as a resolved capability.

### DIFF
--- a/lua/lspconfig/solargraph.lua
+++ b/lua/lspconfig/solargraph.lua
@@ -13,6 +13,7 @@ configs.solargraph = {
         diagnostics = true,
       },
     },
+    init_options = { formatting = true },
     filetypes = { "ruby" },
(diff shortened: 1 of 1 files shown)


mjlbach:
Seems like we could send this in settings instead?

CKolkey:
Thats what I tried initially, and it had no effect. :\

```lua
    settings = {
      solargraph = {
        diagnostics = true,
        formatting = true,
      },
    }
```

mjlbach:
Ah sorry, I misunderstood the last comment on that thread.

CKolkey:
To be honest, I did too at first. But this configuration works :)

PR #1051, opened 2021-07-09
fix: fix buffer without file detection

raw1z:
we also need to check buftype value in case we need the filetype to be
set to a value different from "nofile"

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -161,7 +161,8 @@ function configs.__newindex(t, config_name, config_def)
 
     function manager.try_add(bufnr)
       bufnr = bufnr or api.nvim_get_current_buf()
-      if vim.api.nvim_buf_get_option(bufnr, "filetype") == "nofile" then
+      if vim.api.nvim_buf_get_option(bufnr, "filetype") == "nofile"
+         or vim.api.nvim_buf_get_option(bufnr, "buftype") == "nofile" then
         return
       end
       local root_dir = get_root_dir(api.nvim_buf_get_name(bufnr), bufnr)


mjlbach:
Can you clarify the usecase?

raw1z:
I have in my vim config a custom command which create a buffer without file attached but with a file type. This the code of the function I use for that command:

```vim
    function! g:StartDraftBuffer(open_command, ...)
      " if the filetype isnt set then
      " use current buffer's filetype
      let draft_filetype = &ft
      if len(a:1) > 0
        let draft_filetype = a:1
      endif
    
      execute(a:open_command)
    
      nmap <buffer> <c-o> :echo "noop"<cr>
      nmap <buffer> <c-i> :echo "noop"<cr>
    
      setlocal buftype=nofile bufhidden=wipe noswapfile nobuflisted nomodified
      exe 'set ft='.draft_filetype
    endfunction

    command! -nargs=? -complete=filetype Draft call g:StartDraftBuffer("new", <q-args>)
```

As you can see, the filetype can't be nofile but the the buftype is. nvim-lspconfig will try to start the language server and fail because, it can't identify a root project since there is no file attached to the buffer. That's the reason why I'm submitting this pull request.

I think this change is not only relevant for my use case but in general, we should also check if the buffer is linked to a file instead of checking only the filetype.

mjlbach:
Is it EFM? Most of the configurations already have a filetype associated with them and should not start otherwise.

raw1z:
Excuse-me, what do you mean by EFM ?

mjlbach:
EFM language server, which server is trying to attach to the empty buffer?

raw1z:
I'm facing the issue for all the language servers I have configured (vim, ruby, python, lua, typescript, elm, elixir). It is not related to one particular server. The problem is that nvim-lspconfig can't identify a root project for the buffer because there is no file attached to it.

davidsierradz:
> Can you clarify the usecase?

 Hi, my usecase would be to use [iamcco/diagnostic-languageserver](https://github.com/iamcco/diagnostic-languageserver) with ([clj-kondo/clj-kondo](https://github.com/clj-kondo/clj-kondo/)) and [Olical/conjure](https://github.com/Olical/conjure) to lint clojure files. Conjure creates a log-type buffer with `filetype=clojure` and `buftype=nofile`. `diagnosticls` attaches to this buffer and shows linter errors (It's a valid clojure file, but I don't want to lint this particular buffer).

I noticed that the original commit (#307) uses `buftype` but it was changed in [Add option to disable autostart · neovim/nvim-lspconfig@687c293](https://github.com/neovim/nvim-lspconfig/commit/687c293cc20fd2bc2b6bd60cfe1037578e0703a2#diff-0e1bd346624fa540fa153e867fbc2036ced5c83d5cd7d79d9308df1328b7c879L135-R154) and checking the manual (`:help 'buftype'`), I can see that `nofile` is here (I don't think `nofile` is a valid `filetype` option <or yes?>)

I think this PR should not check the two conditions and revert to the old behavior (only check that `buftype` is `nofile`). Or is there some plugin that uses a `nofile` filetype?

Maybe someone wants to attach an lsp client to a buffer with nofile buftype? I could see the reasoning for a yes (If the buffer is editable, maybe you want lsp intellisense; or pop documentation with `vim.lsp.buf.hover`)

mjlbach:
> I noticed that the original commit (#307) uses buftype but it was changed in Add option to disable autostart · neovim/nvim-lspconfig@687c293 and checking the manual (:help 'buftype'), I can see that nofile is here (I don't think nofile is a valid filetype option <or yes?>)

Yep, I think that was just a mistake on my part. This PR should change back to testing buftype. I believe this also fixes @raw1z's issue

mjlbach:
superseded by https://github.com/neovim/nvim-lspconfig/pull/1064

PR #1059, opened 2021-07-11
[WIP] Allow for per-config tests with `leanls`/`lean3ls` as the first imple…

rish987:
…mentation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,6 +38,20 @@ jobs:
     - name: Run test with building Nvim
       run: |
         make test
+    - name: Install per-config test dependencies
+      run: |
+        mkdir packpath
+        git clone --depth 1 https://github.com/nvim-lua/plenary.nvim packpath/plenary.nvim
+    - name: leanls install
+      run: |
+          curl https://raw.githubusercontent.com/Kha/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain "leanprover-community/lean:3.30.0" -y
+          echo "$HOME/.elan/bin/" >> $GITHUB_PATH
+          PATH="$PATH:$HOME/.elan/bin/"
+          cd test/fixtures/leanls/ && leanpkg configure && leanpkg build
+    - name: lean3ls install
+      run: |
+          sudo npm install -g lean-language-server
+          cd test/fixtures/lean3ls/ && leanpkg configure && leanpkg build
   macos:
     runs-on: [macos-latest]
     steps:
@@ -51,4 +65,3 @@ jobs:
     - name: Run test with building Nvim
       run: |
         make test
-
--- a/test/fixtures/lean3ls/leanpkg.toml
+++ b/test/fixtures/lean3ls/leanpkg.toml
@@ -1,0 +1,7 @@
+[package]
+name = "test"
+version = "0.1"
+lean_version = "leanprover-community/lean:3.30.0"
+path = "."
+
+[dependencies]
--- a/test/fixtures/lean3ls/test.lean
+++ b/test/fixtures/lean3ls/test.lean
@@ -1,0 +1,13 @@
+import test.test1
+
+def test1 : nat := test
+
+theorem test2 {p : Prop} {q : Prop} : p ∨ q → q ∨ p :=
+begin
+  intro h,
+  cases h with h1 h2,
+    apply or.inr,
+    exact h1,
+    apply or.inl,
+    assumption,
+end
--- a/test/fixtures/lean3ls/test/test1.lean
+++ b/test/fixtures/lean3ls/test/test1.lean
@@ -1,0 +1,1 @@
+def test : nat := 5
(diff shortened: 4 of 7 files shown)


rish987:
(whoops, still working on this, meant to PR this to my fork, ignore)

mjlbach:
You should keep working on this, but I am somewhat mixed on the idea of bringing language server CI into lspconfig. We're going to have to start maintaining 100+ language server piplines, including installation + keeping up with server response changes + adding an external dependency on plenary... 

Lspconfig is meant to be a catch-all fallback when there does not exist a good dedicated language server specific plugin like lean/nvim-jdtls/nvim-metals. Much of our CI will only be able to test a limited subset of spec compliant functionality. 

Anyways, just my 2c, we can see what others think too :)

rish987:
(for context: Julian/lean.nvim#91)

Yeah, I get that per-config testing would be a lot to maintain. Were the plans for this repo to eventually upstream to neovim core as "out-of-the-box" language server support? If so per-config tests would seem necessary. If not, then IMO we should probably (as soon as possible) stop including default configs here and outsource those to their respective plugins.

If a per-config test fails because some dependency (in this case, for example `elan`) broke, then clearly this plugin does not work as advertised - so we could just mark those tests as pending, and create an issue for it.

Mentioning @Julian in case they have any opinion on this.

mjlbach:
> Yeah, I get that per-config testing would be a lot to maintain. Were the plans for this repo to eventually upstream to neovim core as "out-of-the-box" language server support? I

This has been discussed, but I don't think it's practical. Language server settings change too fast, servers get deprecated, etc.

The plan (IMO), is to remove everything *but* the configs from lspconfig, and have each config refer to the respective "language-specific plugin"

>  If not, then IMO we should probably (as soon as possible) stop including default configs here and outsource those to their respective plugins.

I can't reasonably maintain 70+ LSP configurations. At the same time, a ton of people want a configuration that just works OOTB (like polyglot for syntax highlighting). I think the ideal is that each language server specific plugin contributes back their configuration (sans custom, off-spec handlers) upstream, and that we boost the visibility of these plugins in the documentation for each server. We currently have a wiki section on these.

Generally one goal of mine for the 0.6 release cycle is to get the built-in lsc ecosystem off the ground with more tailored plugins based on the nvim-metals/nvim-jdtls/fluttertools/lean template.

rish987:
Gotcha, I totally agree with your vision of keeping this as some kind of OOTB support.

The maintenance I believe could be scripted, and so would require ~0 extra work on your part. We could do periodic workflow runs, and if something breaks, we could automatically mark it as pending and create an issue. I'd be happy to look into this in this PR. IMO this would totally be worth it for the extra assurance that everything in this plugin works as intended.

rish987:
(^presumably this could be done using some state external to the repo, and so marking a test pending won't actually create a new commit)

rish987:
OK here's an idea -- could we create a new repository like `nvim-lspconfig-test` for writing per-config tests to and `nvim-lspconfig-test-state` for storing the pending/active states? That way we could have this kind of testing with zero changes to this repo. Will try this (so ignore this PR for now).

mjlbach:
@rish987 I was thinking the same thing, in fact, I started working on it last night :)

mjlbach:
@rish987 I'm down to work on this a bit, I think it should be in it's own repo. Maybe we do nvim-lua for now.

rish987:
Awesome, in that case I'll hold off on my attempt and let you work on it. Would be happy to offer a first per-config test implementation for lean.nvim as soon as you're ready 😀

rish987:
See [this file](https://github.com/rish987/lean.nvim/blob/ceff5e4d01e3f5d0ca896f90379fd62ac5c10bd2/lua/tests/lsp/lsp/lsp_spec.lua) for roughly what I expect such a per-config test to look like

mjlbach:
I think we could make a good case that neovim is an excellent platform for doing automated CI on your language server and try to get buy-in from upstream :) We can make the repo into a GH action that runs on the latest lspconfig when it's done (or core actually, very little needs to get tested here).

rish987:
Not sure I understand correctly -- by "upstream" do you mean neovim core or the Lean 3/4 repos, and which repo did you want to turn into a Github action, this one or a new one specifically for testing?

rish987:
The Lean 3/4 language servers are already well-tested I believe. The reason I think these tests are necessary is specifically to make sure the language server works with nvim in particular (e.g. what #1041 was meant to fix, which is untested at the moment, and all of the untested features from the default configs for all the other languages)

rish987:
Ah I think I get it, you're saying that we can try to convince neovim upstream to allow us to use the `plenary.nvim` dependency for these tests (as opposed to pure `busted`), CMIIW. If so, that sounds good to me. I'm a bit of a newbie so forgive me for missing some context here...

mjlbach:
No, I'm saying there are a ton of language servers (that aren't lean) without CI against a client, for language servers that are not MS developed, it would be nice to show as a PoC how you can test against neovim.

rish987:
(closing in favor of [rish987/nvim-lspconfig-test](https://github.com/rish987/nvim-lspconfig-test))

PR #1062, opened 2021-07-12
docs: add a direct link to lean.nvim (for leanls and lean3ls)

Julian:
Better than going back and forth on gitter I suppose :D @mjlbach.

Possibly this is a decent enough way to let people know something exists, though I'm not sure how many Leanfolk even know about this repo yet (hopefully that changes soon!).

### DIFF
--- a/lua/lspconfig/lean3ls.lua
+++ b/lua/lspconfig/lean3ls.lua
@@ -28,6 +28,10 @@ Once Lean is installed, you can install the Lean 3 language server by running
 ```sh
 npm install -g lean-language-server
 ```
+
+Note: that if you're using [lean.nvim](https://github.com/Julian/lean.nvim),
+that plugin fully handles the setup of the Lean language server,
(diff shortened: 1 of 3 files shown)


lua/lspconfig/lean3ls.lua
```diff
@@ -28,6 +28,10 @@ Once Lean is installed, you can install the Lean 3 language server by running
 ```sh
 npm install -g lean-language-server
 ```
+
+Note: that if you're using [lean.nvim](https://github.com/Julian/lean.nvim),
+that plugin fully handles the setup of the Lean language server,
+and you shouldn't set up `lean3ls` both with it and this plugin.
```
mjlbach, line 34: maybe this plugin -> lspconfig just to be more explicit

Julian: Done, thanks (as well as for `nvim-metals`, which I copied this from :D)

Issue #1028, opened 2021-07-03
NEED TESTERS: fallback to interactive root picker if no root directory detected


mjlbach:
There are basically three usability complaints with lspconfig

1. we don't automatically provide language servers (solved by lspinstall)
2. we don't set up autocompletion + snippet for you (solved by nvim-compe + luasnip/vnsip)
3. Occasionally servers fail to start

3 is because lspconfig tries to find the correct root directory for the project, many language servers are picky about this, but some will still provide some utility without the correct directory.

I'm going to add a fallback hook that can be customized to either:
1. Prompts you interactively to choose the root directory or
2. Prints a warning that no non-parent root pattern has been detected, and tries to start the server in the parent directory

WIP:
![image](https://user-images.githubusercontent.com/13316262/125319558-bc59e080-e2ef-11eb-9092-8636d7a9781a.png)

mizlan:
I think an lsp-mode style prompt is better for usability and troubleshooting. Granted, users will have to (be patient enough to) read through the different options selected in the prompt.

mjlbach:
We got a lot of complaints last time I put an interactive prompt, so mixed.

mjlbach:
Please test #1066 

I had to make sweeping changes to root directory detection (mainly disabling working directory and parent directory fallback) to support this change.

You will be able to override the fallback behavior, but I think this is generally a better default for everyone.

dundargoc:
I can't get the interactive root directory picker to trigger/start, but the root directory isn't being detected either. At least I don't think so, maybe I missed something. It only works when I create a softlink build/.....command.json (something like that) to root manually.

mjlbach:
That's strange, if you `mkdir test` && `cd test` && `touch main.c` && `nvim main.c`, it immediately pops up for me. If you run `:LspInfo` do you see a root? The server can still detect a root and fail to (appropriately) start because compile_command.json is missing. In the event *any* root is detected, lspconfig will never show you the popup.

mizlan:
You should make it more obvious what this is for, i.e., the popup should mention LSP somewhere. This is important because "Ignore this file" has no context otherwise (you may also want to reword it to "Don't enable LSP on this file" or something of the sort).

mizlan:
[...]

PR #1066, opened 2021-07-13
[WIP] Feat/interactive root picker

mjlbach:
Closes https://github.com/neovim/nvim-lspconfig/issues/1028

### DIFF
--- a/lua/lspconfig/efm.lua
+++ b/lua/lspconfig/efm.lua
@@ -7,9 +7,7 @@ local bin_name = "efm-langserver"
 configs[server_name] = {
   default_config = {
     cmd = { bin_name },
-    root_dir = function(fname)
-      return util.root_pattern ".git"(fname) or util.path.dirname(fname)
-    end,
+    root_dir = util.root_pattern ".git",
   },
 
   docs = {
@@ -18,9 +16,7 @@ https://github.com/mattn/efm-langserver
 
 General purpose Language Server that can use specified error message format generated from specified command.
 ]],
-    default_config = {
-      root_dir = [[util.root_pattern(".git")(fname) or util.path.dirname(fname)]],
-    },
   },
 }
+
 -- vim:et ts=2 sw=2
--- a/lua/lspconfig/gopls.lua
+++ b/lua/lspconfig/gopls.lua
@@ -15,9 +15,7 @@ https://github.com/golang/tools/tree/master/gopls
 
 Google's lsp server for golang.
 ]],
-    default_config = {
-      root_dir = [[root_pattern("go.mod", ".git")]],
-    },
   },
 }
+
 -- vim:et ts=2 sw=2
--- a/lua/lspconfig/jedi_language_server.lua
+++ b/lua/lspconfig/jedi_language_server.lua
@@ -1,12 +1,19 @@
 local configs = require "lspconfig/configs"
+local util = require "lspconfig/util"
+
+local root_files = {
+  "setup.py",
+  "pyproject.toml",
+  "setup.cfg",
+  "requirements.txt",
+  ".git",
+}
 
 configs.jedi_language_server = {
   default_config = {
     cmd = { "jedi-language-server" },
     filetypes = { "python" },
-    root_dir = function(fname)
-      return vim.fn.getcwd()
-    end,
+    root_dir = util.root_pattern(unpack(root_files)),
   },
   docs = {
     description = [[
@@ -14,9 +21,6 @@ https://github.com/pappasam/jedi-language-server
 
 `jedi-language-server`, a language server for Python, built on top of jedi
     ]],
-    default_config = {
-      root_dir = "vim's starting directory",
-    },
   },
 }
 -- vim:et ts=2 sw=2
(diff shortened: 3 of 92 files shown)


lua/lspconfig/ui/prompt.lua
```diff
@@ -0,0 +1,54 @@
...
+	["p"]=function() return parent_dir end,
+	["c"]=M.directory_picker,
+	["i"]=function() return nil end,
+    }
+
+    local user_input
+    local user_choice
+    local root_directory
+
+    while true do
```
mjlbach, line 36: @folke Do you have a better solution for blocking input? I need to return a result from this function so I can't just map the callbacks and be done with it.

folke: That's pretty much what I do in WhichKey too.

The only thing I would change is:

```lua
  local ok, n = pcall(vim.fn.getchar)
 if not ok then break end
```

That properly handles keyboard interrupts too (user pressing `<control>c`)

folke:
Let me know once it makes sense to test and do a code review. Looking good!

mjlbach:
> CONFIG.md is auto-generated: edit the lua source file instead (and re-read the PR template).

you have made an enemy today, my child.

lua/lspconfig/efm.lua
```diff
@@ -7,9 +7,7 @@ local bin_name = "efm-langserver"
 configs[server_name] = {
   default_config = {
     cmd = { bin_name },
-    root_dir = function(fname)
-      return util.root_pattern ".git"(fname) or util.path.dirname(fname)
-    end,
+    root_dir = util.root_pattern ".git",
```
WhyNotHugo, line 10: This change breaks for me. Using the current directory if outside a git repository makes EFM work for things that are standalone scripts and alike.

lua/lspconfig/gopls.lua
```diff
@@ -15,9 +15,7 @@ https://github.com/golang/tools/tree/master/gopls
 
 Google's lsp server for golang.
 ]],
-    default_config = {
-      root_dir = [[root_pattern("go.mod", ".git")]],
```
WhyNotHugo, line 19: The previous version here seems pretty solid, given that `go.mod` exists in the root of all modern `golang` projects.

lua/lspconfig/jedi_language_server.lua
```diff
@@ -1,22 +1,26 @@
...
+}
 
 configs.jedi_language_server = {
   default_config = {
     cmd = { "jedi-language-server" },
     filetypes = { "python" },
-    root_dir = function(fname)
-      return vim.fn.getcwd()
-    end,
+    root_dir = util.root_pattern(unpack(root_files)),
```
WhyNotHugo, line 16: Nice! :+1:

lua/lspconfig/jsonls.lua
```diff
@@ -11,7 +11,7 @@ configs[server_name] = {
     init_options = {
       provideFormatter = true,
     },
-    root_dir = util.root_pattern(".git", vim.fn.getcwd()),
+    root_dir = util.find_git_ancestor,
```
WhyNotHugo, line 14: Having `cwd` as a fallback is great here. The `JSON` LSP doesn't even have to deal with things like imports or alike, it _just_ checks the current file.

lua/lspconfig/solargraph.lua
```diff
@@ -30,9 +30,7 @@ You can install solargraph via gem install.
 gem install --user-install solargraph
 ```
     ]],
-    default_config = {
-      root_dir = [[root_pattern("Gemfile", ".git")]],
```
WhyNotHugo, line 34: `Gemfile` as a fallback here made more sense that not having it.

WhyNotHugo (commented):
I'm not a fan of having the interactive picker become the default. It's very invasive: when opening a file, I generally want to start reading / editing right away, not configuring an LSP or dealing with plugins prompting for data.

Configurating neovim and using neovim are two very distinct activities for me, and I don't want to deal with under-the-hood nuances when working on something else.

It should definitely be configurable option, and, IMHO, disabled by default.

Some of the auto-detection fallbacks that are being dropped here were pretty solid, and, personally, I think we'd be dropping some very useful code by deleting them. Is there any rationale for not keeping them, especially the ones that work fine?

lua/lspconfig/gopls.lua
mjlbach: This is documentation only, it's autogenerated now and does not affect current behavior in any way.

lua/lspconfig/jsonls.lua
mjlbach: get_cwd() is completely incorrect, it should be, at best, util.path.dirname

lua/lspconfig/solargraph.lua
mjlbach: This is documentation only, it's autogenerated now and does not affect current behavior in any way.

mjlbach:
For those leaving drive-by comments, removing `default_config` from the *documentation* section does not change current behavior. It uses the auto-generated documentation pipeline I pulled to extract the function and paste the text. Many of our currently documented root directories do *not* match the actual code.

PR #1072, opened 2021-07-14
feat: added global_config hooks

folke:
For the workspace stuff I'm working on, it would be good to have two ways to hook into lspconfig:

1. global `on_setup`. Every client config will be passed through this global function during setup if it exists.
2. `config.enabled`. When `false`, the client won't start. Use case is to disable certain lsp servers in a workspace, or programmatically based on other conditions in `config.on_new_config` (same as coc)

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -50,6 +50,10 @@ function configs.__newindex(t, config_name, config_def)
 
     config = tbl_extend('keep', config, default_config)
 
+    if util.on_setup then
+      pcall(util.on_setup, config)
+    end
+
     local trigger
     if config.filetypes then
       trigger = 'FileType ' .. table.concat(config.filetypes, ',')
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -84,7 +84,7 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     -- Used by :LspInfo
-    M.get_root_dir = config.root_dir
+    M.get_root_dir = get_root_dir
     M.filetypes = config.filetypes
     M.handlers = config.handlers
     M.cmd = config.cmd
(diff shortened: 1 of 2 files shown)


lua/lspconfig/configs.lua
```diff
@@ -84,7 +88,7 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     -- Used by :LspInfo
-    M.get_root_dir = config.root_dir
+    M.get_root_dir = get_root_dir
```
mjlbach, line 91: This needs to be reverted right?

lua/lspconfig/util.lua
```diff
@@ -222,6 +225,10 @@ function M.server_per_root_dir_manager(_make_config)
     local client_id = clients[root_dir]
     if not client_id then
       local new_config = _make_config(root_dir)
+      -- do nothing if the client is not enabled
+      if new_config.enabled == false then
```
mjlbach, line 229: more lu-onic to do if not new_config.enabled imo

lua/lspconfig/configs.lua
folke: I think it was a bug in the code tbh. `get_config_dir`  is used in the code, but we expose `config.root_dir`. Probably best to expose `get_root_dir` instead? 

Based on the current code, you'll get the same result, so doesn't really matter. Thought I'd change it for consistency

lua/lspconfig/util.lua
folke: `new_config.enabled` will normally not be set, so it will be `nil`. That's why a check for `false` is needed here

lua/lspconfig/configs.lua
mjlbach: Ah, that's fine. can you put it under it's own commit though with the message?

lua/lspconfig/util.lua
mjlbach: ah, got it

lua/lspconfig/configs.lua
folke: done

PR #1079, opened 2021-07-16
[julia] improvements to current project detection.

fredrikekre:
Based on discussions from https://discourse.julialang.org/t/neovim-languageserver-jl/37286

cc @kdheepak @MichaelHatherly

### DIFF
--- a/lua/lspconfig/julials.lua
+++ b/lua/lspconfig/julials.lua
@@ -7,16 +7,31 @@ local cmd = {
   '--history-file=no',
   '-e',
   [[
-    using Pkg;
+    using Pkg
     Pkg.instantiate()
-    using LanguageServer; using SymbolServer;
+    using LanguageServer
     depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
-    project_path = dirname(something(Base.current_project(pwd()), Base.load_path_expand(LOAD_PATH[2])))
-    # Make sure that we only load packages from this environment specifically.
-    @info "Running language server" env=Base.load_path()[1] pwd() project_path depot_path
-    server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path);
-    server.runlinter = true;
-    run(server);
+    project_path = let
+        dirname(something(
+            ## 1. Finds an explicitly set project (JULIA_PROJECT)
+            Base.load_path_expand((
+                p = get(ENV, "JULIA_PROJECT", nothing);
+                p === nothing ? nothing : isempty(p) ? nothing : p
+            )),
+            ## 2. Look for a Project.toml file in the current working directory,
+            ##    or parent directories, with $HOME as an upper boundary
+            Base.current_project(),
+            ## 3. First entry in the load path
+            get(Base.load_path(), 1, nothing),
+            ## 4. Fallback to default global environment,
+            ##    this is more or less unreachable
+            Base.load_path_expand("@v#.#"),
+        ))
+    end
+    @info "Running language server" VERSION pwd() project_path depot_path
+    server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path)
+    server.runlinter = true
+    run(server)
   ]],
 }
 
@@ -57,16 +72,31 @@ require'lspconfig'.julials.setup{
         "--startup-file=no",
         "--history-file=no",
         "-e", [[
-          using Pkg;
+          using Pkg
           Pkg.instantiate()
-          using LanguageServer; using SymbolServer;
+          using LanguageServer
           depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
-          project_path = dirname(something(Base.current_project(pwd()), Base.load_path_expand(LOAD_PATH[2])))
-          # Make sure that we only load packages from this environment specifically.
-          @info "Running language server" env=Base.load_path()[1] pwd() project_path depot_path
-          server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path);
-          server.runlinter = true;
-          run(server);
+          project_path = let
+              dirname(something(
+                  ## 1. Finds an explicitly set project (JULIA_PROJECT)
+                  Base.load_path_expand((
+                      p = get(ENV, "JULIA_PROJECT", nothing);
+                      p === nothing ? nothing : isempty(p) ? nothing : p
+                  )),
+                  ## 2. Look for a Project.toml file in the current working directory,
+                  ##    or parent directories, with $HOME as an upper boundary
+                  Base.current_project(),
+                  ## 3. First entry in the load path
+                  get(Base.load_path(), 1, nothing),
+                  ## 4. Fallback to default global environment,
+                  ##    this is more or less unreachable
+                  Base.load_path_expand("@v#.#"),
+              ))
+          end
+          @info "Running language server" VERSION pwd() project_path depot_path
+          server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path)
+          server.runlinter = true
+          run(server)
         \]\]
     };
       new_config.cmd = cmd


MichaelHatherly:
Thanks @fredrikekre, works well for me locally.

kdheepak:
https://github.com/neovim/nvim-lspconfig/blob/386c9230c681f7bb11e3bfd0e466ecba05483676/lua/lspconfig/julials.lua#L44

I think this line might need to be updated too. Maybe even this block is no longer needed? We could also update it such that it doesn’t install languageserver in a global environment, and installs it in a nvim’s cache folder or data folder.

clason:
Standard location for language servers not globally installed is under `XDG_CACHE_DIR`: `stdpath('cache')..'lspconfig'..repo_checkout`, see https://github.com/neovim/nvim-lspconfig/blob/9d49032a64d8a275123ad7f84e9c699d279307fe/lua/lspconfig/sumneko_lua.lua#L39

clason:
But just to make it clear: installing language servers is no longer within the scope of this project; users should either 

* do this themselves;
* use an installer like https://github.com/kabouzeid/nvim-lspinstall;
* use a language(server)-specific plugin like https://github.com/mfussenegger/nvim-jdtls.

(I know the situation is a bit different here since the language server is not a separate binary, but still.)

fredrikekre:
> I think this line might need to be updated too.

Yea, I was gonna do that in a separate PR and limit this one just to project discovery, but I can include it here too.

> We could also update it such that it doesn’t install languageserver in a global environment, and installs it in a nvim’s cache folder or data folder.

> But just to make it clear: installing language servers is no longer within the scope of this project; users should either ....


Is there an easy way to configure the install path? I imagine something like
```
require'lspconfig'.julials.setup{
    languageserver_install_path = "path where I installed LanguageServer.jl"
}
```
or do you have to override the entire `cmd`? I think it would make sense, from the perspective of this repo, to assume LanguageServer.jl to be installed globally if that setting is not set and then users can override that (e.g. I used `~/.config/nvim/lsp-julia` in https://discourse.julialang.org/t/neovim-languageserver-jl/37286). Then an installer like https://github.com/kabouzeid/nvim-lspinstall would simply provide this setting in a more automatic way.

clason:
Yes, users (of nvim-lspconfig) are expected to provide the full `cmd` needed to start the language server. But the config can specify a default location and build that programmatically if needed; this is exactly what the config for `sumneko/lua-language-server` does (which is exactly why I linked it in my first comment).

clason:
(And I'm also not sure why the `on_new_config` is needed...)

fredrikekre:
I see. Would it make sense if the default configuration tried to load LanguageServer.jl from first `vim.fn.stdpath('cache')..'/lspconfig/julia/` and if it is not found there from the global environment? That would make installation instructions pretty easy:

> To install, either make sure LanguageServer.jl is available in the global env, or install it to the `/home/user/.cache/nvim/lspconfig/julia` directory, e.g. with
> ```
> $ julia --project=/home/user/.cache/nvim/lspconfig/julia -e 'using Pkg; Pkg.add("LanguageServer")'
> ```

otherwise custom installation requires duplicating all of the `cmd` IIUC (and that means updates and improvements to the `cmd` in this repo won't propagate to users?).


> (And I'm also not sure why the on_new_config is needed...)

This is for my question in the config in https://discourse.julialang.org/t/neovim-languageserver-jl/37286/60 ? My guess is that since there is a default `on_new_config` here https://github.com/neovim/nvim-lspconfig/blob/9d49032a64d8a275123ad7f84e9c699d279307fe/lua/lspconfig/julials.lua#L26 it overrides it an uses the `cmd` from https://github.com/neovim/nvim-lspconfig/blob/9d49032a64d8a275123ad7f84e9c699d279307fe/lua/lspconfig/julials.lua#L25 but I have no idea how lspconfig works really.

fredrikekre:
Alternatively, I suppose the default `cmd` could take an optional argument for an install path, and then to use a custom install path you can just use
```
require'lspconfig'.julials.setup{
        table.insert(new_config.cmd, "/path/to/LanguageServer.jl/installation")
    end
}
```

clason:
> but I have no idea how lspconfig works really.

Black magic :)

On a more serious note, that pattern (`on_new_config`) was introduced by @mjlbach in https://github.com/neovim/nvim-lspconfig/pull/547, but he didn't comment why he made that change. (Maybe he can do so now?)

My personal feeling is that we should assume that `julia` is in the user's path, but it might make sense to allow specifying the location of the julia _project_ the language server gets installed to. I also think that it's completely OK to leave users to configure the `cmd` to their specific setting as long as there's a default in the docs that is fairly straightforward to adapt.

But none of this should block this PR, which looks like a fairly minor change (as far as code structure goes) that is a strict improvement (and actually removing breakage, if I understand correctly?)

fredrikekre:
> make sense to allow specifying the location of the julia project the language server gets installed to

Right, this is what I have talked about from the start.

> But none of this should block this PR, which looks like a fairly minor change (as far as code structure goes) that is a strict improvement (and actually removing breakage, if I understand correctly?)

Okay, I will make a follow up PR with the other changes. This PR should not be breaking anything, it only improves the situation for non-standard configurations (for example, `LOAD_PATH[2]` from the old configuraiton might not exist).

clason:
(Really appreciate you contributing these fixes, by the way! ❤️ )

mjlbach:
Can someone please type of a summary of why these changes are necessary, why SymbolServer is no longer required, etc? I can read through the discourse, but a summary will expedite my review :)

fredrikekre:
1. Loading `SymbolServer` has never been necessary to load from the cmd snippet. It is still used by `LanguageServer.jl` internally of course, but `LanguageServer` itself takes care of loading it. (So this is just a cleanup).
2. For `project_path`: The old configuration assumed that user have not configured anything. This new one should play along better with users that configure things via `JULIA_PROJECT` and `JULIA_LOAD_PATH` environment variables. Some examples:
   1. This would ignore the user specified `JULIA_PROJECT` and just use the project from `pwd`:
      ```
      $ export JULIA_PROJECT=/some/path
      $ nvim  ...
      ```
   2. This would give an error (out of bounds access in `LOAD_PATH[2]` in the old code):
      ```
      $ export JULIA_LOAD_PATH=/some/path
      $ nvim  ...
      ```
   This PR makes sure that `project_path` becomes `/some/path` in both cases, and from what I have seen this should work out of the box with most (all?) peoples workflow and certainly good enough for beeing a default.

kdheepak:
@mjlbach I believe this was required is not anymore

https://github.com/julia-vscode/SymbolServer.jl/issues/54#issuecomment-643667218

> In the symbol server process, we are serializing things onto disc using Julia serialization. When we deserialize in the LS process, it probably is looking for the types that should be materialized in the Main.SymbolServer module, and not in Main.LanguageServer.SymbolServer.

The relevant PRs in those issues have been merged and new releases have been made.

mjlbach:
Terrific, thank you so much!

I've also just tested and it seems to work fine for me on some old Julia projects, including the new environmental variables.

One thing I've wondered, is if we should also add instructions to the Julia configuration for precompiling a system image with LanguageServer.jl to improve the startup times. I haven't investigated whether this would help, but I previously did this for plots.jl

I'll merge as is.

fredrikekre:
Thanks!

> I believe this was required is not anymore: julia-vscode/SymbolServer.jl#54 (comment)

Ah, nice find. I'll keep an eye open for any errors related to that but  hopefully it is fixed now.

MichaelHatherly:
> One thing I've wondered, is if we should also add instructions to the Julia configuration for precompiling a system image with LanguageServer.jl to improve the startup times. I haven't investigated whether this would help, but I previously did this for plots.jl

It _kind of_ did the last time I built one, though not hugely, probably due to not exercising enough code during tracing. If someone manages to come up with a script that precompiles enough stuff that would be great.

mjlbach:
Related: https://github.com/julia-vscode/LanguageServer.jl/issues/950

If someone wants to follow-up/look into this, that would be great. I've heard people comment on how "slow" our client is for Julia, but in reality it's that the "time to first" problem for Julia also applies to LanguageServer.jl :(

clason:
And, as opposed to the VS Code extension, we don't show progress messages on startup so people think it's _our_ fault ;)

(But isn't the TTFP issue almost fully mitigated by putting `LanguageServer.jl` in its own project, as the instructions -- which still contain `]add SymbolServer`, by the way -- suggest?)

clason:
> And, as opposed to the VS Code extension, we don't show progress messages on startup so people think it's _our_ fault ;)

Sumneko does something nice, by the way: While it's spinning up, all responses for hover, completion etc. are replaced by "Indexing workspace x/y". That is helpful feedback without requiring non-spec extensions.

mjlbach:
> (But isn't the TTFP issue almost fully mitigated by putting LanguageServer.jl in its own project, as the instructions -- which still contain ]add SymbolServer, by the way -- suggest?)

I don't think so, you still need to use `PackageCompiler` after running `LanguageServer.jl` such that the functions you care about are called/JAOT compiled, and them compile a `SysImage`such that the state of the running Julia process is seralized/stored so all of the cached/compiled code can be reused. Minimizing the size of the sysimage by putting LanguageServer.jl in its own project does bring some performance benefits AFAIK, but a SysImage should eliminate the startup time completely (apart from the actual runtime cost of indexing the project etc.). This is also my naive understanding of Julia's compilation/caching model as I only use it for hobby projects :)

clason:
Yes, I am aware of all that; my impression was just that the _real_ latency on startup was from LanguageServer indexing the global project -- (and having to start from scratch every time there was an error). The TTFP latency (AOT compiling LanguageServer.jl) shouldn't be as terrible with the latest version. (As in, a second maybe, but that's not more than sumneko take spooling up.)

mjlbach:
Ah, sorry for the overly verbose summary. Let's see what people's experience is and hopefully this change is sufficient 😄

mjlbach:
Can someone confirm that there is or is not a regression here? 

https://github.com/neovim/nvim-lspconfig/issues/1106

Issue #1067, opened 2021-07-13
Extended JDTLS support


eruizc-dev:
![image](https://user-images.githubusercontent.com/58179604/125474999-8456d09c-8ff2-4dca-8e12-55d2a6e734ad.png)

### Clarification

This is something I'd like to do and contribute. I'm creating an issue to see if there is interest from the community and it stays within this project's scope before I spend time on something that won't be merged.

### Issue

JDTLS is a non-conventional language server, I've been using it for 3 months and it works fine but sometimes errors pop up, code generations always use incorrect indentation and I know there's a lot of missing features.

I know there's an amazing plugin ([nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls)) that takes care of this, but it more than just a language server client including testing, debugging, bytecode inspection and more...

There's also an issue on how it's launched, the command is huge (you gotta spin up jvm with lots of arguments) and there's no convention. Also, the default cmd has no Lombok support which is a **must** for writing modern java.

### Proposal

 - Add some extended client capabilities
 - Autodetect if user installed jdtls via [AUR](https://aur.archlinux.org/packages/java-language-server/) or [JDTLS-LAUNCHER](https://github.com/eruizc-dev/jdtls-launcher)
 - Allow Lombok setup with default cmd
 - Add better documentation

mjlbach:
I'm in favor of anything that does not substantially add to the complexity of the jdtls configuraiton in nvim-lspconfig, however, I am not in favor of anything that is too opinionated (does adding lombok support result in strange behavior for those not using lombok?) or anything that is tailored to a specific installation method (I would consider the AUR to be an unreliable source generally).

Re: extending client capabilities which capabilities were you thinking of? If we advertise something core does not support it may result in undefined behavior from the language server.

I really want to push people to use/improve nvim-jdtls, is there an issue with not using the additional features it affords?

mfussenegger:
> but it more than just a language server client including testing, debugging, bytecode inspection and more...

All of that functionality is based on the language server. The debug-adapter is an extension to the language server, and doesn't work without it. The bytecode inspection is using the language server to retrieve the classpath of your project. Besides - having that functionality in it has no overhead at runtime. Most of it also doesn't have an overhead at import time, because it lives in separate modules.

> There's also an issue on how it's launched, the command is huge (you gotta spin up jvm with lots of arguments) and there's no convention. Also, the default cmd has no Lombok support which is a must for writing modern java.

[...]

PR #1082, opened 2021-07-17
Improve JDTLS defaults and improve documentation

eruizc-dev:
After the discussion in #1067 this is what came up:

 - Improve default root_dir: 
    - Some people complaint that projects without package managers didn't work, so I added Makefile and .git
    - The multi-file project root was unnecessary, it also included `.kts` files that are for Kottlin projects
 - Improve cmd autodetect:
    - Removed `cmd_env` property that did nothing
    - Made `JAVA_HOME` and `WORKSPACE` env variables optional
    - Unified `JAR` and `JDTLS_CONFIG` into `JDTLS_HOME`
    - Preserved `JAR` and `JDTLS_CONFIG` with a message to avoid breaking existing configurations
    - Added `.exe` for windows support
 - Update documentation
    - Add links to point in the right direction
    - Mention jdtls-launcher as unofficial installer/launcher
    - Mention nvim-jdtls as complete client implementation
    - Provide actual configuration examples

### DIFF
--- a/lua/lspconfig/jdtls.lua
+++ b/lua/lspconfig/jdtls.lua
@@ -1,52 +1,44 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 local handlers = require 'vim.lsp.handlers'
-local path = util.path
 
 local server_name = 'jdtls'
 
-local cmd = {
-  util.path.join(tostring(vim.fn.getenv 'JAVA_HOME'), '/bin/java'),
-  '-Declipse.application=org.eclipse.jdt.ls.core.id1',
-  '-Dosgi.bundles.defaultStartLevel=4',
-  '-Declipse.product=org.eclipse.jdt.ls.core.product',
-  '-Dlog.protocol=true',
-  '-Dlog.level=ALL',
-  '-Xms1g',
-  '-Xmx2G',
-  '-jar',
-  tostring(vim.fn.getenv 'JAR'),
-  '-configuration',
-  tostring(vim.fn.getenv 'JDTLS_CONFIG'),
-  '-data',
-  tostring(vim.fn.getenv 'WORKSPACE'),
-  '--add-modules=ALL-SYSTEM',
-  '--add-opens java.base/java.util=ALL-UNNAMED',
-  '--add-opens java.base/java.lang=ALL-UNNAMED',
+local sysname = vim.loop.os_uname().sysname
+local env = {
+  HOME = vim.loop.os_homedir(),
+  JAVA_HOME = os.getenv 'JAVA_HOME',
+  JDTLS_HOME = os.getenv 'JDTLS_HOME',
+  WORKSPACE = os.getenv 'WORKSPACE',
 }
 
---- The presence of one of these files indicates a root directory.
---
--- We search for configuration files of the most common Java build systems. We
--- could add more build systems, such as Make, but let's keep things simple for
--- now.
-local root_files = {
-  -- Single-module projects
-  {
-    'build.xml', -- Ant
-    'pom.xml', -- Maven
-    'settings.gradle', -- Gradle
-    'settings.gradle.kts', -- Gradle
-  },
-  -- Multi-module projects
-  { 'build.gradle', 'build.gradle.kts' },
-}
+local function get_java_executable()
+  local executable = env.JAVA_HOME and util.path.join(env.JAVA_HOME, 'bin', 'java') or 'java'
 
---- Callback function for the `language/status` notification.
---
--- The server sends a non-standard notification when the status of the language
--- server changes. This can be used to display progress as the server is
--- starting up.
+  return sysname:match 'Windows' and executable .. '.exe' or executable
+end
+
+local function get_workspace_dir()
+  return env.WORKSPACE and env.WORKSPACE or util.path.join(env.HOME, 'workspace')
+end
+
+local function get_jdtls_jar()
+  return vim.fn.expand '$JDTLS_HOME/plugins/org.eclipse.equinox.launcher_*.jar'
+end
+
+local function get_jdtls_config()
+  if sysname:match 'Linux' then
+    return util.path.join(env.JDTLS_HOME, 'config_linux')
+  elseif sysname:match 'Darwin' then
+    return util.path.join(env.JDTLS_HOME, 'config_mac')
+  elseif sysname:match 'Windows' then
+    return util.path.join(env.JDTLS_HOME, 'config_win')
+  else
+    return util.path.join(env.JDTLS_HOME, 'config_linux')
+  end
+end
+
+-- Non-standard notification that can be used to display progress
 local function on_language_status(_, _, result)
   local command = vim.api.nvim_command
   command 'echohl ModeMsg'
@@ -54,8 +46,8 @@ local function on_language_status(_, _, result)
   command 'echohl None'
 end
 
--- If the text document version is 0, set it to nil instead so that Neovim
--- won't refuse to update a buffer that it believes is newer than edits.
+-- TextDocument version is reported as 0, override with nil so that
+-- the client doesn't think the document is newer and refuses to update
 -- See: https://github.com/eclipse/eclipse.jdt.ls/issues/1695
 local function fix_zero_version(workspace_edit)
   if workspace_edit and workspace_edit.documentChanges then
@@ -69,12 +61,40 @@ local function fix_zero_version(workspace_edit)
   return workspace_edit
 end
 
+local root_files = {
+  -- Single-module projects
+  {
+    'build.xml', -- Ant
+    'pom.xml', -- Maven
+    'settings.gradle', -- Gradle
+    'settings.gradle.kts', -- Gradle
+  },
+  -- Multi-module projects
+  { 'build.gradle', 'build.gradle.kts' },
+}
+
 configs[server_name] = {
   default_config = {
-    cmd = cmd,
-    cmd_env = {
-      JAR = vim.fn.getenv 'JAR',
-      GRADLE_HOME = vim.fn.getenv 'GRADLE_HOME',
+    cmd = {
+      get_java_executable(),
+      '-Declipse.application=org.eclipse.jdt.ls.core.id1',
+      '-Dosgi.bundles.defaultStartLevel=4',
+      '-Declipse.product=org.eclipse.jdt.ls.core.product',
+      '-Dlog.protocol=true',
+      '-Dlog.level=ALL',
+      '-Xms1g',
+      '-Xmx2G',
+      '-jar',
+      get_jdtls_jar(),
+      '-configuration',
+      get_jdtls_config(),
+      '-data',
+      get_workspace_dir(),
+      '--add-modules=ALL-SYSTEM',
+      '--add-opens',
+      'java.base/java.util=ALL-UNNAMED',
+      '--add-opens',
+      'java.base/java.lang=ALL-UNNAMED',
     },
     filetypes = { 'java' },
     root_dir = function(fname)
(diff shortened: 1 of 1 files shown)


eruizc-dev:
I've been using it all day with no issues whatsoever (in linux), here's my config:

Neovim config:
https://github.com/eruizc-dev/eruizc-dev/blob/jdtls-pr-test/dotfiles/xdg_config_home/nvim/lua/eruizc/lsp.lua#L95

Environment:
https://github.com/eruizc-dev/eruizc-dev/blob/jdtls-pr-test/dotfiles/.zshrc#L79-L81

lua/lspconfig/jdtls.lua
```diff
@@ -129,26 +140,50 @@ configs[server_name] = {
...
+IMPORTANT: If you want all the features jdtls has to offer, [nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls)
+is highly recommended. If all you need is diagnostics, completion, imports, gotos and formatting and some code actions
+you can keep reading here.
+
+For manual installation you can download precompiled binaries from the
+[official downloads site](http://download.eclipse.org/jdtls/snapshots/?d)
 
-Due to the nature of java, the settings for eclipse jdtls cannot be automatically
-inferred. Please set the following environmental variables to match your installation. You can set these locally for your project with the help of [direnv](https://github.com/direnv/direnv). Note version numbers will change depending on your project's version of java, your version of eclipse, and in the case of JDTLS_CONFIG, your OS.
+Due to the nature of java settings cannot be inferred, please set the following
```
mjlbach, line 154: You removed a comma that should probably not have been removed

lua/lspconfig/jdtls.lua
```diff
@@ -129,26 +140,50 @@ configs[server_name] = {
...
+# Optional:
+export JAVA_HOME=/path/to/java_home # In case you don't have java in path or want to use a version in particular
+export WORKSPACE=/path/to/workspace # Defaults to $HOME/workspace
+```
+```lua
+  -- init.lua
+  require'lspconfig'.jdtls.setup{}
 ```
+
+For automatic instllation you can use the following unofficial installers/launchers under your own risk:
```
mjlbach, line 172: instllation

lua/lspconfig/jdtls.lua
```diff
@@ -129,26 +140,50 @@ configs[server_name] = {
...
+  - [jdtls-launcher](https://github.com/eruizc-dev/jdtls-launcher) (Includes lombok support by default)
```
mjlbach, line 173: Until `curl -s https://raw.githubusercontent.com/eruizc-dev/jdtls-launcher/master/install.sh | sudo bash` is removed from that launcher, I can't recommend it.

lua/lspconfig/jdtls.lua
```diff
@@ -71,44 +82,44 @@ end
...
-    root_dir = function(fname)
-      for _, patterns in ipairs(root_files) do
-        local root = util.root_pattern(unpack(patterns))(fname)
-        if root then
-          return root
-        end
-      end
-      return vim.fn.getcwd()
-    end,
+    root_dir = util.root_pattern({
```
mjlbach, line 102: People asked for the mult-root fallback, so I'm concerned about changing it.

lua/lspconfig/jdtls.lua
```diff
@@ -1,61 +1,72 @@
...
---- Callback function for the `language/status` notification.
---
--- The server sends a non-standard notification when the status of the language
--- server changes. This can be used to display progress as the server is
--- starting up.
+  if vim.fn.getenv('JDTLS_HOME') == vim.NIL then
+    error('JDTLS_HOME env variable not set')
+  end
+
+  if vim.fn.has('unix') ~= 0 then
```
mjlbach, line 48: I would use luv's os_uname here instead

lua/lspconfig/jdtls.lua
```diff
@@ -1,61 +1,72 @@
...
+    print('Using old $JDTLS_CONFIG environment variable for JDTLS, refer to documentation and update to JDTLS_HOME')
+    return vim.fn.getenv('JDTLS_CONFIG')
+  end
 
---- Callback function for the `language/status` notification.
---
--- The server sends a non-standard notification when the status of the language
--- server changes. This can be used to display progress as the server is
--- starting up.
+  if vim.fn.getenv('JDTLS_HOME') == vim.NIL then
```
mjlbach, line 44: os.getenv

lua/lspconfig/jdtls.lua
```diff
@@ -1,61 +1,72 @@
...
+  if vim.fn.getenv('JAR') ~= vim.NIL then
+    print('Using old $JAR environment variable for JDTLS, refer to documentation and update to JDTLS_HOME')
+    return vim.fn.getenv('JAR')
+  end
+
+  if vim.fn.getenv('JDTLS_HOME') == vim.NIL then
+    error('JDTLS_HOME env variable not set')
+  end
+
+  return vim.fn.expand('$JDTLS_HOME/plugins/org.eclipse.equinox.launcher_*.jar')
```
mjlbach, line 34: os.getenv + concatenation

mjlbach: Uh, I get you are doing a glob... maybe it's fine

lua/lspconfig/jdtls.lua
eruizc-dev: Fixed

lua/lspconfig/jdtls.lua
eruizc-dev: Fixed

lua/lspconfig/jdtls.lua
```diff
@@ -1,33 +1,35 @@
+local os = require 'os'
```
mjlbach, line 1: this is unecessary

lua/lspconfig/jdtls.lua
```diff
@@ -1,33 +1,35 @@
...
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 local handlers = require 'vim.lsp.handlers'
 
 local server_name = 'jdtls'
 
 local function get_java_executable()
-  local executable = vim.fn.getenv('JAVA_HOME') ~= vim.NIL
-    and vim.fn.expand('$JAVA_HOME/bin/java')
+  local executable = os.getenv('JAVA_HOME') ~= vim.NIL
```
mjlbach, line 10: I would just allocate a local variable for home here. Also you can test against nil, not vim.nil, and simplify this to if home then...

lua/lspconfig/jdtls.lua
eruizc-dev: I didn't saw the getcwd and added it back. I don't see why the multi root would be necessary, are there any issues related to that?

mjlbach: https://github.com/neovim/nvim-lspconfig/issues/455

mjlbach: Also I got asked about this on matrix IIRC and dm'd a few times

lua/lspconfig/jdtls.lua
eruizc-dev: I've been working on this and now sudo is no longer necessary.
https://github.com/eruizc-dev/jdtls-launcher#linux--macos--wsl

lua/lspconfig/jdtls.lua
eruizc-dev: I couldn't find this function anywhere, where should I look for this?

lua/lspconfig/jdtls.lua
eruizc-dev: Thank you for pointing out, didn't know that!

lua/lspconfig/jdtls.lua
mjlbach: With neovim, you have access to all of luv's functions (bindings to libuv) under vim.loop

https://github.com/luvit/luv/blob/master/docs.md#uvos_uname

vim.loop == uv for those examples

mjlbach: You can grep the repo for how we match against windows/other systems

lua/lspconfig/jdtls.lua
eruizc-dev: I couldn't find another way of achieving this without iterate files and doing tons of checks. The filename goes like this: `org.eclipse.equinox.launcher_1.6.200.v20210416-2027.jar`.

mjlbach: it's probably fine, you could list the directory with luv, iterate and :match on the predicates org.eclipse.equinox.launcher to avoid going through vimscript, but it's ok for now to use vim.fn.expand

mjlbach:
A couple high level comments for cleanup:

* store envvars in local variables and reuse those, it makes the code much cleaner
* vim.NIL is not nil in lua land
* I'm still not sure about all of these deprecation notices, we have the breaking changes issue now

eruizc-dev:
I addressed all things you mentioned, let me know what you think!

eruizc-dev:
Sry I had to force push, I forgot something as unstaged in the last commit. It's fixed now and ready for review!

eruizc-dev:
I fixed the lint ci but commit-lint will fail again. Should I rename all commits, squash them or is it okay to leave like this? You can always squash at the moment of merging.

mjlbach:
I do always squash if a PR doesn't follow our commit guidelines, I added the CI recently with the help of dundargoc to encourage users to squash their own commits so I don't have to guesstimate the correct squashed message :)

eruizc-dev:
There ya go, commits squashed!

lua/lspconfig/jdtls.lua
```diff
@@ -1,61 +1,53 @@
...
-    'settings.gradle', -- Gradle
-    'settings.gradle.kts', -- Gradle
-  },
-  -- Multi-module projects
-  { 'build.gradle', 'build.gradle.kts' },
-}
+local function get_java_executable()
+  local executable = env.JAVA_HOME and env.JAVA_HOME .. '/bin/java' or 'java'
+
+  return sysname:match 'Windows' and executable .. '.exe' or executable
```
mfussenegger, line 18: Will the path concattenation above actually work under windows? Before this used `util.path.join` 
(Same also happens in other places)

eruizc-dev: I'm currently doing some tests in Windows, will confirm once it works

eruizc-dev:
Regarding Windows, I managed to get it to execute with the correct params but I get an exit code 13... This is what LspInfo tells me, there are no logs whatsoever.

According to my google search, this relates to Java environment not being properly set up. I've never worked with Java in Windows and there's something I probably did wrong.

image

mfussenegger:
@mjlbach as far as I can tell this now looks okay.

bndbsh:
@eruizc-dev this breaks when jdtls is installed by a package manager. jdtls wants write access to the config directory. Previously, it was possible to set the config directory separately from path to the jar. Now it's mandatory to install jdtls in your home directory in order to use this. This should at least be noted in the instructions, though ideally we'd revert to being able to set the paths separately.

eruizc-dev:
@bndbsh I wasn't aware of any package managers that included jdtls. And I didn't know they wouldn't give you access to config directories. I'll fix that, just tell me what package manager you use!

bndbsh:
@eruizc-dev package managers generally install to a system-wide directory. I installed [this](https://aur.archlinux.org/packages/jdtls/).

I think it might be enough to just allow overriding the config directory with `JDTLS_CONFIG`, and if it's not present keep the current behavior.

mjlbach:
@bndbsh Please open another issue to track this

PR #1087, opened 2021-07-18
Added pascal language server

cryptolake:
From The language server [repo](https://github.com/genericptr/pascal-language-server):

### Features
The implementation is still incomplete.

### Implemented Methods

- textDocument
    - declaration
    - implementation
    - references
    - signatureHelp
    - documentSymbol
        - Only SymbolInformation is implemented
    - documentHighlight
    - completion
- window
    - showMessage
- workspace
    - symbol
- diagnostics

### DIFF
--- a/lua/lspconfig/pasls.lua
+++ b/lua/lspconfig/pasls.lua
@@ -1,0 +1,22 @@
+local configs = require 'lspconfig.configs'
+local util = require 'lspconfig.util'
+
+configs.pasls = {
+	default_config = {
+		cmd = { 'paslsp' },
+		cmd_env = {
+			PP = 'fpc',
+			FPCDIR = '/usr/lib/fpc/src',
+			LAZARUSDIR =  '',
+			FPCTARGET = '',
+			FPCTARGETCPU = ''
+		},
+		filetypes = { 'pascal' },
+		root_dir = util.path.dirname,
+	},
+	docs = [[
+	https://github.com/arjanadriaanse/pascal-language-server
+
+	An LSP server implementation for Pascal variants that are supported by Free Pascal, including Object Pascal. It uses CodeTools from Lazarus as backend.
+	]],
+}
--- a/lua/lspconfig/pasls.lua
+++ b/lua/lspconfig/pasls.lua
@@ -3,20 +3,20 @@ local util = require 'lspconfig/util'
 
 configs.pasls = {
 	default_config = {
-		cmd = { 'paslsp' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/pasls.lua
```diff
@@ -0,0 +1,23 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.pasls = {
+  default_config = {
+    cmd = { 'pasls' },
+    cmd_env = {
+      PP = '/usr/lib/fpc/3.2.2/ppcx64',
```
mjlbach, line 8: This should be documented at the least, because I doubt this paths are portable (especially not to mac and windows)

cryptolake: I added Documentation to everyone of those

lua/lspconfig/pasls.lua
```diff
@@ -0,0 +1,50 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.pasls = {
+  default_config = {
+    cmd = { 'pasls' },
+    cmd_env = {
+      FPCDIR = '/usr/lib/fpc/src',
+      PP = '/usr/lib/fpc/3.2.2/ppcx64',
```
mjlbach, line 9: These paths are too specific. Does the language server really not work without these?

cryptolake: The only Path required for the server to work is `FPCDIR` as i specified in the documentation now.

lithammer: Wouldn't it be better if these were inherited from the parent environment rather than set by us? That feels more flexible to me, and kind of the point of environment variables. So you'd just stick `export FPCDIR='...'` in your `.bashrc` or similar.

cryptolake: I was going to make it that way at first but now that i think about it it's better this way.

lua/lspconfig/pasls.lua
```diff
@@ -0,0 +1,39 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.pasls = {
+  default_config = {
+    cmd = { 'pasls' },
+    filetypes = { 'pascal' },
+    root_dir = util.path.dirname,
```
mjlbach, line 8: this should include at the very least `.git`

cryptolake: added now

PR #1090, opened 2021-07-20
remove: deprecated haskell lsps

mTvare6:
Ghcide, Hie, dhall all are archived and tell people to use haskell language server instead


mjlbach:
Doesn't seem dead? https://github.com/dhall-lang/dhall-haskell/tree/master/dhall-lsp-server

teto:
also maybe display a message "use haskell-language-server instead" instead of just removing hie/ghcide.

mjlbach:
This is still incorrect, the dhall file was removed and dhall added back in the ghcide file... feel free to make a new PR when this is sorted

PR #1091, opened 2021-07-20
feat: add stylua check to make lint

teto:
it is checked on the CI but tehre is no target to reproduce locally


teto:
I just added the Makefile entry, the github action doesn't use it since the action seems to use stylua in all cases. Alternative was to install stylua via cargo instead and call the make target but looks like to much work for just a `stylua --check . ` call

PR #1094, opened 2021-07-22
feat: add vdmj language server

donbex:
This PR adds a configuration for the VDMJ language server. Below I highlighted a couple of small issues I would like some help with before merging.

I tested this configuration on Arch Linux and MacOS Catalina, but I have no way to test it on Windows. I expect some minor modification will be needed to make it work on Windows.

Also, for this to work properly https://github.com/neovim/neovim/commit/1dab9357de586b08431175f19a1b44bb294866d6 is needed. What is the appropriate way to mention this in the docstring (if needed at all)?

### DIFF
--- a/lua/lspconfig/vdmj.lua
+++ b/lua/lspconfig/vdmj.lua
@@ -1,0 +1,135 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+if vim.fn.has 'nvim-0.5.1' == 0 then
+  vim.notify('The VDMJ language server requires nvim > 0.5', vim.log.levels.ERROR)
+  return
+end
+
+local server_name = 'vdmj'
+
+local mavenrepo = util.path.join(vim.env.HOME, '.m2', 'repository', 'com', 'fujitsu')
+
+local function get_jar_path(config, package, version)
+  return util.path.join(config.options.mavenrepo, package, version, package .. '-' .. version .. '.jar')
+end
+
+local function with_precision(version, is_high_precision)
+  return is_high_precision and version:gsub('([%d.]+)', '%1-P') or version
+end
+
+local function get_latest_installed_version(repo)
+  local path = util.path.join(repo, 'lsp')
+  local sort = vim.fn.sort
+
+  local subdirs = function(file)
+    local stat = vim.loop.fs_stat(util.path.join(path, file))
+    return stat.type == 'directory' and 1 or 0
+  end
+
+  local candidates = vim.fn.readdir(path, subdirs)
+  local sorted = sort(sort(candidates, 'l'), 'N')
+  return sorted[#sorted]
+end
+
+local function find_vscode_ancestor(startpath)
+  return util.search_ancestors(startpath, function(path)
+    if util.path.is_dir(util.path.join(path, '.vscode')) then
+      return path
+    end
+  end)
+end
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'java' },
+    filetypes = { 'vdmsl', 'vdmpp', 'vdmrt' },
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname) or find_vscode_ancestor(fname) or util.path.dirname(fname)
+    end,
+    options = {
+      java = vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java',
+      java_opts = { '-Xmx3000m', '-Xss1m' },
+      annotation_paths = {},
+      mavenrepo = mavenrepo,
+      version = get_latest_installed_version(mavenrepo),
+      logfile = util.path.join(vim.fn.stdpath 'cache', 'vdm-lsp.log'),
+      debugger_port = -1,
+      high_precision = false,
+    },
+  },
+  docs = {
+    description = [[
+https://github.com/nickbattle/vdmj
+
+The VDMJ language server can be installed by cloning the VDMJ repository and
+running `mvn clean install`.
+
+Various options are provided to configure the language server (see below). In
+particular:
+- `annotation_paths` is a list of folders and/or jar file paths for annotations
+that should be used with the language server;
+- any value of `debugger_port` less than zero will disable the debugger; note
+that if a non-zero value is used, only one instance of the server can be active
+at a time.
+
+More settings for VDMJ can be changed in a file called `vdmj.properties` under
+`root_dir/.vscode`. For a description of the available settings, see
+[Section 7 of the VDMJ User Guide](https://raw.githubusercontent.com/nickbattle/vdmj/master/vdmj/documentation/UserGuide.pdf).
+
+Note: proof obligations and combinatorial testing are not currently supported
+by neovim.
+]],
+    default_config = {
+      cmd = 'Generated from the options given',
+      root_dir = "Git root or vim's starting directory",
+      options = {
+        java = '$JAVA_HOME/bin/java',
+        java_opts = { '-Xmx3000m', '-Xss1m' },
+        annotation_paths = {},
+        mavenrepo = '$HOME/.m2/repository/com/fujitsu',
+        version = 'The latest version installed in `mavenrepo`',
+        logfile = "path.join(vim.fn.stdpath 'cache', 'vdm-lsp.log')",
+        debugger_port = -1,
+        high_precision = false,
+      },
+    },
+  },
+  on_new_config = function(config, root_dir)
+    local version = with_precision(
+      config.options.version or get_latest_installed_version(config.options.mavenrepo),
+      config.options.high_precision
+    )
+
+    local classpath = table.concat({
+      get_jar_path(config, 'vdmj', version),
+      get_jar_path(config, 'annotations', version),
+      get_jar_path(config, 'lsp', version),
+      util.path.join(root_dir, '.vscode'),
+      unpack(config.options.annotation_paths),
+    }, ':')
+
+    local java_cmd = {
+      config.options.java,
+      config.options.java_opts,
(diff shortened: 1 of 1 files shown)


lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,99 @@
...
+    },
+  },
+  on_new_config = function(config)
+    local version = get_version(config)
+    config.cmd = {
+      config.options.java,
+      -- TODO: how to unpack java_opts here? unpack(config.options.java_opts)
+      -- results in only the first option being used.
+      '-Xmx3000m',
+      '-Xss1m',
```
donbex, line 84: As per comment, I would like to give the user the ability to override the default Java options.

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,99 @@
...
+      table.concat({
+        getJarPath(config, 'vdmj', version),
+        getJarPath(config, 'annotations', version),
+        getJarPath(config, 'lsp', version),
+      }, ':'),
+      'lsp.LSPServerStdio',
+      '-' .. config.options.dialect,
+      '-dap',
+      -- TODO: should be unique for each instance
+      tostring(config.options.debugger_port),
```
donbex, line 96: A new server will fail to start if the port is already in use. Is there a way to increment the port number either for each new configuration or if the port is already in use?

donbex:
By the way, is there a way to replace the generated "Snippet to enable the language server" with
```lua
require'lspconfig'.vdmj.setup {
  options = {
    version = '4.4.3-SNAPSHOT',
    dialect = 'vdmsl',
    ...
  }
  ...
}
```
or at least to suppress its generation altogether? Since it isn't possible to provide sensible defaults for `version` and `dialect`, those options are required during setup, so `require'lspconfig'.vdmj.setup{}` will fail.

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,106 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'vdmj'
+
+local mavenrepo = util.path.join(tostring(vim.fn.getenv 'HOME'), '/.m2/repository/com/fujitsu')
```
lithammer, line 6: Use `vim.env` instead of `vim.fn.getenv`.

```suggestion
local mavenrepo = util.path.join(vim.env.HOME, '/.m2/repository/com/fujitsu')
```

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,106 @@
...
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'java' },
+    filetypes = { 'vdmsl', 'vdmpp', 'vdmrt' },
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname) or vim.fn.getcwd()
+    end,
+    options = {
+      java = (tostring(vim.fn.getenv 'JAVA_HOME') or '/usr') .. '/bin/java',
```
lithammer, line 38:
```suggestion
      java = (vim.env.JAVA_HOME or '/usr') .. '/bin/java',
```

lithammer:
> Also, for this to work properly [neovim/neovim@1dab935](https://github.com/neovim/neovim/commit/1dab9357de586b08431175f19a1b44bb294866d6) is needed. What is the appropriate way to mention this in the docstring (if needed at all)?

This is a tough one. I'm quite sure we'll get bug reports about this from people running v0.5. Could this be handled in nvim-lspconfig somehow?

donbex:
> Could this be handled in nvim-lspconfig somehow?

The only thing I can think of is hiding the configuration behind a version check and throwing an error message if the version requirement isn't met. I don't know what the most appropriate way to achieve this would be, though.

lua/lspconfig/vdmj.lua
lithammer: Usually binding to 0 picks a random available port. Not sure if that's the case here though. Otherwise I think the only way is to attempt to bind to a port (see `:h tcp-server` for examples).

Is it necessary/important to bind to a port at all?

donbex: > Usually binding to 0 picks a random available port. Not sure if that's the case here though.

It is not the case at the moment with the VDMJ LS, unfortunately.

> Is it necessary/important to bind to a port at all?

Binding a port is only necessary to start the debugger adapter with the LS. It is not necessary if the DA is not needed.

mjlbach:
The version check is fine to me, just do an early return and print an error message at the top of the file

lua/lspconfig/vdmj.lua
donbex: > Usually binding to 0 picks a random available port.

Based on this, I changed the configuration to disable the debugger for negative port values. I don't know whether it is better to have the logic for using a random port in the configuration or in the LS itself, so I think it is best to leave it out for the moment.

donbex:
@lithammer I have added the version check.

lithammer (approved):
Looks good. Just squash the commits and we'll merge.

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,127 @@
...
+local util = require 'lspconfig/util'
+
+if vim.fn.has 'nvim-0.5.1' == 0 then
+  vim.notify('The VDMJ language server requires nvim > 0.5', vim.log.levels.ERROR)
+  return
+end
+
+local server_name = 'vdmj'
+
+local mavenrepo = util.path.join(vim.env.HOME, '/.m2/repository/com/fujitsu')
```
mjlbach, line 11: let's not do '/.m2/repository/com/fujitsu', but split this up into individual arguments to path join

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,127 @@
...
+  local sorted = sort(sort(candidates, 'l'), 'N')
+  return sorted[#sorted]
+end
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'java' },
+    filetypes = { 'vdmsl', 'vdmpp', 'vdmrt' },
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname) or vim.fn.getcwd()
```
mjlbach, line 40: remove cwd

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,127 @@
...
+    end,
+    options = {
+      java = (vim.env.JAVA_HOME or '/usr') .. '/bin/java',
```
mjlbach, line 43: shouldn't we use util.path.join here as well?

lua/lspconfig/vdmj.lua
```diff
@@ -0,0 +1,127 @@
...
+at a time.
+
+More settings for VDMJ can be changed in a file called `vdmj.properties` under
+`root_dir/.vscode`. For a description of the available settings, see
+[Section 7 of the VDMJ User Guide](https://raw.githubusercontent.com/nickbattle/vdmj/master/vdmj/documentation/UserGuide.pdf).
+
+Note: proof obligations and combinatorial testing are not currently supported
+by neovim.
+]],
+    default_config = {
```
mjlbach, line 75: You don't need defualt_config under documentation

donbex: I think I do, since a couple of options aren't really meaningful without a custom description:
- https://github.com/neovim/nvim-lspconfig/blob/72e5b3f3952fa29d45a28a12257b6a6faf8dfc8f/lua/lspconfig/vdmj.lua#L47 will not have a meaningful value unless VDMJ is installed on the system
- https://github.com/neovim/nvim-lspconfig/blob/72e5b3f3952fa29d45a28a12257b6a6faf8dfc8f/lua/lspconfig/vdmj.lua#L37 is not representative of the actual default command generated in `on_new_config`

lua/lspconfig/vdmj.lua
donbex: This was a conscious choice: VDM specifications are often enough distributed as-is, outside of a Git repository -- at the very least, that is the case at my workplace, where we have a mix of both.

I am happy to change this if there is a better solution, but `util.find_git_ancestor(fname)` (or looking for some sort of build files) is not enough.

lua/lspconfig/vdmj.lua
mjlbach: Just put that under documentation then, as it stands you are repeating options that are auto-populated.

lua/lspconfig/vdmj.lua
mjlbach: You should remove cwd, and suggest users add util.path.dirname(filename) in the instructions then.

cwd will lead to highly undesirable behavior (you open neovim in $HOME, open a subfolder the server indexes $HOME slowing down the entire computer. you open neovim in a subfolder of a project, open a file in a parent folder, neovim cannot index that file)

lua/lspconfig/vdmj.lua
donbex: If I leave only those two descriptions in the `default_config` table under documentation, then no other values are auto-populated and the other options do not appear in `CONFIG.md`, which is why I manually copied them over. I tried looking through `docgen.lua`, but I don't quite understand it and I can't figure out if there is some other way to direct it to generate the documentation for the other options.

lua/lspconfig/vdmj.lua
donbex: While there are no "standard" files that would mark the root of a VDM specification, the VDMJ language server stores some (optional) configuration under the `.vscode` subfolder of the workspace root. Therefore I changed this to return either the Git ancestor, the VSCode ancestor, or the current file's directory, in order of preference.

PR #1097, opened 2021-07-22
Respect the setting of $FLUTTER_SDK

CodeLongAndProsper90:
On some systems, notably, NixOS, the Flutter SDK does not reside with the executable.
`$FLUTTER_SDK` is used to set a new path for the SDK, so this will search there first.
This solves the workaround of manually setting `cmd` in the config for dartls.

### DIFF
--- a/lua/lspconfig/dartls.lua
+++ b/lua/lspconfig/dartls.lua
@@ -5,13 +5,13 @@ local server_name = 'dartls'
 local bin_name = 'dart'
 
 local find_dart_sdk_root_path = function()
-  if os.getenv 'FLUTTER_SDK' ~= nil then
+  if os.getenv 'FLUTTER_SDK' then
     local flutter_path = os.getenv 'FLUTTER_SDK'
-    return util.path.path_join(flutter_path, '/cache/dart-sdk/bin/dart')
+    return util.path.path_join(flutter_path, 'cache', 'dart-sdk', 'bin', 'dart')
   elseif vim.fn['executable'] 'flutter' == 1 then
     local flutter_path = vim.fn['resolve'](vim.fn['exepath'] 'flutter')
     local flutter_bin = vim.fn['fnamemodify'](flutter_path, ':h')
-    return flutter_bin .. '/cache/dart-sdk/bin/dart'
+    return util.path.path_join(flutter_bin, 'cache', 'dart-sdk', 'bin', 'dart')
   elseif vim.fn['executable'] 'dart' == 1 then
     return vim.fn['resolve'](vim.fn['exepath'] 'dart')
   else
@@ -21,7 +21,7 @@ end
 
 local analysis_server_snapshot_path = function()
   local dart_sdk_root_path = vim.fn['fnamemodify'](find_dart_sdk_root_path(), ':h')
-  local snapshot = dart_sdk_root_path .. '/snapshots/analysis_server.dart.snapshot'
+  local snapshot = util.path.path_join(dart_sdk_root_path, 'snapshots', 'analysis_server.dart.snapshot')
 
   if vim.fn['has'] 'win32' == 1 or vim.fn['has'] 'win64' == 1 then
     snapshot = snapshot:gsub('/', '\\')
--- a/lua/lspconfig/dartls.lua
+++ b/lua/lspconfig/dartls.lua
@@ -5,7 +5,10 @@ local server_name = 'dartls'
 local bin_name = 'dart'
 
 local find_dart_sdk_root_path = function()
-  if vim.fn['executable'] 'flutter' == 1 then
+  if os.getenv 'FLUTTER_SDK' ~= nil then
+    local flutter_path = os.getenv 'FLUTTER_SDK'
+    return util.path.path_join(flutter_path, '/cache/dart-sdk/bin/dart')
+  elseif vim.fn['executable'] 'flutter' == 1 then
     local flutter_path = vim.fn['resolve'](vim.fn['exepath'] 'flutter')
     local flutter_bin = vim.fn['fnamemodify'](flutter_path, ':h')
     return flutter_bin .. '/cache/dart-sdk/bin/dart'


lua/lspconfig/dartls.lua
```diff
@@ -5,7 +5,10 @@ local server_name = 'dartls'
 local bin_name = 'dart'
 
 local find_dart_sdk_root_path = function()
-  if vim.fn['executable'] 'flutter' == 1 then
+  if os.getenv 'FLUTTER_SDK' ~= nil then
+    local flutter_path = os.getenv 'FLUTTER_SDK'
+    return flutter_path .. '/cache/dart-sdk/bin/dart'
```
mjlbach, line 10: this is not windows compatible, we have some path utils for joining

CodeLongAndProsper90: I was going off line 14, where it also concatenates a string to append to the path.

mjlbach: Yes, I didn't write the original implementation so I just see this as an opportunity for some cleanup :)

CodeLongAndProsper90: Okay, working on it now

mjlbach: https://github.com/neovim/nvim-lspconfig/blob/4f72377143fc0961391fb0e42e751b9f677fca4e/lua/lspconfig/util.lua#L140-L143

mjlbach: that's not quite it :) You need to separate out the directories into a table.

CodeLongAndProsper90:
I'm trying to rebase all these commits and fix their names but it's just generating new ones...

mjlbach:
(This is an automated reply)

Please squash your commits, this ensures a clean git history . If you're not familiar with how to do this, you can do the following:

Assuming you are on a feature branch, use interactive rebase on the master branch:

`git rebase -i master`

You should see something like the following:

```
pick 2fad63b Add Crystalline (Crystal)
pick 1917167 Fix wrong root_uri
pick 5ea0995 clangd: "can’t" -> "cannot" in switchSourceHeader failure message
pick 30cf918 Update issue template
pick bcfca98 fix(docgen): remove codeql package.json

# Rebase 29ea30f..bcfca98 onto 29ea30f (5 commands)
#
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup <commit> = like "squash", but discard this commit's log message
# x, exec <command> = run command (the rest of the line) using shell
# b, break = stop here (continue rebase later with 'git rebase --continue')
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
# .       create a merge commit using the original merge commit's
# .       message (or the oneline, if no original merge commit was
# .       specified). Use -c <commit> to reword the commit message.
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
```
For each commit **that is not the first commit you have added**, change `pick` to `squash`. For the first commit, ensure the commit message follows the [conventional commits guidelines](https://www.conventionalcommits.org/en/v1.0.0/) 

```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```

mjlbach:
forgot I had that saved :)

CodeLongAndProsper90:
It's helpful all the same- I've got it all squashed but I can't can't push because it's not a fast-foward and pulling overwrites all my squashing :/

mjlbach:
it's ok to force push in this case :)

CodeLongAndProsper90:
Oh right, just had to prepend a `+`

CodeLongAndProsper90:
Oh and now commitlint doesn't like the big merge message

lua/lspconfig/dartls.lua
```diff
@@ -5,7 +5,10 @@ local server_name = 'dartls'
 local bin_name = 'dart'
 
 local find_dart_sdk_root_path = function()
-  if vim.fn['executable'] 'flutter' == 1 then
+  if os.getenv 'FLUTTER_SDK' ~= nil then
+    local flutter_path = os.getenv 'FLUTTER_SDK'
+    return util.path.path_join(flutter_path, '/cache/dart-sdk/bin/dart')
```
mjlbach, line 10: Still needs to change

CodeLongAndProsper90: Change how?

mjlbach: `util.path.path_join(flutter_path, 'cache', 'dart-sdk', 'bin', 'dart')`

lua/lspconfig/dartls.lua
```diff
@@ -5,7 +5,10 @@ local server_name = 'dartls'
...
+  elseif vim.fn['executable'] 'flutter' == 1 then
     local flutter_path = vim.fn['resolve'](vim.fn['exepath'] 'flutter')
     local flutter_bin = vim.fn['fnamemodify'](flutter_path, ':h')
     return flutter_bin .. '/cache/dart-sdk/bin/dart'
```
mjlbach, line 14: same here: `util.path.path_join(flutter_path, 'cache', 'dart-sdk', 'bin', 'dart')`

lua/lspconfig/dartls.lua
lithammer, line 10: Use `vim.env.FLUTTER_SDK` instead of `os.getenv('FLUTTER_SDK')`.

mjlbach: why?

lithammer: Because they're already defined, while `os.getenv` needs to do a lookup.

mjlbach: It's basically free, I would generally prefer avoiding calling vimscript functions if possible.

mjlbach:
You tested this and everything works?

PR #1101, opened 2021-07-25
ci: run codespell on pull requests

dundargoc:
I've added two files related to codespell:

- `.codespellignorewords`: if you want to disable a word globally then just add it here.
- `.codespellignorelines`: if you want to stop codespell from warning in one specific instance, but not globally, then add the entire string (aka the full line) to ignore it. I don't know how to explain it well but just check the file and I think you'll get what I mean.

It's also possible to ignore files/directories, but that has to be done with the `--skip` flag to codespell.

### DIFF
--- a/.codespellignorewords
+++ b/.codespellignorewords
@@ -1,0 +1,3 @@
+als
+edn
+esy
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,21 @@
+name: codespell
+on: [pull_request]
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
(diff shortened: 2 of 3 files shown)


dundargoc:
I've added two files related to codespell:

- `.codespellignorewords`: if you want to disable a word globally then just add it here.
- `.codespellignorelines`: if you want to stop codespell from warning in one specific instance, but not globally, then add the entire string (aka the full line) to ignore it. I don't know how to explain it well but just check the file and I think you'll get what I mean.

It's also possible to ignore files/directories, but that has to be done with the `--skip` flag to codespell.

mjlbach:
We should probably ignore CONFIG.md since it's auto-generated. Can you fix the couple typos it detected? Thanks for doing this :)

dundargoc:
Done. Check the .codespellignorewords if there are any words that *shouldn't* be ignored. I've also removed .codespellignorelines for now since there didn't seem any need for ignoring words locally.

PR #1108, opened 2021-07-27
update ccls root pattern

abgeana:
CCLS searches for a `.ccls` file to identify the root folder for a project. This patch adds `.ccls` to the root pattern in `ccls.lua`.

### DIFF
--- a/lua/lspconfig/ccls.lua
+++ b/lua/lspconfig/ccls.lua
@@ -6,7 +6,8 @@ configs.ccls = {
     cmd = { 'ccls' },
     filetypes = { 'c', 'cpp', 'objc', 'objcpp' },
     root_dir = function(fname)
-      return util.root_pattern('compile_commands.json', 'compile_flags.txt', '.git')(fname) or util.path.dirname(fname)
+      return util.root_pattern('compile_commands.json', '.ccls', 'compile_flags.txt', '.git')(fname)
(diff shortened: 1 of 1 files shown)


lithammer:
You have a couple of lint errors.

abgeana:
Updated the patch to fix the lua lint errors.

There were also a few git commit lint errors, but not sure what they were about.

lithammer:
> There were also a few git commit lint errors, but not sure what they were about.

The commit message needs to follow the 'conventional commit' format. So something like this:

    fix(ccls): Add .ccls as root pattern

    CCLS searches for a .ccls file to identify the root folder for a project. This
    patch adds .ccls to the root pattern in ccls.lua.

https://github.com/conventional-changelog/commitlint/#what-is-commitlint

lithammer:
Not entirely sure why the commit linter is still complaining. Possibly that it has to be all lowercase...?

```diff
-fix(ccls): Add .ccls as root pattern
+fix(ccls): add .ccls as root pattern
```

CC @mjlbach

mjlbach:
I think it's the capitalization as well (sentence-case?)

clason:
yes, has to be lower case

abgeana:
Sorry for my delayed reaction. I updated the commit message.

PR #1114, opened 2021-07-30
Set Ctrl-] as jump-to-definition as well

mcepl:
`Ctrl-]` is a standard vim shortcut for jumping to definition, so it seems like a good idea to add it as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ local on_attach = function(client, bufnr)
   -- See `:help vim.lsp.*` for documentation on any of the below functions
   buf_set_keymap('n', 'gD', '<Cmd>lua vim.lsp.buf.declaration()<CR>', opts)
   buf_set_keymap('n', 'gd', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
+  buf_set_keymap('n', '<C-]>', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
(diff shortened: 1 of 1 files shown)


gegoune:
I am bit fan of those principles https://github.com/nanotee/nvim-lsp-basics/issues/1 which your proposal violates.

mcepl:
> I am bit fan of those principles [nanotee/nvim-lsp-basics#1](https://github.com/nanotee/nvim-lsp-basics/issues/1) which your proposal violates.

Yes, we just have to agree to disagree, and let actual maintainers of this package to decide. I think that [using your finger memory](https://web.archive.org/web/20130703010939/http://www.oreilly.com/news/zenclavier_1299.html) is more important than an elusive possibility of using tags while using LSP (that seems really weird).

polyzen:
An alternative consideration is https://github.com/weilbith/nvim-lsp-smag which aims "to simplify the whole process of either using a tag file or the `vim.lsp.client`."

teto:
I am one of those that uses both on a same codebase (tags are faster to be generated and LSP servers can be capricious as well, unavailable, crashing etc) so I wouldn't like the override. https://github.com/weilbith/nvim-lsp-smag looks like a good solution.

justinmk:
The readme is just an example and the `gd` example already shows the idea.

A more general solution for a builtin set of default mappings is a potential idea for core in the future but is a different topic.

PR #1124, opened 2021-08-05
Fix julials

uncomfyhalomacro:
I noticed that the default config for julials, `server_path` is not a path but a file, `print(Base.find_package("LanguageServer"))` prints `"/home/uncomfy/.julia/packages/LanguageServer/JrIEf/src/LanguageServer.jl"` - which is a file. I edited it to `print(dirname(something(Base.current_project(Base.find_package("LanguageServer")))))` so it will give `"/home/uncomfy/.julia/packages/LanguageServer/JrIEf"`. If this small change is not merged, the server will not run due to an error because of `server_path` needs to be a path, and not a file.

### DIFF
--- a/lua/lspconfig/julials.lua
+++ b/lua/lspconfig/julials.lua
@@ -9,7 +9,7 @@ local cmd = {
   [[
     using Pkg
     Pkg.activate(Base.current_project())
-    using LanguageServer, SymbolServer, StaticLint
+    using LanguageServer
     depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
     project_path = let
         dirname(something(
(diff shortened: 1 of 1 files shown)


ReneganRonin:
Also noticed that you need to import SymbolServer to main according to this:
> Note
>
>  Due to a current bug
>  (https://github.com/julia-vscode/LanguageServer.jl/issues/750), SymbolServer must
> be imported into Main.

So I edited a few things.

clason:
Did you read through that issue? The latest comment refers to a (merged) PR that fixes the issue in `SymbolServer.jl`, so this workaround is no longer required.

See https://github.com/neovim/nvim-lspconfig/pull/1079 for (a lot of) previous discussion.

ReneganRonin:
> Did you read through that issue? The latest comment refers to a (merged) PR that fixes the issue in `SymbolServer.jl`, so this workaround is no longer required.
> 
> See #1079 for (a lot of) previous discussion.

oh my bad i never did. i should have

PR #1138, opened 2021-08-11
Update docs description for sumneko_lua

MrAdityaAlok:
- Add instructions for termux installation


mjlbach:
I don't think this is a big enough pool of users to warrant a separate installation instructions, and it isn't really lspconfig specific.

MrAdityaAlok:
Okay closing then.

PR #1139, opened 2021-08-12
feat: add pyre language server.

nemith:
Pyre is yet another static type checker for python. https://pyre-check.org/

### DIFF
--- a/lua/lspconfig/pyre.lua
+++ b/lua/lspconfig/pyre.lua
@@ -1,0 +1,23 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.pyre = {
+  default_config = {
+    cmd = { 'pyre', 'persistent' },
+    filetypes = { 'python' },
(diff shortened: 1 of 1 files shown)


mjlbach:
Thanks! Can you add some installation instructions, and possibly copy some of the root detection logic from pyright?

mjlbach:
Currently when I run this, I get
```
LSP[pyre]: Error NO_RESULT_CALLBACK_FOUND: {                               
  id = 0,                                                                  
  jsonrpc = "2.0",                                                         
  result = {                                                               
    capabilities = {}                                                      
  }                                                                        
}
```

Which indicates pyre might be doing something nonstandard, need to look into it

nemith:
I have only tested this on pyre initialized repositories (i.e pyre init) has been ran.  I haven't tested at all on code that isn't initialized and may just fail like above.  Also why i didn't include other root detection logic.

mjlbach:
Yes, I'm having some issues getting it to work even after initializing pyre for a project. Missing imports, dying language server, etc... Pyright works fine with the same project.

PR #1140, opened 2021-08-13
Add glob for graphqlrc detection

keith:
This file can have a ton of different extensions that are supported https://graphql-config.com/introduction


keith:
FYI this says to read the PR template but AFAICT there is not one?

PR #1141, opened 2021-08-13
feat(graphql): add more supported config files

keith:
This file can have a ton of different extensions that are supported
https://graphql-config.com/introduction

### DIFF
--- a/lua/lspconfig/graphql.lua
+++ b/lua/lspconfig/graphql.lua
@@ -8,7 +8,7 @@ configs[server_name] = {
   default_config = {
     cmd = { bin_name, 'server', '-m', 'stream' },
     filetypes = { 'graphql' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/graphql.lua
```diff
@@ -8,7 +8,7 @@ configs[server_name] = {
   default_config = {
     cmd = { bin_name, 'server', '-m', 'stream' },
     filetypes = { 'graphql' },
-    root_dir = util.root_pattern('.git', '.graphqlrc'),
+    root_dir = util.root_pattern('.git', '.graphqlrc*'),
```
lithammer, line 11: Seems like it would make sense to add `graphql.config.*` as well according to the link the pull request description.

PR #1144, opened 2021-08-14
fix: lua/ folder in default root pattern

pedro757:
I think this should be the default root pattern for sumneko


mjlbach:
I don't agree with this, you could have an arbitrarily nested lua directory in a project, and this would set this as project root.

PR #1148, opened 2021-08-15
[jdtls] Use subdirectories in workspace directory

tors42:
If one launches multiple eclipse.jdt.ls servers for different projects,
they should use unique data directories to not interfere with eachother.
This commit adds a project subdirectory in the workspace directory,
based on the directory name of the project.

This allows starting multiple servers without them overwriting the data of the previously launched servers.

### DIFF
--- a/lua/lspconfig/jdtls.lua
+++ b/lua/lspconfig/jdtls.lua
@@ -19,7 +19,9 @@ local function get_java_executable()
 end
 
 local function get_workspace_dir()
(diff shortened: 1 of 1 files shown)


mjlbach:
We shouldn't merge anything that has a dependency on cwd, as this is brittle

tors42:
Ah, I see.
My skills are limited in this area so I will probably not be able to propose an alternative solution. 

Thanks for the quick feedback!

PR #1150, opened 2021-08-16
fix(pyright): use pyrightconfig.json like sign of root directory 

defntvdm:


lua/lspconfig/pyright.lua
```diff
@@ -13,6 +13,7 @@ local root_files = {
   'setup.cfg',
   'requirements.txt',
   '.git',
+  'pyrightconfig.json',
```
lithammer, line 16: Not that it matters much because it's a breadth-first search, but I think you should put it before `.git`.

lithammer:
You're gonna have to update the commit message to make commitlint happy.

https://github.com/conventional-changelog/commitlint/#what-is-commitlint

lua/lspconfig/pyright.lua
defntvdm: thanks, done :)

defntvdm:
Should I squash commits and rename it like PR?

lithammer:
Thanks!

PR #1151, opened 2021-08-16
[denols] Fix root_dir

e-kwsm:
At the moment denols may [fail to start since root directory is not detected](https://github.com/neovim/nvim-lspconfig/blob/b4661ec91e4675a339289d5662a93e646aeabe6f/lua/lspconfig/configs.lua#L71-L73).

I referred to
https://github.com/neovim/nvim-lspconfig/blob/b4661ec91e4675a339289d5662a93e646aeabe6f/lua/lspconfig/pylsp.lua#L8-L17

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -103,7 +103,13 @@ configs[server_name] = {
   default_config = {
     cmd = { 'deno', 'lsp' },
     filetypes = { 'javascript', 'javascriptreact', 'javascript.jsx', 'typescript', 'typescriptreact', 'typescript.tsx' },
-    root_dir = util.root_pattern('package.json', 'tsconfig.json', '.git'),
+    root_dir = function(fname)
+      local root_files = {
+        'package.json',
+        'tsconfig.json',
+      }
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname) or util.path.dirname(fname)
+    end,
     init_options = {
       enable = true,
       lint = false,


lua/lspconfig/denols.lua
```diff
@@ -103,7 +103,14 @@ configs[server_name] = {
   default_config = {
     cmd = { 'deno', 'lsp' },
     filetypes = { 'javascript', 'javascriptreact', 'javascript.jsx', 'typescript', 'typescriptreact', 'typescript.tsx' },
-    root_dir = util.root_pattern('package.json', 'tsconfig.json', '.git'),
+    root_dir = function(fname)
+      local root_files = {
+        'package.json',
+        'tsconfig.json',
+        '.git',
```
lithammer, line 110: This `.git` is redundant since you have `find_git_ancestor` below.

mjlbach:
I'm not in favor of adding util.path.dirname as a fallback, given most deno projects should have a well defined structure. As otherwise noted, you added a redundant git check

lua/lspconfig/denols.lua
e-kwsm: Thank you for your suggestion.

e-kwsm:
Without this modification, deno doesn’t start when a user writes the very first code on a project or a snippet, which is a bit inconvenient.
The behaviour is consistent with pylsp.

mjlbach:
That's not true, the server will not start if the user does not have a root directory which is a different matter.

e-kwsm:
I’m not sure why other servers such as pylsp and clangd are allowed to start under the cwd.

Is there something disadvantageous?

mjlbach:
Because the configs are community contributed, and I wasn't historically as strict about undefined behavior as I'm being now.

I'll be removing those fallbacks shortly.

oblitum:
@e-kwsm

> Is there something disadvantageous?

* https://github.com/scalameta/metals/discussions/3079#discussioncomment-1255446
   * https://github.com/scalameta/metals-feature-requests/issues/231
* https://github.com/neovim/neovim/pull/15430#issuecomment-906179810

The spec asks for LSP clients to send `null` when a workspace root can't  be detected/configured, so that servers acknowledge that fact and possibly provide services without a workspace (tl;dr, workspaces are not a requirement in the protocol).

> At the moment denols may fail to start since root directory is not detected.

Clients shouldn't be expected to embed heuristics to enforce the existence of a workspace root. It's denols's job to work or not without a workspace (e.g. https://github.com/scalameta/metals-feature-requests/issues/231).

Issue #1106, opened 2021-07-26
Julia LanguageServer.jl lsp throws Missing reference Warning


SirLukeSchande:
- `nvim --version`:
`NVIM v0.5.0
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/config -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/src -I/usr/local/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/src/nvim/auto -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/include
Übersetzt von brew@BigSur`
- nvim-lspconfig version(commit hash): `b8ab3b84e9ebff631136b5c3e0f5ed3a17c1f3ca`
- What language server (If the problem is related to a specific language server): julials
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): no
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): no
- Is the problem isolated to a particular language server: yes
- Operating system/version: macos big Sur

After startup of the julia language server I get a Missing Reference Warning for every Package I import including Standard Library packages like Dates.jl. I tried it both in the current stable as well as neovim nightliy v0.6. 

The problem first emerged after updating Julia to version 1.6.2. Both the standard configuration and a manual configuration using a dedicated run.jl file together with dedicated Project.toml and Manifest.toml. 

[...]

PR #1153, opened 2021-08-17
[julia]: Better configurability for LanguageServer.jl install location.

fredrikekre:
I think after this patch, no user configuration is necessary unless you want to have LanguageServer.jl installed into a custom package environment.

### DIFF
--- a/lua/lspconfig/julials.lua
+++ b/lua/lspconfig/julials.lua
@@ -7,9 +7,15 @@ local cmd = {
   '--history-file=no',
   '-e',
   [[
-    using Pkg
-    Pkg.instantiate()
+    # Load LanguageServer.jl: attempt to load from ~/.julia/environments/nvim-lspconfig
+    # with the regular load path as a fallback
+    ls_install_path = joinpath(
+        get(DEPOT_PATH, 1, joinpath(homedir(), ".julia")),
+        "environments", "nvim-lspconfig"
+    )
+    pushfirst!(LOAD_PATH, ls_install_path)
     using LanguageServer
+    popfirst!(LOAD_PATH)
     depot_path = get(ENV, "JULIA_DEPOT_PATH", "")
     project_path = let
         dirname(something(
@@ -51,13 +57,20 @@ configs.julials = {
     description = [[
 https://github.com/julia-vscode/julia-vscode
 
-`LanguageServer.jl` can be installed with `julia` and `Pkg`:
+LanguageServer.jl can be installed with `julia` and `Pkg`:
 ```sh
-julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
+julia --project=~/.julia/environments/nvim-lspconfig -e 'using Pkg; Pkg.add("LanguageServer")'
 ```
-This installs LanguageServer.jl into your global julia environment.
+where `~/.julia/environments/nvim-lspconfig` is the location where
+the default configuration expects LanguageServer.jl to be installed.
 
-In order to have LanguageServer.jl pick up installed packages or dependencies in a Julia project, you must first instantiate the project:
+To update an existing install, use the following command:
+```sh
+julia --project=~/.julia/environments/nvim-lspconfig -e 'using Pkg; Pkg.update()'
+```
+
+Note: In order to have LanguageServer.jl pick up installed packages or dependencies in a
+Julia project, you must make sure that the project is instantiated:
 ```sh
 julia --project=/path/to/my/project -e 'using Pkg; Pkg.instantiate()'
 ```


lua/lspconfig/julials.lua
```diff
@@ -38,9 +46,6 @@ local cmd = {
 configs.julials = {
   default_config = {
     cmd = cmd,
-    on_new_config = function(new_config, root_dir)
-      new_config.cmd_cwd = root_dir
```
fredrikekre, line 42: When is this necessary? I couldn't find a case where it was necessary.

lua/lspconfig/julials.lua
```diff
@@ -51,13 +56,31 @@ configs.julials = {
...
-`LanguageServer.jl` can be installed with `julia` and `Pkg`:
+LanguageServer.jl can be installed with `julia` and `Pkg`:
+```sh
+julia -e 'using Pkg; Pkg.add("LanguageServer")'
+```
+This installs LanguageServer.jl into your global julia environment where the default launch
+command will look for the package.
+
+It is possible to use a separate package environment for the LanguageServer.jl installation.
+For example, to install in a folder `~/.config/nvim/lsp-julia` use the following command:
```
clason, line 67: Better to use `~/.local/share/nvim/lspconfig`, I would say.

lua/lspconfig/julials.lua
mjlbach: I added this. 

Try launching neovim in home, and opening some child folder (/home/test/my_julia_project) which contains a project.toml. If cmd_cwd is not set to the folder, or child folder, of the directory containing project.toml, then the current project resolution in the julia launch script will fail, since it uses the cwd determined by neovim's starting location.

fredrikekre: Okay, I believe that from the perspective of the julia process running the language server it won't matter (@pfitzseb ?). Is it used for something other than spawning the julia process?

lua/lspconfig/julials.lua
mjlbach: Is there not a common convention in Julia for storing user julia environments (eg ~/.virtualenvs for python)?

lua/lspconfig/julials.lua
mjlbach: Yes, this is only necessary for Base.current_project() to find the correct project.toml, it should not be removed.

clason: (It shouldn't matter for Julia, but it will matter for neovim...)

mjlbach: it does matter for julia, as finding the project.toml is relative to the cwd of the julia instance

fredrikekre: Yea. Will add it back, I simply never use that workflow. In any case, that means that the custom config for using a custom install location also need to include it, right?

mjlbach: Yes, if you override on_new_config, you would need to also include that snippet.

I think this is a pretty common usecase, as sometimes users will switch between projects with telescope. Without this, you would need to manually changed the cwd of neovim before initiating the telescope project switcher (see also #1106)

fredrikekre: Yea okay. It just seems a bit annoying that if the default configuration uses `on_new_config` it needs to be duplicated in a user-overridden one.

mjlbach:
Should we just set the current hardcoded cmd to what is recommended in the installation instructions in this PR? Is there a reason a user would ever prefer a global installation of LanguageServer.jl? That would simplify things. We might also include instructions for compiling a system image wtih PackageCompiler.jl

fredrikekre:
If you are okay then I believe that is "best practice" since the language server process is an isolated thing and it is good to have an isolated package environment for it.

lua/lspconfig/julials.lua
mjlbach: couldn't we have users override cmd instead of overriding on_new_config? I'm not sure why this needs to be done in on_new_config as opposed to appending to cmd ` --project=/path/to/my/julia/environment`

mjlbach:
I think it's way easier, and we're already providing installation instructions so why not follow best practice?

lua/lspconfig/julials.lua
fredrikekre: There is `.julia/environments/*` where I guess we could put it. Something like `.julia/environments/nvim-lspconfig/Project.toml` then I guess. I don't have a preference

mjlbach: I think that makes sense to me. I would like it to be obvious to users where to wipe the environment, so keeping it with other julia environments is probably easiest

fredrikekre: Sure, will update.

lua/lspconfig/julials.lua
```diff
@@ -51,13 +57,20 @@ configs.julials = {
     description = [[
 https://github.com/julia-vscode/julia-vscode
 
-`LanguageServer.jl` can be installed with `julia` and `Pkg`:
+LanguageServer.jl can be installed with `julia` and `Pkg`:
 ```sh
-julia -e 'using Pkg; Pkg.add("LanguageServer"); Pkg.add("SymbolServer")'
+julia --project=~/.julia/environments/nvim-lspconfig -e 'using Pkg; Pkg.add("LanguageServer")'
```
fredrikekre, line 62: In Julia 1.7 and above this can be spelled
```
julia --project=@nvim-lspconfig -e 'using Pkg; Pkg.add("LanguageServer")'
```
but lets wait (at least) until that is released before recommending that.

lua/lspconfig/julials.lua
```diff
@@ -7,9 +7,15 @@ local cmd = {
...
-    Pkg.instantiate()
+    # Load LanguageServer.jl: attempt to load from ~/.julia/environments/nvim-lspconfig
+    # with the regular load path as a fallback
+    ls_install_path = joinpath(
+        get(DEPOT_PATH, 1, joinpath(homedir(), ".julia")),
+        "environments", "nvim-lspconfig"
+    )
+    pushfirst!(LOAD_PATH, ls_install_path)
     using LanguageServer
+    popfirst!(LOAD_PATH)
```
mjlbach, line 18: Why do we still need all of this if we are passing `--project` to the julia process?

fredrikekre: We are not passing `--project` actually. It is a bit tricky, because that would interfere with the project discovery on the lines after. I think the only options are the PR as is, or to juggle with `Base.ACTIVE_PROJECT[]` instead in an equally "hacky" way.

mjlbach: Sorry, I meant to imply "why shouldn't we just pass project?" 

I think I understand now, Base.active_project() would then return the project.toml containing directory for LanguageServer.jl vs the one containing the project meant to be analyzed/passed to the LanguageServer.jl instance.

For now, this is good. I still think there's a little bit of duplication in functionality between our root detection in lspconfig and the project.toml detection for figuring out the appropriate julia environment.

mjlbach:
@fredrikekre is this ready now?

fredrikekre:
Yes. It shouldn't (:crossed_fingers:) break for anyone since the config still fall backs to the global env. Only difference should be that we look in `.julia/environments/nvim-lspconfig` first.

PR #1156, opened 2021-08-17
feat(leanls,lean3ls): add support for Lean 3/4 standard libraries

rish987:
Enabled by leanprover/elan#36. Pending review from @gebner, @Julian

### DIFF
--- a/lua/lspconfig/lean3ls.lua
+++ b/lua/lspconfig/lean3ls.lua
@@ -6,12 +6,22 @@ configs.lean3ls = {
     cmd = { 'lean-language-server', '--stdio', '--', '-M', '4096', '-T', '100000' },
     filetypes = { 'lean3' },
     root_dir = function(fname)
-      return util.root_pattern 'leanpkg.toml'(fname) or util.find_git_ancestor(fname) or util.path.dirname(fname)
+      -- check if inside elan stdlib
+      local stdlib_dir
+      do
+        local _, endpos = fname:find(util.path.sep .. util.path.join('lean', 'library'))
+        if endpos then
+          stdlib_dir = fname:sub(1, endpos)
+        end
+      end
+
+      return util.root_pattern 'leanpkg.toml'(fname)
+        or util.root_pattern 'leanpkg.path'(fname)
+        or stdlib_dir
+        or util.find_git_ancestor(fname)
+        or util.path.dirname(fname)
     end,
     on_new_config = function(config, root)
-      if not util.path.is_file(root .. '/leanpkg.toml') then
-        return
-      end
       if not config.cmd_cwd then
         config.cmd_cwd = root
       end
--- a/lua/lspconfig/leanls.lua
+++ b/lua/lspconfig/leanls.lua
@@ -6,12 +6,21 @@ configs.leanls = {
     cmd = { 'lean', '--server' },
     filetypes = { 'lean' },
     root_dir = function(fname)
-      return util.root_pattern 'leanpkg.toml'(fname) or util.find_git_ancestor(fname) or util.path.dirname(fname)
+      -- check if inside elan stdlib
+      local stdlib_dir
+      do
+        local _, endpos = fname:find(util.path.sep .. util.path.join('lib', 'lean'))
+        if endpos then
+          stdlib_dir = fname:sub(1, endpos)
+        end
+      end
+
+      return util.root_pattern 'leanpkg.toml'(fname)
+        or stdlib_dir
+        or util.find_git_ancestor(fname)
+        or util.path.dirname(fname)
     end,
     on_new_config = function(config, root)
-      if not util.path.is_file(root .. '/leanpkg.toml') then
-        return
-      end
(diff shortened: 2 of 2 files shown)


lua/lspconfig/lean3ls.lua
```diff
@@ -6,12 +6,27 @@ configs.lean3ls = {
...
+        end
+      end
+
+      return util.root_pattern 'leanpkg.toml'(fname)
+        or stdlib_dir
+        or util.find_git_ancestor(fname)
+        or util.path.dirname(fname)
     end,
     on_new_config = function(config, root)
-      if not util.path.is_file(root .. '/leanpkg.toml') then
```
rish987, line 12: Had this check before assuming the server would only work with a `leanpkg.toml` file next to it, but clearly that isn't the case.

gebner:
I think checking for the `.elan` directory is the wrong approach here.  It won't work for anything that is not installed inside the default elan toolchain directory (e.g. linked development builds, custom `$ELAN_HOME`, there are even people who use lean without elan).

For Lean 3 I would suggest the following heuristic:
  - If the filename is of the form `$a/lean/library/$b`, then use `$a/lean/library` as the root.  This works for toolchains installed via elan.
  - If there exists a `leanpkg.path` file in an ancestor directory, then use that directory as the root.  This works for git checkouts of lean.

For Lean 4 I would suggest the following:
 - If the filename is of the form `$a/lib/lean/$b`, then use `$a/lib/lean` as the root.  This works for toolchains installed via elan.
 - If the filename is of the form `$a/src/$b` and the file `$a/src/Lean.lean` exists, then use `$a/src` as the root.  This should work for git checkouts.

For Lean 4 it is slightly less important to reduce the number of server processes, since the Lean 4 server will start one process per file anyhow.  As long as the server is started somewhere in the toolchain/checkout directory, it should be fine.

gebner:
> does elan already work with checkouts

It's not automatic but still pretty easy to do by hand:
1. `elan toolchain link mytoolchain ~/mycheckout/build/release/stage1`
2. `cd ~/mycheckout && elan override set mytoolchain`

rish987:
@gebner done as you suggested, let me know what you think!

lua/lspconfig/leanls.lua
```diff
@@ -6,12 +6,29 @@ configs.leanls = {
...
+      end
+
+      -- check if inside checkout stdlib
+      local checkout_stdlib_dir
+      do
+        local _, endpos = fname:find(util.path.sep .. 'src')
+        if endpos then
+          checkout_stdlib_dir = fname:sub(1, endpos)
+        end
+      end
```
gebner, line 25: I think this might have too many false positives, `src` directories are pretty common.  It's no big deal if we start too many lean 4 server processes, so maybe we should just remove this check.

rish987: Done.

rish987: Also note that in this case it should fallback to the root directory of the checkout as a git ancestor.

rish987:
In general yeah this string matching is a bit fragile, it would be great if we could do something like `elan --get-root <filename>` that could handle these different situations for us. But this should work well for now, assuming standalone lean files whose paths also happen to match `/lib/lean` or `/lean/library` are pretty uncommon.

gebner (approved):

mjlbach:
I'm still not sure how I feel about adding root detection for standard libraries. Vscode doesn't do this, not sure we should either.

rish987:
(@gebner CMIIW) I think VSCode doesn't have simultaneous support for lean 3/lean 4 like we do? I think it just uses a single server for every UI session.

gebner:
> Vscode doesn't do this, not sure we should either.

vscode requires the user to open vscode in the correct directory, which is then used as the LSP root.  It doesn't have any root detection at all (at least in the Lean extensions).

rish987:
@mjlbach thoughts? This enables simultaneous Lean 3/Lean 4 stdlib LSP, something that is specifically _not_ supported in VSCode.

Julian:
Perhaps the right question is --

@mjlbach the full *plugin* (i.e. `lean.nvim`) wants to do this -- and right now has obviously the defaults for LSP configuration via what's here.

I can probably personally understand being weary of adding stdlib detection across all languages if VSCode doesn't do so, but I guess the immediate question is for Lean (i.e. us), we just need to know if this should go here, or if not then it probably makes sense for lean.nvim to fully manage the LSP config so that we can do this downstream without having half the config there and half here, which means obviously things get further away from this repo for Lean specifically.

If it makes it easier to push on a broader decision, maybe it makes sense to merge this for Lean specifically where there's a central downstream plugin at the minute, without making a broader decision on whether to do it for other languages, just to punt? Obviously understandable either way, just let us know what you think!

mjlbach:
If you all want this I'll merge it since you've been good about maintaining the configuration :) Just let me know.

rish987:
I'd say merge it, yes I will definitely update this if/when `elan` root detection is implemented.

PR #1159, opened 2021-08-18
Better default root_dir for fortls.

lkhphuc:
Require an explicit `.fortls` file in every project to activate Fortran LSP is not very intuitive I think, so I add either a git folder or the current working directory.

### DIFF
--- a/lua/lspconfig/fortls.lua
+++ b/lua/lspconfig/fortls.lua
@@ -6,7 +6,7 @@ configs.fortls = {
     cmd = { 'fortls' },
     filetypes = { 'fortran' },
     root_dir = function(fname)
-      return util.root_pattern('.fortls') or util.find_git_ancestor(fname) or util.path.dirname(fname)
+      return util.root_pattern('.fortls') or util.find_git_ancestor(fname)
     end,
     settings = {
       nthreads = 1,
--- a/lua/lspconfig/fortls.lua
+++ b/lua/lspconfig/fortls.lua
@@ -5,7 +5,7 @@ configs.fortls = {
   default_config = {
     cmd = { 'fortls' },
     filetypes = { 'fortran' },
-    root_dir = util.root_pattern '.fortls',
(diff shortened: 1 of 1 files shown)


lua/lspconfig/fortls.lua
```diff
@@ -5,7 +5,7 @@ configs.fortls = {
   default_config = {
     cmd = { 'fortls' },
     filetypes = { 'fortran' },
-    root_dir = util.root_pattern '.fortls',
+    root_dir = return util.root_pattern('.fortls') or util.find_git_ancestor(fname) or util.path.dirname(fname),
```
lithammer, line 8: You can't just use `return` like that, you need to wrap the whole thing in a function.

lkhphuc, line 8:
```suggestion
    root_dir = function(fname) 
      return util.root_pattern('.fortls') or util.find_git_ancestor(fname) or util.path.dirname(fname)
    end,
```

lkhphuc: Sorry fixed.

lua/lspconfig/fortls.lua
```diff
@@ -5,7 +5,9 @@ configs.fortls = {
   default_config = {
     cmd = { 'fortls' },
     filetypes = { 'fortran' },
-    root_dir = util.root_pattern '.fortls',
+    root_dir = function(fname) 
+      return util.root_pattern('.fortls') or util.find_git_ancestor(fname) or util.path.dirname(fname)
+    end,
```
lkhphuc, line 10:
```suggestion
    root_dir = function(fname)
      return util.root_pattern('.fortls') or util.find_git_ancestor(fname) or util.path.dirname(fname)
    end,
```

mjlbach:
You can verify this works with the new single_file_support if you want the server to start in an arbitrary non-project, otherwise will merge the git change for now

mjlbach:
Rebased and merged sans dirname in #1424, we have a `single_file_support` option that can be flagged for config that you can test and enable in a follow-up PR which has the same effect from an end user perspective (usually) without some of the drawbacks.

Issue #1155, opened 2021-08-17
Angularls bad behaviour on pre-Ivy projects


mactep:
angularls cmd needs the `--viewEngine` argument when working on Angular projects with a version lesser than 9, otherwise it'll not work correctly. I solved it manually by appending the argument in my config, but an automatic way would be better.

My solution for reference:
```lua
local languageServerPath = vim.fn.stdpath("config").."/languageserver"
local cmd = {"ngserver", "--stdio", "--tsProbeLocations", languageServerPath , "--ngProbeLocations", languageServerPath, "--viewEngine"}

lspconfig.angularls.setup{
    on_attach = on_attach,
    capabilities = capabilities,
    cmd = cmd,
    on_new_config = function(new_config, new_root_dir)
        new_config.cmd = cmd
    end,
}
```

mjlbach:
Maybe this should just be documented in CONFIG.md?

mactep:
I don't like this approach since other clients do this automatically. I'll poke around after work to see what I can get, but I remember an issue on the language serve repository where someone proposed a version verification. I'll see what I can do.

mactep:
Alright I landed on something.

``` lua
local languageServerPath = vim.fn.stdpath("config").."/languageserver"
local cmd = {"ngserver", "--stdio", "--tsProbeLocations", languageServerPath , "--ngProbeLocations", languageServerPath}

local file = io.open(vim.fn.getcwd()..'/node_modules/@angular/core/package.json', 'r')
if file then
    local file_contents = file:read('*all')
    if tonumber(file_contents.match(file_contents, [["version": "(%d+).%d+.%d+"]])) < 9 then
        table.insert(cmd, "--viewEngine")
    end
end

lspconfig.angularls.setup{
    on_attach = on_attach,
    capabilities = capabilities,
    cmd = cmd,
    on_new_config = function(new_config, new_root_dir)
        new_config.cmd = cmd
    end,
}
```

There's room for lots of improvements as value and error checking.

mjlbach:
This can be done in on_new_config for angular.

Fabyao:
@mactep My setup works fine with the latest Angular version. However for legacy Angular version (V7), I keep getting:

`Angular extension might not work correctly because ngcc operation failed`

I tried the workaround you suggested using `--viewEngine` but it seems LSP is not applying my custom config. I am using the config you posted above: 
<details>
 <summary>My Angular LSP Config</summary>

[...]

PR #1161, opened 2021-08-18
fix(angularls): add support for legacy projects

mactep:
Solves #1155

### DIFF
--- a/lua/lspconfig/angularls.lua
+++ b/lua/lspconfig/angularls.lua
@@ -33,6 +33,15 @@ configs[server_name] = {
   on_new_config = function(new_config, new_root_dir)
     local new_probe_dir = get_probe_dir(new_root_dir)
 
+    -- Checking if we should provide legacy support
+    local file = io.open(new_probe_dir..'/@angular/core/package.json', 'r')
+    if file then
+        local file_contents = file:read('*all')
+        if tonumber(file_contents.match(file_contents, [["version": "(%d+).%d+.%d+"]])) < 9 then
+            table.insert(new_config.cmd, "--viewEngine")
+        end
+    end
+
     -- We need to check our probe directories because they may have changed.
     new_config.cmd = {
       'ngserver',


lua/lspconfig/angularls.lua
```diff
@@ -33,6 +33,15 @@ configs[server_name] = {
   on_new_config = function(new_config, new_root_dir)
     local new_probe_dir = get_probe_dir(new_root_dir)
 
+    -- Checking if we should provide legacy support
+    local file = io.open(new_probe_dir..'/@angular/core/package.json', 'r')
+    if file then
+        local file_contents = file:read('*all')
+        if tonumber(file_contents.match(file_contents, [["version": "(%d+).%d+.%d+"]])) < 9 then
+            table.insert(new_config.cmd, "--viewEngine")
```
mjlbach, line 41: How is this not overwritten by https://github.com/neovim/nvim-lspconfig/blob/6062ab4a5ee0a2b0ccc1fe9531f8ee713114ea7c/lua/lspconfig/angularls.lua#L46-L53

mactep: I couldn't get it to work the other way around, but I just saw what I got wrong.

mactep: Well, actually I'm missing something either Lua level or at the language server level. In the README of the vscode implementation of the client they say that the backend is always shipped with the latest version, but the way we do here is by probing the project the user is currently working on. I guess that our decision is to make this a fallback.

In my case, I'm using Angular 6 and couldn't get to work with a "compatible" version of the language server since it asked for a minimum version 4.2 of typescript, which is not compatible with legacy projects.

If the user choose to use a custom cmd, providing a custom probing directory, my code does the trick, without breaking the usage with the default cmd. I'm not a Lua programmer, so I don't know exactly whats going on.

Should we just add this to the CONFIG.md and let the user copy the code if needed? I thought it would be nice to avoid the boilerplate.

mjlbach:
Sorry, I don't really do angular so it would be great if someone other than me could review/comment on this.

lua/lspconfig/angularls.lua
mjlbach: Let's move view_engine to a local variable and construct the table at the end.

mjlbach:
Can you try out https://github.com/neovim/nvim-lspconfig/pull/1462?

I can't test because angular crashes for me regardless.

PR #1166, opened 2021-08-22
feat(sixtyfps): add support for SixtyFPS ls

RustemB:

### DIFF
--- a/lua/lspconfig/sixtyfps.lua
+++ b/lua/lspconfig/sixtyfps.lua
@@ -1,0 +1,30 @@
+local configs = require 'lspconfig/configs'
+
+configs.sixtyfps = {
+  default_config = {
+    cmd = { 'sixtyfps-lsp' },
+    filetypes = { 'sixtyfps' },
+    single_file_support = true,
+  },
+  docs = {
+    description = [=[
+https://github.com/sixtyfpsui/sixtyfps
+`SixtyFPS`'s language server
+
+You can build ans install `sixtyfps-lsp` binary with `cargo`:
+```sh
+cargo install sixtyfps-lsp
+```
+
+Vim does not have built-in syntax for the `sixtyfps` filetype currently.
+
+This can be added via an autocmd:
+
+```lua
+vim.cmd [[ autocmd BufRead,BufNewFile *.60 set filetype=sixtyfps ]]
+```
+
+or by installing a filetype plugin such as https://github.com/RustemB/sixtyfps-vim
+]=],
+  },
+}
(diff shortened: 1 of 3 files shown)


lua/lspconfig/sixtyfps.lua
```diff
@@ -0,0 +1,21 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.sixtyfps = {
+    default_config = {
+        cmd = {'sixtyfps-lsp'},
+        filetypes = {'sixtyfps'},
```
mjlbach, line 7: Is there a filetype plugin users can install so this is recognized?

lua/lspconfig/sixtyfps.lua
```diff
@@ -0,0 +1,21 @@
...
+        root_dir = function(filename) return util.path.dirname(filename) end
```
mjlbach, line 8: There has to be a better root marker than this

RustemB: I don't think that it has it
It's just a markup language, that can be used in Rust, C++ and NodeJS projects

lua/lspconfig/sixtyfps.lua
RustemB: Not yet, but can be added manually at this moment, 'till language development in active progress

lua/lspconfig/sixtyfps.lua
mjlbach: I'm just generally opposed to adding roots that could be invalid, I may change the mechanism such that servers launch without project roots after 0.5.1 is released.

lua/lspconfig/sixtyfps.lua
RustemB: https://github.com/RustemB/sixtyfps-vim

mjlbach: Can you mention that you need this filetype plugin in the docs?

RustemB: ok

lua/lspconfig/sixtyfps.lua
```diff
@@ -0,0 +1,24 @@
...
```
lithammer, line 8: I'd say replace this with `util.find_git_ancestor` and then add `single_file_support = true`. On the other hand, I'm not sure how SixtyFPS handles a `null` root.

 ```suggestion
        root_dir = util.find_git_ancestor,
        single_file_support = true,
```

mjlbach: That was my concern too :/ I need to test

mjlbach:
I'm closing this PR in favor of #1430 for the sole reason that since you used master instead of a feature branch I am now locked in a battle with our autogenerated doc

PR #1184, opened 2021-08-25
feat(ui): lspinfo: better layout

kylo252:
# Description
 
Various improvements to LspInfo 

- Allow better readability by applying some extra paddings and a border
- Remove redundant information, e.g. if a client is attached then it's already active so no need to display the full info about it multiple times
- Break-out the arguments part of the client command
- Refactor LspInfo and extract some utility functions to allow for better maintainability

This also means we now have more available space to display any other useful information.

### before
image

### after
image

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -1,5 +1,117 @@
 local configs = require 'lspconfig/configs'
 local windows = require 'lspconfig/ui/windows'
+local util = require 'lspconfig/util'
+
+local indent = '  '
+local cmd_not_found_msg = 'False. Please check your path and ensure the server is installed'
+
+local function trim_whitespace(cmd)
+  local trimmed_cmd = {}
+  for _, str in pairs(cmd) do
+    table.insert(trimmed_cmd, str:match '^%s*(.*)')
+  end
+  return trimmed_cmd
+end
+
+local function indent_lines(lines, offset)
+  return vim.tbl_map(function(val)
+    return offset .. val
+  end, lines)
+end
+
+local function remove_newlines(cmd)
+  cmd = trim_whitespace(cmd)
+  cmd = table.concat(cmd, ' ')
+  cmd = vim.split(cmd, '\n')
+  cmd = trim_whitespace(cmd)
+  cmd = table.concat(cmd, ' ')
+  return cmd
+end
+
+local function autostart_to_str(autostart)
+  local autostart_status = 'true'
+  if autostart == false then
+    autostart_status = 'false'
+  end
+  return autostart_status
+end
+
+local function make_config_info(config)
+  local config_info = {}
+  config_info.name = config.name
+  if config.cmd then
+    config_info.cmd = remove_newlines(config.cmd)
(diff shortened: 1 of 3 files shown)


justinmk:
nice improvement, thanks!

side note: weird that lspconfig has a ui/ subsystem, that is rather an overreach for a "bunch of configurations".

gegoune:
Not everyone like borders, Justin really has a good point, so I assume it will never become configurable?

clason:
bad bot

clason:
The end goal is for lspconfig to indeed be nothing but configs. The info should be upstreamed and folded into `:checkhealth` as part of this.

(Which is getting unwieldy; putting efforts into structuring it better would be welcome.)

mjlbach:
I just woke up. I'm not happy this was merged. 

* The title is redundant and wastes vertical space
* It doesn't make sense to split cmd/args, this is an arbitrary distinction
* The point of listing active client IDs, is it's important for users to know which particular client is attached to a buffer, and the particular active root directories associated with each client. This is critical for debugging, especially in the case the root directories do not match. Also many operations (get_client_by_id, for example) rely on the specific client ID attached to the current buffer.
* The padding is ridiculously large
* Why are we adding rounded borders by default? This looks awful with the gap
* The code was originally extracted from popup.nvim so the providence is correct

mjlbach:
You can file a new PR and I'll review it.

kylo252:
> * It doesn't make sense to split cmd/args, this is an arbitrary distinction (we don't have the notion of args in a setting, so people will ask "Where is args defined? Should I add this in setup?" People will copy/paste this onto the command line to reproduce launching the language server. Before it was pretty easy to yank onto the command line.

I understand this point, but which is more common: 
1. get an overview of the flags that are passed to the server
2. get the literal command that you can copy paste somewhere else

We don't necessarily need to add the "args" distinction, it is more about printing them in a table form, _even though I didn't do the last part._

> * The point of listing active client IDs, is it's important for users to know which particular client is attached to a buffer, and the particular active root directories associated with each client. This is critical for debugging, especially in the case the root directories do not match. Also many operations (get_client_by_id, for example) rely on the specific client ID attached to the current buffer.

I'm not sure I did anything to contradict this? I kept the full listing of the active clients and removed the other ones **when they're duplicated**. I'll add a few more screenshots in the new PR.

dhruvmanila:
> Not everyone likes borders, Justin really has a good point, so I assume it will never become configurable?

You can update the window configuration in a `ftplugin` file. So, in `{after}/ftplugin/lspinfo.lua`:
```lua
vim.api.nvim_win_set_config(vim.api.nvim_get_current_win(), { border = "single" })
```

mjlbach:
> I understand this point, but which is more common:

    get an overview of the flags that are passed to the server
    get the literal command that you can copy paste somewhere else

The latter, as it's explicitly recommended as a debugging step in the lspconfig readme which I also wrote :) 

> I'm not sure I did anything to contradict this? I kept the full listing of the active clients and removed the other ones when they're duplicated. I'll add a few more screenshots in the new PR.

Add the client IDs next to the printed server name in the first section and you're good.

PR #1187, opened 2021-08-25
Add the VSCode package.json link for the puppet LSP

dhollinger:
See description


mjlbach:
Did you verify locally that this does not break docgen?

PR #1189, opened 2021-08-25
Add Lua to .editorconfig

polyzen:
- Complements modeline removal:
  https://github.com/neovim/nvim-lspconfig/commit/08a698d652f3d1fc6195b4efd9d23ecaced2c879

- Matches StyLua config


polyzen:
This could be reverted. I had overlooked that *.lua was already covered by the * catch-all.

PR #1203, opened 2021-08-29
fix(solargraph): Add dirname to root_dir

uplus:
Add dirname to solargraph's root_dir

The reasons are as follows.
* `solargraph` can work as a standalone ruby script without a Gemfile
* I want to use `solargraph` in snippets (e.g. to check how it works)
* It's intuitive (especially beginner-friendly)

### DIFF
--- a/lua/lspconfig/solargraph.lua
+++ b/lua/lspconfig/solargraph.lua
@@ -15,7 +15,9 @@ configs.solargraph = {
     },
     init_options = { formatting = true },
     filetypes = { 'ruby' },
-    root_dir = util.root_pattern('Gemfile', '.git'),
+    root_dir = function(fname)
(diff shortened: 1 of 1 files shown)


mjlbach:
I'm removing all util.path.dirname fallbacks, you can add this to your local config if you would like.

uplus:
> I'm removing all util.path.dirname fallbacks

Oops, I got it 😅 

By the way, what is the reason for this?
Would you mind telling me that for my information?

oblitum:
@uplus https://github.com/neovim/nvim-lspconfig/pull/1151#issuecomment-911023310

Gonna stop with this. Too many PRs doing the same...

uplus:
Thanks for the information!

PR #1206, opened 2021-08-30
Use extracted vscode-json-languageserver

entropitor:


mjlbach:
Why? We're already using https://github.com/hrsh7th/vscode-langservers-extracted which is more up to date.

justinmk (approved):

mjlbach (changes requested):
see my comment asking for justification for the change

entropitor:
Ah okay, from the name it looks like it was extracted out before microsoft extracted it out officially, but so the "fork" is still more up to date than the official extraction?

mjlbach:
As far as I can tell, https://github.com/vscode-langservers is **not** in any way affiliated with microsoft (and probably shouldn't be using the vscode logo)

PR #1214, opened 2021-09-01
Add psalm support

ftassi:


ftassi:
Looks like I got the commit message wrong. Do you want me to amend it or do you want to do it yourself?

lithammer:
It's preferable if you amend it yourself, thanks.

mjlbach:
I'm confused as to why you added the prefer cwd logic, that seems brittle.

ftassi:
@mjlbach I basically copy pasted here https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/phpactor.lua . I thought it could be a good starting point.

There is a missing bit here though (It's missing because my lua is not good enough yet :wink:): we might want  to update cmd based on the actual position of  psalm.xml file. If you are working on a mono repo your project root may be one level up your php code. I have a project at hand were I need this custom cmd `cmd = {'backend/bin/psalm.phar', '--language-server', '-r', 'backend'},` I'm customizing this in my vimrc but it would be nice if lspconfig could figure out this on its own. 
I sent the pr anyway because I think this is good enough to have psalm language server for most cases (it is in fact  the very config I'm using right now) and we could leave the rest to future improvements. 

disclaimer: I am not fully aware of how this `root_dir` works in the first place and how it should be set working with multiple language server for the same project. As I said I just assume that what was done for phpactor could be fine here too.

ftassi:
@lithammer  done :wink:

lua/lspconfig/psalm.lua
```diff
@@ -0,0 +1,28 @@
...
+local bin_name = 'bin/psalm-language-server'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { bin_name },
+    filetypes = { 'php' },
+    root_dir = function(pattern)
+      local cwd = vim.loop.cwd()
+      local root = util.root_pattern('psalm.xml', 'composer.json', '.git')(pattern)
+      -- prefer cwd if root is a descendant
```
lithammer, line 14: Why is this preferred? I feel like there's just too many edge cases that will resolve to an incorrect root.

For example if `cwd` is the root of a monorepo, but the Psalm project is a few directories deep. Or if I'm opening a file from while standing in my home directory (I'm quite sure I don't want the root to resolve to `~`).

lua/lspconfig/psalm.lua
```diff
@@ -0,0 +1,28 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'psalm'
+local bin_name = 'bin/psalm-language-server'
```
lithammer, line 5: Remove the `bin/` prefix.

lua/lspconfig/psalm.lua
```diff
@@ -0,0 +1,28 @@
...
+
+configs[server_name] = {
+  default_config = {
+    cmd = { bin_name },
+    filetypes = { 'php' },
+    root_dir = function(pattern)
+      local cwd = vim.loop.cwd()
+      local root = util.root_pattern('psalm.xml', 'composer.json', '.git')(pattern)
```
lithammer, line 13: There's now another open pull request for Psalm support (GH-1260) that mentions `psalm.xml.dist`.

lithammer:
Closing this in favour of GH-1260 since it's more straight forward and [supposedly more in line with the VS Code plugin and coc-psalm](https://github.com/neovim/nvim-lspconfig/pull/1260#discussion_r708029678). The `root_dir` approach here is a bit controversial in my opinion.

However feel free to open a new pull request to discuss the details of `root_dir`. Maybe @gbprod has some input on that (author of the other pull request). I don't do any PHP myself, so...

lithammer:
There's also the fact that Psalm doesn't even start without a `psalm.xml[.dist]` file. So looking for other root markers seem unnecessary.

PR #1222, opened 2021-09-05
Add zk lsp

wurosh:


mjlbach:
You'll need to update the handler style, we want to use a compatibility layer though, so see my comment https://github.com/neovim/nvim-lspconfig/pull/1233#issuecomment-914633260

wurosh:
@mjlbach I think that fixes it.

Also, I assume the mergify bot is complaining because it looks like i edited the `CONFIG.md` file but I just rebased my branch to get the utils. Or am I doing something else wrong? Either way could we re-open this?

wurosh:
Scratch that, seems like my commit didn't move to the top

wurosh:
I'll just open a new PR for this in the meantime

PR #1223, opened 2021-09-05
Add volar

3nuc:
After the release of Vue 3 there arose a need of a new language server that suppors the new syntax.
We already have `vuels` in the repo (so that's the one that existed pre-Vue 3) and the new `volarvuels` I'm adding here that supports the new syntax.

https://github.com/johnsoncodehk/volar

https://github.com/johnsoncodehk/volar#communitys-language-client-implements - it's supported in many vim lsp configuration providing extensions, I'm just adding it also here :P

See https://github.com/johnsoncodehk/volar/discussions/441

Problems:
- [x] <s>If this is merged we'll have `vls`, `vuels` and `volarvuels` which makes things even more confusing, since `vuels` is normally `vls`, but `vls` is taken by the V language server, so this repo uses `vuels` instead. But now we have `volarvuels`. AND there's also https://github.com/vuedx/languagetools/tree/main/packages/vue-languageservice which is a third language server for Vue </s> Decided on just volar
- [x] It's actually crucial to pass serverPath. I'll need to look at angularls's code to detect it dynamically
- [x] Improve docs -  I assume as long as the nvim set-rtp command from CONTRIBUTING.md has the word 'volar' I'm good. Since it's not modifying the CONFIG.md for me locally
- [x] <s>bin_name uses volar-server, which I got from the AUR. Better wait for https://github.com/johnsoncodehk/volar/issues/458</s> - fixed because newest alpha version exposes vue-server binary. <s>Let's still wait for this to stabilize</s> Volar's maitainer's words sound like it's gonna be a while until this changes
- [x]  <s>Run linters, right now it probably has lots of lint errors</s>

### DIFF
--- a/lua/lspconfig/volar.lua
+++ b/lua/lspconfig/volar.lua
@@ -1,0 +1,89 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local function get_typescript_server_path(root_dir)
+  local project_root = util.find_node_modules_ancestor(root_dir)
+  return project_root and (util.path.join(project_root, 'node_modules', 'typescript', 'lib', 'tsserverlibrary.js'))
+    or ''
+end
+
+local server_name = 'volar'
+local bin_name = 'volar-server'
+
+-- https://github.com/johnsoncodehk/volar/blob/master/packages/shared/src/types.ts
+local volar_init_options = {
+  typescript = {
+    serverPath = '',
+  },
+  languageFeatures = {
+    -- not supported - https://github.com/neovim/neovim/pull/14122
+    semanticTokens = false,
+    references = true,
+    definition = true,
+    typeDefinition = true,
+    callHierarchy = true,
+    hover = true,
+    rename = true,
+    renameFileRefactoring = true,
+    signatureHelp = true,
+    codeAction = true,
+    completion = {
+      defaultTagNameCase = 'both',
+      defaultAttrNameCase = 'kebabCase',
+    },
+    schemaRequestService = true,
+    documentHighlight = true,
+    documentLink = true,
+    codeLens = true,
+    diagnostics = true,
+  },
+  documentFeatures = {
+    -- not supported - https://github.com/neovim/neovim/pull/13654
+    documentColor = false,
+    selectionRange = true,
+    foldingRange = true,
+    linkedEditingRange = true,
+    documentSymbol = true,
+    documentFormatting = {
+      defaultPrintWidth = 100,
+    },
+  },
+}
+
+configs[server_name] = {
+  default_config = {
+    cmd = { bin_name, '--stdio' },
+    filetypes = { 'vue' },
+    root_dir = util.root_pattern 'package.json',
+    init_options = volar_init_options,
+    on_new_config = function(new_config, new_root_dir)
+      if
+        new_config.init_options
+        and new_config.init_options.typescript
+        and new_config.init_options.typescript.serverPath == ''
+      then
+        new_config.init_options.typescript.serverPath = get_typescript_server_path(new_root_dir)
+      end
+    end,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/johnsoncodehk/volar/master/package.json',
+    description = [[
+https://github.com/johnsoncodehk/volar/tree/master/packages/server
+
+Volar language server for Vue
+Volar can be installed via npm
+```sh
+npm install -g @volar/server
+```
+
+With Vue 3 projects - it works out of the box.
+
+With Vue 2 projects - requires [additional configuration](https://github.com/johnsoncodehk/volar#using)
+
+Do not run `vuels` and `volar` at the same time.
+
+To check which language servers are running, open a `.vue` file and run the `:LspInfo` command.
+]],
+  },
+}


lua/lspconfig/volarvuels.lua
```diff
@@ -0,0 +1,88 @@
...
+local util = require 'lspconfig/util'
+
+local server_name = 'volarvuels'
+
+-- TODO: This is the name if you download volar-server from the AUR
+-- need to wait for https://github.com/johnsoncodehk/volar/issues/458
+local bin_name = 'volar-server'
+
+-- TODO: Figure out how to load this dynamically
+local typescript_server_path = '/home/artur/.npm-global/lib/node_modules/typescript/lib/tsserverlibrary.js'
```
3nuc, line 11: Path to tsserver is hardcoded. I need to figure out how to use the $NPM_PREFIX to look for a global installation of tssesrver OR look in the current workspace.

Another problem is that supposedly Volar doesn't work very well with TypeScript 4.4?

lua/lspconfig/volarvuels.lua
```diff
@@ -0,0 +1,88 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'volarvuels'
+
+-- TODO: This is the name if you download volar-server from the AUR
+-- need to wait for https://github.com/johnsoncodehk/volar/issues/458
+local bin_name = 'volar-server'
+
```
3nuc, line 9: This comes from https://aur.archlinux.org/packages/volar-server/ . The npm package @volar/server doesn't expose a binary, so you need to refer to it by `node $(npm get prefix)/lib/node_modules/@volar/server/out/index.js`.

I guess it's better to wait for the upstream volar issue 458 to be implemented so we don't have to implement a wrapper script for getting the path.

lua/lspconfig/volarvuels.lua
```diff
@@ -0,0 +1,88 @@
...
+     typeDefinition = true,
+     callHierarchy = true,
+     hover = true,
+     rename = true,
+     renameFileRefactoring = true,
+     signatureHelp = true,
+     codeAction = true,
+     completion={
+       defaultTagNameCase='both',
+       defaultAttrNameCase='kebabCase',
```
3nuc, line 29: The lack of spaces between '=' will probably trigger a linting error. I didn't run the linters yet

lua/lspconfig/volarvuels.lua
```diff
@@ -0,0 +1,88 @@
...
+  docs = {
+    package_json = 'https://github.com/johnsoncodehk/volar/master/packages/server/package.json',
+    description = --TODO
+    [[
+      https://github.com/johnsoncodehk/volar/tree/master/packages/server
+      Volar language server for Vue
+      Volar can be installed via NPM
+      ```sh
+      npm install -g @volar/server
+      ```
```
3nuc, line 85: Improve docs

1. Mention that you can't run vuels and volarvuels at the same time
2. If we decide to not wait for Volar issue 453 (add executable) we need to mention that there's extra configuration needed (add your own volar-server .sh wrapper to PATH)

lua/lspconfig/volarvuels.lua
3nuc: I could put some .lua script that'd scan the filesystem in one of the on_setup hooks nvim-lspconfig has. I'm just pretty sure I won't be able to call shell commands like `npm get prefix` or `npx` to know where to look for node_modules (global and local respectively)

lua/lspconfig/volarvuels.lua
3nuc: This is now fixed & closed. The executable name is vue-server. This change is only published in an alpha version so it still may be worth waiting a little bit. I'll change bin-name to `vue-server` in a second.

lua/lspconfig/volarvuels.lua
```diff
@@ -0,0 +1,85 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'volarvuels'
+local bin_name = 'vue-server'
```
3nuc, line 5: `vue-server` executable is only exposed in @volar/server@alpha so it may change in future releases. Let's wait a bit.

3nuc: It's called `volar-server` now in latest alpha.

3nuc:
Is the language server in this PR slow for anyone else? I open a .vue file and errors appear after around 12 seconds. Did I enable some resource-heavy options?

johnsoncodehk:
> Is the language server in this PR slow for anyone else? I open a .vue file and errors appear after around 12 seconds. Did I enable some resource-heavy options?

Initialization time is proportional to your project scale, at least need a few seconds, Volar exchanges this for better runtime performance.

lua/lspconfig/volarvuels.lua
3nuc: Sounds like this will be the bin name for a while (based on maintainer's comments)

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
...
+local function get_typescript_server_path(root_dir)
+  local project_root = util.find_node_modules_ancestor(root_dir);
+  return project_root and (project_root .. '/node_modules/typescript/lib/tsserverlibrary.js') or '';
+end
+
+local server_name = 'volar'
+local bin_name = 'volar-server'
+
+-- https://github.com/johnsoncodehk/volar/blob/master/packages/shared/src/types.ts
+local volar_init_options = {
```
mjlbach, line 13: Do all of these options need to be explicitly set, or are these overriding defaults?

johnsoncodehk: Yes, need explicitly set to enable these features. Otherwise features was disabled.

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
...
+      defaultPrintWidth = 100,
+    },
+  }
+}
+
+-- https://github.com/johnsoncodehk/volar/blob/master/package.json#L56
+local volar_settings = {
+  volar= {
+    codeLens= {
+      references = true,
```
3nuc, line 54: @mjlbach Regarding your comment above about "is it necessary to set those explicitly". init_options need to be explicit, but volar_settings contains some things that basically set some settings to their default value. Here I'm setting references to true, but it's already true by default: https://github.com/johnsoncodehk/volar/blob/26b36fb01a83a68c94fcedd06144041876a7ae69/package.json#L78

I assume you'd like a minimal config that overrides defeaults only when it's necessary to make the LS work with nvim?

On the other hand, I'm worried people won't know about the settings = {} prop in nvim_lsp.volar.setup{}, since most LSs use init_options

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
...
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/johnsoncodehk/volar/master/packages/server/package.json',
+    description = [[
+    https://github.com/johnsoncodehk/volar/tree/master/packages/server
+    Volar language server for Vue
+
+    Volar can be installed via npm
+    ```sh
+    npm install -g @volar/server
+    vue-server --stdio
```
mjlbach, line 86: Do you need to run vue-server --stdio during installation for some reason?  That's how this reads.

lua/lspconfig/volar.lua
3nuc: Would it be ok to do something like
```lua
init_options = volar_init_options,
-- https://github.com/johnsoncodehk/volar/blob/master/package.json
settings = {} 
```
?

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
...
+    ```
+
+    With Vue 3 projects - it works out of the box.  
+    With Vue 2 projects - requires [additional configuration](https://github.com/johnsoncodehk/volar#using)
+
+    Do not run `vuels` and `volarvuels` at the same time. 
+    To check which language servers are running, open a `.vue` file and run the `:LspInfo` command.
+    ]],
+    default_config = {
```
mjlbach, line 95: What's in default_config should get automatically pulled during docgen, you can test locally but I think it's unnecessary here

lua/lspconfig/volar.lua
mjlbach: I would just use the package.json to autogenerate the documentation for these optiosn (see pyright or other servers, it's just one line with the url to the package.json). I find people get extremely confused when we duplicate default settings.

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local function get_typescript_server_path(root_dir)
+  local project_root = util.find_node_modules_ancestor(root_dir);
+  return project_root and (project_root .. '/node_modules/typescript/lib/tsserverlibrary.js') or '';
```
mjlbach, line 6: I would use path.join from utils

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,103 @@
...
+    codeAction = true,
+    completion={
+      defaultTagNameCase='both',
+      defaultAttrNameCase='kebabCase',
+    },
+    schemaRequestService  = true ,
+    documentHighlight = true,
+    documentLink = true ,
+    codeLens = true,
+    semanticTokens = true,
```
mjlbach, line 35: We don't support semanticTokens currently, or documentColor (if it is what I think it is

3nuc:
It's not needed for installation. I thought it'd be helfpul if I mentioned the server's executable name. But I see other LS's don't do that. I'll remove

On September 7, 2021 9:17:18 AM GMT+02:00, Michael Lingelbach ***@***.***> wrote:
***@***.*** commented on this pull request.

lua/lspconfig/volar.lua
3nuc: Set both of those to false. Added links to relvant neovim PRs

lua/lspconfig/volar.lua
```diff
@@ -0,0 +1,77 @@
...
+  return project_root and (util.path.join(project_root, 'node_modules', 'typescript', 'lib', 'tsserverlibrary.js')) or '';
+end
+
+local server_name = 'volar'
+local bin_name = 'volar-server'
+
+-- https://github.com/johnsoncodehk/volar/blob/master/packages/shared/src/types.ts
+local volar_init_options = {
+  typescript = {
+    serverPath = get_typescript_server_path(vim.fn.getcwd()),
```
mjlbach, line 15: this vim.fn.getcwd seems brittle, should it not use rootdir? This can be done in on_new_config if necessary.

lua/lspconfig/volar.lua
3nuc: done

lua/lspconfig/volar.lua
mjlbach: Thanks! If you could, since you are the expert, please let me know if anything needs to change in this PR regarding settings or init_options.

lua/lspconfig/volar.lua
3nuc: done

lua/lspconfig/volar.lua
3nuc: https://github.com/neovim/nvim-lspconfig/pull/1223/commits/86431d9b5d09d1fd6d8be892d207b543fbe72265

Done. I also changed the link to package.json in a later commit, since the previous one didn't have a "contributes" property

lua/lspconfig/volar.lua
3nuc: I checked locally. Indeed it gets fetched :)

https://github.com/neovim/nvim-lspconfig/pull/1223/commits/ce7a7e059282cd0ca7f600e7b27e1cc79b5aaad1

lua/lspconfig/volar.lua
3nuc: This comes from https://github.com/neovim/nvim-lspconfig/blob/08a698d652f3d1fc6195b4efd9d23ecaced2c879/lua/lspconfig/angularls.lua#L15 , but perhaps it's warranted in case of angularls's probedir but weird for detecting typescript.

I'll try with on_new_config later

lua/lspconfig/volar.lua
johnsoncodehk: The options look good to me! And no need to specifically set false for unsupported features, it will just ignore if unsupported.

johnsoncodehk:
@mjlbach if possible, I hope we can split features to multiple language server, because the DX will be much better. Can nvim-lsp do that?

References:
- https://github.com/johnsoncodehk/volar/discussions/393#discussioncomment-1213736
- https://github.com/yaegassy/coc-volar/blob/master/src/index.ts#L158-L203

mjlbach:
Yes, but the API for doing it declaratively is a bit annoying, we basically have a hook after start_client to disable detected server capabilities in`on_attach`

lua/lspconfig/volar.lua
3nuc: Just wanted to be explicit

lua/lspconfig/volar.lua
3nuc: Done https://github.com/neovim/nvim-lspconfig/pull/1223/commits/f170b0cbf23ff10efb4ee0755259befd29fff263

3nuc: I'm not sure if straight up not providing serverPath in init_options to Volar is a mistake. Anyhow, I preferred the user to get an error from Volar in that case rather than get some vague lua error about "index is nil".

Hence the if chain check

3nuc:
> Yes, but the API for doing it declaratively is a bit annoying, we basically have a hook after start_client to disable detected server capabilities in`on_attach`

Is there an example of this somewhere in the repo? I can't wrap my head around what that'd look like

My first reflex when thinking about mapping these lines https://github.com/yaegassy/coc-volar/blob/master/src/index.ts#L82-L86 from the coc-volar implementation to nvim-lspconfig would be 

```lua
configs['volar-server-1-api'] = {
  init_options = {languageFeatures = { --[[ all languageFeatures ]]--}}
}

configs['volar-server-2-document'] = {
  init_options = {languageFeatures = { --[[ all documentFeatures ]]--}}
}
```

but that feels naive and probably would add unnecesary entries to CONFIG.md

mjlbach:
When you say "multiple language servers", do you mean mutiple instances of volar? That might be kind of challenging. I'd have to think about it a bit, as it's not something with explicitly support.

Hypothetically you could do something like (contrived example):

```lua
vim.lsp.volar.setup{
  on_attach = function(client)
    client.resolved_capabilities.document_formatting = true
  end;
}
vim.lsp.volar.setup{
  on_attach = function(client)
    client.resolved_capabilities.document_formatting = false
  end;
}
```

But that shouldn't work with our current autocommands

mjlbach:
Can you address lint/style lint/commit lint issues and then I'll merge? You can just run stylua in the base directory, and then interactive git rebase (or I'll do it for you)

mjlbach:
Seems like a bad rebase :)

3nuc:
@mjlbach do you mean "use git rebase --interactive to drop the merge commits because here we prefer rebases"? :P not sure if I understand correctly

mjlbach:
No, I mean I want you to squash your commits and write a "conventional commits" style commit message haha.

3nuc:
Thanks for your supervision guys :)

@johnsoncodehk Can you take a look at https://github.com/neovim/nvim-lspconfig/pull/1223#issuecomment-915330392

I understood that there should be multiple instances of volar that handle different features (eg. if you open htop/task manger you'd see 2-3 volar processes) - is this correct?

johnsoncodehk:
> Thanks for your supervision guys :)
> 
> @johnsoncodehk Can you take a look at [#1223 (comment)](https://github.com/neovim/nvim-lspconfig/pull/1223#issuecomment-915330392) ?
> 
> I understood that there should be multiple instances of volar that handle different features (eg. if you open htop/task manger you'd see 2-3 volar processes) - is this correct?

If my understanding is correct, it means that we cannot simply configure it. We need to wait for nvim-lsp to improve its support.

mjlbach:
To clarify, it's an lspconfig thing not a built-in client thing (they are separate, you don't need to use this repo to use the built-in client)

PR #1224, opened 2021-09-05
feat(ci): use luacheck without relying on apt-get

kylo252:
This should be an improvement over relying on `apt`.

- no need to run as `sudo`
- make it a bit more stable and reproducible since there's no `apt update` anymore

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,14 +11,20 @@ jobs:
     steps:
     - name: Checkout sources
       uses: actions/checkout@v2
-    - name: Setup luacheck
-      run: |
-        sudo apt update &&
-        sudo apt install -y lua5.1 luarocks &&
-        sudo luarocks install luacheck
+
+    - name: Use lua
+      uses: leafo/gh-actions-lua@v8
+
+    - name: Use luarocks
+      uses: leafo/gh-actions-luarocks@v4
+
+    - name: Install luacheck
+      run: luarocks install luacheck
+
     - name: Run luacheck
       run: |
-        sudo make lint
+        make lint
+
   style-lint:
     runs-on: [ubuntu-latest]
     steps: 


mjlbach:
Doesn't this replace stability (relying on debian packaging) with relying on a random github action?

kylo252:
> Doesn't this replace stability with relying on a random github action?

Interesting point.

It's just that updating the environment before running a CI workflow will have unpredictable results and is generally not ideal. I realize that the `luarocks` version itself might be updated in the github action but the important package is `luacheck` which is always built _manually_ anyway. 

It should also be faster and less risky than some weird system-package getting updated. so I would say that it's still worth going for. 

> (relying on debian packaging, probably the most reliable packaging ecosystem of all linux distros) 

Someone at Redhat is getting triggered right now.. 😄

teto:
using the actions is much more opaque than what we have right now so -1 from me. If the sudo can be removed than all the better, but it runs on CI so it's ok.

kylo252:
> using the actions is much more opaque than what we have right now so -1 from me. 

Not sure what you mean by opaque here. As I mentioned earlier, we're always installing `luacheck` manually, I just made it so that we can skip relying on apt-get.

> If the sudo can be removed than all the better, but it runs on CI so it's ok.

The purpose is to avoid running a system-wide update for the runner which isn't ideal for ensuring consistent runs.

---

That said, there's alternative which I've gotten a little bit more familiar with recently, and which could also be quite nice for this: https://github.com/mpeterv/hererocks.

teto:
> Not sure what you mean by opaque here. As I mentioned earlier, we're always installing luacheck manually, I just made it so that we can skip relying on apt-get.

Most people understand  what `sudo apt update &&  sudo apt install -y lua5.1 luarocks` does straightaway.
I have no idea what `leafo/gh-actions-lua@v8` does without checking their repo. hererocks is an whole other can of problems , it needs python and all. 
btw apt update doesn't do a system wide update, that would be `upgrade`. Remove the sudo and let's move on.

kylo252:
> Most people understand what `sudo apt update && sudo apt install -y lua5.1 luarocks` does straightaway. I have no idea what `leafo/gh-actions-lua@v8` does without checking their repo.

I totally agree with sentiment, if we're talking about a user-facing command. What's the point of GitHub actions if not to relieve the burden on the runner by using some pre-baked environments? 

>  hererocks is an whole other can of problems , it needs python and all. 

Oh I had assumed that these come with Python already. I'm not actually sure: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions

> btw apt update doesn't do a system wide update, that would be `upgrade`. Remove the sudo and let's move on.

Oh of course 🤦‍♂️ 

In my experience, `luarocks` is a bit annoying when it comes to setting up a local install, you need to set it up with `--local`, or `--tree`, or maybe both. So I believe that this gh-action actually takes care of setting the environment variable. So I don't think you can really get rid of the `sudo` requirement unless you set that up.

It really doesn't matter much in this case, since these system-packages don't have that many dependencies. So overall, it's a pretty minor optimization that I thought could be useful. Thanks for the feedback!

mjlbach:
I'll need to be strongly convinced of the advantage of this, for now I'm closing.

PR #1233, opened 2021-09-07
Update denols to support new LSP handler signature

jason0x43:
This PR updates the denols config to support the [new LSP handler signature](https://github.com/neovim/neovim/commit/df17d7844ed7dedcb80f9405f7078a046a12524a) in recent Neovim 0.6 nightlies. It maintains compatibility with the signature used in Neovim 0.5 and previous versions of 0.6 nightly.

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -65,7 +65,18 @@ local function virtual_text_document(uri)
   virtual_text_document_handler(uri, result)
 end
 
-local function denols_handler(err, method, result)
+local function denols_handler(err, methodOrResult, resultOrContext)
+  local is_new_signature = resultOrContext.method ~= nil
+  local result
+
+  -- LSP handler signature was changed in
+  -- https://github.com/neovim/neovim/commit/df17d7844ed7dedcb80f9405f7078a046a12524a
+  if is_new_signature then
+    result = methodOrResult
+  else
+    result = resultOrContext
+  end
+
   if not result or vim.tbl_isempty(result) then
     return nil
   end
@@ -79,7 +90,14 @@ local function denols_handler(err, method, result)
     end
   end
 
-  lsp.handlers[method](err, method, result)
+  if is_new_signature then
+    local context = resultOrContext
+    local method = resultOrContext.method
+    lsp.handlers[method](err, result, context)
+  else
+    local method = methodOrResult
+    lsp.handlers[method](err, method, result)
+  end
 end
 
 local function denols_definition()


mjlbach:
Can you just add this: https://github.com/neovim/neovim/issues/14090#issuecomment-913198455

to utils and wrap the old handler?

jason0x43:
That snippet wraps a new handler to work with v0.5.0 calls. This PR updated the existing v0.5.0 denols handler to work with the new call format. I could update the deno handler to use the new format by default and wrap it.

However, I think the deno handler might need to know whether it's being called with the new or old arguments, because it calls another handler. Simply an updated handler doesn't work in both 0.5 and the nightly because the handler being called by deno (`lsp.handlers[method](err, method, result)`) may or may not itself be configured to handle both call formats.

jason0x43:
Updating the deno handler is easy. The issue I ran into when doing that is that the deno handler [calls another handler](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/denols.lua#L82). That handler isn't wrapped -- if the deno handler calls it using the 0.5.0 arguments, it works in neovim 0.5.0 but breaks in the nightly. If it's called with the new arguments, it breaks in 0.5.0 but works in the nightly.

clason:
Well, then _that_ handler should be wrapped, too ;)

jason0x43:
I haven't really dug into it, though, so I may be missing something obvious.

jason0x43:
It should, yes. 😄

jason0x43:
I believe the handler deno is calling, by default, is the one in neovim's runtime, in which case wrapping isn't really an option.

jason0x43:
I updated denols_handler to use the new format and added the wrapper to lspconfig.util. I modified the wrapper slightly; it adds an `is_legacy_call` field to the context if it had to convert the call format. This lets receivers that care, like denols, know how they should call other handlers.

clason:
There is already a wrapper in `lspconfig.util`; please use that instead of adding yet another one.

clason:
(You need to rebase on master.)

lua/lspconfig/denols.lua
```diff
@@ -79,7 +79,11 @@ local function denols_handler(err, method, result)
     end
   end
 
-  lsp.handlers[method](err, method, result)
+  if context.is_legacy_call then
+    lsp.handlers[context.method](err, context.method, result)
+  else
+    lsp.handlers[context.method](err, result, context)
+  end
```
clason, line 86: why can't you just do
```lua
util.compat_handler(lsp.handlers[method])(err, result, context)
```

jason0x43:
D'oh, I was rebased on master, but I didn't see the handler wrapper was already there 🤦

lua/lspconfig/denols.lua
jason0x43: It's the same problem I mentioned earlier -- the wrapper works when you're making a new-style handler work with Neovim 0.5, but not the other way around. If the updated denols config is used with Neovim 0.5, the target handler will be using legacy arguments, so that wrapper won't do anything for it.

clason: Now I get it. Maybe add a `TODO` comment before it so it's easier to find and revert when the handler change is in a release?

jason0x43:
I'm using the existing compat_handler, but I added the `is_legacy_call` to its context. I'm definitely open to not doing that if there's a better way, but I think that somehow the deno handler will have to change its behavior based on what call syntax is used in the current environment.

lua/lspconfig/denols.lua
```diff
@@ -65,7 +65,7 @@ local function virtual_text_document(uri)
   virtual_text_document_handler(uri, result)
 end
 
-local function denols_handler(err, method, result)
+local function denols_handler(err, result, context)
```
mjlbach, line 68: the convention is ctx

mjlbach:
Wouldn't it be cleaner instead of doing the "is legacy" checks to just check if the user `has` nvim-0.5.1?

jason0x43:
Version checking would work, as long as it's clear what version the new signature should come in at. Checking how the handler is actually being called removes the need to know that, though, so it's less fragile.

jason0x43:
In any case, I'm fine with doing a version check if that's preferable.

mjlbach:
As I don't use deno, testing and saying this fixes your issue is the fastest way to get this merged :)

David-Else:
@mjlbach `0.6 master`  appears to be working fine without it, @jason0x43 am I missing something?

Nvim master made a very important fix adding `codeAction/resolve` that makes Deno LSP now work properly, so it is very important all is well when using master :)

jason0x43:
I haven't checked with the most recent 0.6. I'll take a look...

mjlbach:
This PR is necessary to adapt the custom handler overrides that deno uses. You probably aren't invoking a command that would use them

jason0x43:
This is still an issue as of NVIM v0.6.0-dev+dd7812ec6 (from Oct 5) with the current master of lspconfig. I'll update neovim to the latest commit and check that.

jason0x43:
Ah, what @mjlbach said 😁 

Also, I updated the PR to do a version check since 0.5.1 does include the new handler signature (I wasn't aware that was definitely going to be the case when you mentioned it earlier).

mjlbach:
Thanks, I just approved the CI run. I'll merge after this, and we'll remove the compatibility shims once 0.6 is released.

PR #1235, opened 2021-09-07
rust-analyzer: Ignore stderr when invoking cargo metadata

jonhoo:
Vim's `system` output [includes both STDOUT and STDERR](https://github.com/neovim/neovim/blob/3c497e214f48ee1433d759f5a56c028df5186f24/src/nvim/os/shell.c#L773-L776), which means it may contain all sorts of debug output, especially if the user is using custom rustup toolchains.

To reproduce, set up neovim to use `lsp-config`, and then:
```console
$ cargo new foo
     Created binary (application) `foo` package
$ cd foo/
$ mkdir -p toolchain/bin/
$ $EDITOR toolchain/bin/cargo
$ cat toolchain/bin/cargo
#!/bin/bash
echo "X" >&2
"$(env -u RUSTUP_TOOLCHAIN rustup +stable which cargo)" "$@"
exit 0
$ chmod +x toolchain/bin/cargo
$ ln -s "$(rustup +stable which rustc)" toolchain/bin/rustc
$ echo '[toolchain]' > rust-toolchain.toml
$ echo "path = \"$PWD/toolchain\"" >> rust-toolchain.toml
$ cargo check
X
    Checking foo v0.1.0 (/Users/jonhoo/dev/foo)
    Finished dev [unoptimized + debuginfo] target(s) in 0.91s
$ nvim src/main.rs
Error detected while processing FileType Autocommands for "rust":
E5108: Error executing lua Vim:E474: Unidentified byte: X
{"packages":[{"name":"foo","version":"0.1.0","id":"foo 0.1.0 (path+file:///Users/jonhoo/dev/foo)","license":null,"license_f
ile":null,"description":null,"source":null,"dependencies":[],"targets":[{"kind":["bin"],"crate_types":["bin"],"name":"foo",
"src_path":"/Users/jonhoo/dev/foo/src/main.rs","edition":"2018","doc":true,"doctest":false,"test":true}],"features":{},"man
ifest_path":"/Users/jonhoo/dev/foo/Cargo.toml","metadata":null,"publish":null,"authors":[],"categories":[],"keywords":[],"r
eadme":null,"repository":null,"homepage":null,"documentation":null,"edition":"2018","links":null}],"workspace_members":["fo
o 0.1.0 (path+file:///Users/jonhoo/dev/foo)"],"resolve":null,"target_directory":"/Users/jonhoo/dev/foo/target","version":1,
"workspace_root":"/Users/jonhoo/dev/foo","metadata":null}
```

With this fix, the error no longer occurs, and `rust-analyzer` starts correctly.

With this change, the plugin will also inform of `cargo metadata` failures. Replace `exit 0` with `exit 1` in `toolchain/bin/cargo`, and run `nvim src/main.rs`, and you'll see
```
cmd "cargo metadata --no-deps --format-version 1 --manifest-path /Users/jonhoo/dev/foo/Cargo.toml" is not executable:
X
```
Previously, this error would be entirely swallowed.

### DIFF
--- a/lua/lspconfig/rust_analyzer.lua
+++ b/lua/lspconfig/rust_analyzer.lua
@@ -18,14 +18,36 @@ configs.rust_analyzer = {
     filetypes = { 'rust' },
     root_dir = function(fname)
       local cargo_crate_dir = util.root_pattern 'Cargo.toml'(fname)
-      local cmd = 'cargo metadata --no-deps --format-version 1'
+      local cmd = { 'cargo', 'metadata', '--no-deps', '--format-version', '1' }
       if cargo_crate_dir ~= nil then
-        cmd = cmd .. ' --manifest-path ' .. util.path.join(cargo_crate_dir, 'Cargo.toml')
+        cmd[#cmd + 1] = '--manifest-path'
+        cmd[#cmd + 1] = util.path.join(cargo_crate_dir, 'Cargo.toml')
       end
-      local cargo_metadata = vim.fn.system(cmd)
+      local cargo_metadata = ''
+      local cargo_metadata_err = ''
+      local cm = vim.fn.jobstart(cmd, {
+        on_stdout = function(j, d, e)
+          cargo_metadata = table.concat(d, '\n')
+        end,
+        on_stderr = function(j, d, e)
+          cargo_metadata_err = table.concat(d, '\n')
+        end,
+        stdout_buffered = true,
+        stderr_buffered = true,
+      })
+      if cm > 0 then
+        cm = vim.fn.jobwait({ cm })[1]
+      else
+        cm = -1
+      end
       local cargo_workspace_dir = nil
-      if vim.v.shell_error == 0 then
+      if cm == 0 then
         cargo_workspace_dir = vim.fn.json_decode(cargo_metadata)['workspace_root']
+      else
+        vim.notify(
+          string.format('cmd [%q] failed:\n%s', table.concat(cmd, ' '), cargo_metadata_err),
+          vim.log.levels.Warning
+        )
       end
       return cargo_workspace_dir
         or cargo_crate_dir


jonhoo:
Arguably, we shouldn't be using `system` in the first place, and instead something more like `execve` that avoids argument parsing and gives us control over head stream of the spawned process individually, but that doesn't appear simple in Lua land (there's [`io.popen`](http://www.lua.org/manual/5.1/manual.html#pdf-io.popen), but it doesn't seem like it has any knobs).

mjlbach:
If/when plenary jobs are moved into core, we could use the wrappers around uv_spawn. Otherwise could now probably use `jobstart`/`jobwait`

jonhoo:
@mjlbach I've updated the code to use `jobstart`/`jobwait` instead.

mjlbach:
Can you run stylua to fix style lint? Thanks!

jonhoo:
I don't have that installed, but made the fixes suggested by CI 😅

jonhoo:
Separately, I've observed (with both `system` and `jobwait`) that if `cargo metadata` takes a long time to execute, which it sometimes does for a never-built project, that prevents Neovim from drawing the opened buffer. I'm guessing that's because the code in this callback gets executed synchronously. Not sure if there's currently a workaround for that, or whether these necessarily have to be synchronous given the current architecture?

jonhoo:
Seems it didn't like `rust_analyzer` as the module name, even though that's what it is in the code 🤔 
Updated the commit message.

jonhoo:
@mjlbach Could you poke CI again please?

jonhoo:
I don't understand the complaint from the commit linter here. It says that the subject (`rust-analyzer`) may not be sentence-case, but what would you prefer instead? Should it not be the module name?

jonhoo:
I pushed a new commit that just uses `rust` as the subject instead, which I assume will then work, but that particular policy seems perhaps misguided in this instance?

mjlbach:
It's the capital S it doesn't like

jonhoo:
Oh! I think it's the reference to `sentence-case` that threw me off. Fixed (and restored `rust-analyzer`) in the latest commit. Thanks!

mjlbach:
> Separately, I've observed (with both system and jobwait) that if cargo metadata takes a long time to execute, which it sometimes does for a never-built project, that prevents Neovim from drawing the opened buffer.

Currently we don't have a good way to do this. I would like to asyncinfy the entire start process so that the language server will just start/attach eventually,; that might be easier with plenary jobs, but I'm also working on avoiding any async blocking function calls in the lspconfig rewrite/prototype I am working on.

jonhoo:
So, [custom rustup toolchains](https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains) let you completely replace which `cargo`/`rustc`/etc gets invoked when run by the user in a particular directory (usually project). Most commonly, this is used to run additional logic before or after Cargo runs, which may include running _other_ tools (like other build tools), which themselves may print to stderr. Not to mention that `cargo metadata` could _itself_ print to stderr, such as if it wants to issue a warning about a newer JSON format version being available.

I'd be happy to write up some variant of ^ into the commit message tomorrow, but really there are a whole host of reasons why `cargo metadata` may print to STDERR, and no one case is more important than the others.

d0u9:
image

jonhoo:
The code ignores errors from `cargo metadata` because it has sensible fallbacks, but still prints the error so that you can debug if you _didn't_ expect it to fail. It sounds like in your case you want a way to suppress the warning? Not entirely sure what precedent is for something like that — that'd be a new feature rather than fixing a bug :)

PR #1240, opened 2021-09-09
fix(texlab): update to handler signature change

clason:
Update the handlers for `textDocument/build` and `textDocument/forwardSearch` to the upstream breaking signature change.

### DIFF
--- a/lua/lspconfig/texlab.lua
+++ b/lua/lspconfig/texlab.lua
@@ -18,27 +18,26 @@ local texlab_forward_status = vim.tbl_add_reverse_lookup {
 
 -- add compatibility shim for breaking signature change
 local function mk_handler(fn)
-   return function(...)
-     local config_or_client_id = select(4, ...)
-     local is_new = type(config_or_client_id) ~= 'number'
-     if is_new then
-       fn(...)
-     else
-       local err = select(1, ...)
-       local method = select(2, ...)
-       local result = select(3, ...)
-       local client_id = select(4, ...)
-       local bufnr = select(5, ...)
-       local config = select(6, ...)
-       fn(err, result, { method = method, client_id = client_id, bufnr = bufnr }, config)
-     end
-   end
- end
+  return function(...)
+    local config_or_client_id = select(4, ...)
+    local is_new = type(config_or_client_id) ~= 'number'
+    if is_new then
+      return fn(...)
+    else
+      local err = select(1, ...)
+      local method = select(2, ...)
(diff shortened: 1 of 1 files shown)


mjlbach:
We still need to maintain backwards compat/should wrap these: https://github.com/neovim/nvim-lspconfig/pull/1233#issuecomment-914633260

clason:
Isn't that a bit tedious to do for _every_ server? Can't we just add a wrapper in `lspconfig/util.lua`?

clason:
Specifically, upstream https://github.com/mfussenegger/nvim-lsp-compl/commit/29a81f3747373f42d002775b1f79dcd19fada2e4 (@mfussenegger)?

mfussenegger:
Not sure if it is relevant for the handlers here but in nvim-jdtls I discovered that I had some handlers that require a response, so I had to adapt the shim:

```
@@ -7,19 +7,19 @@ function M.mk_handler(fn)
   return function(...)
     local count = select('#', ...)
     local config_or_client_id = select(4, ...)
     local is_new = type(config_or_client_id) ~= 'number' or count == 4
     if is_new then
-      fn(...)
+      return fn(...)
     else
       local err = select(1, ...)
       local method = select(2, ...)
       local result = select(3, ...)
       local client_id = select(4, ...)
       local bufnr = select(5, ...)
       local config = select(6, ...)
-      fn(err, result, { method = method, client_id = client_id, bufnr = bufnr }, config)
+      return fn(err, result, { method = method, client_id = client_id, bufnr = bufnr }, config)
     end
   end
 end
 ```

clason:
Any thoughts on moving this to `util.lua` and using this wherever it's needed? I can do this in a followup PR (tomorrow or this weekend).

kylo252:
> Any thoughts on moving this to `util.lua` and using this wherever it's needed? I can do this in a followup PR (tomorrow or this weekend).

I was just planning on doing the same thing before I saw your comment. It's currently needed/required in `clangd`, `deno` and `texlab`.

mjlbach:
> Any thoughts on moving this to `util.lua` and using this wherever it's needed? I can do this in a followup PR (tomorrow or this weekend).
> 
> 
> 
> That would make it easier to rip out again once 0.5.1 (or whatever release contains the handler change) is released.

@clason thats what my comment I linked to suggested :)

clason:
Ah, didn't notice the "util" there :) Go ahead and merge this; I'll tackle this tomorrow then (unless somebody else wants to do it quicker).

PR #1247, opened 2021-09-10
fix: ClangdSwitchSourceHeader for current nightly

### LINKED ISSUE #1239: ClangdSwitchSourceHeader causes lua error on Nvim 0.6.x due to signature change of vim.lsp.buf_request
- `nvim --version`: NVIM v0.6.0-dev+227-gc1f573fbc
- nvim-lspconfig version(commit hash): 01153c5
- What language server (If the problem is related to a specific language server): clangd
- Can you reproduce this behavior on other language server clients (vscode, languageclient-neovim, coc.nvim, etc.): -
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? (pyls -> pyright): -
- Is the problem isolated to a particular language server: vim.lsp.buf_request also appears in rust_analyzer and texlab, so I imagine the same problem exists there.
- Operating system/version: Linux

- Go to a directory containing a C++ project with two files, test.h and test.cpp
- Open test.h
- When the language server has successfully attached, run `:ClangdSwitchSourceHeader`

### Actual behaviour
The error `Error executing vim.schedule lua callback: .../tools/neovim/install/share/nvim/runtime/lua/vim/uri.lua:96: attempt to call method 'match' (a nil value)` is raised

### Expected behaviour
The file test.cpp is opened for editing.

### Minimal init.vim or init.lua and code sample
```
require('packer').startup(function(use)
  use 'wbthomason/packer.nvim'
  use {
    'neovim/nvim-lspconfig',
    config = function()
      local lspconfig = require('lspconfig')
      lspconfig.clangd.setup({
        cmd = { 'clangd', '--background-index'},
      })
  }
end)
```

### Health check
<details>
<summary>Checkhealth result</summary>

```
health#lspconfig#check
  - INFO: clangd: configuration checked.
```
</details>

### LSP log
I'll skip this.

## Diagnosis
Apparently, the signature of `vim.lsp.buf_request` has changed from the 0.5.x release to the current 0.6.x nightly build. In particular, the function handler signature is now `function(err, result, ctx, config)` (see `:help lsp-handler`), while it was `function(err, method, result, client_id, bufnr, config)` in 0.5.x.

As a result, the function `switch_source_header` in `clangd.lua` passes the `ctx` param to `vim.uri_to_fname` instead of the intended `result` param, which leads to aforementioned error.

Conni2461:
Closes #1239

0.5 is still supported in this solution. I dont know how to handle compatibility with 0.5 and nightly here. If this is wrong feel free to close it :)

### DIFF
--- a/lua/lspconfig/clangd.lua
+++ b/lua/lspconfig/clangd.lua
@@ -5,7 +5,14 @@ local util = require 'lspconfig/util'
 local function switch_source_header(bufnr)
   bufnr = util.validate_bufnr(bufnr)
   local params = { uri = vim.uri_from_bufnr(bufnr) }
-  vim.lsp.buf_request(bufnr, 'textDocument/switchSourceHeader', params, function(err, _, result)
+  vim.lsp.buf_request(bufnr, 'textDocument/switchSourceHeader', params, function(err, result_1, result_2)
+    local result = (function()
+      if type(result_1) == 'table' then
+        return result_1
+      end
+      return result_2
+    end)()
+
     if err then
       error(tostring(err))
     end


clason:
There's a compatibility shim in https://github.com/neovim/nvim-lspconfig/pull/1240 which I was going to move to `util.lua` and use for all applicable servers (certainly clangd). If it's not urgent, I'd suggest waiting until this afternoon.

(You can also do it yourself if you prefer -- if _I_ can do it, anybody can ;) But at this point I think it's better to make on big push to fix all servers rather than one at a time.)

Conni2461:
Yeah makes sense. I just fixed it locally and thought why not open a PR :rofl:

mjlbach:
Yes, can you move what clason added in his PR to utils like I asked for on that PR and adjust that file, and this file?

clason:
Too late: https://github.com/neovim/nvim-lspconfig/pull/1248 :P

clason:
Sorry for stealing your thunder ;) I've made some changes to my PR; can you test again?

Conni2461:
All good :)

1248 seems to work on `v0.6.0-dev+223-g384a43fe7` (havent build today)

PR #1249, opened 2021-09-11
Feature/zk

wurosh:
#1222 but properly rebased

### DIFF
--- a/lua/lspconfig/zk.lua
+++ b/lua/lspconfig/zk.lua
@@ -1,0 +1,43 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+configs.zk = {
+  default_config = {
+    cmd = { 'zk', 'lsp' },
+    filetypes = { 'markdown' },
+    root_dir = util.root_pattern '.zk',
+  },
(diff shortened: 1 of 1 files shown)


mjlbach:
@wurosh how exactly are you using zk.new and init?

wurosh:
`zk.index` updates the internal database to match what's on disk when you do some things manually (such as create a file the old-fashioned way). `zk.new` is a way for you to create a new note and let zk handle all the nasty stuff like hashes and respecting naming conventions you set up in your `config.toml`. Personally, I wrap it in a utility function that does so based on the link at point, but I didn't add that here because:
1. Some people use the built-in markdown plugin
2. Some people use `plasticboy/vim-markdown`
3. Some use wiki links instead of markdown links
4. Some people use a different format altogether

mjlbach:
Thanks! My question was more why you set the references to the functions as entries in the config table rather than using our commands interface (I'm doing a big refactor of configs and this stuck out to me)

wurosh:
I wasn't aware of the commands interface but I'm happy to do a refactor. Any good examples to look at?

PR #1251, opened 2021-09-11
feat(csharp_ls): add support for C# using csharp-ls language server

razzmatazz:
As on the tin

I used fsautocomplete code as a template for csharp-ls.. so I am not 100% if I adhere to project standarts for PRs.

### DIFF
--- a/lua/lspconfig/csharp_ls.lua
+++ b/lua/lspconfig/csharp_ls.lua
@@ -1,0 +1,26 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'csharp_ls'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'csharp-ls' },
+    root_dir = util.root_pattern('*.sln', '*.csproj', '.git'),
+    filetypes = { 'cs' },
+    init_options = {
+      AutomaticWorkspaceInit = true,
+    },
+  },
+  docs = {
+    description = [[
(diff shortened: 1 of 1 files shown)


mjlbach:
I was just thinking about this as I was struggling to debug a user's issues with omnisharp (which I would love to pick your brain about on matrix given your expertise :) )

razzmatazz:
> I was just thinking about this as I was struggling to debug a user's issues with omnisharp (which I would love to pick your brain about on matrix given your expertise :) )

No problem, you can bug me about it, however I am just a random guy trying to make/fix C# tooling for myself (and others if that goes along the way) by no means I am an expert on C# or LSP :)

mjlbach:
The main issue we've run into is it seems `textDocument/didOpen` is not sufficient for new files to be recognized by omnisharp, the new files show missing imports (even if identical to a previous file opened by the project). We don't send  `workspace/didCreateFiles` notifications, which lsp-mode does, which I thought might be the issue, but I wouldn't expect this to be required.

razzmatazz:
> The main issue we've run into is it seems `textDocument/didOpen` is not sufficient for new files to be recognized by omnisharp, the new files show missing imports (even if identical to a previous file opened by the project). We don't send `workspace/didCreateFiles` notifications, which lsp-mode does, which I thought might be the issue, but I wouldn't expect this to be required.

Have you tried it with csharp-ls -- does it work better? csharp-ls is alpha-quality s/w but maybe things work better?

PR #1253, opened 2021-09-12
chore: improve issue templates

clason:
* The feature request template now has to be in  `ISSUE_TEMPLATE`.
* Add a link to the Neovim Discourse for support questions.
* Disable blank issue templates to prevent support "issues".
* Turn issue template into Github Form.


mjlbach:
2 comments looking at this now that it's merged:

1. Maybe we should do the same for feature requests
2. Maybe we should link directly to the language server subsection of the discourse for quesitons

clason:
3. Sneak an `@mjlbach` somewhere in the prefilled text :P 

I was thinking the same thing, actually, but wanted to take it one step at a time.

Specifically, I couldn't really think of any useful (mandatory or optional) fields for feature requests?

clason:
Maybe we should rename "Language servers" to "LSP (Language Server Protocol)" to make it more inclusive (and easier to find since according to the dashboard, most people search for `lsp`)?

mjlbach:
I changed it, but I made it Language Server Protocol (LSP), I don't like people using the shorthand as it's often used incorrectly

PR #1254, opened 2021-09-12
feat: add missing package.json entries

kylo252:
# Description

Dynamically retrieve [missing] `package.json` URLs for all the configurations. 

This is a simplified break-down of the process:

1. parse the official list of [implemented language servers](https://microsoft.github.io//language-server-protocol/implementors/servers), currently using a Lua table (`servers.lua`) that was converted from the markdown page.
2. get a list of all servers with TypeScript implementation
3. use `github` API to retrieve the download URL for the `package.json` file.
4. check against all the configurations with `package_json` attribute in their doc table.
5. get the list of missing matches

I'm not yet settled on what to do with the output, but I think the API should be pretty flexible enough at this point that it should, hopefully, be useful.

### How to run

```shell
bash scripts/package_json_gen.sh
```

### List of the confiurations with missing `package_json` entries

<details>
<summary>Details</summary>
https://raw.githubusercontent.com/admvx/as2-language-support/main/package.json
https://raw.githubusercontent.com/XVincentX/vscode-apielements/master/package.json
https://raw.githubusercontent.com/bash-lsp/bash-language-server/master/package.json
https://raw.githubusercontent.com/rokucommunity/brighterscript/master/package.json
https://raw.githubusercontent.com/phil294/coffeesense/master/package.json
https://raw.githubusercontent.com/SPYGlassMC/SPYGlass/main/package.json
https://raw.githubusercontent.com/rcjsuen/dockerfile-language-server-nodejs/master/package.json
https://raw.githubusercontent.com/elm-tooling/elm-language-server/main/package.json
https://raw.githubusercontent.com/ember-tooling/ember-language-server/master/package.json
https://raw.githubusercontent.com/sourcegraph/javascript-typescript-langserver/master/package.json
https://raw.githubusercontent.com/jonnyboyC/kos-language-server/master/package.json
https://raw.githubusercontent.com/tintoy/msbuild-project-tools-vscode/master/package.json
https://raw.githubusercontent.com/openvalidation/openvalidation-languageserver/master/package.json
https://raw.githubusercontent.com/bmewburn/vscode-intelephense/master/package.json
https://raw.githubusercontent.com/recca0120/vscode-phpunit/master/package.json
https://raw.githubusercontent.com/Polymer/polymer-editor-service/master/package.json
https://raw.githubusercontent.com/OGoodness/VSCode-PowerPC-Syntax/main/package.json
https://raw.githubusercontent.com/rescript-lang/rescript-vscode/master/package.json
https://raw.githubusercontent.com/tomi/vscode-rf-language-server/master/package.json
https://raw.githubusercontent.com/jamesbirtles/svelte-language-server/master/package.json
https://raw.githubusercontent.com/iamcco/vim-language-server/master/package.json
https://raw.githubusercontent.com/chemzqm/wxml-languageserver/master/package.json
https://raw.githubusercontent.com/adamvoss/vscode-yaml-languageservice/master/package.json
</details>


mjlbach:
What is the advantage of adding this complexity vs manually checking on merge there is a package.json? We also validate that these are somewhat reasonable and not entirely vscode specific.

kylo252:
> What is the advantage of adding this complexity vs manually checking on merge there is a package.json? We also validate that these are somewhat reasonable and not entirely vscode specific.

It doesn't necessarily need to go into the repo at its current form, but here are some of the Pros:
- It gave us a static list of what's missing, which I think should be good enough on its own merits.
- This will actually allow you to drop that hard-coded `package_json` or at least find a way to check if they get outdated every once in a while. (if you use the script in the CI, for example)
- This is only matching against `package.json`, but it's actually more flexible that we can use it to get the binary names from any other mainstream packaging manifest, e.g. Python(pip), Ruby(gem), Rust(crate), Go(pkg).

scripts/servers.lua
```diff
@@ -0,0 +1,827 @@
+return {
+  {
+    '1C Enterprise',
+    '[Contributors](https://github.com/1c-syntax/bsl-language-server/graphs/contributors)',
+    '[BSL Language Server](https://github.com/1c-syntax/bsl-language-server)',
+    'Java',
```
justinmk, line 6: if we actually want this info, it should live in the configs. Do not introduce a new, parallel source of truth.

kylo252: Yeah I fully agree, that's why I said I'm not sure how and if this table should live in the repo. I added it here to show a "parsable" version the list of this https://github.com/Microsoft/language-server-protocol/blob/gh-pages/_implementors/servers.md, and to give some suggestions about maybe creating an "update" script that can validate these URLs regularly.

Furthermore, there are about 100 servers that are currently supported by lspconfig, I would imagine going through these 100 files and updating the entries by hand to be not only tedious, but also error prone. But maybe reviewing it will take more effort 🤷‍♂️ 

I actually wanted to have the markdown table parsed from directly from the URL but I spent way too much time trying to figure out a reasonable way to do it before I gave up because it takes like 10s to fix it with a macro instead of setting up a bunch of libraries to do it. The rabbit holes can start with either https://github.com/jgm/lunamark or https://github.com/msva/lua-htmlparser if anyone would like to explore.

kylo252:
I have no idea why github is insisting I have added all those commits. Anyway, I wanted to show some big progress with this!

Take a look at the example from `phpactor.lua` and how we can generate all this server-info and feed it into `docgen.lua` with hardly any changes to the latter.

lua/lspconfig/util.lua
```diff
@@ -373,4 +373,24 @@ function M.get_other_matching_providers(filetype)
...
+  end)
+end
+
+function M.read_async(path)
+  local fd = assert(vim.loop.fs_open(path, "r", 438))
+  local stat = assert(vim.loop.fs_fstat(fd))
+  local data = assert(vim.loop.fs_read(fd, stat.size, 0))
+  assert(vim.loop.fs_close(fd))
+  return data
+end
```
mjlbach, line 394: this is not async.

mjlbach:
I still do not understand the point of this PR. We already have method/doc generation pipeline built around explicitly annotating the configurations with validated package.json's. I don't want to add another source of truth, and I don't understand what you are trying to accomplish.

lua/lspconfig/util.lua
kylo252: > I still do not understand the point of this PR. We already have method/doc generation pipeline built around explicitly annotating the configurations with validated package.json's. I don't want to add another source of truth, and I don't understand what you are trying to accomplish.

I should've done another example with an already configured server. The idea is to replace the `package_json` with the `url` entry. The file with Lua table `servers.lua` is basically a cache file that isn't going to leave this PR.

mjlbach:
@mergfiy rebase

kylo252:
Here's the main and (only) command that is needed, the same used https://github.com/kylo252/nvim-lspconfig/blob/d6e5dd99c8532d307d5da029cd20831f6ef7a980/lua/lspconfig/util/server-info.lua#L120:

```shell
gh api repos/bmewburn/vscode-intelephense/contents/package.json --jq .download_url | xargs curl -LSs | jq 'if has("contributes") then .contributes else .configuration end'
```

The code that you see is used to prepare a **reference list of valid servers** that have **valid** `package.json` that you can then selectively add to each server's config.

How do you think I found out about `phractor` from `bmewburn/vscode-intelephense`? why doesn't it already have an entry? is it not just waiting for someone to figure it out?

There are probably 10 or more of those servers.

PR #1256, opened 2021-09-12
Fix svelteserver on Windows

al3xfischer:
On Windows the call to the LS fails with the msg: 'E5108: Error executing lua ...ps\neovim\current\share\nvim\runtime\lua\vim\lsp\rpc.lua:384: start `svelteserver` failed: ENOENT: no such file or directory'.

It works quite fine if I use the custom cmd "svelterserver.cmd".
Other LS (e.g. pyright) also use this check.


mjlbach:
We should probably add this to utils (wrap_cmd_windows) and just call this for all the servers we run into instead of adding the check everywhere.

PR #1260, opened 2021-09-14
Introduce psalm language server support

gbprod:
This PR introduce [Psalm](https://psalm.dev) language server support.

### DIFF
--- a/lua/lspconfig/psalm.lua
+++ b/lua/lspconfig/psalm.lua
@@ -1,0 +1,27 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'psalm'
+local bin_name = 'psalm-language-server'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { bin_name },
+    filetypes = { 'php' },
+    root_dir = util.root_pattern('psalm.xml', 'psalm.xml.dist'),
+  },
+  docs = {
+    description = [[
+https://github.com/vimeo/psalm
+
+Can be installed with composer.
+```sh
+composer global require vimeo/psalm
+```
+]],
+    default_config = {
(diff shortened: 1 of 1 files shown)


gbprod:
Just seen tha there already was a [PR](https://github.com/neovim/nvim-lspconfig/pull/1214) maybe I should close this one ?

lua/lspconfig/psalm.lua
```diff
@@ -0,0 +1,28 @@
...
+local util = require 'lspconfig/util'
+
+local server_name = 'psalm'
+local bin_name = 'psalm-language-server'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { bin_name },
+    filetypes = { 'php' },
+    root_dir = util.root_pattern('psalm.xml', 'psalm.xml.dist'),
```
lithammer, line 11: What about `composer.json`? Noticed the other pull request has it. Probably makes sense to add a fallback to `util.find_git_ancestor(fname)` as well.

gbprod: I don't think that composer or git fallback are required because the language server will not works if the psalm.xml[.dist] file doesn't exists. We will get this error  :

    $ psalm-language-server
    Could not locate a config XML file in path /home/gilles/.gbvim/. Have you run 'psalm --init' ?

If we look at the [VS Code plugin](https://github.com/psalm/psalm-vscode-plugin/blob/25d12f47f22a1a2ae18b0db17a5120c00014f0e5/src/ConfigurationService.ts#L55) and the [coc-psalm](https://github.com/yaegassy/coc-psalm), it works like this.

lithammer: Cool, nevermind then! 👍

reisnobre:
Hi guys, how plugins would/should work with psalm while using it like this?

PR #1268, opened 2021-09-19
Gracefully abort on incompatible Neovim versions

amarshall:
This is helpful when trying to use the same configuration across many machines, but being, for whatever reason, stuck with an older version somewhere.

I understand this behavior might be confusing because it’s “silently failing” in a way, so certainly open to adjusting to have a clear error message or requiring a variable setting to opt-in to silently ignoring.

### DIFF
--- a/plugin/lspconfig.vim
+++ b/plugin/lspconfig.vim
@@ -2,6 +2,9 @@ if exists('g:lspconfig')
   finish
 endif
 let g:lspconfig = 1
+if !has("nvim-0.5")
+  finish
+endif
 
 lua << EOF
 lsp_complete_configured_servers = function()


plugin/lspconfig.vim
```diff
@@ -2,6 +2,9 @@ if exists('g:lspconfig')
   finish
 endif
 let g:lspconfig = 1
+if !has("nvim-0.5")
+  finish
```
justinmk, line 6: This is not needed. The conventional pattern with Vim plugins, is to `let g:lspconfig = 1` in your configuration, if you have need for this.

amarshall: Didn’t really think of that, seems fine. Feel free to close.

PR #1273, opened 2021-09-23
feat: add eslint language server

entropitor:


lua/lspconfig/eslint.lua
```diff
@@ -0,0 +1,119 @@
...
+    filetypes = {
+      'javascript',
+      'javascriptreact',
+      'javascript.jsx',
+      'typescript',
+      'typescriptreact',
+      'typescript.tsx',
+    },
+    root_dir = util.root_pattern('.eslintrc.json', '.eslintrc.js', 'package.json', 'tsconfig.json', '.git'),
+    settings = {
```
mjlbach, line 56: Are all of these settings necessary?

David-Else:
Maybe the server could be added to https://github.com/hrsh7th/vscode-langservers-extracted ? @hrsh7th is keeping that updated nicely, and it would be very convenient :)

hrsh7th:
I've added vscode-eslint-language-server to vscode-langservers-extracted. (also other servers upated now)

entropitor:
Thanks @hrsh7th! I've updated the PR to reference that package as well now, like the other servers in nvim-lspconfig

entropitor:
@gegoune 
1. This is an LSP and as far as I can see `eslint_d` is not
2. The reason I want this over ALE or the like is that I was jealous of VS Code peeps having a code action to disable a rule / apply a code fix for a specific diagnostic error. I think that's the main benefit of using this LSP

williamboman:
> Neither of that is happening with this LSP. Do I need to configure that somehow?

I think I've found some issues with the provided settings, will open a PR soon, just have to clean some things up and refresh my memory of what each setting actually does

PR #1275, opened 2021-09-23
style: format files with stylua

lithammer:
Pull requests are currently failing because these files don't pass the StyLua lint check.

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -102,7 +102,14 @@ end
 configs[server_name] = {
   default_config = {
     cmd = { 'deno', 'lsp' },
-    filetypes = { 'javascript', 'javascriptreact', 'javascript.jsx', 'typescript', 'typescriptreact', 'typescript.tsx' },
+    filetypes = {
+      'javascript',
+      'javascriptreact',
+      'javascript.jsx',
+      'typescript',
+      'typescriptreact',
+      'typescript.tsx',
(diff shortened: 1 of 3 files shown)


lithammer:
I suspect this is because of the new StyLua v0.11.0 release...?

JohnnyMorganz:
> I suspect this is because of the new StyLua v0.11.0 release...?

Hey, someone pointed this issue out to me in passing, and mentioned that their CI suddenly fails whenever stylua updates, pointing out yours as well.
I naively created the action with the `version` field being optional in the workflow.yml, so by default the action runs using the latest build - this can be annoying when CI suddenly fails like this 🤦!

I'd recommend pinning the version in your `lint.yml` file, to have something like
```yml
  style-lint:
    runs-on: [ubuntu-latest]
    steps: 
    - name: Checkout sources
      uses: actions/checkout@v2
    - name: Lint with stylua
      uses: JohnnyMorganz/stylua-action@1.0.0
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
        version: 0.11.0
        # CLI arguments
        args: --check .
```
The disadvantage being that you'll have to manually change it as you bump your stylua version for new updates, but I think thats better than letting it auto update as current.
I tried to make the README more obvious in the action repo but since the workflow is already out in the wild most workflows have the version field omitted.
Apologies for the hassle (to you and anyone else reading!)

Issue #1278, opened 2021-09-24
gopls can't start if not have go.mod and .git


fcying:
### Description

sometimes I write single go file as a script.
if there not have go.mod and .git, gopls can't start.

### Neovim version

NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system and version

debian 11

### Affected language servers

gopls

### Steps to reproduce

1. nvim a.go      # gopls not start
2. :q
3. mkdir .git
4. nvim a.go     #gopls start

### Actual behavior

gopls not start.

### Expected behavior

start gopls when I edit a go file

### Minimal config

```Lua
require'lspconfig'.gopls.setup{}
```

### Health check

health#lspconfig#check
========================================================================
## Checking language server protocol configuration
  - INFO: vimls: configuration checked.
  - INFO: clangd: configuration checked.
  - INFO: pyright: configuration checked.
  - INFO: gopls: configuration checked.

### LSP log

https://gist.github.com/fcying/dfb09a53b69231ba4fda8bb1f7660f09

justinmk:
this is a general problem. some configs for some lang servers try to address it, but see also https://github.com/neovim/nvim-lspconfig/issues/1221

> The LSP specification does afford not any particular distinction between editing single files and editing a file in a project

send a PR for the gopls config

PR #1279, opened 2021-09-24
gopls root_dir add dirname for support single go file.

fcying:
for #1278


mjlbach:
I'm trying to remove these fallbacks (or at least not proliferate them). Lspconfig only supports projects for now.

PR #1286, opened 2021-09-28
Nimls: state the correct default root directory in CONFIG.md

gwerbin:


mjlbach:
Fails style lint, also this can be autogenerated if removed (pulls the lines of code)

PR #1288, opened 2021-09-28
fix(json): binary was renamed to vscode-json-languageserver

joelazar:


mjlbach:
...are you sure? https://github.com/hrsh7th/vscode-langservers-extracted

joelazar:
> ...are you sure? https://github.com/hrsh7th/vscode-langservers-extracted

Yeah, it looks like the binary was renamed in that package, but it's `vscode-json-languageserver` everywhere else:
https://archlinux.org/packages/community/any/vscode-json-languageserver/
https://github.com/vscode-langservers/vscode-json-languageserver

+ it looks like that npm package is more popular than hrsh7th's fork:
https://www.npmjs.com/package/vscode-json-languageserver
vs.
https://www.npmjs.com/package/vscode-langservers-extracted

mjlbach:
Right, but we officially recommend hrsh7th's npm package. The package you're referring to on npm is extremely out of date.

polyzen:
I wonder why @hrsh7th decided to rename the binaries. It would be nice to get them changed back, and wouldn't take too much effort downstream. I personally created symlinks to make lspconfig happy.

mjlbach:
I'd file an upstream issue and ask, these work well and resolved issues people had with the stale binary distribution so I'm hesitant to change anything

polyzen:
Welp the issue tracker for his repo has been disabled.

hrsh7th:
It's a bad idea that different packages provide the same bin name.

I also closed the issue tracker.
I'm just extracting the server and I think the problem should be reported upstream.

If you think `should provide same name bin`, you can create your own one (because it's just only simple scripts).

polyzen:
>  I'm just extracting the server and I think the problem should be reported upstream.

The names chosen by upstream are `vscode-json-languageserver` and such:
https://github.com/microsoft/vscode/blob/main/extensions/json-language-features/server/package.json#L11

polyzen:
> If you think `should provide same name bin`, you can create your own one (because it's just only simple scripts).

I'm the maintainer for the [vscode-languageservers Arch packages](https://archlinux.org/packages/community/any/vscode-languageservers/). Having different binary names gets in the way of tools that  provide out-of-the-box support, like lspconfig. You have done a great service of providing the popular vscode-langservers-extracted npm package once vscode-langservers's server binaries went unmaintained. These tools use the name of the binaries you chose, but which differ from what upstream had chosen.

PR #1291, opened 2021-09-30
fix: add single_file_support for dockerls

kylo252:
Allow `dockerls` to attach to a dockerfiles regardless if the default `Dockerfile` is found.
This is useful for editing files with custom names, e.g. `Dockerfile.release`

### DIFF
--- a/lua/lspconfig/dockerls.lua
+++ b/lua/lspconfig/dockerls.lua
@@ -9,6 +9,7 @@ configs[server_name] = {
     cmd = { bin_name, '--stdio' },
     filetypes = { 'dockerfile' },
     root_dir = util.root_pattern 'Dockerfile',
+    single_file_support = true,
   },
   docs = {
     description = [[


lua/lspconfig/dockerls.lua
```diff
@@ -8,7 +8,9 @@ configs[server_name] = {
   default_config = {
     cmd = { bin_name, '--stdio' },
     filetypes = { 'Dockerfile', 'dockerfile' },
-    root_dir = util.root_pattern 'Dockerfile',
+    root_dir = function(fname)
+      return util.root_pattern 'Dockerfile'(fname) or util.path.dirname(fname)
```
lithammer, line 12: I was thinking. Is there a point in setting this to anything but `util.path.dirname(fname)`?

kylo252: Not sure tbh.
I just went through their docs over at https://github.com/rcjsuen/dockerfile-language-server-nodejs, and I can't seem to find anything that seems dependent on this.

kylo252:
tested with `release.Dockerfile` according to the rules here:
https://github.com//neovim/neovim/blob/29cec32f44db358b0dfcd560d2ba8228b9d9de39/runtime/filetype.vim#L396

PR #1295, opened 2021-10-04
add emmet_ls

williamboman:
Ping @aca 👋 :)


lua/lspconfig/emmet_ls.lua
```diff
@@ -0,0 +1,27 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'emmet_ls'
```
williamboman, line 4: I followed the name suggested in https://github.com/aca/emmet-ls, but this kind of deviates from the convention of not having it snake-cased. Should this change to `emmetls` perhaps?

lua/lspconfig/emmet_ls.lua
```diff
@@ -0,0 +1,27 @@
...
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'emmet-ls', '--stdio' },
+    filetypes = { 'html', 'css' },
```
williamboman, line 9: Not sure if these filetypes should be expanded? What about XML and JSX `typescriptreact`, `javascriptreact`? @aca any thoughts?

aca: I think I need to check whether it is working fine on react project. 

By the way, thanks for opening the issue.

williamboman:
https://xkcd.com/927/

Would be nice if we could align on one going forward :D! What does VSCode use for their Emmet integration? Is it proprietary?

PR #1296, opened 2021-10-05
fix(checkhealth): rename check_health to check

teto:
as mandated by https://github.com/neovim/neovim/pull/15259#issuecomment-934241413

### DIFF
--- a/autoload/health/lspconfig.vim
+++ b/autoload/health/lspconfig.vim
@@ -1,4 +1,4 @@
 function! health#lspconfig#check()
   call health#report_start('Checking language server protocol configuration')
-  lua require 'lspconfig/health'.check_health()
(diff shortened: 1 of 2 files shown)


clason:
This will still run the `autoload#health` check, right? So you get duplicated output?

teto:
don't merge as we decide on the way forward.

@clason it will so maybe we can tweak neovim directly to run just one of the healthchecks instead as you mentioned.

justinmk:
why not delete `autoload#health` ? We would just need to wait until Nvim 0.5.2 is released.

clason:
So the Lua healthcheck will be backported?

justinmk:
> So the Lua healthcheck will be backported?

seems low risk, why not?

PR #1299, opened 2021-10-12
fix(eslint): provide extra settings to make it work outside of a VSCode context

williamboman:


williamboman:
> @williamboman Does autoformatting in eslint also fix all issues? Or does it only do prettier issues?

When I've been debugging the server, my understanding has been that `textDocument/formatting` and `eslint.applyAllFixes` currently produce the same results:
- https://github.com/microsoft/vscode-eslint/blob/6437dfff199c8f6fc7293056f1529b72e3e11eb4/server/src/eslintServer.ts#L2371-L2380
- https://github.com/microsoft/vscode-eslint/blob/6437dfff199c8f6fc7293056f1529b72e3e11eb4/server/src/eslintServer.ts#L2084-L2104
- https://github.com/williamboman/nvim-lsp-installer/issues/41#issuecomment-886302945

However, @dbaeumer <sup>:wave:</sup>  corrected me [here](https://github.com/microsoft/vscode-eslint/pull/1307#issuecomment-896639213), claiming that they should not be used interchangeably.

While taking the above into consideration, in my experience they are the same. I've never really gotten `eslint.applyAllFixes` to actually fix any issues - I have to go through regular code actions for that. Maybe your experience is different?

> If so, is there a guide on how to set up autoformatting on save?

To go through `textDocument/formatting`, you need to currently override the resolved client capabilities for `eslint`. This is because the ESLint server relies on dynamic registration to set this up (neovim LSP client does not support this). This is best done in the `on_attach` callback:

```lua
  client.resolved_capabilities.document_formatting = true
end
```

Once this has been set to true, neovim will start considering the eslint server as a candidate for `vim.lsp.buf.formatting*()` calls. There are three variants:
- `vim.lsp.buf.formatting()` and `vim.lsp.buf.formatting_sync()` will prompt you which attached server you want to issue the request to
- `vim.lsp.buf.formatting_seq_sync()` will issue the command in sequence (blocking calls) to each attached server that supports `textDocument/formatting`

You can incorporate this to your workflow as you see fit, but one example would be:

```vim
autocmd BufWritePre lua vim.lsp.buf.formatting_seq_sync()
```

gegoune:
@williamboman Yes, it works with manually set capabilities. Thanks for very clear write up and all work on this configuration.

NullVoxPopuli:
does this still work?

my on_attach is doing
```lua
        if client.name == 'eslint' then
          client.server_capabilities.document_formatting = true
        end
```
and i'm using the newer format api
```vim
nnoremap <leader>ff :lua vim.lsp.buf.format({ async = true })<CR>
```

entropitor:
@NullVoxPopuli You might also need to read https://github.com/jose-elias-alvarez/null-ls.nvim/issues/344#issuecomment-1114036678 as you might want to change the `resolved_capabilities.document_formatting` to `server_capabilities.documentFormattingProvider` in the latest neovim version (master, not released)

PR #1316, opened 2021-10-16
docs(volar): Add Take Over Mode docs, Add example for global typescript detection

sethidden:
Volar now has Take Over Mode(TM) which gives it ability to provide language features to .ts, .js files as tsserver does. https://github.com/johnsoncodehk/volar/discussions/471

Opening as draft as I'm not really sure if it's applicable to neovim. The Take Over Mode discussion mentions VSCode a lot, so I feel as if I need to think on it more, as neovim may not have the performance problems that VSCode has

Obviously I'll squash the commits once this PR has the green light

### DIFF
--- a/doc/server_configurations.md
+++ b/doc/server_configurations.md
@@ -6744,6 +6744,22 @@ require'lspconfig'.sumneko_lua.setup {
 This server accepts configuration via the `settings` key.
 <details><summary>Available settings:</summary>
 
+- **`Lua.IntelliSense.traceBeSetted`**: `boolean`
+
+  null
+
+- **`Lua.IntelliSense.traceFieldInject`**: `boolean`
+
+  null
+
+- **`Lua.IntelliSense.traceLocalSet`**: `boolean`
+
+  null
+
+- **`Lua.IntelliSense.traceReturn`**: `boolean`
+
+  null
+
 - **`Lua.color.mode`**: `enum { "Grammar", "Semantic", "SemanticEnhanced" }`
 
   Default: `"Semantic"`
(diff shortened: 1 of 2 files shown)


mjlbach:
The documentation in general seems a little verbose now, but I don't have concrete suggestions on how to cut it down.

sethidden:
Removed a few sections, let me know what do you think. Probably should've kept the previous commit so you could see the delta :/

doc/server_configurations.md
```diff
@@ -7695,14 +7695,50 @@ Volar can be installed via npm
...
+    return local_tsserverlib
+  else
+    return global_tsserverlib
+  end
+end
+
+require'lspconfig'.volar.setup{
+  config = {
+    on_new_config = function(new_config, new_root_dir)
+      new_config.init_options.typescript.serverPath = get_typescript_server_path(new_root_dir)
```
mjlbach, line 7736: Is there any advantage of automating this other than just having users fill this in directly? Adds a lot of code/ complexity for an example. I feel like hardcoding the string is easier.

sethidden: The global setting of TS helps when you open a .vue file that's not part of any project. If you created a .vue file in your home directory just to quickly draft something the default config will make the volar's binary throw an error and close because it didn't find TypeScript - it's unlikely there's node_modules with TypeScript in your home dir. Setting it globally fixes that problem.

The "use global if local TS not found" is useful when your global TS version is eg. super new, like TypeScript 4.5, but the one in your workspace is older - e.g. TypeScript 3.7.

mjlbach: Right, I was suggesting shifting the responsibility of setting this correctly per project onto the user rather than embedding the utility function :)

doc/server_configurations.md
```diff
@@ -7695,14 +7695,50 @@ Volar can be installed via npm
 npm install -g @volar/server
 ```
 
-With Vue 3 projects - it works out of the box.
+Vue 3 works by default, but Vue 2 needs [additional configuration](https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/README.md?plain=1#L28-L63)
```
mjlbach, line 7698: missing period

doc/server_configurations.md
```diff
@@ -7695,14 +7695,50 @@ Volar can be installed via npm
...
 
-With Vue 2 projects - requires [additional configuration](https://github.com/johnsoncodehk/volar#using)
+`vuels` and `volar` shouldn't be running at the same time
```
mjlbach, line 7700: avoid contractions in the documentation

mjlbach: Also maybe add explanation as to why they should not be running at the same time (don't call setup{} on both) etc.

sethidden: I've rewritten it with explanation and removed contraction

doc/server_configurations.md
sethidden: Users could theoretically set per-project init.lua files in their projects but I believe this is brittle and something only the non-mortal users of nvim do. 

I assume you don't want the CONFIG.md to baloon in size due to every LS providing their own user-experience improving tricks? I could:
1. just change the second bullet point (the one with the big code block) to something like "merge the existing TS detection function and the above global TS example to use global TS only when local TS is not found" and remove the codeblock
2. move the codeblock to a gist and link it in the docs

doc/server_configurations.md
```diff
@@ -7695,14 +7695,50 @@ Volar can be installed via npm
 npm install -g @volar/server
 ```
 
-With Vue 3 projects - it works out of the box.
+Vue 3 works by default, but Vue 2 needs [additional configuration](https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/README.md?plain=1#L28-L63).
```
mjlbach, line 7698: can you change "works by default" to something more explicitly. Works with volar I'd guess.

doc/server_configurations.md
```diff
@@ -7695,14 +7695,50 @@ Volar can be installed via npm
...
 
-With Vue 2 projects - requires [additional configuration](https://github.com/johnsoncodehk/volar#using)
+`volar` succeeds `vuels` (Vetur), so certain language features overlap between the two. To avoid getting duplicate completions, code actions etc. from two servers at once make sure `vuels` is not running alongside `volar`
```
mjlbach, line 7700: volar is the successor to? it's by the same author?

dundargoc:
> @dundargoc why is my autoclose server_configurations.md action not acting with impunity (even though this PR should not be closed)

Workflows aren't triggered by changes from github-actions. The `docgen.yml` works on the `master` branch but it doesn't take into account which `master`. Since @sethidden pushed to their own master then `docgen.yml` created the changes even though it was the wrong master.

doc/server_configurations.md
sethidden: Changed to

Volar in Vue 3 projects doesn't need any additional configuration and works out of the box. Vue 2 projects need [additional configuration](https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/README.md?plain=1#L28-L63).

doc/server_configurations.md
sethidden: Changed to

`volar` succeeds `vuels` (Vetur) as the main language server in the Vue ecosystem, so certain language features overlap between the two. To avoid getting duplicate completions, code actions etc. from two servers at once make sure `vuels` is not running alongside `volar`

mjlbach:
I pushed some changes here: https://github.com/neovim/nvim-lspconfig/pull/1448

PR #1319, opened 2021-10-17
update Vala Language Server plugin

Prince781:


mjlbach:
We added `single_file_support` as an option to `config`, which sends null when the server is started and we can't reliably detect a project pattern. I removed the `vim.fn.cwd` fallback and am merging your PR, but you can test if your server works with single_file_support = true and if so send a PR enabling that (which should have the desired effect).

PR #1321, opened 2021-10-17
deno_ls: Add new Deno configuration files to root_dir

David-Else:
This update is for the new configuration file format: https://deno.land/manual@v1.15.1/getting_started/configuration_file

### DIFF
--- a/lua/lspconfig/denols.lua
+++ b/lua/lspconfig/denols.lua
@@ -116,7 +116,7 @@ configs[server_name] = {
       'typescriptreact',
       'typescript.tsx',
     },
-    root_dir = util.root_pattern('package.json', 'tsconfig.json', '.git'),
+    root_dir = util.root_pattern('deno.json', 'deno.jsonc', 'package.json', 'tsconfig.json', '.git'),
     init_options = {
       enable = true,
       lint = false,


David-Else:
Whoops! That should be correct now.

David-Else:
@mjlbach @jason0x43 The deno config file `deno.json` or `deno.jsonc` is now automatically loaded in the latest version of deno https://deno.com/blog/v1.18#auto-discovery-of-the-config-file . I am sure that having a config file now will become extremely common. I don't think there is any point in looking for `package.json`, `tsconfig.json`, or `.git` anymore for the root pattern.

The official docs for Neovim https://deno.land/manual@v1.18.0/getting_started/setup_your_environment#neovim-06-and-nvim-lspconfig now recommend the following to seperate deno and node projects:
```
nvim_lsp.denols.setup {
  on_attach = on_attach,
  root_dir = nvim_lsp.util.root_pattern("deno.json"),
  init_options = {
    lint = true,
  },
}

nvim_lsp.tsserver.setup {
  on_attach = on_attach,
  root_dir = nvim_lsp.util.root_pattern("package.json"),
  init_options = {
    lint = true,
  },
}
```
I think if we made `util.root_pattern('deno.json', 'deno.jsonc',)` it would simplify things a lot.

I don't know how other people are using Deno, but I think this would make sense. Would be good to get feedback.

I also think adding `lint = true,` would be good, I am not sure why anyone would not want that by default. It can be turned off using comments in files.

jason0x43:
I can definitely see value in using a Deno config as a discriminator. It always bothered me that there wasn't an implicit way to tell a Deno project from a TS one, so that I'd often have both servers attach to TS files (before I made my config a bit smarter).

I'm not sure about  _only_ using a Deno config, though. I mean, it is a reasonable choice, but on the other hand, the TS language server doesn't only use `tsconfig.json` to denote the root. Deno also makes a big deal about not needing a config in most situations.

It'd be nice if there were exclude rules for the root dir, so that a directory with a tsconfig could explicitly _not_ be a Deno project by default.

David-Else:
@jason0x43 I hear what you are saying, but why would we want `package.json`, `tsconfig.json` or `.git` to denote root for deno? `tsconfig.json` has been obsoleted by `deno.json` for typescript settings, `package.json` is not used in deno, and `.git` adds nothing.

There has to be some way to say that it is a deno project, maybe even putting an empty `deno.json` is the best way to do this?

The new node compatibility  `--compat` mode makes deno work in a normal node folder which will probably have `package.json` etc, so there still needs to be a way to say it is a deno project, even when it is really a node project that is running deno as a replacement.

> It'd be nice if there were exclude rules for the root dir, so that a directory with a tsconfig could explicitly not be a Deno project by default.

What do you think the ultimate rules based solution would be, including running deno in a node project environment to replace node? Maybe that node project needs `package.json`, `tsconfig.json` deleting and converting to the deno way of things first, or maybe there is value in just using deno as a drop in replacement?

All this talk seem to be quite heavy, it really begs the question what is deno?! :)

jason0x43:
Currently, deno only looks for `deno.json`, `deno.jsonc`, `tsconfig.json`, or `.git` to determine the root. I think dropping `tsconfig.json` makes sense because Deno now has it's own preferred config file name. Having `.git` seems reasonable enough because projects will often be in git repos. I could even see enabling `single_file` mode by default for the Deno LS since making simple CLI tools is (theoretically) a standard use case.

My thought on "exclude rules" was simply that it would be nice if the presence of a `tsconfig.json` would indicate that a project was _not_ a Deno project so that the Deno server wouldn't try to attach to buffers in a TS project. That used to happen all the time for me, and it was kind of annoying.

In any case, I'm not a decision maker here, I just contributed an update at some point. 😄

PR #1328, opened 2021-10-19
Add tf filetype in terraformls and tflint default configs

kevinrambaud:
I was wondering why I had 0 client attached on my buffer although it was named `main.tf` and after checking what was in the `LspInfo` popup, it appears that's because the detected filetype is `tf` and it's not included in the default config of both terraformls and tflint so here it is!

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -7261,7 +7261,7 @@ require'lspconfig'.terraformls.setup{}
   
   Default Values:
     cmd = { "terraform-ls", "serve" }
-    filetypes = { "terraform", "tf" }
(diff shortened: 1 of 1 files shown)


lithammer:
This is a duplicate of https://github.com/neovim/nvim-lspconfig/pull/757. Neither Vim nor Neovim supports Terraform out of the box. You need [vim-terraform](https://github.com/hashivim/vim-terraform) or [vim-polyglot](https://github.com/sheerun/vim-polyglot) for that. The `tf` filetype is completely unrelated (see [TinyFugue](http://tinyfugue.sourceforge.net/)).

PR #1339, opened 2021-10-25
feat(lua-language-server): simplify configuration

mjlbach:
The vast majority of our users that are using lua-language-server, are using it for neovim development. I think we should ship the default configuration including the neovim runtime path. It's very easy for users who do *not* want this to opt-out. (just override the settings key).

Furthermore, `lua-language-server` is now the canonical command name, and it's getting packaged by more repositories. I personally packaged it for nix, and I know it's on the AUR now.

This should cut a good 20 lines out of everyone's config.


mjlbach:
I've soured on this until we have support for project local settings as a first-class option.

PR #1345, opened 2021-10-27
Update options key for dartls server

romariolopezc:
`init_options` didn't work, I tried settings and it work as expected


mjlbach:
@akinsho I'm adopting a subset of flutter-tools settings (thanks!) If you ever got bored and want to review our dart configuration/update it again please do 😅

PR #1352, opened 2021-10-31
change signature_help keymap to insert mode

sychen52:
I think `<C-k>` is intended to be used in the insert mode.


mjlbach:
This wasn't how I intended it, it's pretty easy for users to make their own judgement call on when/where they would like these keymaps mapped. The ones listed are just my own personal eccentricities.

PR #1353, opened 2021-10-31
Update the `on_attach` hook example

dmos62:
Previously the example had the user specifying their LSP servers manually. This change looks at available servers and attaches to all of them. This makes sense if it's more likely that the user will be interested in attaching this hook to all available servers, not just a subset of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,11 +123,11 @@ local on_attach = function(client, bufnr)
 
 end
 
--- Use a loop to conveniently call 'setup' on multiple servers and
+-- Use a loop to conveniently call 'setup' on all available servers and
 -- map buffer local keybindings when the language server attaches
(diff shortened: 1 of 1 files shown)


mjlbach:
This isn't going to work, because those servers are only listed in the available servers table once they have been added to the global config table via setup.

Users should declaratively list in their config which servers they would like enabled. There are duplicates, so it is important users are explicit about which servers should be enabled.

dmos62:
@mjlbach This works if you perform the LSP-specific setup, and then loop over the `.available_servers()` defining the `on_attach` callback. The problem then is that the new setup overrides the options set in the previous setup, right?

mjlbach:
Sure, but in that case then it ends up being more verbose than status quo

PR #1354, opened 2021-10-31
feat: support using `minimal_init.lua` on windows

kylo252:
- let `packpath` use the correct temp folder per platform
- use the appropriate path-separator per platform

### DIFF
--- a/test/minimal_init.lua
+++ b/test/minimal_init.lua
@@ -1,8 +1,25 @@
-vim.cmd [[set runtimepath=$VIMRUNTIME]]
-vim.cmd [[set packpath=/tmp/nvim/site]]
+local on_windows = vim.loop.os_uname().version:match 'Windows'
 
-local package_root = '/tmp/nvim/site/pack'
-local install_path = package_root .. '/packer/start/packer.nvim'
+local function join_paths(...)
+  local path_sep = on_windows and '\\' or '/'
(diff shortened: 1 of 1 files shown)


kylo252:
I figured this could be useful if anyone needs it. 

Feel free to add it to the wiki instead, if you don't care for adding Windows support in `minimal_init.lua`.

mjlbach:
It's probably a good idea to add windows support, even though I despise the fact that 20% of my once minimal template is now devoted to windows.

clason:
Why not just put everything in `stdpath('cache')` instead of `tmp` -- then it's platform-independent? E.g., like I did in https://github.com/clason/neovim/blob/minit.lua/contrib/minimal.lua

kylo252:
> Why not just put everything in `stdpath('cache')` instead of `tmp` -- then it's platform-independent? E.g., like I did in [https://github.com/clason/neovim/blob/minit.lua/contrib/minimal.lua](https://github.com/clason/neovim/blob/minit.lua/contrib/minimal.lua?rgh-link-date=2021-10-31T10%3A41%3A02Z)

I had considered it since it will remove the reliance on shell variables, but you still have the problem of the path separators either way.

mjlbach:
I think I like tmp better because I like the fact that it is wiped on reboot, but happy to discuss in follow-up PR

PR #1356, opened 2021-10-31
fix: simplify handling of config.autostart

kylo252:
- check for `true` instead of `not false`
- replace `trigger` with the conventional `event` and `pattern`
- (performance) scope `autostart` variables


kylo252:
This now fixes the issue mentioned in #1382

PR #1364, opened 2021-11-02
feat(ltexls): simplify configuration

mjlbach:
We should redirect users to grammarguard if they want something more opinionated. This seems to work fine for me.

### DIFF
--- a/lua/lspconfig/ltex.lua
+++ b/lua/lspconfig/ltex.lua
@@ -1,37 +1,41 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
+local language_id_mapping = {
+  bib = 'bibtex',
+  plaintex = 'tex',
+  rnoweb = 'sweave',
+  rst = 'restructuredtext',
+  tex = 'latex',
+  xhtml = 'xhtml',
+}
+
 configs.ltex = {
   default_config = {
     cmd = { 'ltex-ls' },
-    filetypes = { 'tex', 'bib', 'markdown' },
+    filetypes = { 'bib', 'markdown', 'org', 'plaintex', 'rst', 'rnoweb', 'tex' },
     root_dir = util.find_git_ancestor,
-    settings = {
-      ltex = {
-        enabled = { 'latex', 'tex', 'bib', 'markdown' },
-        checkFrequency = 'edit',
-        language = 'en',
-        diagnosticSeverity = 'information',
-        setenceCacheSize = 2000,
-        additionalRules = {
-          enablePickyRules = true,
-          motherTongue = 'en',
-        },
-        dictionary = {},
-        disabledRules = {},
-        hiddenFalsePositives = {},
-      },
-    },
+    get_language_id = function(_, filetype)
+      local language_id = language_id_mapping[filetype]
+      if language_id then
+        return language_id
+      end
+    end,
   },
   docs = {
     package_json = 'https://raw.githubusercontent.com/valentjn/vscode-ltex/develop/package.json',
-    description = [[
+    description = [=[
 https://github.com/valentjn/ltex-ls
 
 LTeX Language Server: LSP language server for LanguageTool 🔍✔️ with support for LaTeX 🎓, Markdown 📝, and others
 
 To install, download the latest [release](https://github.com/valentjn/ltex-ls/releases) and ensure `ltex-ls` is on your path.
 
-]],
+To support org files or R sweave, users can define a custom filetype autocommand (or use a plugin which defines these filetypes):
+
+```lua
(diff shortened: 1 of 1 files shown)


mjlbach:
@David-Else want to test this?

lua/lspconfig/ltex.lua
```diff
@@ -4,24 +4,13 @@ local util = require 'lspconfig/util'
...
-        },
-        dictionary = {},
-        disabledRules = {},
-        hiddenFalsePositives = {},
-      },
-    },
+    get_language_id = function(_, filetype)
+      if vim.tbl_contains({ 'tex', 'plaintex' }, filetype) then
+        return 'latex'
+      end
```
valentjn, line 12: Mapping plain T<sub>E</sub>X to `latex` is against the [language IDs recommended by the LSP spec](https://microsoft.github.io/language-server-protocol/specification#textDocumentItem), and it's a bit of a hack. *If* you want enable users to check plain T<sub>E</sub>X as well, then send `tex` and not `latex` (it would also be OK to send `plaintex`, but then LT<sub>E</sub>X LS needs to be changed first). Users then need to add `tex` (or `plaintex`) to [`ltex.enabled`](https://valentjn.github.io/ltex/settings.html#ltexenabled).

(The reason LT<sub>E</sub>X is not enabled for `tex` by default is that most plain T<sub>E</sub>X documents are packages, class files, etc., which shouldn't be checked.)

clason: Note that Vim treats both TeX and LaTeX as `filetype=tex` -- these are _Vim's_ internal filetypes, not LSP language IDs.

lua/lspconfig/ltex.lua
```diff
@@ -4,24 +4,13 @@ local util = require 'lspconfig/util'
 configs.ltex = {
   default_config = {
     cmd = { 'ltex-ls' },
-    filetypes = { 'tex', 'bib', 'markdown' },
+    filetypes = { 'plaintex', 'tex', 'bib', 'markdown' },
```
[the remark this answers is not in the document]
mjlbach: The main issue is we don't have filetype plugins for all of the supported languages built-in to vim/neovim, these are just used to populate a [filetype](http://vimdoc.sourceforge.net/htmldoc/autocmd.html#FileType) [autocmd](http://vimdoc.sourceforge.net/htmldoc/autocmd.html). We could add support for `org` and `bibtex`, but it would be up to the users to manually match against the extension to the filetype, or to find a filetype plugin (I don't think they exist for these languages).

lua/lspconfig/ltex.lua
clason: There's also `context`.

lua/lspconfig/ltex.lua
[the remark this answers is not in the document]
clason: (`bibtex` is supported by the builtin `filetype.vim`)

lua/lspconfig/ltex.lua
mjlbach: isn't bib the filetype for bibtex? So we should remap this to bibtex too right?

clason: Yes, the filetype for BibTeX is (only!) `bib`.

mjlbach: I went ahead and mapped plaintex -> tex, bib -> bibtex, and tex -> latex which should be correct with my understanding of vim's filetype settings

mjlbach: I went ahead and also added rst and xhtml support as those filetype plugins are also built into neovim, re: org/rsweave would need to ask users to define an autocommand as there are not available plugins

clason: you could map `rnoweb` to Sweave

mjlbach: I did :)

David-Else:
I was away from the computer. I tried this update just now, but I am afraid ltexls is not running on my Markdown file that was working. I get this from LspInfo:
```
 Client: ltex (id: 2, pid: 72659, bufnr: [1])
 	filetypes:       bib, markdown, org, plaintex, rst, rnoweb, tex
 	autostart:       true
 	root directory:  /home/david/Downloads
 	cmd:             ltex-ls
```
I am definitely in a git repo. I have:
```
require('lspconfig').ltex.setup {}
```
to configure it. This is what I had before:
```
-- require('lspconfig').ltex.setup {
--   settings = {
--     ltex = {
--       language = 'en-US',
--       additionalRules = {
--         enablePickyRules = true,
--         motherTongue = 'en',
--         -- languageModel = '~/',
--       },
--     },
--   },
-- }
```

mjlbach:
should be fixed now

David-Else:
Brilliant, all working now :) THANKS for all the work you put into getting this sorted!

PR #1368, opened 2021-11-03
Add .gpr and .adc root marker files for als

TamaMcGlinn:
It's a no-brainer. The comment saying what to do to fix this was already there.


lithammer:
You need to fix the commit message according to [commitlint](https://github.com/conventional-changelog/commitlint).

PR #1374, opened 2021-11-04
fix(jdtls): moved add-opens and add-modules java options before jar option

ankurs47:
The issue is --add-opens and --add-modules are arguments to jvm. If we put them after the -jar option, they are passed to the java program, not the jvm. Similar fix was done in https://github.com/williamboman/nvim-lsp-installer/pull/242 and kabouzeid/nvim-lspinstall#117


mjlbach:
@mfussenegger does this need to be fixed in the nvim-jdtls wiki? https://github.com/mfussenegger/nvim-jdtls/wiki/Sample-Configurations#cjgratacos-configuration

mfussenegger:
What's there matches the upstream docs: https://github.com/eclipse/eclipse.jdt.ls#running-from-the-command-line
If this needs to be changed I think it should be brought up there too

mjlbach:
@ankurs47 do you mind following up with upstream to correct the documentation? Please tag me in the issue :)

PR #1379, opened 2021-11-06
fix: normalize drive letter in windows path

### LINKED ISSUE #1168: case sensitivity causes multiple servers running in the same directory
I am having a bug where two servers are started for the same workspace. After debugging, i discovery that a diference in the case of the path of the file cause the bug. This is my `:messages` output after adding a `nvim_err_writeln` in [this line](https://github.com/neovim/nvim-lspconfig/blob/5b0fa84ee35006e06142f98b8a5b28d79cfa5000/lua/lspconfig/util.lua#L226).

image

As you can see, for some reason, there is a difference in the case of the drive D:\\, which cause the bug. I don't investigate further to know why this happen. Running `:buffers` show that the path of the file that has this difference:
```
:buffers
  1  h   "src\style.rs"                 linha 1
  3  h   "src\text.rs"                  linha 1
  6  a   "examples\main.rs"             linha 1
 10  h   "d:\repos\rust\crui\src\graphics.rs" linha 1
 11  h   "src\render.rs"                linha 1
```

Maybe the `root_dir` should be normalized?

- `nvim --version`: NVIM v0.6.0-dev+166-g8331cd13c
- nvim-lspconfig version(commit hash): 5b0fa84
- What language server (If the problem is related to a specific language server): rust-analyzer
- Can you reproduce this behavior on other language servers offered in the nvim-lspconfig repo? Not checked.
- Operating system/version: Windows 10.0.19042

### Health check
<details>
<summary>Checkhealth result</summary>
health#lspconfig#check
  - INFO: clangd: configuration checked.
  - INFO: sumneko_lua: configuration checked.
  - INFO: rust_analyzer: configuration checked.
</details>

Rodrigodd:
Closes #1168

As @aqez has investigated, a LSP server, when sends a windows path, normalizes the drive letter to lowercase. This allowed the linked issue to happen, in which lspconfig spawned two servers to the same path (one starting with `C:\` and the other with `c:\`). This also happens if you directly open a file with a lower case drive letter, e.g., `:e c:\my_root_dir\my_file.txt`.

This PR solves the problem by normalizing the drive letter of the `root_dir` before checking and storing it in the `clients` list in the `server_per_root_dir_manager`.

Another solution for this problem would be normalizing all paths inside Neovim, which would also solve some inconsistencies in the `:buffers` list, but this would be harder to do.

### DIFF
--- a/doc/server_configurations.md
+++ b/doc/server_configurations.md
@@ -2599,6 +2599,12 @@ This server accepts configuration via the `settings` key.
   
   Specifies how modifications on build files update the Java classpath\/configuration
 
+- **`java.configuration.workspaceCacheLimit`**: `null|integer`
+
+  Default: `vim.NIL`
+  
+  The number of days \(if enabled\) to keep unused workspace cache data\. Beyond this limit\, cached workspace data may be removed\.
+
 - **`java.contentProvider.preferred`**: `string`
 
   Default: `vim.NIL`
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -164,6 +164,14 @@ M.path = (function()
     return result
   end
 
+  -- For now, this only normalizes the drive letter of a Windows path to lowercase
+  local function normalize_path(path)
+    local has_windows_drive_letter = path:match '^%a:'
+    if has_windows_drive_letter then
+      return path:sub(1,1):lower()..path:sub(2)
+    end
+  end
+
   -- Traverse the path calling cb along the way.
   local function traverse_parents(path, cb)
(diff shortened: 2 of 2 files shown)


mjlbach:
I'll think about this and try to review by the end of the week. I need to double check how I did things in core around windows pathing.

Rodrigodd:
Moved the path normalization above the single_file_mode check that was added in other PR, as @petrosAth had said in the linked issue.

mjlbach:
Thanks! closing in favor of https://github.com/neovim/nvim-lspconfig/pull/1593

PR #1380, opened 2021-11-07
fix: correctly handle windows paths & paths with spaces in autocommand

adrianmgg:
autocommands always want slash delimited paths, even on windows.

from `:help file-pattern` -
> ```
> Note that for all systems the '/' character is used for path separator (even
> Windows). This was done because the backslash is difficult to use in a pattern
> and to make the autocommands portable across different systems.
> ```

it seems like it usually works anyways and the autocommand will register fine despite its path containing backslashes, but there are a few cases where it will actually cause an error (and even if there weren't, it's still probably a good idea to format the path according to what the docs say)

in the case where i encountered the issue, i had a folder with a name that started with two underscores (`D:\__foo\`...), which caused this error:
```
E5108: Error executing lua ...ck\packer\start\nvim-lspconfig\configs.lua:84: Vim(autocmd):E877: (NFA regexp) Invalid character class: 95
```
, which was presumably due to the `_` character class expecting the character following it to be a valid character class. from `:help /character-classes` -
> ```
> Character classes:					*/character-classes*
>       magic   nomagic	matches 
> ...
> |/\_|	\_x	\_x	where x is any of the characters above: character
> 			class with end-of-line included
> ```

----

this PR fixes that issue with a `normalize_for_autocmd(path)` function in `util.path` that replaces `\` with `/` in the path if on windows

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -77,7 +77,7 @@ function configs.__newindex(t, config_name, config_def)
       api.nvim_command(
         string.format(
           "autocmd %s unsilent lua require'lspconfig'[%q].manager.try_add_wrapper()",
-          'BufReadPost ' .. util.path.normalize_for_autocmd(root_dir) .. '/*',
+          'BufReadPost ' .. vim.fn.fnameescape(util.path.to_unix_style(root_dir)) .. '/*',
           config.name
         )
       )
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -77,7 +77,7 @@ function configs.__newindex(t, config_name, config_def)
       api.nvim_command(
         string.format(
           "autocmd %s unsilent lua require'lspconfig'[%q].manager.try_add_wrapper()",
-          'BufReadPost ' .. root_dir .. '/*',
+          'BufReadPost ' .. util.path.normalize_for_autocmd(root_dir) .. '/*',
           config.name
         )
       )
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -215,8 +215,7 @@ M.path = (function()
     return dir == root
   end
 
-  -- autocommands always want '/' delimited paths, even on windows
-  local function normalize_for_autocmd(path)
+  local function to_unix_style(path)
     if is_windows then
       return (path:gsub('\\', '/'))
     else
@@ -235,7 +234,7 @@ M.path = (function()
     traverse_parents = traverse_parents,
     iterate_parents = iterate_parents,
     is_descendant = is_descendant,
-    normalize_for_autocmd = normalize_for_autocmd,
+    to_unix_style = to_unix_style,
   }
 end)()
 
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -215,6 +215,15 @@ M.path = (function()
     return dir == root
   end
 
+  -- autocommands always want '/' delimited paths, even on windows
+  local function normalize_for_autocmd(path)
+    if is_windows then
+      return (path:gsub('\\', '/'))
+    else
+      return path
+    end
+  end
+
   return {
     is_dir = is_dir,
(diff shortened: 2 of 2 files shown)


lua/lspconfig/configs.lua
```diff
@@ -77,7 +77,7 @@ function configs.__newindex(t, config_name, config_def)
       api.nvim_command(
         string.format(
           "autocmd %s unsilent lua require'lspconfig'[%q].manager.try_add_wrapper()",
-          'BufReadPost ' .. root_dir .. '/*',
+          'BufReadPost ' .. util.path.normalize_for_autocmd(root_dir) .. '/*',
```
kylo252, line 80:
```suggestion
          'BufReadPost ' .. util.path.join(root_dir, '*'),
```

lua/lspconfig/util.lua
```diff
@@ -215,6 +215,15 @@ M.path = (function()
...
 
+  -- autocommands always want '/' delimited paths, even on windows
+  local function normalize_for_autocmd(path)
+    if is_windows then
+      return (path:gsub('\\', '/'))
+    else
+      return path
+    end
+  end
+
```
kylo252, line 226:
```suggestion
 ```

File: lua/lspconfig/util.lua [hunk 3]
```diff
@@ -226,6 +235,7 @@ M.path = (function()
     traverse_parents = traverse_parents,
     iterate_parents = iterate_parents,
     is_descendant = is_descendant,
+    normalize_for_autocmd = normalize_for_autocmd,
```
kylo252, line 238:
```suggestion
```

kylo252 (changes requested):
I think this should be enough since `root_dir` should already be normalized.

adrianmgg:
> I think this should be enough since `root_dir` should already be normalized.

the way the paths need to be normalized for autocommands specifically is to have forward slashes *regardless of* the os, the existing `util.path` functions are for more general path manipulation so they use an os-specific path separator
> ```lua
> local path_sep = is_windows and '\\' or '/'
> ```

as a test, i put in your suggested changes along with some debug prints right above where the command is run, then tested that with `/` and `\` delimited paths (using `set shellslash`)

```vim
e hello_world.py

set shellslash
LspStop
LspStart
" root_dir                        : D:/__temp/foo/hello_world.py
" util.path.normalize_for_autocmd : D:/__temp/foo/hello_world.py/*
" util.path.join (using this one) : D:/__temp/foo/hello_world.py\*

set noshellslash
LspStop
LspStart
" root_dir                        : D:\__temp\foo
" util.path.normalize_for_autocmd : D:/__temp/foo/*
" util.path.join (using this one) : D:\__temp\foo\*
" E5108: Error executing lua ...ck\packer\start\nvim-lspconfig\lua\lspconfig\configs.lua:80: Vim(autocmd):E877: (NFA regexp) Invalid character class: 95
```

as you can see, `util.path.join` (correctly) keeps `root_dir`'s `\`s as `\`s, which is usually what you would want but isn't correct when working with autocommands

(also, looking back at those paths from the test above it turns out there's a problem with multiple of the `util.path` functions assuming the input path will be exclusively `\` delimited on windows even though `/` is usually also valid, i'll create an issue for that)

adrianmgg:
also, spaces in the path need to be backslash escaped in an autocommand, `normalize_for_autocmd` should handle that as well

adrianmgg:
added a call to `fnameescape` to make the autocommand work with paths that contain spaces, renamed the new `util.path.normalize_for_autocmd` to `util.path.to_unix_style` since that's more descriptive of what it actually does

kylo252 (approved):

mjlbach:
let's lump in the capitalization of the drive letter and fnamescape into the function, and I'd prefer "normalize_path", I think we can reuse the normalization here for the internal logic to fix the issue with buf_add (see the other pr)

mjlbach:
Closing in favor of https://github.com/neovim/nvim-lspconfig/pull/1593

PR #1186, opened 2021-08-25
feat(ui): better layout

kylo252:
# Description
 
Various improvements to LspInfo 

- Allow better readability by applying some extra paddings and a border
- Remove redundant information, e.g. if a client is attached then it's already active so no need to display the full info about it multiple times
- Add additional information about attached buffers numbers
- Refactor LspInfo and extract some utility functions to allow for better maintainability

This also means we now have more available space to display any other useful information.

## Screenshots

WindowsTerminal_KdRSL6JSj1
WindowsTerminal_VNTHKMCc4K
WindowsTerminal_CA6yLDorfB
WindowsTerminal_qyTmXoXgWf
WindowsTerminal_1nvIn5swCK

PR #1382, opened 2021-11-07
fix(lspinfo): fix autostart value reporting

tomtomjhj:
`config.autostart` is a function in `make_config_info`.
https://github.com/neovim/nvim-lspconfig/blob/3a5d94df2d61992496d1fc986d2433b944584a8d/lua/lspconfig/configs.lua#L71-L97

Note: this line reports the correct value.
https://github.com/neovim/nvim-lspconfig/blob/3a5d94df2d61992496d1fc986d2433b944584a8d/lua/lspconfig/ui/lspinfo.lua#L78

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -47,7 +47,7 @@ local function make_config_info(config)
 
   local buffer_dir = vim.fn.expand '%:p:h'
   config_info.root_dir = config.get_root_dir(buffer_dir) or 'NA'
-  config_info.autostart = (config.autostart and 'true') or 'false'
+  config_info.autostart = (config._autostart and 'true') or 'false'
   config_info.handlers = table.concat(vim.tbl_keys(config.handlers), ', ')
   config_info.filetypes = table.concat(config.filetypes or {}, ', ')
 


kylo252:
I think you might be confusing `(lspconfig/configs).autostart()` with this one:

https://github.com/neovim/nvim-lspconfig/blob/6224c54a9945a52bf43a8bc1a42a112084590c0b/lua/lspconfig/configs.lua#L63

Which I have attempted to make it look better btw 😄 
https://github.com/neovim/nvim-lspconfig/pull/1356

tomtomjhj:
Hmm I think that PR doesn't seem to be relevent to this PR. The issue this PR fixes is that even if the configuration has `autostart = false`, LspInfo always reports `autostart = true` because `(config.autostart and 'true') or 'false'` always evaluates to `'true'` since `config.autostart` is a function, not a boolean value. Sorry for not explaining this point earlier.

kylo252:
I think you missed my point. There's some overlap with the naming here, so there are two separate `autostart` instances:
1.  this is the function one that you're referring to
```lua
require("lspconfig/configs").autostart()
```
2. this is specific to each client, In the case of `LspInfo` we only care about this one.
You can set it to `false` to tell `lspconfig` that it shouldn't start it automatically
```lua
client.config={
---
autostart = false
---
}
```
You can pass it yourself in the settings and see how it works. Furthermore, you can see an example here: https://github.com/jose-elias-alvarez/null-ls.nvim/blob/64b269b51c7490660dcb2008f59ae260f2cdbbe4/lua/null-ls/lspconfig.lua#L21

tomtomjhj:
I have something like
```
lspconfig.ltex.setup { autostart = false }
```
in my config, but LspInfo says `autostart: true` when run in a buffer with matching filetype (in this case markdown or tex).
If I manually run `LspStart ltex`, then LspInfo says `autostart: false`. Is this expected behavior? 
Also, all the occurrences of `_autostart` field (introduced in #964, IIUC in order to keep the value passed in `setup`) were removed in #1186.

kylo252:
Apologies, I didn't read the post carefully the first time, and I completely forgot that there are two different functions for this. 

However, I still stand by my point that `LspInfo` is not really doing anything wrong here, it's the `configs` manager that's introducing this needless inconsistency instead of storing `config.autostart` as is.

mjlbach:
Closed by 1356

PR #1383, opened 2021-11-07
fix(powershell_es): don't overwrite cmd when it is present in the new config

Drllap:
1. Now the cmd in "new_config" is only generated with "make_cmd" if
       there is no cmd already in "new_config"
    2. make_cmd now takes "new_config" as an argument instead of just
       "bundle_path"
    3. Support support changing the execution shell by setting the
       "shell" member in "new_config"

### DIFF
--- a/lua/lspconfig/powershell_es.lua
+++ b/lua/lspconfig/powershell_es.lua
@@ -4,20 +4,20 @@ local util = require 'lspconfig/util'
 local server_name = 'powershell_es'
 local temp_path = vim.fn.stdpath 'cache'
 
-local function make_cmd(bundle_path)
-  if bundle_path ~= nil then
+local function make_cmd(new_config)
+  if new_config.bundle_path ~= nil then
     local command_fmt =
       [[%s/PowerShellEditorServices/Start-EditorServices.ps1 -BundledModulesPath %s -LogPath %s/powershell_es.log -SessionDetailsPath %s/powershell_es.session.json -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
-    local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
-    return { 'pwsh', '-NoLogo', '-NoProfile', '-Command', command }
+    local command = command_fmt:format(new_config.bundle_path, new_config.bundle_path, temp_path, temp_path)
+    return { new_config.shell, '-NoLogo', '-NoProfile', '-Command', command }
   end
 end
 
 configs[server_name] = {
   default_config = {
+    shell = 'pwsh',
     on_new_config = function(new_config, _)
-      local bundle_path = new_config.bundle_path
-      new_config.cmd = make_cmd(bundle_path)
+      new_config.cmd = make_cmd(new_config)
     end,
     filetypes = { 'ps1' },
     root_dir = util.find_git_ancestor,
--- a/lua/lspconfig/powershell_es.lua
+++ b/lua/lspconfig/powershell_es.lua
@@ -17,8 +17,12 @@ configs[server_name] = {
   default_config = {
     shell = 'pwsh',
     on_new_config = function(new_config, _)
-      new_config.cmd = make_cmd(new_config)
+      -- Don't overwrite cmd if already set
+      if not new_config.cmd then
+        new_config.cmd = make_cmd(new_config)
+      end
     end,
+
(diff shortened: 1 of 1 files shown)


lua/lspconfig/powershell_es.lua
```diff
@@ -4,20 +4,27 @@ local util = require 'lspconfig/util'
...
       [[%s/PowerShellEditorServices/Start-EditorServices.ps1 -BundledModulesPath %s -LogPath %s/powershell_es.log -SessionDetailsPath %s/powershell_es.session.json -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
-    local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
-    return { 'pwsh', '-NoLogo', '-NoProfile', '-Command', command }
+    command = command_fmt:format(new_config.bundle_path, new_config.bundle_path, temp_path, temp_path)
   end
+  if new_config.shell == nil then
+    -- defaulting to PowerShell Core
+    new_config.shell = 'pwsh'
+  end
+  new_config.cmd = { new_config.shell, '-NoLogo', '-NoProfile', '-ExecutionPolicy', 'Unrestricted', '-Command', command }
```
Drllap, line 18:
```suggestion
  new_config.cmd = { new_config.shell, '-NoLogo', '-NoProfile', '-Command', command }
```

lua/lspconfig/powershell_es.lua
```diff
@@ -4,20 +4,27 @@ local util = require 'lspconfig/util'
...
 end
 
 configs[server_name] = {
   default_config = {
     on_new_config = function(new_config, _)
-      local bundle_path = new_config.bundle_path
-      new_config.cmd = make_cmd(bundle_path)
+      if new_config.cmd == nil then
+        -- Only create a new cmd if it isn't specified in the new config
+        make_cmd(new_config)
```
mjlbach, line 26: I prefer making this assignment explicit, return cmd like before and set new_config.cmd = cmd.

lua/lspconfig/powershell_es.lua
```diff
@@ -4,20 +4,27 @@ local util = require 'lspconfig/util'
...
+local function make_cmd(new_config)
+  local command = ""
+  if new_config.bundle_path ~= nil then
     local command_fmt =
       [[%s/PowerShellEditorServices/Start-EditorServices.ps1 -BundledModulesPath %s -LogPath %s/powershell_es.log -SessionDetailsPath %s/powershell_es.session.json -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
-    local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
-    return { 'pwsh', '-NoLogo', '-NoProfile', '-Command', command }
+    command = command_fmt:format(new_config.bundle_path, new_config.bundle_path, temp_path, temp_path)
   end
+  if new_config.shell == nil then
```
mjlbach, line 14: Rather than settings new_config.shell == pwsh, just have a default value for shell (pwsh) which is overridden by new_config.shell if it exists.

lua/lspconfig/powershell_es.lua
```diff
@@ -4,20 +4,27 @@ local util = require 'lspconfig/util'
 local server_name = 'powershell_es'
 local temp_path = vim.fn.stdpath 'cache'
 
-local function make_cmd(bundle_path)
-  if bundle_path ~= nil then
+local function make_cmd(new_config)
+  local command = ""
```
mjlbach, line 8: Why the empty string? I preferred the old style...

lua/lspconfig/powershell_es.lua
Drllap: Added `shell` member to `default_config` with the default value of `pwsh`. Is that what you were asking for?

lua/lspconfig/powershell_es.lua
```diff
@@ -4,21 +4,25 @@ local util = require 'lspconfig/util'
...
 end
 
 configs[server_name] = {
   default_config = {
+    shell = 'pwsh',
     on_new_config = function(new_config, _)
-      local bundle_path = new_config.bundle_path
-      new_config.cmd = make_cmd(bundle_path)
+      -- Don't overwrite the cmd if it is already set
+      if new_config.cmd == nil then
```
mjlbach, line 21: you can do `if not new_config.cmd`, it's more readable

PR #1401, opened 2021-11-11
fix(lspinfo): handle single file mode

mjlbach:


lua/lspconfig/ui/lspinfo.lua
```diff
@@ -46,7 +46,11 @@ local function make_config_info(config)
   end
 
   local buffer_dir = vim.fn.expand '%:p:h'
-  config_info.root_dir = config.get_root_dir(buffer_dir) or 'NA'
+  if config.get_root_dir then
+    config_info.root_dir = config.get_root_dir(buffer_dir) or 'NA'
+  else
+    config_info.root_dir = 'Running in single file mode.'
+  end
```
kylo252, line 53: This will always be true
https://github.com/neovim/nvim-lspconfig/blob/edca6c107027111ed02abe17a231d1cf0d353adf/lua/lspconfig/configs.lua#L108

Let's use the default string instead if it's available.

```suggestion
  config_info.root_dir = config.document_config.docs.default_config.root_dir or 'NA'
```

kylo252 (changes requested):

mjlbach:
How does this actually break for you again?

kylo252:
> How does this actually break for you again?

Set `autostart = false`. Here's a comparison between your change and my suggestion:

image

lua/lspconfig/ui/lspinfo.lua
clason: Big -1 from me -- I want to see the _actual_ resolved root directory to figure out what's going on.

kylo252: It is not resolved yet, that's the point. Furthermore, if you look at the comparison below, it's currently showing the wrong `root_dir`. So either don't show it until the server starts, or show what it _should_ be.

PR #1404, opened 2021-11-11
docs: use find_git_ancestor more consistently

kylo252:


kylo252:
We should probably add `single_file_mode` into the docs.

---

@dundargoc, the CI action should take into account additions or deletions.

dundargoc:
> We should probably add `single_file_mode` into the docs.
> 
> ---
> 
> @dundargoc, the CI action should take into account additions or deletions. 

I think it already does. You removed the string dirname in one or two places. Not sure if it's the correct behavior though.

mjlbach:
> We should probably add single_file_mode into the docs.

It is, see `:help lspconfig-single-file-suppport` and `:help lspconfig-setup`

kylo252:
> It is, see `:help lspconfig-single-file-suppport` and `:help lspconfig-setup`

I meant adding _which_ server supports it 😄 

But if the intention is that it should work on all of them, then we can do a blacklist style and just indicate which servers it will _definitely_ not work with.

mjlbach:
It's an allowlist, I'm fine adding it to docgen.

PR #1405, opened 2021-11-11
fix(ci): avoid failing "dirname"-cleanup PRs

kylo252:
Re-wrote the check so that it will only fail on additions.

And not deletions, see #1404


mjlbach:
cc @dundargoc

.github/workflows/check-for-dirname.yml
```diff
@@ -11,8 +11,10 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
       - run: |
-          if ! git log -G"dirname" --exit-code origin/$GITHUB_BASE_REF..$(git branch --show-current) -- . ':!lua/lspconfig/configs.lua'; then
+          # checks for added lines that contain the string 'dirname' and prints them
+          if git diff --pickaxe-all -S "dirname" origin/$GITHUB_BASE_REF..HEAD -- '*.lua' | grep '^+' | grep 'dirname'; then
```
dundargoc, line 15: The `-- '*.lua'` looks a bit suspicious. It means that this workflow will only work for lua-files and not for files of any other types. Maybe that's the intention but the behavior is so different from before (which was that the workflow worked on all files *except* for `lua/lspconfig/configs.lua`) that I'm unsure.

@mjlbach: what's the correct behavior?

dundargoc:
Looks good overall, need only input from @mjlbach on which files this should work on.

.github/workflows/check-for-dirname.yml
mjlbach: Any luafile which adds dirname or vim.fn.cwd should show red and the message

mjlbach: except  lua/lspconfig/{configs.lua,utils.lua}

kylo252: well, this is awkward,... I thought I just fixed it 😢

kylo252: It was just using the wrong reference. I'll give it a test to see if it works correctly.

PR #1406, opened 2021-11-11
fix: update error message

williamboman:
🤓

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -260,7 +260,7 @@ function M.server_per_root_dir_manager(_make_config)
       if not new_config.cmd then
         print(
           string.format(
-            'Error: cmd not defined for %q. You must manually set cmd in the setup{} call according to CONFIG.md.',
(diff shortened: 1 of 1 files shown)


mjlbach:
I think we should just call it "sever_configurations.md" and say `see :help lspindex`

lua/lspconfig/util.lua
```diff
@@ -260,7 +260,7 @@ function M.server_per_root_dir_manager(_make_config)
       if not new_config.cmd then
         print(
           string.format(
-            'Error: cmd not defined for %q. You must manually set cmd in the setup{} call according to CONFIG.md.',
+            'Error: cmd not defined for %q. You must manually set cmd in the setup{} call according to server_configurations.md, see :help lspindex.',
```
mjlbach, line 263: `:help lspconfig-index` :)

williamboman: And here I am spamming `:helptags ALL` wondering why it won't work haha

PR #1407, opened 2021-11-11
feat: spawn the language server in resolved root directory

mjlbach:
This is kind of niche, but it was requested and seems harmless.


neovim-discourse:
This pull request has been mentioned on **Neovim Discourse**. There might be relevant details there:

https://neovim.discourse.group/t/pyright-unable-to-resolve-imports-that-are-in-my-pythonpath/1289/6

jfly (approved):
Loving it, thank you! :shipit:

mjlbach:
fyi @jfly this is not a reliable method to rely on, because the relative root will depend on your project.  Regardless of this PR, you should use the absolute path in direnv/whatever you are using.

PR #1414, opened 2021-11-12
revert: disable single file support

jspricke:
EFM does support NULL root directories.

This reverts commit 0d2722a63e7e79bfc97ea10322ecbbfd48154e77.

### DIFF
--- a/lua/lspconfig/efm.lua
+++ b/lua/lspconfig/efm.lua
@@ -8,9 +8,7 @@ configs[server_name] = {
   default_config = {
     cmd = { bin_name },
     root_dir = util.find_git_ancestor,
-    -- EFM does not support NULL root directories
-    -- https://github.com/neovim/nvim-lspconfig/issues/1412
-    single_file_support = false,
+    single_file_support = true,
   },
 
   docs = {


mjlbach:
Has the new release been cut? Also I'd rather hold off until this hits package managers.

jspricke:
Yes: https://github.com/mattn/efm-langserver/releases/tag/v0.0.38

I'm fine with holding this off (users can simply change their config)
Though I've fixed it in Debian already ;).

jspricke:
>Would also be good to verify that the internal EFM root directory it infers with the root pattern setting that can be sent via workspace/didChangeConfiguraiton still works well.

How would you do that? (It seems fine when looking at the sources)

mjlbach:
@mattn Sorry for the ping. Would you mind commenting on EFM's expectation?

I added support for launching EFM in the parent directory of the open file, without sending rootDirectory or workspaceFolders during initialization (sending NULL) if we can't detect the current file is a child of a git repository.

I see there is also `root-markers` in the EFM configuration file. I'm wondering if sending NULL disables any features/how EFM handles locating `root-markers`

Thank you!

mjlbach:
I went ahead and directly committed this to master with a note about backwards compatibility.

PR #1415, opened 2021-11-12
fix(ansiblels): add yaml.ansible filetype and add ".git" to root_dir pattern

williamboman:
The current root pattern seems a bit off?

The `yaml.ansible` filetype seems to be a generally accepted filetype for ansible yaml files:

- https://github.com/sheerun/vim-polyglot/blob/b147123070e5d7418fe67e315a53599cfde38d58/autoload/polyglot/init.vim#L1833-L1837
- https://github.com/pearofducks/ansible-vim

### DIFF
--- a/lua/lspconfig/ansiblels.lua
+++ b/lua/lspconfig/ansiblels.lua
@@ -22,9 +22,9 @@ configs[server_name] = {
         },
       },
     },
-    filetypes = { 'yaml' },
+    filetypes = { 'yaml', 'yaml.ansible' },
     root_dir = function(fname)
-      return util.root_pattern { '*.yml', '*.yaml' }(fname)
+      return util.root_pattern('*.yml', '*.yaml', '.git')(fname)
     end,
   },
   docs = {
--- a/lua/lspconfig/ansiblels.lua
+++ b/lua/lspconfig/ansiblels.lua
@@ -23,9 +23,7 @@ configs[server_name] = {
       },
     },
     filetypes = { 'yaml', 'yaml.ansible' },
-    root_dir = function(fname)
-      return util.root_pattern('*.yml', '*.yaml', '.git')(fname)
-    end,
+    root_dir = util.root_pattern('*.yml', '*.yaml', '.git'),
   },
   docs = {
     description = [[


lua/lspconfig/ansiblels.lua
```diff
@@ -22,10 +22,8 @@ configs[server_name] = {
         },
       },
     },
-    filetypes = { 'yaml' },
-    root_dir = function(fname)
-      return util.root_pattern { '*.yml', '*.yaml' }(fname)
-    end,
+    filetypes = { 'yaml', 'yaml.ansible' },
+    root_dir = util.root_pattern('*.yml', '*.yaml', '.git'),
```
lithammer, line 26: There was recently a pull request (GH-1404) merged to more consistently use `find_git_ancestor`. So I suggest sticking to that convention.

The `filetype` change looks good though 👍

williamboman: Do you mean 

```lua
root_dir = function (fname) return util.find_git_ancestor(fname) or util.root_pattern("*.yml", "*.yaml") end
```

lithammer: Yes. But I would probably put the Git last. Or on second thought... Is using yaml even a good root marker? Normally, Ansible playbooks are structured something like this:

```
/
└── roles
    └── nginx
        ├── handlers
        │   └── main.yml
        ├── tasks
        │   └── main.yml
        └── templates
            └── nginx.conf.j2
```

And I definitely wouldn't want to resolve the root to `/roles/nginx/tasks/` instead of `/` when opening `/roles/nginx/tasks/main.yml`.

`ansible.cfg` is probably a good marker to have as well since those are usually in the root of an Ansible playbook.

lithammer: And I also suspect that many times an Ansible playbook doesn't live in the root of a repo, but rather under some subfolder like `ansible/` or `deploy/` or whatever.

mjlbach: Also the order here matters a bit, not sure if git should be included on each step up the filesystem or should be a fallback like lithammer suggests, I think the latter

mjlbach:
superseded by #1423

PR #1416, opened 2021-11-12
Add support for boot in clojure_lsp.

Ultra-Hax:
Boot is another build system for Clojure. 

Here is the link to boot: https://boot-clj.github.io/


lua/lspconfig/clojure_lsp.lua
```diff
@@ -14,7 +14,7 @@ https://github.com/snoe/clojure-lsp
 Clojure Language Server
 ]],
     default_config = {
-      root_dir = [[root_pattern("project.clj", "deps.edn", ".git")]],
+      root_dir = [[root_pattern("project.clj", "build.boot", "deps.edn", ".git")]],
```
mjlbach, line 17: the order of these should be consistent as it matters

PR #1420, opened 2021-11-13
ci: skip workflow if mjlbach triggered it

dundargoc:


mjlbach:
I feel like it's kinda weird to harcode my gh id, can we special case core people by team?

dundargoc:
Pretty sure you need special read access for the organization in order to get members of a team, which I cannot create. Closing.

PR #1421, opened 2021-11-13
docs(yammlls): add examples for schemas

mjlbach:
Closes https://github.com/neovim/nvim-lspconfig/issues/1207

We get a lot of questions about settings schemas, probably worth embedding in the documentation.

cc @longwuyuan @kylo252 

I'm still not sure I understand using the yaml store url instead of the direct url path, as from the documentation I got the impression you need to explicitly associate a document uri with a schema.


mjlbach:
I'm going to merge but feel free to file a follow-up PR clarifying things.

PR #1425, opened 2021-11-14
vala_ls: add single file support

Prince781:


mjlbach:
Thanks for the follow-up PR! Merged.

PR #1431, opened 2021-11-16
docs: update server_configurations.md

lcrockett:
skip-checks: true


clason:
> skip-checks: true

Nice try... ;)

lcrockett:
Apologies, I figured that was the standard for docs. Just found out we're supposed to change the actual docs table in the Lua file, will follow that (correct) approach instead. Cheers !

clason:
Be aware that any "use current/home directory" fallbacks are deprecated, so unless Groovy is a very special case, that PR won't be accepted as is (and rather should include a _proper_ root directory pattern).

PR #1432, opened 2021-11-16
groovyls: Add Jenkinsfile as `root_dir`

lcrockett:


lua/lspconfig/groovyls.lua
```diff
@@ -12,7 +12,10 @@ configs[name] = {
       bin_name,
     },
     filetypes = { 'groovy' },
-    root_dir = util.find_git_ancestor or vim.loop.os_homedir(),
+    root_dir = function(fname)
+      local root_files = { 'Jenkinsfile' }
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
```
lithammer, line 17: Just inline `{ 'Jenkinsfile' }` imo.

clason: Without the table ;)

You can probably further compress it as `util.root_pattern('Jenkinsfile') or util.find_git_ancestor`?

lcrockett: Yep, I figured a table in case people named their Jenkins file `jenkinsfile` in all lower-case, however the correct filename is `Jenkinsfile` with upper case `J`, so fine with this.

clason:
Don't forget to run Stylua (using the `.stylua.toml` from the repo root) when you're done, and squash your commits.

lua/lspconfig/groovyls.lua
```diff
@@ -12,7 +12,7 @@ configs[name] = {
       bin_name,
     },
     filetypes = { 'groovy' },
-    root_dir = util.find_git_ancestor or vim.loop.os_homedir(),
+    root_dir = util.root_pattern 'Jenkinsfile' or util.find_git_ancestor,
```
lithammer, line 15: This won't work, it will just always assign the _function_ `util.root_pattern` to `root_dir`.

lithammer: You're basically doing this:

```console
> function foo() end
> function bar() end
> a = foo or bar
> print(a == foo)
true
> print(b == bar)
false
```

clason: Ah, I was afraid of that. (I briefly tested, but with an inconclusive example ;))

Looks like you have to follow this pattern: https://github.com/neovim/nvim-lspconfig/blob/a4deb0c770bed049218f183f6a0de249c52d1d69/lua/lspconfig/robotframework_ls.lua#L10-L12

lcrockett: Ah, I think we can do this then: `util.root_pattern 'Jenkinsfile'(fname) or util.find_git_ancestor(fname)`

clason: But wrap it in a function, like for `robotframework_ls`.

lua/lspconfig/groovyls.lua
```diff
@@ -41,7 +43,9 @@ require'lspconfig'.groovyls.setup{
         bin_name,
       },
       filetypes = { 'groovy' },
-      root_dir = util.root_pattern 'Jenkinsfile' or util.find_git_ancestor,
+      root_dir = function(fname)
+        return util.root_pattern('Jenkinsfile')(fname) or util.find_git_ancestor(fname)
+      end,
```
lithammer: I think you can omit changing this and it will be automatically updated.

lua/lspconfig/groovyls.lua
clason: Ah, that was bad info -- that is only the case for table literals, not string literals. (I opened an issue about that, but no changes were made yet.) Sorry!

lithammer:
Also, this isn't really a `docs` change, but rather `fix`... I think? 🤔

lcrockett:
Initially it was a fix indeed, by now it evolved in an update of the root_dir to new standards I would say. But happy to change the commit message to fix if you deem it a fix.

clason:
> Initially it was a fix indeed, by now it evolved in an update of the root_dir to new standards I would say. But happy to change the commit message to fix if you deem it a fix.

I would consider that a fix ;) (Commit message can be changed while squashing, so not a problem.)

Why did you close the PR?

clason:
I'd say `fix(groovyls): update root_dir` is good enough as a commit message. Let me know when you're done and I'll squash and merge.

lcrockett:
Commit changed and squashed, should be good to merge. Cheers to both of you for your help and guidance.

PR #1437, opened 2021-11-17
feat(jsonnet): add JSONNET_PATH environement variable

chaudum:
`<CWD>/lib` and `<CWD>/vendor` are common jsonnet library paths.
The jsonnet language server can use the `JSONNET_PATH` env variable to
resolve referenced files within these folders.

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

### DIFF
--- a/lua/lspconfig/jsonnet_ls.lua
+++ b/lua/lspconfig/jsonnet_ls.lua
@@ -4,8 +4,8 @@ local util = require 'lspconfig/util'
 -- common jsonnet library paths
 local function jsonnet_path(root_dir)
   local paths = {
-    root_dir .. '/lib',
-    root_dir .. '/vendor',
+    util.path.join(root_dir, 'lib'),
+    util.path.join(root_dir, 'vendor'),
   }
   return table.concat(paths, ':')
 end
--- a/lua/lspconfig/jsonnet_ls.lua
+++ b/lua/lspconfig/jsonnet_ls.lua
@@ -10,18 +10,20 @@ local function jsonnet_path(root_dir)
   return table.concat(paths, ':')
 end
 
+local function get_root_dir(fname)
+  return util.root_pattern('jsonnetfile.json')(fname) or util.find_git_ancestor(fname)
+end
+
 configs.jsonnet_ls = {
   default_config = {
     cmd = { 'jsonnet-language-server' },
     cmd_env = { JSONNET_PATH = vim.fn.getcwd() }, -- will be overwritten by `on_new_config`
     filetypes = { 'jsonnet', 'libsonnet' },
-    root_dir = vim.fn.getcwd(), -- will be overwritten by `on_new_config`
+    root_dir = get_root_dir,
     on_new_config = function(new_config, file_dir)
-      local root_dir = util.root_pattern('jsonnetfile.json')(file_dir) or util.find_git_ancestor(file_dir)
       new_config.cmd_env = {
-        JSONNET_PATH = jsonnet_path(root_dir)
+        JSONNET_PATH = jsonnet_path(get_root_dir(file_dir))
       }
-      new_config.root_dir = root_dir
     end
   },
   docs = {
--- a/lua/lspconfig/jsonnet_ls.lua
+++ b/lua/lspconfig/jsonnet_ls.lua
@@ -2,10 +2,10 @@ local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
 -- common jsonnet library paths
-local function jsonnet_path()
+local function jsonnet_path(root_dir)
   local paths = {
-    vim.fn.getcwd() .. '/lib',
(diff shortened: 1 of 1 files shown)


lua/lspconfig/jsonnet_ls.lua
```diff
@@ -1,9 +1,19 @@
 local configs = require 'lspconfig/configs'
 local util = require 'lspconfig/util'
 
+-- common jsonnet library paths
+local function jsonnet_path()
+  local paths = {
+    vim.fn.getcwd() .. '/lib',
+    vim.fn.getcwd() .. '/vendor',
```
lithammer, line 8: Shouldn't this be relative to the resolved `root_dir` rather than `cwd`?

mjlbach: Also use path.join instead of string concatenation.

lua/lspconfig/jsonnet_ls.lua
```diff
@@ -1,13 +1,28 @@
...
+
 configs.jsonnet_ls = {
   default_config = {
     cmd = { 'jsonnet-language-server' },
+    cmd_env = { JSONNET_PATH = vim.fn.getcwd() }, -- will be overwritten by `on_new_config`
     filetypes = { 'jsonnet', 'libsonnet' },
-    root_dir = function(fname)
-      return util.root_pattern 'jsonnetfile.json'(fname) or util.find_git_ancestor(fname)
-    end,
+    root_dir = vim.fn.getcwd(), -- will be overwritten by `on_new_config`
```
mjlbach, line 18: Keep the original root dir

mjlbach (approved):

lua/lspconfig/jsonnet_ls.lua
```diff
@@ -1,13 +1,28 @@
...
+    on_new_config = function(new_config, file_dir)
+      local root_dir = util.root_pattern('jsonnetfile.json')(file_dir) or util.find_git_ancestor(file_dir)
+      new_config.cmd_env = {
+        JSONNET_PATH = jsonnet_path(root_dir)
+      }
+      new_config.root_dir = root_dir
```
mjlbach, line 24: Why do you have to replace root dir in on_new_config?

mjlbach (changes requested):

lua/lspconfig/jsonnet_ls.lua
```diff
@@ -10,18 +10,20 @@ local function jsonnet_path(root_dir)
...
     cmd = { 'jsonnet-language-server' },
     cmd_env = { JSONNET_PATH = vim.fn.getcwd() }, -- will be overwritten by `on_new_config`
     filetypes = { 'jsonnet', 'libsonnet' },
-    root_dir = vim.fn.getcwd(), -- will be overwritten by `on_new_config`
+    root_dir = get_root_dir,
     on_new_config = function(new_config, file_dir)
-      local root_dir = util.root_pattern('jsonnetfile.json')(file_dir) or util.find_git_ancestor(file_dir)
       new_config.cmd_env = {
-        JSONNET_PATH = jsonnet_path(root_dir)
+        JSONNET_PATH = jsonnet_path(get_root_dir(file_dir))
```
mjlbach, line 25: Why do you have to do this again? file_dir is root_dir (I would rename this)

chaudum: What I get as root_dir is the directory of the opened file :shrug:

mjlbach: then the root pattern is failing to find a root and you should get the same result again

mjlbach: Or there is a bug, this is where it's supposed to be passed 

https://github.com/mjlbach/nvim-lspconfig/blob/c1af9bd8296b1627ed0517b1915c76545ea7cee9/lua/lspconfig/configs.lua#L208-L209

chaudum: You're right. I had an erroneous override of the default config in the setup function :man_facepalming:

chaudum:
@mjlbach Should I rebase and squash commits?

lua/lspconfig/jsonnet_ls.lua
```diff
@@ -1,13 +1,28 @@
...
+    util.path.join(root_dir, 'lib'),
+    util.path.join(root_dir, 'vendor'),
+  }
+  return table.concat(paths, ':')
+end
+
 configs.jsonnet_ls = {
   default_config = {
     cmd = { 'jsonnet-language-server' },
+    cmd_env = { JSONNET_PATH = vim.fn.getcwd() }, -- will be overwritten by `on_new_config`
```
mjlbach, line 16: Why does this need to be a placeholder?

mjlbach:
Need to run stylua in the root of lspconfig

PR #1442, opened 2021-11-18
fix(julials): add language-specific root marker

clason:
Proper Julia packages have a `Project.toml` in the root directory. Use that as root marker (but keep `.git` as a fallback).

### DIFF
--- a/lua/lspconfig/julials.lua
+++ b/lua/lspconfig/julials.lua
@@ -48,7 +48,9 @@ configs.julials = {
       new_config.cmd_cwd = root_dir
     end,
     filetypes = { 'julia' },
-    root_dir = util.find_git_ancestor,
+    root_dir = function(fname)
+      return util.root_pattern 'Project.toml'(fname) or util.find_git_ancestor(fname)
+    end,
     single_file_support = true,
   },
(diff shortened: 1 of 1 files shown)


mjlbach:
Does julia ever have projects with submodules/nested project.tomls? in which case we could do something like I did (and reverted) in https://github.com/neovim/nvim-lspconfig/pull/1399

clason:
"Submodules" are a thing, but I haven't ever seen "subpackages" in the wild. That means little, though...

mjlbach:
Ok, this seems reasonable so I'm merging, but if someone brings this up we can redirect them at #1399 for inspiration

fredrikekre:
> I haven't ever seen "subpackages"

This is supported since some time ago, see e.g. https://github.com/timholy/SnoopCompile.jl which has `SnoopCompileBot` and `SnoopCompileCore` as sub-packages with their own `Project.toml`. I don't know which project root would make more sense here though.

mjlbach:
In that case I should probably factor out the logic in #1399 into a root utility function. Comments/PRs welcome if there is a better/more reliable way to generally solve this.

PR #1451, opened 2021-11-19
 feat(clojure_lsp) add support for shadow-cljs in clojure_lsp 

Loreo13:
Shadow-cljs is a build tool for cljs with many features such as npm integration and live reloading.
Link: https://github.com/thheller/shadow-cljs


Loreo13:
> Do you actually need this as a root marker? Is there a situation where you would have this root marker and not the others?
Yes, there are situations where you are using shadow-cljs standalone without boot, deps.edn/clojure cli, and leiningen.

PR #1455, opened 2021-11-19
feat: add spectral-language-server support

luizcorreia:
This adds support for [Spectral](https://github.com/stoplightio/spectral), a flexible JSON/YAML linter for creating automated style guides, with baked in support for OpenAPI v2 & v3.

### DIFF
--- a/lua/lspconfig/spectral.lua
+++ b/lua/lspconfig/spectral.lua
@@ -1,0 +1,32 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'spectral'
+local bin_name = 'spectral-language-server'
+
+configs[server_name] = {
+  default_config = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/spectral.lua
```diff
@@ -0,0 +1,32 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'spectral'
```
mjlbach, line 4: I've been enforcing a convention recently where x-language-server type names get postfixed with _ls (spectral_ls), up for discussion though

mjlbach:
Thanks for the contribution! Looks great, left one comment

lua/lspconfig/spectral.lua
luizcorreia: It's good to me, put the convention on the contrib file. 
I'll change that.

PR #1459, opened 2021-11-20
feat: add ruby typeprof support

mutecipher:
In Ruby 3.1, TypeProf adds experimental LSP support. This adds a configuration profile for using that LSP in NeoVim.

### DIFF
--- a/lua/lspconfig/typeprof.lua
+++ b/lua/lspconfig/typeprof.lua
@@ -1,0 +1,22 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'typeprof'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'typeprof', '--lsp', '--stdio' },
+    filetypes = { 'ruby', 'eruby' },
+    root_dir = util.root_pattern('Gemfile', '.git'),
+  },
+  docs = {
+    description = [[
+https://github.com/ruby/typeprof
+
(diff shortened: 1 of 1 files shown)


lua/lspconfig/typeprof.lua
```diff
@@ -0,0 +1,22 @@
+local configs = require 'lspconfig/configs'
+local util = require 'lspconfig/util'
+
+local server_name = 'typeprof'
+
+configs[server_name] = {
+  default_config = {
+    cmd = { 'typeprof', '--lsp', '--stdio' },
+    filetypes = { 'ruby', 'eruby' },
+    root_dir = util.root_pattern('Gemfile', '.git'),
```
mjlbach, line 10: do you want the search to search for `.git` at each step, or do you want to first search up the filesystem for gemfile and subsequently search for .git? Either way is fine. If it's the latter use find_git_ancestor.

lua/lspconfig/typeprof.lua
```diff
@@ -0,0 +1,22 @@
...
+  },
+  docs = {
+    description = [[
+https://github.com/ruby/typeprof
+
+`typeprof` is the built-in analysis and LSP tool for Ruby 3.1+.
```
mjlbach, line 16: Any instructions for installing?

mutecipher: It's actually an included gem. No additional install required as long as you're running 3.1.

PR #1460, opened 2021-11-20
fix(health): fix problem that could return empty results

erw7:
Ref. neovim/neovim#16374.

### DIFF
--- a/lua/lspconfig/health.lua
+++ b/lua/lspconfig/health.lua
@@ -2,6 +2,9 @@ local M = {}
 function M.check()
   local configs = require 'lspconfig/configs'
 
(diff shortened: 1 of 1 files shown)


lua/lspconfig/health.lua
```diff
@@ -2,6 +2,9 @@ local M = {}
 function M.check()
   local configs = require 'lspconfig/configs'
 
+  if #configs then
```
mjlbach, line 5: Are you checking for nil or 0 here?

mjlbach:
Thanks! I couldn't push to your branch, so I made a small code style change and merged in https://github.com/neovim/nvim-lspconfig/pull/1461

PR #1464, opened 2021-11-20
ci: close PRs from default branches

dundargoc:


theHamsta:
Why? Most first time contributors do their PR from master. I don't see any reason to discourage them. I don't see any bad effects for this project except that the contributors can only have one PR when they are using their master.

mjlbach:
I asked for this, it's bad practice and extremely annoying because it triggers docgen on their local forks (which could ofc also be changed). I also want to push to the feature branch in the (often) case where the PR is not updated after review.

theHamsta:
:smile: I have no problem to push to other people's master in that case but I get your point. I often told the contributors the same after they got their PR merged.

mjlbach:
I think this is pretty gentle. I wish there was a better resource for informing people, but if (which would be extremely bad) I pushed to master and they were pulling from their own  fork  (on master) for their config I would feel mildly bad when I broke things pushing WIP fixes (although I guess that in itself is a good deterrent about developing on master :))

vatosarmat:
One doesn't simply `gh pr create` from their 'nvim/site/pack/packer'

PR #1465, opened 2021-11-21
fix(lean3ls): use utf-32 offset encoding

rish987:
@Julian @gebner this should fix things for us following neovim/neovim#16382.


mjlbach:
I assume this should be merged now?

rish987:
Yes it should. Thank you!

PR #1466, opened 2021-11-21
fix(health): fix problem of always showing warning

erw7:
It seems that my last change was not reflected in #1461. Sorry, the first `#configs` was wrong. It seems to be a table without an array part, so it needs to be `not next(configs)`.


mjlbach:
let's just check vim.tbl_count == 0

mjlbach:
Thanks!

PR #1477, opened 2021-11-24
chore: remove executable check

mjlbach:
* start_client handles checking if the server is executable, it is
  redundant to check this in lspconfig
* vim.fn.executable does not respect cmd_env, which can allow injecting
  servers via a local PATH override

cc @williamboman

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -258,15 +258,12 @@ function M.server_per_root_dir_manager(_make_config)
         return
       end
       if not new_config.cmd then
-        print(
+        vim.notify(
           string.format(
-            'Error: cmd not defined for %q. You must manually set cmd in the setup{} call according to server_configurations.md, see :help lspconfig-index.',
+            'Error: cmd not defined for %q. Manually set cmd in the setup {} call according to server_configurations.md, see :help lspconfig-index.',
             new_config.name
           )
         )
-        return
-      elseif vim.fn.executable(new_config.cmd[1]) == 0 then
-        vim.notify(string.format('cmd [%q] is not executable.', new_config.cmd[1]), vim.log.levels.Error)
         return
       end
       new_config.on_exit = M.add_hook_before(new_config.on_exit, function()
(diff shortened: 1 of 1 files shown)


lua/lspconfig/util.lua
```diff
@@ -258,16 +258,13 @@ function M.server_per_root_dir_manager(_make_config)
...
           string.format(
-            'Error: cmd not defined for %q. You must manually set cmd in the setup{} call according to server_configurations.md, see :help lspconfig-index.',
+            'Error: cmd not defined for %q. Manually set cmd in the setup {} call according to server_configurations.md, see :help lspconfig-index.',
             new_config.name
           )
         )
         return
-      elseif vim.fn.executable(new_config.cmd[1]) == 0 then
-        vim.notify(string.format('cmd [%q] is not executable.', new_config.cmd[1]), vim.log.levels.Error)
-        return
```
ikalnytskyi, line 270: Hm.. it seems now I've started to experience the following error on NeoVim start:

```
Error detected while processing FileType Autocommands for "python":
E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:326: The given command "pylsp" is not executable.
Press ENTER or type command to continue
```

Yes, I don't have `pylsp` installed on that machine. Is this expected behaviour from now on? I mean to error if a LSP server is missing but configured?

mjlbach: The error message is less intrusive on the latest nightly

ikalnytskyi: Thanks. So with NeoVim 0.6 release, the behavior will remain essentially the same.

mjlbach: Yes, in fact I think the error message is even more explicit

ikalnytskyi: > Yes, in fact I think the error message is even more explicit

What about the use case when the same config is used on different machines with different LSPs installed? Making every single `lspconfig.<LSP>.setup()` conditional (checking that the LSP is installed because it may not) is sub-optimal in my opinion. :thinking:

mjlbach: it's not checkingif  it's installed, lspconfig is calling `start_client`, the client fails to start, and we report (in neovim core) the error that the client has failed to start

PR #1478, opened 2021-11-25
feature: add a pre-attach callback that allows stopping the attach

pfeiferj:
Add an optional pre_attach callback configuration option that controls whether the language server attaches to the buffer. This would be useful because it enables a user to create custom logic to prevent language server instances from being spawned for files they don't care about or to prevent adding files they don't care about to the currently running language server instance.

A couple examples of when this could be useful to a user:
* A user has triggered a diff using fugitive. Current behavior causes an `Autostart for server failed: matching root directory not detected.` message. With the feature the user could now add a simple check of the buffer's file path to prevent the attach attempt from proceeding and thereby prevent this message from appearing.
* A user goes to the definition of a symbol that brings them to a library file. Current behavior would either spawn a new language server instance or add the file to the current language server instance depending on the root of the library file. With the feature the user could now add a check for library files to prevent the attach process and thereby prevent the additional hardware resource usage of the language server.
* A user has that one project that brings the language server to its knees. Currently the user would have to disable auto start entirely for that language server to prevent it from attaching to that project. With the feature the user could create custom logic that only prevents that one project from starting the language server and still use the auto-start functionality on other projects.

Open to thoughts and suggestions on this. Documentation probably needs to be added before this is merged as well if it's decided the feature is valuable.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -10,6 +10,7 @@ function configs.__newindex(t, config_name, config_def)
     default_config = { config_def.default_config, 't' },
     on_new_config = { config_def.on_new_config, 'f', true },
     on_attach = { config_def.on_attach, 'f', true },
+    pre_attach = { config_def.pre_attach, 'f', true },
     commands = { config_def.commands, 't', true },
   }
   if config_def.commands then
@@ -198,6 +199,10 @@ function configs.__newindex(t, config_name, config_def)
     end)
 
     function manager.try_add(bufnr)
+      if config.pre_attach and not config.pre_attach(bufnr) then
+        return
+      end
+
       bufnr = bufnr or api.nvim_get_current_buf()
 
       if vim.api.nvim_buf_get_option(bufnr, 'buftype') == 'nofile' then
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -199,11 +199,11 @@ function configs.__newindex(t, config_name, config_def)
     end)
 
     function manager.try_add(bufnr)
+      bufnr = bufnr or api.nvim_get_current_buf()
+
       if config.pre_attach and not config.pre_attach(bufnr) then
         return
       end
-
-      bufnr = bufnr or api.nvim_get_current_buf()
 
       if vim.api.nvim_buf_get_option(bufnr, 'buftype') == 'nofile' then
         return


vatosarmat:
This looks useful, thanks. But is not it a fugitive issue that it reports diff-buffers as usual files? And if no, maybe LSP should not require a callback from user to check if file actually exists before attaching it?

pfeiferj:
> This looks useful, thanks. But is not it a fugitive issue that it reports diff-buffers as usual files? And if no, maybe LSP should not require a callback from user to check if file actually exists before attaching it?

So I think the behavior of both fugitive and lsp-config are reasonable here.

Fugitive is taking advantage of vim's "hypertext link" handling to create virtual files that then allows it to do behavior on writes without creating a file in the filesystem (couldn't find good help text describing vim's hypertext link behavior, closest I could find is :help gf).

Then lsp-config's behavior when it gets this hypertext link is that it can't find what the server config declares as a root directory because there's really not any parents in a hypertext link path so there's no way to find a .git folder or what have you. So lsp-config then informs you that it can't start because it needs a valid root, which is reasonable behavior to prevent the user from thinking lsp commands should be working when they can't reasonably work.

That being said, it may be reasonable for lsp-config to detect hypertext link paths and give a message that says it doesn't support that style of path to make things more clear for the user. It might also be reasonable to add a setting to turn off the error message. Finally, I'm not sure what happens if you're using a language server with "single_file_support". If a language server has "single_file_support" I suspect that you wouldn't get an error message and that the attach will actually succeed.

vatosarmat:
If language servers don't attach to these non-existent "virtual" files, then maybe something like that is reasonable?
```lua
      if  vim.api.nvim_buf_get_option(bufnr, 'buftype') == 'nofile' or 
          vim.fn.filereadable(buf_name) ~= 1 then
        return
      end

```

pfeiferj:
> If language servers don't attach to these non-existent "virtual" files, then maybe something like that is reasonable?
> ```lua
>       if  vim.api.nvim_buf_get_option(bufnr, 'buftype') == 'nofile' or 
>           vim.fn.filereadable(buf_name) ~= 1 then
>         return
>       end
> 
> ```

So these are distinctly different to nofile buffer in that vim still sees it as having a file, just not backed by the filesystem. This file path syntax is there so that netrw can edit files over ftp or what have you. I also think that language servers with single file support are able to attach to files with this type of file path without any issue, I just haven't tested that. Even if those types of language servers can't, you can't really just check for the file on the file system because language servers generally allow you to add files you haven't written to disk yet. Also, what vim-lspconfig is doing here isn't wrong, the language server can't attach and vim-lspconfig tells you why. There is an argument to be made that the message could be more specific here, but at the end of the day this is a pretty small user experience improvement and not really an issue.

I think this is a little off topic though. The point of including the fugitive example wasn't to say that it was an issue, it was to say that there could be incompatibilities with plugins and that while it isn't vim-lspconfigs job to fix every plugin compatibility issue, it is its job to provide tooling to make integration easier. 

The type of discussion I'm really looking for is whether this feature is useful? Is there something about it that I've overlooked? Could it cause any issues? Does it overly complicate configuration? Is there something that could be added in that would make it more useful? Maybe it should allow overriding the config? Are there any other events that a callback could be added for? Maybe there could be an on_error callback?

If we want to continue discussion on how vim-lspconfig handles hyperlink format file paths then maybe we should open an issue?

mjlbach:
I have actually been thinking about this scenario for the lspconfig rewrite, I think having a pre-attach callback is a bit hacky. I'd rather just run a regex against components of the filepath as a way to restrict which files (or root directories) a given estup applies to.

pfeiferj:
> I have actually been thinking about this scenario for the lspconfig rewrite, I think having a pre-attach callback is a bit hacky. I'd rather just run a regex against components of the filepath as a way to restrict which files (or root directories) a given estup applies to.

A regex wouldn't be bad, A regex would work reasonably well and would probably cover 90%+ of scenarios, but callbacks are always going to be more flexible. Just brainstorming a little, one situation that a regex wouldn't cover is if a user decided that they didn't want to attach to a file that's larger than some size. Another could be that a user might want to disable attaching based off of a hint string in the file. So the advantage of a callback is really that it allows users the flexibility to solve issues that weren't anticipated.

However, your suggestion of an ignore list does have the advantage of having less complexity during the configuration. If you were to go down that path it might also be nice to support ignore files directly as well so that on setup a user could say that if there's (for example) a .gitignore file the lsp shouldn't attach to anything listed in it.

I also just saw that there were some changes in the last couple days to expose the default config files. That means that a user can really just override the root_dir function with a wrapper around the default config's function that returns no root directory and get most of the behavior I would want. The biggest issue with using an override of the root_dir function is that the user would have to disable single_file_support to actually prevent an attach to the buffer. Another more minor thing is that the no root directory error message would still show up, but I would prefer suppressing error/warning messages to be a setting anyways.

vatosarmat:
Is 'hacky' really so bad in a Vim realm? One of the main reasons people use Vim instead of some proprietary closed IDE is that Vim provides more control on what happens under the hood. While easy to implement callback gives the full control allowing users to decide for themselves when to attach or not to attach based on their own personal reasoning

mjlbach:
>Is 'hacky' really so bad in a Vim realm? 

Hacky != hackable. I'm working on a project local configuration module right now (as in, literally right now) which IMO is a cleaner solution.

frangio:
One more potential use case: TypeScript files are a match for both `tsserver` and `denols`, but for a given file only one of those should be used. A pre-attach callback could allow cancelling the wrong one.

mjlbach:
I'm aware of the usecases this attempts to solve, but I think that "cancelling" a configuration is not a good design pattern.

pfeiferj:
I'm not sure that I agree that cancelling a configuration is a bad design pattern. It seems like a more straight forward thing to do than the things allowed by on_new_config. In a lot of ways this is just a conditional override of a matched file type similar to how on_new_config allows conditional overrides of cmd or what have you.

Now that being said, this is different from on_new_config in the sense that i want to control something that technically happened before the event fired. So maybe another way to do this is to allow setting a custom function to be fired in the auto command. As long as we have access to the manager then we could do custom logic that then conditionally fires the manager's try_add function.

pfeiferj:
I'm going to go ahead and close this PR since it's not making any movement. Will create a feature request ticket to track any new conversation.

pfeiferj:
Was going to create a new enhancement issue but instead i'll just suggest people look at #1620. It's not a perfect fit but [this comment](https://github.com/neovim/nvim-lspconfig/issues/1620#issuecomment-1001663056) in the issue in particular will likely solve the things this pr was trying to deal with.

PR #1479, opened 2021-11-25
feat: expose configs

### LINKED ISSUE #1331: feature request: expose configurations for direct use
# Description

It's currently not easy to retrieve the default configuration

```lua
print(vim.inspect(require("lspconfig.zls")) -- this probably won't give you anything useful
```

You actually need to interact with the `configs` singleton, 

```lua
print(vim.inspect(require("lspconfig.configs.zls")) -- common gotcha, but won't print anything useful either
print(vim.inspect(require("lspconfig/configs")["zls"]) -- this is the one you want
```

The problem is that this will result in `LspInfo` thinking that you've configured that `zls` server, even though you have only been interested in seeing the defaults #1302. 

But why is this happening? Remember how I said I that `configs` is a singleton? Let's test that. Assume you have a setup function similar to https://github.com/mjlbach/defaults.nvim/blob/cddcfb7821f4799df8767f362fc3d024be4fd7d1/init.lua#L233, so something like this:

```lua
local servers = { "clangd", "sumneko_lua", "bashls", "dockerls", "jsonls", "yamlls", "pyright", "cmake" }

for _, server in ipairs(servers) do
  require("lspconfig")[server].setup()
end
```

What do you think should the result be at startup when you haven't even opened a single file yet? Well, you can just see for yourself.

```lua
print(vim.inspect("lspconfig/configs"))
```

### We can do better

There are people who are a lot smarter than me that can explain why singletons _can_ be an anti-pattern, so I won't dive into that. Instead, I'll just say that there's no technical reason to keep using this design. Maybe it's more practical, but it's generally not needed. Please correct me if I'm wrong. 

### Re-usability

You may want to read some configuration to offer an API for X functionality in your plugin, https://github.com/williamboman/nvim-lsp-installer/pull/117#discussion_r729714058. Or you want to setup something fancier, https://github.com/williamboman/nvim-lsp-installer/issues/184#issuecomment-947848890.

## Suggested solution

This will probably require a big refactor, but some obvious steps could be:
1. break out the `manager` component from `configs`.
2. make sure that the `manager` module is actually independent and does not rely on any global variables/states. 
3. convert each server into their own independent modules. each having something akin to
`myserver.get_default_config()` and `clangd.setup()`. You're free to decide on the implementation details of `myserver.setup()`, in case you prefer to minimize or eliminate breaking the current API.
4. allow LspInfo to operate independently, by completely ignoring the `configs` module and only relying on `vim.lsp.get_active_clients()` and `myserver.get_default_config()`.

mjlbach:
Closes #1331 

This may have a small impact on startup time (5-10 ms)


mjlbach:
@kylo252 @williamboman I think this is what you all wanted.

`:lua print(vim.inspect(require('lspconfig.servers.pyright')))`

lua/lspconfig.lua
```diff
@@ -72,9 +70,12 @@ end
...
 function mt:__index(k)
   if configs[k] == nil then
-    -- dofile is used here as a performance hack to increase the speed of calls to setup({})
-    -- dofile does not cache module lookups, and requires the absolute path to the target file
-    pcall(dofile, script_path .. 'lspconfig/' .. k .. '.lua')
+    local success, config = pcall(require, 'lspconfig/server_configurations/' .. k)
+    if success then
+      configs[k] = config
+    else
+      vim.notify(string.format("Configuration not available for %s", k))
```
kylo252, line 77:
```suggestion
      vim.notify(string.format("Configuration not available for %s", k), vim.log.levels.WARN)
```
Although I'm not really sure what the point here is.

kylo252 (commented):
Why not call the new folder simply `servers`?

kylo252:
> This may have a small impact on startup time (5-10 ms)

I didn't notice any performance degradation following this method in my benchmarks, see the scripts here: https://gist.github.com/kylo252/8ed50f2d69c069d066bd6ef69c9ee182

- Baseline

```shell
hyperfine "./nv-bench" -r 100 --cleanup "rm -f $HOME/.cache/nvim/{lsp.log,log}" --export-markdown results-baseline.md --command-name 'baseline'
```

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `baseline` | 39.2 ± 6.9 | 30.3 | 61.5 | 1.00 |

- Modularized

```shell
hyperfine "./nv-bench" -r 100 --cleanup "rm -f $HOME/.cache/nvim/{lsp.log,log}" --export-markdown results-1479.md --command-name 'modular-configs-1479'
```

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `modular-configs-1479` | 39.3 ± 5.4 | 30.7 | 63.7 | 1.00 |

lua/lspconfig.lua
mjlbach: The point should be obvious by the error message.

mjlbach:
> Why not call the new folder simply servers?

I called it servers originally, I changed it in a later push. These aren't language servers, they are configurations for language servers. I do not want people getting confused.

mjlbach:
Sure, sligh

> > This may have a small impact on startup time (5-10 ms)
> 
> I didn't notice any performance degradation following this method in my benchmarks, see the scripts here: https://gist.github.com/kylo252/8ed50f2d69c069d066bd6ef69c9ee182, where I've added about 50 servers.
> ### Results
> 
>     * Baseline
> 
> 
> ```shell
> hyperfine "./nv-bench" -r 100 --cleanup "rm -f $HOME/.cache/nvim/{lsp.log,log}" --export-markdown results-baseline.md --command-name 'baseline'
> ```
> 
> Command 	Mean [ms] 	Min [ms] 	Max [ms] 	Relative
> `baseline` 	39.2 ± 6.9 	30.3 	61.5 	1.00
> 
>     * Modularized
> 
> 
> ```shell
> hyperfine "./nv-bench" -r 100 --cleanup "rm -f $HOME/.cache/nvim/{lsp.log,log}" --export-markdown results-1479.md --command-name 'modular-configs-1479'
> ```
> 
> Command 	Mean [ms] 	Min [ms] 	Max [ms] 	Relative
> `modular-configs-1479` 	39.3 ± 5.4 	30.7 	63.7 	1.00

Try on slower hardware. I don't think the performance impact is noticeable enough to be a blocker.

lua/lspconfig.lua
williamboman: Maybe the error message should suggest that the provided server name doesn't exist? I'm thinking some users might interpret the current error message as something that can be fixed, as the definition of "available" is a bit open for interpretation. Idk maybe something along the lines of `Unable to retrieve configuration for server %s because the server doesn't exist.`

williamboman: And this message might get noisy for 3rd party code that checks for the existence of servers (I think this is somewhat common for custom servers):

```lua
if not configs.custom_server then
  configs.custom_server = {...}
end
```

mjlbach: > 

Shouldn't apply in this case. This metamethod will only affect indexing the lspconfig table, not the configs table.

mjlbach:
@lithammer @mfussenegger @clason @neovim/lsp Does anyone have objection to this change?

mjlbach:
Ok, I think everyone is in agreement. Would be great if others could also daily drive this and report any breakages, I'll clean up the commits.

strager:
I wrote a config for nvim-lspconfig which doesn't live in the nvim-lspconfig repository. How can I make my config compatible with both older versions and newer versions of nvim-lspconfig?

* At nvim-lspconfig commit 10ad0cca1b83713ed98ed4cb7ea60f2ea8e55c49, `require("lspconfig/configs")` works and `require("lspconfig.configs")` doesn't work ("attempt to index a boolean value" when calling `.setup`).
* At nvim-lspconfig commit 97da7ed12e7e0d86e735e38a8170e941d4ed3e9a, `require("lspconfig.configs")` works and `require("lspconfig/configs") doesn't work ("Cannot access configuration for quick_lint_js. Ensure this server is listed in `server_configurations.md` or added as a custom server.").

mjlbach:
See `:help lspconfig-adding-servers`

strager:
> See `:help lspconfig-adding-servers`

Those instructions don't work with nvim-lspconfig commit 10ad0cca1b83713ed98ed4cb7ea60f2ea8e55c49.

`~/.config/nvim/init.lua`:
```lua
require('plugins') -- Packer

local lspconfig = require 'lspconfig'
local configs = require 'lspconfig.configs'
if not configs.foo_lsp then
  configs.foo_lsp = {
    default_config = {
      cmd = {'quick-lint-js', '--lsp-server'},
      filetypes = {'javascript'};
      root_dir = function(fname)
        return lspconfig.util.find_git_ancestor(fname)
      end;
      settings = {};
    };
  }
end
lspconfig.foo_lsp.setup{}
```

Results:
When I hope a .js file, I see this:
```
Error detected while processing /home/strager/.config/nvim/init.lua:
E5113: Error while calling lua chunk: /home/strager/.config/nvim/init.lua:17: attempt to index field 'foo_lsp' (a nil value)
```

Note: This init.lua does work correctly with nvim-lspconfig commit 97da7ed12e7e0d86e735e38a8170e941d4ed3e9a; I see diagnostics as expected.

mjlbach:
@strager the above snippet you've posted (sans require('plugins')) works fine for me, I dropped it into my config and was able to run quicklint.js

strager:
> @strager the above snippet you've posted (sans require('plugins')) works fine for me

With which version of nvim-lspconfig? 10ad0cca1b83713ed98ed4cb7ea60f2ea8e55c49? Something older?

mjlbach:
I'm using the latest lspconfig.

strager:
> I'm using the latest lspconfig.

The latest nvim-lspconfig works for me. My question was not about only the latest nvim-lspconfig. I want quick-lint-js' config to work with older versions of nvim-lspconfig, not just the latest. How can I do this?

mjlbach:
...why? you can check out `:help lspconfig` from before the refactor. I believe it was:

```
local lspconfig = require 'lspconfig'
local configs = require 'lspconfig.configs'
if not lspconfig.foo_lsp then
  configs.foo_lsp = {
    default_config = {
      cmd = {'quick-lint-js', '--lsp-server'},
      filetypes = {'javascript'};
      root_dir = function(fname)
        return lspconfig.util.find_git_ancestor(fname)
      end;
      settings = {};
    };
  }
end
lspconfig.foo_lsp.setup{}
```

mjlbach:
Ah, nvm, I get what you're asking (still not why). You would have to check the commit and have two different conditionals for each version.

strager:
> ...why?

I don't want people to be forced to upgrade nvim-lspconfig (pulling in a bunch of breaking changes) and also upgrade Neovim (pulling in more breaking changes) just to try my tool (quick-lint-js).

> You would have to check the commit and have two different conditionals for each version.

Yup. I'm asking how to do this.

* There's no version string I can check. (Or is there?)
* `require("lspconfig.configs")` and `require("lspconfig/configs")` return non-`nil` in old and new versions. So I can't check which one will work by checking the result of `require`.

mjlbach:
> I don't want people to be forced to upgrade nvim-lspconfig (pulling in a bunch of breaking changes) and also upgrade Neovim (pulling in more breaking changes) just to try my tool (quick-lint-js).

There's really only two breaking changes between neovim 0.6 and nvim-lspconfig, most of which would not affect normal users.

No version string, there's not a good way you can distinguish the version.

Why don't we just add quick-lint-js to lspconfig? Then it will be 1 line to set it up.

PR #1482, opened 2021-11-25
Add autocmd command to bicep documentation

jhzn:
This issue caused some headscratching from my side and should be added to the docs to help others.


mjlbach:
You need to use [=[ to avoid prematurely truncating the string with ]]  in the autocommand, I rebased and took care of this in https://github.com/neovim/nvim-lspconfig/pull/1493

PR #1484, opened 2021-11-26
Fix eslint `fix_all` sync condition

eloytoro:
The condition to check if eslint's `fix_all` should trigger a sync request or not looks a bit iffy.

### DIFF
--- a/lua/lspconfig/eslint.lua
+++ b/lua/lspconfig/eslint.lua
@@ -20,13 +20,13 @@ local function fix_all(opts)
   end
 
   local request
-  if opts.sync or false then
+  if opts.sync then
     request = function(bufnr, method, params)
-      eslint_lsp_client.request(method, params, nil, bufnr)
+      eslint_lsp_client.request_sync(method, params, nil, bufnr)
     end
   else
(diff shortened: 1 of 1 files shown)


davidmh:
I can confirm, calling `EslintFixAll` through a `BufPreWrite` autocmd does not work as expected at the moment.

The expected behavior is to have the fixes to be applied before the save, but since it's currently running asynchronously, the fixes happen after the save.

clason:
> I can confirm, calling `EslintFixAll` through a `BufPreWrite` autocmd does not work as expected at the moment.
> 
> The expected behavior is to have the fixes to be applied before the save, but since it's currently running asynchronously, the fixes happen after the save.

And does this PR help with that? Testing it and confirming whether it fixes your issue (or not) is the most helpful thing you can add here.

davidmh:
Yes, it does solve the problem.

clason:
Good to hear that! Needs a rebase, though...

mjlbach:
I rebased in https://github.com/neovim/nvim-lspconfig/pull/1492

PR #1495, opened 2021-11-27
feat: add support for sourcery

kylo252:
# Description

Add support for https://github.com/sourcery-ai/sourcery

Related discussions: https://github.com/LunarVim/LunarVim/issues/850 and https://github.com/sourcery-ai/sourcery/issues/43

--- 

#### TODO

- [ ] make sure it's clear that the token is _required_
- [ ] find some testers who already use the service

### DIFF
--- a/lua/lspconfig/server_configurations/sourcery.lua
+++ b/lua/lspconfig/server_configurations/sourcery.lua
@@ -16,13 +16,19 @@ return {
     init_options = {
       editor_version = 'vim',
       extension_version = 'vim.lsp',
-      token = '<YOUR_TOKEN>',
+      token = nil,
     },
     root_dir = function(fname)
       return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
     end,
     single_file_support = true,
   },
+  on_new_config = function(new_config, _)
+    if not new_config.init_options.token then
+      local notify = vim.notify_once or vim.notify
+      notify('[lspconfig] The authentication token must be provided in config.init_options', vim.log.levels.ERROR)
+    end
+  end,
   docs = {
     description = [[
 https://github.com/sourcery-ai/sourcery
--- a/lua/lspconfig/server_configurations/sourcery.lua
+++ b/lua/lspconfig/server_configurations/sourcery.lua
@@ -1,0 +1,49 @@
+local util = require 'lspconfig/util'
+
+local root_files = {
+  'pyproject.toml',
+  'setup.py',
+  'setup.cfg',
+  'requirements.txt',
+  'Pipfile',
+  'pyrightconfig.json',
+}
+
+return {
+  default_config = {
+    cmd = { 'sourcey', 'lsp' },
+    filetypes = { 'python' },
+    init_options = {
+      editor_version = 'vim',
+      extension_version = 'vim.lsp',
+      token = '<YOUR_TOKEN>',
+    },
+    root_dir = function(fname)
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
+    end,
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
+https://github.com/sourcery-ai/sourcery
+
+Refactor Python instantly using the power of AI.
+
+It requires the initializationOptions param to be populated as shown below and will respond with the list of ServerCapabilities that it supports.
+
+init_options = {
+    --- The Sourcery token for authenticating the user.
+    --- This is retrieved from the Sourcery website and must be
+    --- provided by each user. The extension must provide a
+    --- configuration option for the user to provide this value.
+    token = <YOUR_TOKEN>
+
+    --- The extension's name and version as defined by the extension.
+    extension_version = 'vim.lsp'
+
+    --- The editor's name and version as defined by the editor.
+    editor_version = 'vim'
+}
+]],
+  },
+}
--- a/lua/lspconfig/server_configurations/sourcery.lua
+++ b/lua/lspconfig/server_configurations/sourcery.lua
@@ -11,7 +11,7 @@ local root_files = {
 
 return {
   default_config = {
-    cmd = { 'sourcey', 'lsp' },
+    cmd = { 'sourcery', 'lsp' },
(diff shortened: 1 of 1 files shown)


Uzaaft:
I am already using the service if you`d like a tester.

Uzaaft:
Any updates regarding this PR?

kylo252:
@Uzaaft, I was waiting for your tests results 😅

Uzaaft:
Huston, we had a little miscommunication here. xD I'll test it ASAP. Hopefully within 12h.

Uzaaft:
@kylo252 The sourcery token is often located in the path: `~/.config/sourcery/auth.yaml` file. 
I propose that we remove the token part from the config(I did it locally). Makes it more secure as well regarding pushing your config to github.

Uzaaft:
Other than that, it works beautifully <3

lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -0,0 +1,49 @@
...
+}
+
+return {
+  default_config = {
+    cmd = { 'sourcey', 'lsp' },
+    filetypes = { 'python' },
+    init_options = {
+      editor_version = 'vim',
+      extension_version = 'vim.lsp',
+      token = '<YOUR_TOKEN>',
```
Uzaaft, line 19: I think we should a remove the token section, and instead require that people use the `auth.yaml` file in the `~/.config/sourcery/auth.yaml` file. Sourcery even gives the user the command: `sourcery login` to do this.

kylo252: Can it not work with a local `auth.yaml`? otherwise, where's the path for Windows? I can't find that info on their site, could you please share the link?

Uzaaft: I mean, I guess it can. But from what I see, I can't understand where the token is used, besides the variable assignment 
```lua
token = "<YOUR_TOKEN>"
```
Unless I am wrong, and it is used inside the mechanics of lspconfig?

Uzaaft: > Can it not work with a local `auth.yaml`? otherwise, where's the path for Windows? I can't find that info on their site, could you please share the link?

What is considered a "local" `auth.yaml` file? 
Also, output from `sourcery login --help`:
image

Uzaaft: The best [docs](https://pypi.org/project/sourcery-cli/) I found.

kylo252: I was specifically asking about the docs where they mention `~/.config/sourcery/auth.yaml` since I couldn't find it anywhere, because for Windows it would have to be something like `~/AppData/Local/sourcery/..`

I guess the local file would be `.sourcery.yaml`? but I can't find if it has anything to do with storing the auth token.

Uzaaft: I think so. I'll have to look around. But why use the path? Once it is setup(i.e i am logged in), I dont have to use the Path again. Sourcery fetches the token and saves it for future usage. I.e the LSP works everything after the initial setup.

Uzaaft: okay. Turns out there are no docs on this...Which is a bad thing. Opened an issue regarding this.

tcp:
If you need another tester, I can help :)

Uzaaft:
> 

It is easy to test. If you are using packer, just change the repository from which you pull the source code of nvim-lspconfig to the one in the PR.

cjber:
There is a typo `sourcey -> sourcery` which means this branch does not correctly: https://github.com/kylo252/nvim-lspconfig/blob/c45b65667b7603e246f8a53da5e0950ee1397a3e/lua/lspconfig/server_configurations/sourcery.lua#L14

Is there any progress being made on this PR? I agree the `auth.yaml` should be used for the token, but unsure how to implement myself.

kylo252:
> There is a typo `sourcey -> sourcery` which means this branch does not work correctly
> 

Thanks for catching that, fixed! 

> Is there any progress being made on this PR? I agree the `auth.yaml` should be used for the token, but unsure how to implement myself.

Basically waiting on https://github.com/neovim/nvim-lspconfig/pull/1495#discussion_r774138304, unless you guys have some suggestions for auth handling.

Uzaaft:
Auth.yaml hasn't been added to the docs. I've been stalking the docs for a few weeks now.

Uzaaft:
Sent sourcery.ai an email since their maintainer on GitHub didn't answer.

Uzaaft:
@kylo252  I think what you have currently is good enough for usage, despite `auth.yaml` perhaps being a good way to avoid pushing tokens to GitHub. We can't be stuck just because the a maintainer of an external repo does not answer.

lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -0,0 +1,49 @@
...
+}
+
+return {
+  default_config = {
+    cmd = { 'sourcery', 'lsp' },
+    filetypes = { 'python' },
+    init_options = {
+      editor_version = 'vim',
+      extension_version = 'vim.lsp',
+      token = '<YOUR_TOKEN>',
```
mjlbach, line 19: Make token nil, add a default `on_new_config` that asserts token is not nil and throws an error message with notify_once from core.

lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -0,0 +1,54 @@
...
+      token = nil,
+    },
+    root_dir = function(fname)
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
+    end,
+    single_file_support = true,
+  },
+  on_new_config = function(new_config, _)
+    if not new_config.init_options.token then
+      vim.notify_once('The authentication token must be provided in config.init_options', vim.log.levels.ERROR)
```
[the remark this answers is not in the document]
kylo252: BTW, the server would throw the same error message. That's why I hadn't originally added the handling for it.

PR #1496, opened 2021-11-27
chore: tag all vim.notify message with lspconfig

williamboman:
I was playing around a bit with some stuff when I suddenly started getting the `Cannot access configuration for %s.` warning message and at first I had no idea where it originated from and had to grep for it in across all plugins. Figured it'd make sense to prepend the messages with a tag. I think this should be part of core, somehow? Maybe making use of the `opts` param, although that'd add a lot of boilerplate

```lua
vim.notify("This is my error message.", vim.log.levels.ERROR, {
  tag = "lspconfig"
})
```

### DIFF
--- a/lua/lspconfig/server_configurations/vdmj.lua
+++ b/lua/lspconfig/server_configurations/vdmj.lua
@@ -1,7 +1,7 @@
 local util = require 'lspconfig.util'
 
 if vim.fn.has 'nvim-0.5.1' == 0 then
-  vim.notify('The VDMJ language server requires nvim > 0.5', vim.log.levels.ERROR)
+  vim.notify('lspconfig: The VDMJ language server requires nvim > 0.5', vim.log.levels.ERROR)
   return
 end
 
--- a/lua/lspconfig/server_configurations/vdmj.lua
+++ b/lua/lspconfig/server_configurations/vdmj.lua
@@ -1,10 +1,5 @@
 local util = require 'lspconfig.util'
 
-if vim.fn.has 'nvim-0.5.1' == 0 then
-  vim.notify('lspconfig: The VDMJ language server requires nvim > 0.5', vim.log.levels.ERROR)
-  return
-end
-
 local mavenrepo = util.path.join(vim.env.HOME, '.m2', 'repository', 'com', 'fujitsu')
 
 local function get_jar_path(config, package, version)
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -76,7 +76,7 @@ function mt:__index(k)
     else
       vim.notify(
         string.format(
-          'lspconfig: Cannot access configuration for %s. Ensure this server is listed in '
+          '[lspconfig] Cannot access configuration for %s. Ensure this server is listed in '
             .. '`server_configurations.md` or added as a custom server.',
           k
         ),
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -76,7 +76,7 @@ function mt:__index(k)
     else
       vim.notify(
         string.format(
-          'Cannot access configuration for %s. Ensure this server is listed in '
+          'lspconfig: Cannot access configuration for %s. Ensure this server is listed in '
             .. '`server_configurations.md` or added as a custom server.',
           k
         ),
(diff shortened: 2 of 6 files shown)


clason:
I agree that it would be nice to have an (optional) namespace attached to `vim.notify`. (@teto?)

clason:
In the meantime, I'd prefer a more "tag-like" prefix: `[lspconfig] <old message>` (There's too many colons already...)

lua/lspconfig/server_configurations/vdmj.lua
```diff
@@ -1,7 +1,7 @@
 local util = require 'lspconfig.util'
 
 if vim.fn.has 'nvim-0.5.1' == 0 then
```
clason, line 3: That check can be removed, as lspconfig itself should require 0.5.1 (and 0.6.0 next week).

clason (changes requested):

lua/lspconfig/server_configurations/vdmj.lua
mjlbach: We should also concurrently make that clear in the readme, maybe not this PR

clason: That was planned anyway, right?

(Although I'd probably synchronize with the release: "if on Neovim master, _stay_ on lspconfig master; if on 0.6.0, stay on tag 0.1.0; if on 0.5.1 or earlier, you're on your own -- no support")

mjlbach: Yes, I plan to tag on every major release of neovim

williamboman:
> I agree that it would be nice to have an (optional) namespace attached to vim.notify. (@teto?)

To add to this, it might be nice to also allow specifying an identifier for a notification itself, like `"config-not-found"`. This could then be leveraged by the "notification driver" (like https://github.com/rcarriga/nvim-notify) to replace any currently showing notifications with the same ID, instead of stacking them. This is of course already possible today via the `opts` argument, but could be nice if it was specd to drive adoption? 

> In the meantime, I'd prefer a more "tag-like" prefix: [lspconfig] <old message> (There's too many colons already...)

I prefer this too actually haha! Not sure why I went with `lspconfig: `

lua/lspconfig/server_configurations/vdmj.lua
williamboman: Getting OT now but on the topic of tagging lspconfig releases - have you considered how to deal with situations where a language server introduces a change that would necessitate a change in lspconfig (like, package name change or new args, etc.)? Pinning the lspconfig version together with neovim releases might make this difficult to handle. Just a thought!

(I can only remember one case where a server cmd have had a breaking change, [solang](https://github.com/neovim/nvim-lspconfig/pull/1388))

williamboman: I wonder if it'd make sense to find a way to have the server configs always up to date, and at the same time encourage/allow pinning the lspconfig plugin version. Maybe this problem will disappear once things are added to core (I understand this is the plan longterm)?

mjlbach: In practice, it won't matter. Tags will only be guaranteed for the lifetime of the release they correspond to. I won't backport server changes to say the tag corresponding to the neovim 0.6 release when 0.7 comes out.

williamboman: Ah gotcha! So it will be similar to like release branches, which will be subject to updates for as long as the corresponding neovim release is in "LTS" if you will. Makes sense :+1:

mjlbach: Exactly, it would be madness otherwise. Plus people are free to override setup in their configs if they really want to freeze things.

williamboman: >  it would be madness otherwise.

Screenshot 2021-11-27 at 15 51 20

teto:
> I agree that it would be nice to have an (optional) namespace attached to `vim.notify`. (@teto?)

sounds good. Can we do nested namespaces ? would be nice to know that notifications stem from lspconfig + which LSP (either via a nested namespace lspconfig.rust-analyzer or some kind of tag.
The vim.notify was mainly a way to escape the (neo)vim pager hell and there seems to be some motivation to escape that.
We could imagine standardizing some options from https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html for instance.

PR #1501, opened 2021-11-28
fix(yamlls): add yaml.docker-compose to filetypes

williamboman:
This is a filetype used by for example:
 - https://github.com/ekalinin/Dockerfile.vim
 - https://github.com/sheerun/vim-polyglot


williamboman:
I wonder if there should be some changes to lspconfig internals to respect these "compound" filetypes that are composed of more than one filetype. I might be misunderstanding how the `.` notation should be interpreted though (I'm going off of `:h 'filetype'`)

mjlbach:
I'd rather not complicate the filetype handling until it becomes an issue

PR #1503, opened 2021-11-28
Use client.request for unique calls (clangd & texlab)

ranjithshegde:
This PR only addresses clangd and texlab
It uses client.request inplace of vim.lsp.buf_request to call methods
that are outside the lspSpec such as clangd's "textDocument/switchSourceHeader"
This is done to avoid conflicts when using multiple LSPs over the same
buffer.

### DIFF
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -2,10 +2,15 @@ local util = require 'lspconfig.util'
 
 -- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
 local function switch_source_header(bufnr)
+  local clangd_client = nil
+  for _, v in ipairs(vim.lsp.buf_get_clients(0)) do
+    if v.name == 'clangd' then
+      clangd_client = v
+    end
+  end
   bufnr = util.validate_bufnr(bufnr)
   local params = { uri = vim.uri_from_bufnr(bufnr) }
-  vim.lsp.buf_request(
-    bufnr,
+  clangd_client.request(
     'textDocument/switchSourceHeader',
     params,
     util.compat_handler(function(err, result)
@@ -17,7 +22,8 @@ local function switch_source_header(bufnr)
         return
       end
       vim.api.nvim_command('edit ' .. vim.uri_to_fname(result))
-    end)
+    end),
+    bufnr
   )
 end
 
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -3,28 +3,32 @@ local util = require 'lspconfig.util'
 -- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
 local function switch_source_header(bufnr)
   local clangd_client = nil
-  for _, v in ipairs(vim.lsp.buf_get_clients(0)) do
+  for _, v in ipairs(vim.lsp.buf_get_clients(bufnr)) do
     if v.name == 'clangd' then
       clangd_client = v
     end
   end
   bufnr = util.validate_bufnr(bufnr)
   local params = { uri = vim.uri_from_bufnr(bufnr) }
-  clangd_client.request(
-    'textDocument/switchSourceHeader',
-    params,
-    util.compat_handler(function(err, result)
-      if err then
-        error(tostring(err))
-      end
-      if not result then
-        print 'Corresponding file cannot be determined'
-        return
-      end
-      vim.api.nvim_command('edit ' .. vim.uri_to_fname(result))
-    end),
-    bufnr
-  )
+  if clangd_client then
+    clangd_client.request(
+      'textDocument/switchSourceHeader',
+      params,
+      util.compat_handler(function(err, result)
+        if err then
+          error(tostring(err))
+        end
+        if not result then
+          print 'Corresponding file cannot be determined'
+          return
+        end
+        vim.api.nvim_command('edit ' .. vim.uri_to_fname(result))
+      end),
+      bufnr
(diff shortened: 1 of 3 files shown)


lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -2,10 +2,15 @@ local util = require 'lspconfig.util'
 
 -- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
 local function switch_source_header(bufnr)
+  local clangd_client = nil
```
clason, line 5: you probably want to extract that to a `util` function `util.get_buf_client(bufnr, servername)` or some such.

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -2,10 +2,15 @@ local util = require 'lspconfig.util'
...
+  for _, v in ipairs(vim.lsp.buf_get_clients(0)) do
```
clason, line 6: and should use `bufnr` here.

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -2,10 +2,15 @@ local util = require 'lspconfig.util'
...
+    if v.name == 'clangd' then
+      clangd_client = v
+    end
+  end
   bufnr = util.validate_bufnr(bufnr)
   local params = { uri = vim.uri_from_bufnr(bufnr) }
-  vim.lsp.buf_request(
-    bufnr,
+  clangd_client.request(
```
clason, line 13: what happens if `clangd_client` is `nil`?

ranjithshegde: Sure I could add a conditional `if clang_client` if you wish, but seeing as the the command is local to the language server, definied later in the document with `commands = {`, would it cause the clang_client to be nil?

clason: The server could have crashed, or the user could have switched to a different buffer with a different filetype (which doesn't "undefine" the command!).

clason: The important thing is that it handles this case gracefully and doesn't throw a Lua error.

ranjithshegde: Good point. will add the conditional and the previous suggestions! Thanks

clason (approved):
I wonder if these custom buffer requests cannot be further refactored (at least on a per-server basis) to remove even more code duplication.

For example, the texlab functions could be combined into one that takes the request string and a result function/table? Doesn't have to be this PR, though.

lua/lspconfig/util.lua
```diff
@@ -414,4 +414,14 @@ function M.get_clients_from_cmd_args(arg)
   return vim.tbl_values(result)
 end
 
+function M.get_active_client_by_name(bufnr, servername)
+  local client = nil
+  for _, v in ipairs(vim.lsp.buf_get_clients(bufnr)) do
```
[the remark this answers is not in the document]
mjlbach: I would just rename v to client, and return client, that will already implicitly return nil if client is not found

lua/lspconfig/util.lua
```diff
@@ -414,4 +414,14 @@ function M.get_clients_from_cmd_args(arg)
   return vim.tbl_values(result)
 end
 
+function M.get_active_client_by_name(bufnr, servername)
+  for _, client in ipairs(vim.lsp.buf_get_clients(bufnr)) do
+    if client.name == servername then
+      return client
+    else
```
mjlbach, line 421: This doesn't make sense now. The conditional will always return client.

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -2,23 +2,28 @@ local util = require 'lspconfig.util'
 
 -- https://clangd.llvm.org/extensions.html#switch-between-sourceheader
 local function switch_source_header(bufnr)
+  local clangd_client = util.get_active_client_by_name(bufnr, 'clangd')
   bufnr = util.validate_bufnr(bufnr)
```
mjlbach, line 6: I would move this after the validation of the bufnr

lua/lspconfig/server_configurations/texlab.lua
```diff
@@ -16,40 +16,50 @@ local texlab_forward_status = vim.tbl_add_reverse_lookup {
...
+      end),
+      bufnr
+    )
+  else
+    print 'method textDocument/build is not supported by any servers active on the current buffer'
+  end
 end
 
 local function buf_search(bufnr)
+  local texlab_client = util.get_active_client_by_name(bufnr, 'texlab')
```
mjlbach, line 42: Same, move validation check later.

lua/lspconfig/util.lua
```diff
@@ -418,8 +418,6 @@ function M.get_active_client_by_name(bufnr, servername)
   for _, client in ipairs(vim.lsp.buf_get_clients(bufnr)) do
```
mjlbach, line 418: I think ipairs is fine here, because buf_get_clients should be a non-holey list

mjlbach: Actually no

mjlbach: This should be pairs.

mjlbach:
Ok, I think we good now.

mjlbach:
lint failing, you can remove the local lsp var in texlab

neovim-discourse:
This pull request has been mentioned on **Neovim Discourse**. There might be relevant details there:

https://neovim.discourse.group/t/switch-to-h-cpp-via-lsp/813/11

Issue #1504, opened 2021-11-28
The tailwindcss server doesn't work by default since the 0.0.5 upgrade


oryanmoshe:
### Description

On previous versions of `@tailwindcss/language-server` everything worked fine, but after the change to `0.0.5` it broke because the default configuration requires `classAttributes` to be set.
Setting `classAttributes` manually on server startup seems to solve this issue, but it'd be better if there were sensible defaults automatically set.

I personally use [williamboman/nvim-lsp-installer](https://github.com/williamboman/nvim-lsp-installer), so my configuration is a bit different;
But I tested the following configuration (without the installer) and it seems to work as well:

```lua
require('lspconfig').tailwindcss.setup {
  on_attach = on_attach,
  capabilities = capabilities,
  settings = {
    tailwindCSS = {
      classAttributes = { 'class' , 'className' , 'classList' }
    }
  }
}
```

### Neovim version

NVIM v0.5.1
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

d9aa848da3905e0f8153e546d7b630d3d13e0435

### Operating system and version

macOS Monterey 12.1 Beta 3

### Affected language servers

tailwindcss

### Steps to reproduce

0. Installing the tailwindcss language server: `npm i -g @tailwindcss/language-server`
1. Creating test directory: `mkdir twcrepro && cd twcrepro`
2. Initializing an empty `npm` package: `npm init -y`
3. Installing dependencies: `npm i -D tailwindcss@latest solid-js`
4. Downloading the minimal nvim configuration: `curl -L https://gist.githubusercontent.com/oryanmoshe/3dc78935b4650774df505a027e74a749/raw/9c95f5b186f5ad44107275166d71b767f2baa0b2/minimal.broken.lua -o minimal.lua`
5. Creating a dummy `tailwind.config.js` file: `touch tailwind.config.js`
6. Creating a victim file: `printf "import { Component } from 'solid-js';\n\nconst SimpleDiv:Component = () => {\n\treturn (\n\t\t<div class=''></div>\n\t)\n}" > index.tsx`
7. Try to edit the file and add some tailwindcss class (let's say `bg-white`) to the div, wait for autocomplete: `nvim -nu minimal.lua index.tsx`

In short:
```fish
npm i -g @tailwindcss/language-server
mkdir twcrepro && cd twcrepro
npm init -y
npm i -D tailwindcss@latest solid-js
[...]

PR #1505, opened 2021-11-28
fix(tailwindcss): Add default classAttributes

oryanmoshe:
Version `0.0.5` of [`@tailwindcss/language-server`](https://www.npmjs.com/package/@tailwindcss/language-server) requires setting `classAttriubtes` during LSP setup.
nvim-lspconfig didn't set any class atributes by default, so the LSP failed to initialize.

This fix adds these default classAttributes to tailwindcss:
- `class`
- `className`
- `classList`

This change passes all tests ran by `make lint` and `make test`.

Fixes neovim/nvim-lspconfig#1504

### DIFF
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -6,6 +6,7 @@ local class_attributes = {
   'class',
   'className',
   'classList',
+  'ngClass',
 }
 
 return {
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -2,13 +2,6 @@ local util = require 'lspconfig.util'
 
 local bin_name = 'tailwindcss-language-server'
 
-local class_attributes = {
-  'class',
-  'className',
-  'classList',
-  'ngClass',
-}
-
 return {
   default_config = {
     cmd = { bin_name, '--stdio' },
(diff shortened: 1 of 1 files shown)


mjlbach:
This seems like a regression that should be reported upstream?

We don't use emojis in commit messages, please follow conventional commit standards.

oryanmoshe:
> This seems like a regression that should be reported upstream?

Regarding the upstream regression - this seems to work fine on their VSCode extension.
Fixing this on nvim-lspconfig will ensure this LSP works both here and on `williamboman/nvim-lsp-installer` which are the main ways people use tailwindcss-language-server on neovim.

> We don't use emojis in commit messages, please follow conventional commit standards.

Sorry, gitmojis are standard on my personal flow.
Do you want me to create a new commit without the emoji?
Or should I just scrap the PR and report directly to tailwindcss?

mjlbach:
I'm fine merging the workaround but I'm just curious if this is a list now we're going to have to manually maintain.

You can just amend the commit and update the message.

oryanmoshe:
> I'm fine merging the workaround but I'm just curious if this is a list now we're going to have to manually maintain.

Shouldn't be something we'll have to maintain, the defaults I have set here are the defaults in the VSCode extension.
I can't see any reason for someone to create a JS framework with a different class attribute.
And even if they do, they can just override the defaults (:

> You can just amend the commit and update the message.

Done.

mjlbach:
I personally would just add the list directly instead of defining/referencing a variable, not sure why that is preferable.

mjlbach:
style-lint failing, you can run stylua in the root of the lspconfig directory

PR #1511, opened 2021-11-29
feat: add golangcilsp support

desdic:
Combination of both lint server and client

https://github.com/nametake/golangci-lint-langserver
https://github.com/golangci/golangci-lint

once/if merged I'll create a pull request on nvim-lsp-installer to make it easy to install

### DIFF
--- a/lua/lspconfig/server_configurations/golangci_lint_ls.lua
+++ b/lua/lspconfig/server_configurations/golangci_lint_ls.lua
@@ -1,0 +1,34 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'golangci-lint-langserver' },
+    filetypes = { 'go', 'gomod' },
+    init_options = {
+      command = { 'golangci-lint', 'run', '--enable-all', '--disable', 'lll', '--out-format', 'json' },
+    },
+    root_dir = function(fname)
+      return util.root_pattern 'go.work'(fname) or util.root_pattern('go.mod', '.git')(fname)
+    end,
+  },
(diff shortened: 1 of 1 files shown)


desdic:
> Can we rename the file to `golangci_ls.lua`? Also would be good if you can add some installation instructions but otherwise looks good assuming tests pass.

Just renamed it .. I can do documentation no problem but I'm adding it to the nvim-lsp-installer (pull request) so maybe that would be ok ?

mjlbach:
There's a large fraction (I would say the majority) of users who don't use nvim-lsp-installer (including myself).

mjlbach:
Actually sorry, golangci_lint_ls instead of golangci_ls

desdic:
> Actually sorry, golangci_lint_ls instead of golangci_ls, didn't realize the language server's name is golangci_lint

Renamed and added 2 commands to install the binaries needed.

fipso:
4FUnkui
 
am i doing something wrong or is lsp broken right now ?

mjlbach:
Did you read the error message? It tells you what is wrong.

desdic:
> 4FUnkui
> 
> am i doing something wrong or is lsp broken right now ?

I think you found the above code on https://github.com/nametake/golangci-lint-langserver right ? I had problems with it too so I made the above patch so you can replace the above with

```
require'lspconfig'.golangci_lint_ls.setup{}
```

or create a custom server .. I did both but decided to make a pull request to make it available for who ever needed it

fipso:
thanks @desdic

PR #1518, opened 2021-11-30
fix(fsautocomplete): changed cmd to executable of newest version

dkirchhof:


mjlbach:
Thanks! Is this change documented upstream somewhere?

dkirchhof:
> Thanks! Is this change documented upstream somewhere?

Actually i don't find useful information about it.  
I came across the problem yesterday after updating fsautocomplete from 0.46.x to 0.49.4...   
Maybe someone else could confirm it :thinking:

PR #1520, opened 2021-11-30
perf: do not require vim.lsp unless necessary

mjlbach:


clason:
Missed one: https://github.com/neovim/nvim-lspconfig/blob/065d0eee51cc9ee74908786ce195cdaa6107d76c/lua/lspconfig/util.lua#L304

clason:
Also, sorry for the conflict. (Let me know if you want me to take over the PR.)

mjlbach:
I am probably not going to merge this, any module that call vim.lsp will undo the load time improvement (including starting a language server), so in most cases this will have no realistic impact on startup time. I'd rather see an impatient.nvim-esque solution in core.

Issue #1500, opened 2021-11-27
Full windows support for all applicable servers


mjlbach:
### Description

I want to delete [this section](https://github.com/neovim/nvim-lspconfig#windows) from the readme. So lets test all the language servers on windows and update the cmd [like this](https://github.com/neovim/nvim-lspconfig/blob/8c9962bae5ac3b1f8b3d84c18314325f2d8df7fd/lua/lspconfig/server_configurations/pyright.lua#L3-L6)

Please list servers that need to be updated (better yet file PRs)

williamboman:
As I've been going through https://github.com/williamboman/nvim-lsp-installer/pull/283 I've identified the following server configurations that might have to be revisited for Windows:

- erlangls
- ccls
- codeql
- solang

I will probably check these myself once I get around to merging it, just a heads up

PR #1521, opened 2021-12-01
fix(jsonls): update cmd for windows support

agustiawan14:
Try to close: #1500

This is my first contribution ever to any open source so please be kind :)

I tested other lsp like html, cssls, and eslint, but I don't know if it is better to have multiple PR per lsp or one PR with multiple lsp.


mjlbach:
Thanks :)

PR #1522, opened 2021-12-01
fix(npm): resolve the command correctly on windows

### LINKED ISSUE #1500: Full windows support for all applicable servers

kylo252:
Fixes #1500

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,18 @@ The general form of adding a new language server is to start with a minimal skel
 When choosing a server name, convert all dashes (`-`) to underscores (`_`) If the name of the server is a unique name (`pyright`, `clangd`) or a commonly used abbreviation (`zls`), prefer this as the server name. If the server instead follows the pattern x-language-server, prefer the convention `x_ls` (`jsonnet_ls`). 
 
 `default_config` should include, at minimum the following:
-* `cmd`: a list which includes the executable name as the first entry, with arguments constituting subsequent list elements (`--stdio` is common)
+* `cmd`: a list which includes the executable name as the first entry, with arguments constituting subsequent list elements (`--stdio` is common).
+Note that Windows has a limitation when it comes to directly invoking a server that's installed by `npm` or `gem`, so it requires additional handling.
+
+```lua
+local bin_name = 'typescript-language-server'
+local cmd = { bin_name, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
+end
+```
+
 * `filetypes`: a list for filetypes a 
 * `root_dir`: a function (or function handle) which returns the root of the project used to determine if lspconfig should launch a new language server, or attach a previously launched server when you open a new buffer matching the filetype of the server. Note, lspconfig does not offer a dedicated single file mode (this is not codified in the spec). Do not add `vim.fn.cwd` or `util.path.dirname` in `root_dir`. A future version of lspconfig will provide emulation of a single file mode until this is formally codified in the specification. A good fallback is `util.find_git_ancestor`, see other configurations for examples.
 
--- a/README.md
+++ b/README.md
@@ -188,10 +188,6 @@ Please see the [wiki](https://github.com/neovim/nvim-lspconfig/wiki) for additio
 * [Project local settings](https://github.com/neovim/nvim-lspconfig/wiki/Project-local-settings)
 * [Recommended plugins for enhanced language server features](https://github.com/neovim/nvim-lspconfig/wiki/Language-specific-plugins)
 
-## Windows
-
-In order for neovim to launch certain executables on Windows, it must append `.cmd` to the command name. To work around this, manually append `.cmd` to the entry `cmd` in a given plugin's setup{} call.
-
 ## Contributions
 
 If you are missing a language server on the list in [server_configurations.md](doc/server_configurations.md), contributing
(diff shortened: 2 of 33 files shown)


kylo252:
If it turns out that there's a way to handle this in core without invoking `cmd.exe`, then all we have to do is `dap` that `win32` condition and done :)

lithammer:
Is this only Node specific, or could we blanket apply this to every `cmd` in a more central location (instead of every server config file)?

kylo252:
> Is this only Node specific, or could we blanket apply this to every `cmd` in a more central location (instead of every server config file)?

It's probably node-specific, since it depends on creating [shims](https://www.npmjs.com/package/cmd-shim). We can add handling for that either globally or in core, but it's going to take more effort.

kylo252:
@williamboman, here's my attempt to describe the problem more clearly https://github.com/neovim/neovim/issues/16484

mjlbach:
@williamboman this solves the issue of volta, which does not provide the postfixed `.cmd` scripts. It should universally work on windows. I'd rather not inject some modification of the `cmd` passed to start_client to account for windows but fix it locally in lspconfig per server.

kylo252:
We can also prepend the logic once in the command that's sent to `start_client`. It's _probably_ fine :sweat_smile:

mjlbach:
> We can also prepend the logic once in the command that's sent to start_client.

I am saying no 🙃

williamboman:
What do you think of having a set of candidates that will be attempted instead? Like

```lua
local candidates = {
  { "typescript-language-server" },
  is_win and { "typescript-language-server.cmd" },
  is_win and { "cmd.exe", "/C", "typescript-language-server" }
}

return {
  default_config = {
    cmd_candidates = candidates
  }
}
```

mjlbach:
No

kylo252:
For completeness' sake, there's also this ~~atrocity~~ thing which seems like a core dependency for `npm` and has about 1.7M weekly downloads......

> Figure out what a `cmd-shim` is pointing at. This acts as the equivalent of `fs.readlink`.
https://github.com/npm/read-cmd-shim/blob/488b31e867b2f7df6eb44fca6569a5771ee03aee/index.js#L1-L69

mjlbach:
I don't understand why this PR is not sufficient to solve the issue. I think having a chain of cmd is hacky at best. There is no good way to test ahead of time which is executable and I like the simplicity of configurations being *light* sugar over uv_spawn

williamboman:
> I don't understand why this PR is not sufficient to solve the issue. I think having a chain of cmd is hacky at best. There is no good way to test ahead of time which is executable and I like the simplicity of configurations being _light_ sugar over uv_spawn

I'm not opposed to it! I just wanted to raise concerns around using `cmd.exe` as an intermediary. Are signals and pipes properly forwarded, are there any throughput bottlenecks? 🤷‍♂️ One thing I've learned, and especially for Windows, is that things never work in a reliable manner, so this will probably produce a new set of issues haha. We'll see

kylo252:
I should probably add those managed by `yarn` as well.

kylo252 (commented):
@mjlbach, what about these two: 

- https://github.com/neovim/nvim-lspconfig/blob/5b8624c1bcd332e9fd7ae33a2ca910adb31a7ae7/lua/lspconfig/server_configurations/theme_check.lua#L7
- https://github.com/neovim/nvim-lspconfig/blob/5b8624c1bcd332e9fd7ae33a2ca910adb31a7ae7/lua/lspconfig/server_configurations/rescriptls.lua#L24

Do they count?

mjlbach:
I don't think so

kylo252:
I should probably test the ruby ones, ~~but I have no idea if that's even supported on Windows...~~ It is, and it's the same issue as `npm`.

I should add the same fix for all the `gem` managed servers as well 😢

mjlbach:
Out of date installation instructions can be handled in a separate PR

kylo252:
There are probably a few more node modules, but I'd have to use my old script from https://github.com/neovim/nvim-lspconfig/pull/1254 to find them, since they're missing the installation instructions which I've been using so far to classify them.

williamboman:
Sorry for interjecting, but wasn't this only a problem with Volta on Windows (i.e. only Node.js)?

mjlbach:
Well, some of the cmd were missing too, and it seems like this is a universal fix.

kylo252:
I think that should be it! There's still `typeroft`, but it requires ruby 3.1 which didn't get installed by default for me 🤷  https://github.com/neovim/nvim-lspconfig/blob/5b8624c1bcd332e9fd7ae33a2ca910adb31a7ae7/lua/lspconfig/server_configurations/typeprof.lua#L13

mjlbach:
Awesome, thanks for the hard work!

mjlbach:
Also remove the section from readme.md on windows support.

kylo252:
> add a section in CONTRIBUTING.md explaining this.

I'm not very good at writing these kinds of docs, please let me know if you had something else in mind. Maybe @clason can take a look as well.

mjlbach:
I'm going to go ahead and merge this. If there are issues it can be reverted.

PR #1525, opened 2021-12-01
fix: allow launching language server on empty files

mjlbach:


mjlbach:
@clason rough fix

clason:
Thanks, looks good!

lua/lspconfig/configs.lua
```diff
@@ -216,7 +218,13 @@ function configs.__newindex(t, config_name, config_def)
       if root_dir then
         id = manager.add(root_dir, false)
       elseif config.single_file_support then
-        local pseudo_root = util.path.dirname(api.nvim_buf_get_name(0))
+        local bufname = api.nvim_buf_get_name(0)
```
williamboman, line 221: Shouldn't this get the bufname for `bufnr`?

lua/lspconfig/configs.lua
```diff
@@ -216,7 +218,13 @@ function configs.__newindex(t, config_name, config_def)
...
       elseif config.single_file_support then
-        local pseudo_root = util.path.dirname(api.nvim_buf_get_name(0))
+        local bufname = api.nvim_buf_get_name(bufnr)
+        local pseudo_root
+        if bufname ~= "" then
+          pseudo_root = util.path.dirname(bufname)
+        else
+          pseudo_root = vim.loop.cwd()
+        end
         id = manager.add(pseudo_root, true)
```
clason, line 228: @mjlbach I know how you hate `cwd` hackery, but I'd still like to remove the error; I don't really care about the attaching servers to unnamed buffers. Would you consider this PR with a change like the following to be mergable?

```suggestion
        if bufname ~= "" then
          local pseudo_root = util.path.dirname(bufname)
          id = manager.add(pseudo_root, true)
        end
```

PR #1526, opened 2021-12-01
change bin_name to vscode-json-languageserver

hnspn:
if [this ](https://www.npmjs.com/package/vscode-json-languageserver) is the right package
then the bin_name is wrong


mjlbach:
It's not the right package.

PR #1533, opened 2021-12-04
feat: add quick-lint-js

mjlbach:
@strager

### DIFF
--- a/lua/lspconfig/server_configurations/quick_lint_js.lua
+++ b/lua/lspconfig/server_configurations/quick_lint_js.lua
@@ -1,0 +1,18 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'quick-lint-js', '--lsp-server' },
+    filetypes = { 'javascript' },
+    root_dir = util.root_pattern('package.json', 'jsconfig.json', '.git'),
+  },
+  docs = {
+    description = [[
+https://quick-lint-js.com/
+
+quick-lint-js finds bugs in JavaScript programs.
+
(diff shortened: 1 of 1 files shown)


KasonBraley:
Using the default install configuration from [here](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#quick_lint_js) works when a `.js` file is in a folder with a package.json.

But when creating a `.js` file outside of any project folder, the following error is observed when launching neovim on that file.
`lspconfig: Autostart for quick_lint_js failed: matching root directory not detected.`


I was able to fix this issue by using the `root_dir` config that @strager had setup originally in his configuration [here](https://github.com/quick-lint/quick-lint-js/blob/master/plugin/vim/quick-lint-js.vim/lua/lspconfig/quick_lint_js.lua).

```
local util = require("lspconfig.util")

require("lspconfig").quick_lint_js.setup({
  root_dir = function(fname)
    local root = util.path.dirname(fname)
      util.path.traverse_parents(fname, function(dir, _path)
      root = dir
    end)
    return root
  end,
})
```

Should this be the default behavior for `root_dir`?

mjlbach:
Nope, it's expressly forbidden actually to use root_directory in that way (it will fail our CI). You can try enabling single_file_support in setup {} to see if it works, then that can be the default.

KasonBraley:
> Nope, it's expressly forbidden actually to use root_directory in that way (it will fail our CI). You can try enabling single_file_support in setup {} to see if it works, then that can be the default.

Thank you, setting `single_file_support` to true did indeed fix that issue. @strager should be the one to decide if it is the default though.

PR #1537, opened 2021-12-04
docs: update lua-language-server instructions

mjlbach:


lua/lspconfig/server_configurations/sumneko_lua.lua
```diff
@@ -2,6 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
+    cmd = { 'lua-languge-server' },
```
clason, line 5:
```suggestion
    cmd = { 'lua-language-server' },
```

clason (changes requested):

PR #1549, opened 2021-12-10
modify ltex server config

flaviusbuffon:
When using ltex as latex lsp server, it will notice that no root_dir,  I change somethings in ltex.lua


mjlbach:
@clason I feel like we had discussed this and I remember there being a reason not to do this, but I can't recall what it was.

clason:
I think the argument was that `ltex` is in fact _not_ a LaTeX lsp server (although `tex` is one of several filetypes it works well for), so hardcoding this as *primus inter pares* doesn't seem appropriate.

clason:
The right thing here is to check whether `ltex-ls` works with `single_file_support = true`, and if so, add that to the default settings.

mjlbach:
it should AFAIK

mjlbach:
I think this addresses your main concern https://github.com/neovim/nvim-lspconfig/pull/1556

PR #1551, opened 2021-12-10
fix: add missing cmds

williamboman:


mjlbach:
Are these installable via some source (system package manager, npm, etc.)?

williamboman:
Sqlls is distributed via npm. Lemminx I'm not sure about, but I figured lspconfig just assumes things exists on path and that's that? I can remove it if there has to be a standard installation method compatible with the `cmd`

PR #1555, opened 2021-12-10
fix(taplo): add settings table to taplo

timbedard:
Without the settings table present, no use configuration of the server can be done. This has been causing a lot of issues for us Neovim Taplo users. See: https://github.com/tamasfe/taplo/issues/181 and https://github.com/tamasfe/taplo/issues/167


mjlbach:
I'm not sure I understand. This is just the first workspace/didChangeConfiguration that is sent. Users can always send a second settings request later on to add their settings or customize the one in settings. Are you saying we *need* to send an empty settings table for taplo to work? That seems like an upstream issue?

timbedard:
Ugh. Apparently this is hairier than I thought. Thanks for the response. I'm digging deeper.

PR #1559, opened 2021-12-11
fix(jsonnet_ls): merge cmd_env instead of overriding it entirely

williamboman:
Found via tests added in https://github.com/rish987/nvim-lspconfig-test/pull/57

### DIFF
--- a/lua/lspconfig/server_configurations/jsonnet_ls.lua
+++ b/lua/lspconfig/server_configurations/jsonnet_ls.lua
@@ -17,9 +17,12 @@ return {
       return util.root_pattern 'jsonnetfile.json'(fname) or util.find_git_ancestor(fname)
     end,
     on_new_config = function(new_config, root_dir)
-      new_config.cmd_env = {
-        JSONNET_PATH = jsonnet_path(root_dir),
-      }
+      if not new_config.cmd_env then
+        new_config.cmd_env = {}
+      end
+      if not new_config.cmd_env.JSONNET_PATH then
+        new_config.cmd_env.JSONNET_PATH = jsonnet_path(root_dir)
+      end
     end,
   },
   docs = {


mjlbach:
Just set a single table entry, it's cleaner than deep extend.

williamboman:
> Just set a single table entry, it's cleaner than deep extend. I assume this is only problematic for nvim-lsp-installer.

Yeah most likely. It might be problematic if the LSP supports other envs (or if there's a use case to customize JSONNET_PATH)

williamboman:
I guess `cmd_env` is nil by default? So something like this?

```diff
diff --git a/lua/lspconfig/server_configurations/jsonnet_ls.lua b/lua/lspconfig/server_configurations/jsonnet_ls.lua
index 7285a64..bb79493 100644
--- a/lua/lspconfig/server_configurations/jsonnet_ls.lua
+++ b/lua/lspconfig/server_configurations/jsonnet_ls.lua
@@ -17,9 +17,12 @@ return {
       return util.root_pattern 'jsonnetfile.json'(fname) or util.find_git_ancestor(fname)
     end,
     on_new_config = function(new_config, root_dir)
-      new_config.cmd_env = vim.tbl_deep_extend('force', {
-        JSONNET_PATH = jsonnet_path(root_dir),
-      }, new_config.cmd_env or {})
+      local jsonnet_libpath = jsonnet_path(root_dir)
+      if new_config.cmd_env then
+        new_config.cmd_env.JSONNET_PATH = jsonnet_libpath
+      else
+        new_config.cmd_env = { JSONNET_PATH = jsonnet_libpath }
+      end
     end,
   },
   docs = {
```

mjlbach:
```lua
if not new_config.cmd_env then
  new_config.cmd_env = {}
end

new_config.cmd_env = your_thing
```

williamboman:
I think it'd be nice to be able to access the `JSONNET_PATH` env, even when providing a custom `cmd_env` (unless `cmd_env.JSONNET_PATH` is set, in which case it takes precedence). Or do you think that's a bad idea?

williamboman:
Ah nvm I'm slow, havent had my coffee yet 😅

lua/lspconfig/server_configurations/jsonnet_ls.lua
```diff
@@ -17,9 +17,12 @@ return {
...
     on_new_config = function(new_config, root_dir)
-      new_config.cmd_env = {
-        JSONNET_PATH = jsonnet_path(root_dir),
-      }
+      if not new_config.cmd_env then
+        new_config.cmd_env = {}
+      end
+      if not new_config.cmd_env.JSONNET_PATH then
+        new_config.cmd_env.JSONNET_PATH = jsonnet_path(root_dir)
+      end
```
williamboman, line 25: I guess doing it this way might be problematic as we're mutating the `cmd_env` table (if it exists) argument. Might be problematic if the `jsonnet_ls` server is started more than once per session, as `new_config.cmd_env.JSONNET_PATH` will be "permanently" set after first client initializes

PR #1565, opened 2021-12-13
fix(ccls): delete invalid root directory pattern

rhcher:
ccls doesn't support compile_flags.txt, see
https://github.com/MaskRay/ccls/wiki/FAQ#does-ccls-support-compile_flagstxt

### DIFF
--- a/lua/lspconfig/server_configurations/ccls.lua
+++ b/lua/lspconfig/server_configurations/ccls.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'ccls' },
     filetypes = { 'c', 'cpp', 'objc', 'objcpp' },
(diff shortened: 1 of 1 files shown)


mjlbach:
To be fair, the root pattern doesn't imply that `ccls` uses it, it's just a marker for where we set the root of the C/C++ project. Fine by me to remove anyways.

lua/lspconfig/server_configurations/ccls.lua
```diff
@@ -36,7 +36,7 @@ lspconfig.ccls.setup {
 
 ]],
     default_config = {
-      root_dir = [[root_pattern("compile_commands.json", ".ccls", "compile_flags.txt", ".git") or dirname]],
+      root_dir = [[root_pattern("compile_commands.json", ".ccls", ".git") or dirname]],
```
mjlbach, line 39: should remove dirname also (out of date)

PR #1568, opened 2021-12-14
feat: add cssmodules_ls

williamboman:


mjlbach:
See pyright for the appropriate windows command.

lua/lspconfig/server_configurations/cssmodules_ls.lua
```diff
@@ -0,0 +1,32 @@
...
+end
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'javascript', 'javascriptreact', 'typescript', 'typescriptreact' },
+    root_dir = util.find_package_json_ancestor,
+  },
+  docs = {
+    package_json = 'https://github.com/antonk52/coc-cssmodules/blob/master/package.json',
```
mjlbach, line 17: I don't think we should use a coc extension's package.json

williamboman: Yeah now when I look closer it seems like that package.json is providing coc-style settings. It'd probably be better to file an issue/PR in the [standalone language server](https://github.com/antonk52/cssmodules-language-server) repo and try to introduce the LSP metadata in that package.json as well. I'll remove this `package_json` for now

williamboman: Hmm I took a look at a few other package_json entries for reference. It seems like they suffers the same issue, but s/coc/VSCode/ (for example intelephense & solargraph)?  Or maybe I misunderstand the `package_json.contributes` schema

mjlbach: It depends, sometimes these mix vscode and legitimate server settings, in which case I haven't particularly cared.

williamboman: It's actually just good ole JSONSchema https://json-schema.org/

williamboman: The unfortunate thing is that many LSP server devs have ended up merging the LSP server settings with the VSCode-extension settings, which is what ends up being documented in a JSONSchema

williamboman: pylsp is one of few servers that provide a separate schema for the server itself, maybe it could be used as reference - https://github.com/python-lsp/python-lsp-server/blob/develop/pylsp/config/schema.json

antonk52: Thank you for the info. I am going to look into it and updating the package.json accordingly👍

PR #1578, opened 2021-12-16
fix: remove vim.lsp.util.close_preview_autocmd

ii14:
Inline vim.lsp.util.close_preview_autocmd, it was removed from neovim.


ii14:
Duplicate of #1575, I should probably pay more attention lol

Issue #1524, opened 2021-12-01
Autocommand error when setting filetype


clason:
### Description

When setting the filetype of an empty buffer to one for which a server is configured to attach, an error is thrown:
```
Error detected while processing FileType Autocommands for "lua":
E5108: Error executing lua /usr/local/share/nvim/runtime/lua/vim/uri.lua:62: attempt to index local 'path' (a nil value)
stack traceback:
        /usr/local/share/nvim/runtime/lua/vim/uri.lua:62: in function 'uri_from_fname'
        ...ck/packer/start/nvim-lspconfig/lua/lspconfig/configs.lua:191: in function '_make_config'
        .../pack/packer/start/nvim-lspconfig/lua/lspconfig/util.lua:255: in function 'add'
        ...ck/packer/start/nvim-lspconfig/lua/lspconfig/configs.lua:220: in function 'try_add'
        [string ":lua"]:1: in main chunk
```

### Neovim version

NVIM v0.7.0-dev+661-g99f6260f8
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

5b8624c1bcd332e9fd7ae33a2ca910adb31a7ae7

### Operating system and version

macOS 12.0

### Affected language servers

all

### Steps to reproduce

1. `nvim -u minimal_init.lua`
2. `set ft=python`

(with the stock provided `minimal_init.lua`, which sets up `pyright`)

### Actual behavior

error (see above)

### Minimal config

```Lua
curl -fLO https://raw.githubusercontent.com/neovim/nvim-lspconfig/master/test/minimal_init.lua
```

### Health check

lspconfig: require("lspconfig.health").check()
========================================================================
  - INFO: pyright: configuration checked.

### LSP log

N/A

d9d6ka:
The problem is also exists in neovim v0.6.0.
If I open the Python-file via `:e` command then no problem occurs.

But when I try to change Filetype then I receive the same error as @clason 

If I try to open the file via `dirvish` I receive the similar error. `add` and `try_add` functions are there as well.
```
Обнаружена ошибка при обработке FileType Автокоманды для "python":
E5108: Error executing lua ....0-dev-665-g4e352ecc4\share\nvim\runtime/lua/vim/lsp.lua:239: cmd_cwd: expected directory, got D:/git/advent-of-code/2015/day-07/part1.py
stack traceback:
        [C]: in function 'error'
        vim/shared.lua:608: in function 'validate'
[...]

PR #1581, opened 2021-12-16
fix: do not attach to buffer without filename

clason:
also use correct `bufnr` for getting path

more conservative fix for #1524 than #1525 (avoids `cwd` hackery).

@mjlbach


mjlbach:
It's not cwd hackery, it's the correct fix...

clason:
I'm happy to have _your_ fix merged instead of this; I just want to get rid of the error ASAP. 

(Not attaching to unnamed buffers sounds like reasonable behavior to me, though.)

PR #1585, opened 2021-12-18
Update README for setup on_attach

pickfire:
on_attach is required but user might not go through the huge section.


mjlbach:
To clarify, it is not required. I don't think we should add a dummy on_attach to all examples, but if you want to make the keybindings section more explicit, that would be welcome.

pickfire:
I followed it and it didn't work. At the very least people who omit it knows what they are doing, but the people using it for the first time don't so I think we should put it there.

PR #1595, opened 2021-12-23
fix(jedi-language-server) add package_json

ainsleymcgrath:
👉 This change should (re-)enable compatibility with dependents, [LunarVim](https://github.com/LunarVim/LunarVim) and [`nlsp-settings.nvim`](https://github.com/tamago324/nlsp-settings.nvim).

I followed a winding path here. 🛣️ 

LunarVim relies on this repo to install LSP clients, as it should. It also [relies](https://www.lunarvim.org/languages/#server-settings) on `nlsp-settings.nvim` to _configure_ those LSP clients.

`nlsp-settings.nvim` [currently does not support](https://github.com/tamago324/nlsp-settings.nvim/blob/main/schemas/README.md) Jedi 😢 which is sad because I love Jedi.

🙏 **But thankfully**, much of the LSP support in `nlsp-settings.nvim` comes from autogenerated schemas based on _this_ repo. And (it looks to me like) `jedi-language-server` is [not getting picked up due to](https://github.com/tamago324/nlsp-settings.nvim/blob/main/scripts/gen_schemas.lua#L58) a missing `package_json`.

### DIFF
--- a/lua/lspconfig/server_configurations/jedi_language_server.lua
+++ b/lua/lspconfig/server_configurations/jedi_language_server.lua
@@ -16,7 +16,7 @@ return {
     single_file_support = true,
   },
   docs = {
-    package_json = 'https://github.com/pappasam/coc-jedi/blob/main/package.json',
+    package_json = 'https://raw.githubusercontent.com/pappasam/coc-jedi/main/package.json',
     description = [[
 https://github.com/pappasam/jedi-language-server
 
--- a/lua/lspconfig/server_configurations/jedi_language_server.lua
+++ b/lua/lspconfig/server_configurations/jedi_language_server.lua
@@ -16,6 +16,7 @@ return {
     single_file_support = true,
   },
   docs = {
+    package_json = 'https://github.com/pappasam/coc-jedi/blob/main/package.json',
     description = [[
 https://github.com/pappasam/jedi-language-server
 


tamago324:
@mjlbach
By merging these PRs, you will be able to configure jedi-language-server using nlsp-settings.nvim.
Also, wouldn't it be a good idea to add package_json for other servers as well?
If there is a VSCode package.json, I would like to create a path to it, if not, I would like to create a PR to add Coc's extension package.json like stylelintplus, is that ok?

lua/lspconfig/server_configurations/jedi_language_server.lua
```diff
@@ -16,6 +16,7 @@ return {
     single_file_support = true,
   },
   docs = {
+    package_json = 'https://github.com/pappasam/coc-jedi/blob/main/package.json',
```
williamboman, line 19: This seems to be the web URL, here's the raw JSON one:

```suggestion
    package_json = 'https://raw.githubusercontent.com/pappasam/coc-jedi/main/package.json',
```

mjlbach:
I don't think we should add dependencies on 3rd party json.

ainsleymcgrath:
> I don't think we should add dependencies on 3rd party json (this is from coc-jedi, not jedi-language server. I do realize they are maintained by the same person.)

@mjlbach Could you expand on what you mean by "third party?" 

I was also wary about reaching for a different repo to accomplish this.

When I when I looked at the [Elixir config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/elixirls.lua) as an example, though, I saw that the `package.json` was from a different repo than the language server itself. The two repos being [`elixir-lsp/vscode-elixir-ls`](https://github.com/elixir-lsp/vscode-elixir-ls) and [`elixir-lsp/elixir-ls`](https://github.com/elixir-lsp/elixir-ls), respectively. So I figured if that was OK, then this may be as well.

ainsleymcgrath:
Bump on this.

I've seen a large number of similar PRs pop up. (#1612, #1613, #1614 to name a few). I'm still curious about @mjlbach's concerns and about other maintainers' (@williamboman @tamago324) feelings on them.

The disclaimer in the README already states that these configurations might not be 100% airtight perfection.
> Language server configurations are provided on a best-effort basis and are community-maintained.

`jedi` in particular is already pretty niche. I think I'm probably part of a very small (but devoted) user base 😅 I would imagine I'm with the majority of neovimmers in that I'm not super bothered by stuff being mildly hacked together at points.

mjlbach:
I haven't decided what to do on this, I'm leaning 75% towards removing all package.json

ainsleymcgrath:
Could you share your reasoning on that?

If I'm still understanding these mechanisms correctly, that change would pretty much completely kneecap `nlsp-settings.nvim` and its dependent, LunarVim--which is growing substantially in popularity. They only very recently switched to `nlsp-settings.nvim` too...

mjlbach:
They are intermittently accurate at best, if nlsp-settings depends on them they can vendor their own package.json or provide custom schemas.

ainsleymcgrath:
Fair enough. This does look more and more like a concern for the downstream stuff to deal with.

Would removing package.json hurt any of the functionality here? I thought the [configuration](https://github.com/pappasam/coc-jedi/blob/main/package.json#L46-L49) in those files was a baseline requirement to make a language server work at all. (I am relatively new to this stuff fwiw.)

As for offloading responsibility to `nlsp-settings`, I would be happy to brainstorm with @tamago324 on how to do that, if they're open to it.

williamboman:
It is a bit of a hack for sure as it reaches into proprietary VSCode config schemas. On the other hand there is a ton of valuable information in these, and so I think making them readily available to users is great! There's other benefits like signaling to users that settings exist, and it also gives a nice cursory overview of the capabilities and maturity of a server.

For these reasons I actually make these accessible in the UI window in nvim-lsp-installer, which I think is one of the first touchpoint between users and LSP servers:

Screenshot 2022-01-08 at 17 02 09

I think it's positive that more people get exposed to this, it might galvanize people to pay more attention to the parts of the LSP ecosystem that still is very VSCode oriented. Who knows, it might inspire someone to push for editor-agnostic configuration schemas

mjlbach:
> It is a bit of a hack for sure as it reaches into proprietary VSCode config schemas.

It's not only proprietary, it includes a mix of editor specific settings and available server settings. I think users would generally be better served reading the upstream documentation.

> For these reasons I actually make these accessible in the UI window in nvim-lsp-installer, which I think is one of the first touchpoint between users and LSP servers:

You shouldn't rely on these being available.

ainsleymcgrath:
@williamboman's last point is very compelling still.

> I think it's positive that more people get exposed to this, it might galvanize people to pay more attention to the parts of the LSP ecosystem that still is very VSCode oriented. Who knows, **it might inspire someone to push for editor-agnostic configuration schemas**

We're still in the early dawn of Neovim's LSP support. As a  temporary status-quo, I think it does more good than harm to give users an out-of-the-box-_ish_ experience rather than swinging back into a very handrolled direction. I'd rather accept some bugs and troubleshooting than return to the much more manual write-a-lot-of-json experience of CoC and its peers.

Better schemas is almost definitely the Right Way, but (afaik) that's simply not what the landscape looks like right now.

It'd be a substantial pain for active users to have package.json deprecated on them. At the very least, I would advocate for a migration plan to the nascent go-read-upstream-docs paradigm.

mjlbach:
> Who knows, it might inspire someone to push for editor-agnostic configuration schemas

Microsoft is not going to be interested in this. They are the only someone that matters. Nothing about the LSP has ever indicated microsoft has any interest in doing anything that does not directly benefit vscode.

tamago324:
If you want to remove package_json, you can define packag.json for each language server in nlsp-settings.nvim side and generate the schema.
Alternatively, since it is not possible for me to maintain it alone, I will consider creating a repository for each language server and creating a JSON schema for each of them.
If VSCode-specific settings are included, some adjustments will be required, but I think it is possible to create JSON schemas for most settings.

mjlbach:
I've decided (tentatively) to remove package.json from lspconfig. See: https://github.com/neovim/nvim-lspconfig/pull/1656

I'll leave these for a couple weeks so plugins can adapt, either by vendoring their own package.json or preferably by converting downstream server documentation/options into some editor agnostic representation.

PR #1596, opened 2021-12-23
docs: add Chinese README

geometryolife:


justinmk:
Thank you. But, we autogenerate much of the documentation. So this won't be maintainable.

PR #1606, opened 2021-12-24
feat(remark): add remark-language-server

remcohaszing:
This registers the new [remark-language-server](https://github.com/remarkjs/remark-language-server) for linting and formatting markdown files.

/cc @wooorm @ChristianMurphy @Murderlon

### DIFF
--- a/lua/lspconfig/server_configurations/remark_ls.lua
+++ b/lua/lspconfig/server_configurations/remark_ls.lua
@@ -1,0 +1,50 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'remark-language-server'
+local cmd = { bin_name, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
+end
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'markdown' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
+https://github.com/remarkjs/remark-language-server
+
+`remark-language-server` can be installed via `npm`:
+```sh
+npm install -g remark-language-server
+```
+
+`remark-language-server` uses the same
+[configuration files](https://github.com/remarkjs/remark/tree/main/packages/remark-cli#example-config-files-json-yaml-js)
+as `remark-cli`.
+
+This uses a plugin based system. Each plugin needs to be installed locally using `npm` or `yarn`.
+
+For example, given the following `.remarkrc.json`:
+
+```json
+{
+  "presets": [
+    "remark-preset-lint-recommended"
+  ]
+}
+```
+
+`remark-preset-lint-recommended` needs to be installed in the local project:
+
+```sh
+npm install remark-preset-lint-recommended
+```
+
+]],
+  },
+}


ChristianMurphy (approved):

Murderlon:
Thanks for taking the effort in adding this! I will definitely start using this in my neovim setup.

I think [`server_configurations.md`](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md) should have a new `markdown` link at the top and a similar section like the rest on how to get started.

mjlbach:
server_configurations.md is autogenerated

lua/lspconfig/server_configurations/remark_language_server.lua
```diff
@@ -0,0 +1,51 @@
...
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'markdown' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/remarkjs/remark-language-server/main/package.json',
```
mjlbach, line 18: This package.json doesn't contain any relevant information. These are  only used for generating`settings` and `initOptions`

mjlbach:
Also remark_language_server.lua -> remark_ls.lua

lua/lspconfig/server_configurations/remark_ls.lua
```diff
@@ -0,0 +1,50 @@
...
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
+end
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'markdown' },
+    root_dir = util.find_git_ancestor,
```
williamboman, line 14: I wonder if this root_dir should be changed to something like

```lua
root_dir = util.root_pattern(".remarkrc", ".remarkrc.js", "package.json")
```

Preferably it would also peak into `package.json` to ensure that the `.remarkConfig` field exist.

Just installing remark-language-server and opening a markdown file I'm getting these errors:
Screenshot 2022-01-07 at 13 02 10

wooorm: we’re working on a cleaner error when `remark` isn’t installed. But I do think there should be an error?

williamboman: Yeah for sure, if those packages are required to be installed locally to run the server! I'm not sure if there's a convention for reporting these "fatal" errors - usually servers just terminate themselves with an error code. Maybe `$/progress` could be used to report this.

But to expand on my first comment: I think the current trigger mechanism is a bit too broad. Ideally lspconfig should only attempt to start the remark server if it can, with a high degree of certainty, ensure that the server is able to start.  Right now it only checks if the current file is a markdown file, and if it's inside a git repository.

I'm not familiar with remark myself, but from reading the README it seems like the following is a solid set of heuristics for determining this:

- a `.remarkrc` or `.remarkrc.js` file exists in any parent directory
- a `package.json` file exists in any parent directory (and this `.json` file has a `remarkConfig` field)

wooorm: Ah, right, so I understand that any `nvim-lspconfig` user opening markdown now sees this error?

My goal is to add an option to the language server to, configureably, do nothing if `remark` cannot be found. That’s one way to hide the error

williamboman: Only those who:
1) explicitly opt in to using remark_ls (this is done by adding `require("lspconfig").remark_ls.setup({})` in your init files)
2) are editing a markdown file in a non-remark project

I'm also currently looking into [adding remark_ls to nvim-lsp-installer](https://github.com/williamboman/nvim-lsp-installer/pull/404), where users can explore and install servers from within neovim. Remark will be a candidate for the `markdown` language, meaning remark will show up as a candidate when doing `:LspInstall markdown`. Those not familiar with remark will probably be very confused if they try it out (I was).

I think a good start would be to change the root pattern to what I mentioned above, as I think it'd greatly reduce the amount of false positives. I'll open a PR soon!

williamboman: #1641

PR #1611, opened 2021-12-26
fix(denols): add package_json

tamago324:

Please check it.

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -137,6 +137,7 @@ return {
     },
   },
   docs = {
+    package_json = 'https://raw.githubusercontent.com/denoland/vscode_deno/main/package.json',
     description = [[
 https://github.com/denoland/deno
 


mjlbach:
Are you validating these? Most of this looks like options for vscode, not for sending to deno's language server.

tamago324:
I'm sorry. I haven't verified it.
We have not tested other language servers, so we will do so and create the PR again.

mjlbach:
It's fine, just test the already open PRs and let me know if the settings are relevant, and I will merge if that is the case.

tamago324:
Yes, sir.
It may take some time, but I'll verify and get back to you.
Thank you.

tamago324:
I don't know the details of the Rust code, but according to the [deno server code](https://github.com/denoland/deno/blob/0888ba7a8db33ee01f57679308e7e8c4eea108e9/cli/lsp/config.rs#L148), the vscode_deno setting seems to be correct.
Here is the result.
Also, we have not checked all the settings, but we have confirmed that some of them work correctly.

> It is assumed that `YES` is the correct setting value and `NO` is the setting value for VSCode only.

```
YES "deno.cache",
YES "deno.codeLens.implementations",
YES "deno.codeLens.references",
YES "deno.codeLens.referencesAllFunctions",
YES "deno.codeLens.test",
NO "deno.codeLens.testArgs",
YES "deno.config",
YES "deno.enable",
YES "deno.importMap",
YES "deno.internalDebug",
YES "deno.lint",
NO "deno.path",
YES "deno.suggest.autoImports",
YES "deno.suggest.completeFunctionCalls",
YES "deno.suggest.imports.autoDiscover",
YES "deno.suggest.imports.hosts",
YES "deno.suggest.names",
YES "deno.suggest.paths",
YES "deno.unstable"
```

tamago324:
Thank you for merging!

PR #1614, opened 2021-12-26
fix(diagnosticls): add package_json

tamago324:

Please check it.

### DIFF
--- a/lua/lspconfig/server_configurations/diagnosticls.lua
+++ b/lua/lspconfig/server_configurations/diagnosticls.lua
@@ -15,7 +15,8 @@ return {
     filetypes = {},
   },
   docs = {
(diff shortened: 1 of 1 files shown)


tamago324:
@mjlbach 
I did some research on https://github.com/iamcco/diagnostic-languageserver.

Since diagnostic-languageserver is configured using `initializationOptions`, we decided that it was inappropriate to specify `package_json`.
However, I think it is necessary to specify the reason for not setting `package_json` in the code, so I put it in the comments.
If you like, please merge it.

Also, [coc-diagnostic](https://github.com/iamcco/coc-diagnostic) converts the settings in coc-settings.json to `initializationOptions`, so it seems to reflect the settings correctly. ([This code](https://github.com/iamcco/coc-diagnostic/blob/fed2eb53460a610a2b6562dc3c5715466a71d69f/src/index.ts#L70-L81))

PR #1615, opened 2021-12-26
fix(eslint): add package_json

tamago324:

Please check it.

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -148,6 +148,7 @@ return {
     },
   },
   docs = {
+    package_json = 'https://raw.githubusercontent.com/microsoft/vscode-eslint/main/package.json',
     description = [[
 https://github.com/hrsh7th/vscode-langservers-extracted
 


tamago324:
@mjlbach 
I did some research on [eslint](https://github.com/microsoft/vscode-eslint/tree/main/server).

In vscode-eslint, there were many differences in the configuration values from the server.

Those that are in the package.json configuration and are also in the server definition:

```
"eslint.codeAction.disableRuleComment"
"eslint.codeAction.showDocumentation"
"eslint.codeActionsOnSave.mode"
"eslint.codeActionsOnSave.rules"
"eslint.nodePath"
"eslint.onIgnoredFiles"
"eslint.options"
"eslint.packageManager"
"eslint.quiet"
"eslint.run"
"eslint.useESLintClass"
"eslint.validate"
```

Not in the package.json setting, but in the server definition (Value not set in VSCode?):

```
"eslint.format"
"eslint.rulesCustomizations"
"eslint.workspaceFolder"
```

A value that is in the package.json setting but not in the server definition (VSCode's own setting?):

```
"eslint.alwaysShowStatus"
"eslint.autoFixOnSave"
"eslint.debug"
"eslint.enable"
"eslint.execArgv"
"eslint.format.enable"
"eslint.lintTask.enable"
"eslint.lintTask.options"
"eslint.migration.2_x"
"eslint.nodeEnv"
"eslint.probe"
"eslint.provideLintTask"
"eslint.rules.customizations"
"eslint.runtime"
"eslint.trace.server"
"eslint.workingDirectories"
```

mjlbach:
The more you look into these the more I am tempted to just remove all package.json :/ I didn't add the feature, and I think that as the coc plugins show it's better to just curate your own settings list from the server documentation.

tamago324:
Right, it's probably not a good idea to generate documentation based on the configuration list in package.json.
However, it would be very hard to create all the lists manually.
I can't think of a good idea.
Or maybe a script that adds what is missing and removes what is not needed based on the package.json?
This would also be a lot of work, though.

PR #1619, opened 2021-12-26
feat: add gotmpl to gopls

brunnre8:
gopls introduced support for templates:
https://github.com/golang/tools/blob/master/gopls/doc/features.md#template-files

They look for 'tmpl' and 'gotmpl' extensions.
tmpl may be a tad too unspecific but adding gotmpl shouldn't hurt.

### DIFF
--- a/lua/lspconfig/server_configurations/gopls.lua
+++ b/lua/lspconfig/server_configurations/gopls.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'gopls' },
-    filetypes = { 'go', 'gomod' },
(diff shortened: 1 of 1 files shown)


brunnre8:
Why do you care if I send a PR from the master in my fork?
My fork quite literally only exists for that single PR, renaming the master branch to something else had no impact.

mjlbach:
1. It messes up our docgen pipeline which runs against master
2. It means I can push to a user's feature branch if the user doesn't respond to my review in a timely manner
3. It's in our contributing guide, so it's a good filter if someone didn't read our contributing guide carefully

mjlbach:
Hmm maybe it is only on the core contributing guidelines, I should make it more explicit those apply to lspconfig as well.

PR #1624, opened 2021-12-30
fix(nimls): remove erroneous package_json entry

williamboman:
They aren't really the same server, so I really don't believe these settings are valid. Also, https://github.com/williamboman/nvim-lsp-installer/issues/326#issuecomment-996065450 seems to confirm this.

- https://github.com/PMunch/nimlsp (this server)
- https://github.com/pragmagic/vscode-nim

### DIFF
--- a/lua/lspconfig/server_configurations/nimls.lua
+++ b/lua/lspconfig/server_configurations/nimls.lua
@@ -10,7 +10,6 @@ return {
     single_file_support = true,
   },
   docs = {
-    package_json = 'https://raw.githubusercontent.com/pragmagic/vscode-nim/master/package.json',
     description = [[
 https://github.com/PMunch/nimlsp
 `nimlsp` can be installed via the `nimble` package manager:


williamboman:
I've not investigated this further than what's stated above, so I could be wrong.

mjlbach:
I'm seriously contemplating just removing all of these, I'm not sure they add much value, the settings are often incomplete, out-of-date, or include a mix of vscode specific settings.

williamboman:
> I'm seriously contemplating just removing all of these, I'm not sure they add much value, the settings are often incomplete, out-of-date, or include a mix of vscode specific settings.

Yeah I've noticed the same. I definitely think there's a net positive value in keeping these though. If editor-specific settings find their way into these, it's a sign of a broader failure in the LSP initiative, imo. Although to be fair I don't believe this is part of the spec at all, and we are actually reaching into proprietary VSCode mechanisms. Do you know if there are any ongoing discussions about standardizing the settings workflow of LSPs?

Btw I recently implement functionality similar to what exists in lspconfig's `server_configurations.md` within neovim in nvim-lsp-installer: https://github.com/williamboman/nvim-lsp-installer/pull/363

williamboman:
https://en.wikipedia.org/wiki/Embrace,_extend,_and_extinguish

mjlbach:
> If editor-specific settings find their way into these, it's a sign of a broader failure in the LSP initiative, imo. Although to be fair I don't believe this is part of the spec at all, and we are actually reaching into proprietary VSCode mechanisms.

I think you responded to yourself the same way I would have. `package.json` is not a standardized mechanism (and autogenerating docs with it was not my idea and predates me. I'm leaning towards removing them.

> Btw I recently implement functionality similar to what exists in lspconfig's server_configurations.md within neovim in nvim-lsp-installer: williamboman/nvim-lsp-installer#363

This seems like feature creep to me tbh.

PR #1626, opened 2022-01-02
refactor(tests): simpler temp_dir detection

kylo252:
Tested on Windows and Linux (Arch).

### DIFF
--- a/test/minimal_init.lua
+++ b/test/minimal_init.lua
@@ -8,12 +8,7 @@ end
 
 vim.cmd [[set runtimepath=$VIMRUNTIME]]
 
-local temp_dir
-if on_windows then
-  temp_dir = vim.loop.os_getenv 'TEMP'
-else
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
This is not just a simplification but also changing the semantics of the code as it will now also utilize a `TEMP` environment variable on unix if available. Is that intended?

kylo252:
> This is not just a simplification but also changing the semantics of the code as it will now also utilize a `TEMP` environment variable on unix if available. Is that intended?

I don't think that this variable is often used on unix, but even if it is, then it should be pointing to a `tmpfs` mount anyway. at least according to the search results I've found so far 😅

mjlbach:
I'm fine with this

kylo252:
FWIW, I just figured I can restore some minimalism after https://github.com/neovim/nvim-lspconfig/pull/1354

PR #1627, opened 2022-01-02
feat: add sonarlint

hungpt2702:


mjlbach:
I'm going to need more explanation as to why this PR needs to change how we set up commands

hungpt2702:
@mjlbach
Currently, `create_module_commands` function expects the command to be defined as a table. The new command I defined for the server config (`SonarLint.DeactivateRule`) is a function and therefore caused an error. (This command is required to be implemented to handle deactivate rule code action). However, if I redefine the command as a table like `create_module_commands` expects it will also cause an error because the way `neovim` handles commands in [`apply_action`](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/buf.lua): it expects the command to be a function rather than a table. And I think `nvim-lspconfig` should not create module command from commands that are supposed to handle code actions rather than for user to run directly (like `SonarLint.DeactivateRule` command); therefore, I changed `create_module_commands` so that it only creates module command from commands defined as table (i.e. commands that are intended to be run directly by user).

mjlbach:
https://github.com/neovim/nvim-lspconfig/pull/1838 should resolve the issue with user_commands when it is merged.

PR #1630, opened 2022-01-03
refactor(clangd): Remove obsolete --background-index flag

kirillbobyrev:
Clangd background index is enabled by default for a very long while.
In the vast majority of modern versions, the flag is redundant, hence
I think it is better to remove it for clarity.

### DIFF
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -38,7 +38,7 @@ local default_capabilities = vim.tbl_deep_extend(
 
 return {
   default_config = {
-    cmd = { 'clangd', '--background-index' },
+    cmd = { 'clangd' },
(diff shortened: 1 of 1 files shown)


kirillbobyrev:
+cc @sam-mccall @kadircet

mjlbach:
Thanks! My rule of thumb is the version available in the last debian/ubuntu LTS release. Can you confirm this is the case?

sam-mccall:
Yes.
[default clangd package in ubuntu 20.04 focal is clangd-10](https://packages.ubuntu.com/search?keywords=clangd&searchon=names&suite=focal&section=all)
[debian 9 stretch has clangd-11](https://packages.debian.org/stretch/clangd-11)
[--background index is default since clangd-9](https://github.com/llvm/llvm-project/commit/c6fdefa0ee81b7c9c16126cfe27fcd4c963ffd96)

PR #1632, opened 2022-01-04
Update eslint.lua

b-m-f:
Removes `<cmd>` and `<CR>` from autocmd invocation for automatic linting on save.
While not being needed they also break the invocation and must be removed.

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -160,7 +160,7 @@ npm i -g vscode-langservers-extracted
 
 vscode-eslint-language-server provides an EslintFixAll command that can be used to format document on save
 ```vim
-autocmd BufWritePre <buffer> <cmd>EslintFixAll<CR>
+autocmd BufWritePre <buffer> EslintFixAll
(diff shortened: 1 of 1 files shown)


mjlbach:
The semantic of `:help :map-cmd` are different than without. cc @entropitor

lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -160,7 +160,7 @@ npm i -g vscode-langservers-extracted
 
 vscode-eslint-language-server provides an EslintFixAll command that can be used to format document on save
 ```vim
-autocmd BufWritePre <buffer> <cmd>EslintFixAll<CR>
+autocmd BufWritePre <buffer> EslintFixAll
```
rxt30, line 163: I think it would be better to add `silent!`, so that the autocmd would fail silently if the ESLint-Server is not running and therefore `EslintFixAll` is not aviable.

Line would look like this
```
autocmd BufWritePre <buffer> silent! EslintFixAll
```

justinmk:
looks like this was fixed

PR #1634, opened 2022-01-04
feat(leanls): add root dir as command-line arg

gebner:
The argument is ignored by the server, but it is useful because it is shown by `htop`, `ps ax`, etc., so you can figure out which Lean servers are running.

The lean3ls doesn't require a corresponding change: we use the `lean-language-server` wrapper, which appends the root directory itself to the wrapped `lean --server` command.


mjlbach:
You all are in charge of your server but the PID is also available in LspInfo :)

PR #1641, opened 2022-01-07
fix(remark_ls): change root_pattern to remark-specific project files

williamboman:
From discussion here: https://github.com/neovim/nvim-lspconfig/pull/1606#discussion_r780224035

### DIFF
--- a/lua/lspconfig/server_configurations/remark_ls.lua
+++ b/lua/lspconfig/server_configurations/remark_ls.lua
@@ -11,8 +11,16 @@ return {
   default_config = {
     cmd = cmd,
     filetypes = { 'markdown' },
-    root_dir = util.find_git_ancestor,
-    single_file_support = true,
+    root_dir = util.root_pattern(
+      '.remarkc',
+      '.remarkc.json',
+      '.remarkc.cjs',
+      '.remarkc.mjs',
+      '.remarkc.js',
+      '.remarkc.yaml',
+      '.remarkc.yml',
+      'package.json'
+    ),
   },
   docs = {
     description = [[


williamboman:
I found the exhaustive list of project files here: https://github.com/remarkjs/remark/blob/87c50a5884662908cfef15297f109de0102a2c4a/packages/remark-cli/readme.md#example-config-files-json-yaml-js

lua/lspconfig/server_configurations/remark_ls.lua
```diff
@@ -11,7 +11,16 @@ return {
...
-    root_dir = util.find_git_ancestor,
+    root_dir = util.root_pattern(
+      '.remarkc',
+      '.remarkc.json',
+      '.remarkc.cjs',
+      '.remarkc.mjs',
+      '.remarkc.js',
+      '.remarkc.yaml',
+      '.remarkc.yml',
+      'package.json'
```
williamboman, line 22: This should really be `package.json` AND which has a field `remarkConfig`. I don't believe there's a util function for this currently?

I wonder if it'd make sense to make `root_pattern` accept a table where the first index is the filename, and then the rest are keys with extra conditions, like:

```suggestion
      { 'package.json', withKey = { "remarkConfig" } }
```

or (if this were to even be added) it'd deserve its own util function

remcohaszing: This would benefit other tools as well (at least ESLint and Stylelint).

mjlbach:
Can you actually not open remark on a generic markdown file?

williamboman:
> Can you actually not open remark on a generic markdown file?

This might've [changed recently](https://github.com/remarkjs/remark-language-server/pull/3#issue-1091295163). It seems like the server bundles these necessary packages now, will recheck

williamboman:
It seems like the minimum requirement now is that you need to at least have the `remark` npm package installed in the workspace.

remcohaszing:
Yes, `remark-language-server` now tries to import the locally installed `remark` package from `node_modules`. If this fails, it shows a more friendly message.

The pull request that changes this is remarkjs/remark-language-server#5, although really this just implements the new behaviour introduced in https://github.com/unifiedjs/unified-language-server/releases/tag/2.0.0

williamboman:
@remcohaszing Do you think it'd make sense to be able to configure where it should do the node module resolution when locating the `remark` package? The angular LSP allows you to configure which directories it should "probe" when locating the `tsc` compiler and angular language service dependencies. This'd make it possible to bundle the language server with an installation of remark which can be used as a last resort, should the default resolution fail to find a local installation.

glepnir:
What is the progress of this PR?

williamboman:
So the short summary is: The remark language server can only run in locations where it can (via Node module resolution) resolve a locally installed `remark` Node package. Using `.git` as well as `single_file_support = true` is not particularly compatible with this. You can configure remark either through config files (listed in this PR as root dir patterns) or as an entry in a `package.json`, so this PR changes the root dir pattern to those.

williamboman:
> or as an entry in a `package.json`

To add to this; This is quite commonly supported among Node.js tooling btw. Eslint does this as well. Preferably lspconfig would be able to [peek into this JSON file and look for a specific key path](https://github.com/neovim/nvim-lspconfig/pull/1641#discussion_r780245734), but this is not possible today (so `package.json` might trigger false positives).

PR #1643, opened 2022-01-07
Add configuration for "OpenFOAM language server"

FoamScience:
[OpenFOAM language server](https://github.com/FoamScience/foam-language-server) finally reached a decent (usable) state;
Hence, the PR to add its configuration to `lspconfig`.


lua/lspconfig/server_configurations/foam.lua
```diff
@@ -0,0 +1,26 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    -- `foam-ls` can be a shell script executing:
+    -- node /path/to/foam-language-server/lib/foam-ls.js --stdio
+    cmd = { 'foam-ls' },
```
williamboman, line 7: Wouldn't it be better if the npm package provided an executable itself? That way when someone installs it you'll be able to have a common, canonical, name for the executable that everyone can use.

lua/lspconfig/server_configurations/foam.lua
```diff
@@ -0,0 +1,26 @@
...
+    filetypes = { 'foam', 'OpenFOAM' },
+    root_dir = util.root_pattern('system/controlDict'),
```
williamboman, line 9: I don't believe this is a valid pattern. I guess you want to check for the existence of a `controlDict` dir/file inside a `system` directory? The current root dir detection implementation can only traverse the file system in one direction (upwards)

lua/lspconfig/server_configurations/foam.lua
```diff
@@ -0,0 +1,26 @@
...
+    log_level = vim.lsp.protocol.MessageType.Warning,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/FoamScience/foam-language-server/master/package.json',
+    description = [[
+https://github.com/FoamScience/foam-language-server
+
+`foam-language-server` can be installed via `npm`
+```sh
+npm install foam-language-server
```
williamboman, line 19: As long as the npm package provides an executable, I think it'd make sense to suggest installing it globally, making sure it ends up in the user's PATH.

```suggestion
npm install -g foam-language-server
```

lua/lspconfig/server_configurations/foam.lua
```diff
@@ -0,0 +1,26 @@
...
+  default_config = {
+    -- `foam-ls` can be a shell script executing:
+    -- node /path/to/foam-language-server/lib/foam-ls.js --stdio
+    cmd = { 'foam-ls' },
+    filetypes = { 'foam', 'OpenFOAM' },
+    root_dir = util.root_pattern('system/controlDict'),
+    log_level = vim.lsp.protocol.MessageType.Warning,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/FoamScience/foam-language-server/master/package.json',
```
williamboman, line 13: This `package_json` is currently only used to pull the server's configuration schema when generating the `server_configurations.md` docs. It seems like this package.json does not provide any configuration schema, so I think this should be removed.

```suggestion

```

lua/lspconfig/server_configurations/foam.lua
FoamScience: Yes it sure provides one; [package.json#L24](https://github.com/FoamScience/foam-language-server/blob/fe6aa2e5cf9c37c15fc60d0b1164958397eff15f/package.json#L24)
Should I remove the comment?

lua/lspconfig/server_configurations/foam.lua
FoamScience: Yes, I want to check a `controlDict` file inside a `system` dir. If this is not possible, how about checking for just `system`? Because somehow that works in my local setup

williamboman: You could also provide your own matcher function, something like this:

```lua
root_dir = function (fname)
    return util.search_ancestors(fname, function (path)
        if util.path.exists(util.path.join(path, "system", "controlDict")) then
           return path
        end
    end)
end
```

lua/lspconfig/server_configurations/foam.lua
williamboman: For npm packages, the following will also be needed (you can refer to other npm-based configs such as `tsserver.lua`):

```lua
local bin_name = 'foam-ls'
local cmd = { 'foam-ls' }

  cmd = { 'cmd.exe', '/C', bin_name }
end
```

lua/lspconfig/server_configurations/foam.lua
FoamScience: Thanks! will do that

FoamScience:
Is there anything left to do here?

mjlbach:
foam.lua -> foam_ls.lua

lua/lspconfig/server_configurations/foam.lua
```diff
@@ -0,0 +1,33 @@
...
+		cmd = cmd,
+		filetypes = { "foam", "OpenFOAM" },
+		root_dir = function(fname)
+			return util.search_ancestors(fname, function(path)
+				if util.path.exists(util.path.join(path, "system", "controlDict")) then
+					return path
+				end
+			end)
+		end,
+		log_level = vim.lsp.protocol.MessageType.Warning,
```
mjlbach, line 20: Why do you need to set a default log level?

FoamScience: An artifact from early developing phases; Thanks for catching that!

FoamScience:
> foam.lua -> foam_ls.lua

Renaming the file won't have unexpected effects, will it? Like, that changes the name of server; and that's it?

mjlbach:
All renaming the file does is change what users use for the setup call. Typically for servers that are X-language-server, the shortening of that is X_ls in lspconfig. So require('lspconfig').foam_ls.setup{}

FoamScience:
Cool, thanks! Let me educate myself on lint-commits a little bit and reword the failing ones

mjlbach:
It's ok, I will squash and merge.

lua/lspconfig/server_configurations/foam_ls.lua
```diff
@@ -0,0 +1,32 @@
...
+    root_dir = function(fname)
+      return util.search_ancestors(fname, function(path)
+        if util.path.exists(util.path.join(path, 'system', 'controlDict')) then
+          return path
+        end
+      end)
+    end,
+  },
+  docs = {
+    package_json = 'https://raw.githubusercontent.com/FoamScience/foam-language-server/master/package.json',
```
williamboman, line 22: I think this will cause some troubles still

FoamScience: I thought we took it out yesterday 😭 ; I have:
```
"contributes": {
    "configuration": [
    ]
  }
```
so it should pass CI (like dartls does); but do you prefer another PR to remove it?

williamboman: I have a feeling these will be removed from lspconfig soon anyway.  Setting it to an array will at least make nvim-lsp-installer gracefully skip it :+1:. There are some other projects that rely on these containing a valid conf schema, I'm not sure how they'll behave (it'll prob all be fine)

PR #1645, opened 2022-01-08
refactor(info): don't show non-managed clients

kylo252:
**breaking change**

`:LspInfo` will no longer display clients that are externally managed


kylo252:
@mjlbach, I'm a bit conflicted about this, but I generally think it is for the best, since it could be misleading to users that might expect other lspconfig commands to interact with those clients. 

The only non-managed client that I personally use is null-ls, which has its own `NullLsInfo`, so I don't think that @jose-elias-alvarez  would mind. And this info isn't particularly helpful

image

mjlbach:
This is a useful debugging tool for servers launched by calling start_client directly.

PR #1646, opened 2022-01-08
Don't throw an error if config.filetypes is nil

khalidchawtany:
Without this neovim would show an error after each PackerCompile


mjlbach:
I don't get errors on PackerCompile... config.filetypes is explicitly *not* required (servers will attach to all buffers if missing.

khalidchawtany:
This is what I get:
```
packer.compile: Complete
[nvim-lsp-installer] ...pack/packer/opt/nvim-lspconfig/lua/lspconfig/configs.lua:258: bad argument #1 to 'ipairs' (table expected, got nil)

```

mjlbach:
I don't support nvim-lsp-installer, you can file a bug with them.

khalidchawtany:
Makes sense thanks.

PR #1649, opened 2022-01-10
fix(ccls): offset_encoding = utf-32

black-desk:
ccls use codepoint as offset_encoding

check this https://github.com/MaskRay/ccls/commit/66e9cbd9a6fbbc239af1c1014a11bea4ed25ce33

### DIFF
--- a/lua/lspconfig/server_configurations/ccls.lua
+++ b/lua/lspconfig/server_configurations/ccls.lua
@@ -5,6 +5,7 @@ return {
     cmd = { 'ccls' },
     filetypes = { 'c', 'cpp', 'objc', 'objcpp' },
     root_dir = util.root_pattern('compile_commands.json', '.ccls', '.git'),
+    offset_encoding = 'utf-32',
     -- ccls does not support sending a null root directory
     single_file_support = false,
   },


mjlbach:
That commit message explicitly says: "indexer: change Pos computation from byte offset to UTF-8 encoded codePoint". Why are you setting it to utf-32?

https://github.com/MaskRay/ccls/issues/267#issuecomment-464324011

black-desk:
You can have a try with https://github.com/neovim/neovim/pull/17038
that pr and this one actually make ccls behave as expected.

```cpp
int main() {
  /*𐐀*/ int a;
  int b = a + 1;
  return 0;
}
```
Try to jump to the definition of `a`.

---
"code point" basically means a "character".
You can check the definition of code point on [wikipedia](https://en.wikipedia.org/wiki/Code_point)

As far as I know, all Unicode code points can be encoded to a UTF-32 unit, so UTF-32 just means code point here.

clangd's ["Protocol extensions"](https://clangd.llvm.org/extensions.html#utf-8-offsets) also says:

> utf-32: character counts codepoints

mjlbach:
"indexer: change Pos computation from byte offset to **UTF-8 encoded codePoint**".

black-desk:
Can you just have a try? This actually works.

black-desk:
I have asked @MaskRay about this encoding issue. He confirms that ccls does use utf-32 code point as `Position.Character` counting method.

He said you can check:
https://github.com/MaskRay/ccls/blob/c6686be382da46baed4245c7963f07964575e90c/src/clang_tu.cc#L48
and
https://github.com/MaskRay/ccls/blob/31fb0756504ea5f1e95f32f25d29f748c18cb84c/src/working_files.cc#L435
for details.

mjlbach:
If this is actually the case:

1. It should be documented somewhere (note, utf-32 is not present in a single docstring in the ccls repository)
2. It doesn't comply with the language server protocol which states servers must implement utf-16 support, even clangd does this despite having a utf-8 mode
3. There is no offsetEncoding returned in response to the initialization like clangd/rust-analyzer to indicate which encodings ccls supports.

I'm leaning towards removing ccls from lspconfig.

black-desk:
> It doesn't comply with the language server protocol which states servers must implement utf-16 support, even clangd does this despite having a utf-8 mode

As far as I know, this is true. I will open an issue about that.

> There is no offsetEncoding returned in response to the initialization like clangd/rust-analyzer to indicate which encodings ccls supports.

ccls does not implement clangd's LSP extensions.

MaskRay:
The documentation about `offsetEncoding` is unclear. ccls expects that every input file is UTF-8. It expects that `line/character` are in code units. How to negotiate this to the client?

I am unsure code units are "entirely different". Many editors have APIs returning one line as a Unicode string. The string is typically indexed by code points.

MaskRay:
Does https://github.com/MaskRay/ccls/pull/859 work? `offsetEncoding: "utf-32"`

mjlbach:
I'm not sure what you mean by a generic code unit, the concrete numerical representation of a code unit depends on the encoding form, of which there are 3 as defined by the unicode standard (UTF-8, UTF-16, UTF-32). See [definition D77 in Section 3.9, Unicode Encoding Forms](https://www.unicode.org/versions/Unicode14.0.0/ch03.pdf#G7404). 

All that this PR would do is force [Position parameters](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position), to be sent as UTF-32 encoded code-units. Buffers are always represented as utf-8 encodings internally and on-disk.

Currently we default to UTF-16 encoded code units as per the LSP standard, but neovim supports UTF-8, UTF-16, and UTF-32 natively. With https://github.com/MaskRay/ccls/pull/859, neovim will automatically detect the correct code unit encoding from initialization parameters. This is indeed off-spec, but it is the same convention that clangd/rust-analyzer use.

black-desk:
> I believe CoC always sends UTF-16 encoded code units as the position parameters.

Not true. coc also have this issue. It will send position count in code points in some case. check 

https://github.com/neoclide/coc.nvim/commit/bfac5f9416e97b30a1904fa1c3b8ddb8534a29bb#commitcomment-62283500

That's why I switch from coc to neovim's builtin lsp.

mjlbach:
> Not true. coc also have this issue. 

I don't understand, I am *also* saying coc has this issue. Specifically CoC has this issue because it always sends utf-16 encoded code units as opposed to utf-32 code units for ccls

black-desk:
coc has a vimscript function `coc#cursor#position()` (old name `coc#util#cursor()`) which was used to Get cursor position.
 
https://github.com/neoclide/coc.nvim/blob/00ed6677383e56d79cb2812bfe954f424bb15ded/autoload/coc/cursor.vim#L16

You can still find some reference to this function, which return character count in code points such as:

sam-mccall:
@MaskRay 

> The documentation about offsetEncoding is unclear. ccls expects that every input file is UTF-8. It expects that line/character are in code units. How to negotiate this to the client?

MaskRay/ccls#859 is exactly what we intended for the extension, if a server decides to count `character` in codepoints exclusively. :+1: Maybe "codepoints" would have been a better name than "utf-32", but it's not worth trying to change now.

Regarding input files being UTF-8, `offsetEncoding` doesn't have anything to say about that, and you probably have to do anything.
`offsetEncoding` affects operations like `countLSPCharacters(string) -> int`, that operate on unicode strings, not on bytes.
 - for open files, the client is sent over LSP as unicode text (JSON strings) rather than byte buffers, however the client/server stores that internally
 - for non-open files, if the client & server don't agree on how they're encoded, you're going to have a bad time, though not just with positions. I think this is outside LSP's scope though.

What do you think would make the docs clearer here?

(Sorry to jump in late)

PR #1652, opened 2022-01-11
feat: add asm_lsp

williamboman:


lua/lspconfig/server_configurations/asm_lsp.lua
```diff
@@ -0,0 +1,19 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'asm-lsp' },
+    filetypes = { 'asm', 'vmasm' },
```
williamboman, line 6: I'm not entirely sure these are the correct filetypes. The file extensions are `.asm, .s, .S`, and so I tried to map those to the underlying filetypes by checking `filetype.vim`.

PR #1654, opened 2022-01-11
feat: add opencl_ls

williamboman:


williamboman:
Turning this to draft for the time being, seems like the server doesn't handle unimplemented requests very well atm:

Screenshot 2022-01-11 at 19 11 22

PR #1656, opened 2022-01-12
chore: remove all package json

mjlbach:
These are more trouble than they are worth. Users should refer to upstream documentation.


clason:
Would it make sense to replace this with a (user curated) link to the upstream documentation for settings, if it exists?

mjlbach:
That would be my ideal. I would also consider hand-crafted schemas ported from package.json and vendored in lspconfig. I think ultimately server specific plugins should handle validation of settings.

clason:
Sounds good -- as a first step, I'd just put a `settings` key under `docs` that points to any **human**-readable URL (Website, markdown file, json if that is the best) that gets expanded to "for available settings, see <url>" in the generated Markdown.

(Not in this PR, of course.)

williamboman:
I've temporarily moved these to [this gist](https://gist.github.com/williamboman/a01c3ce1884d4b57cc93422e7eae7702) which can be used for the time being (no guarantees).

I'm just wondering, wouldn't it be better to try to push for client-agnostic JSON schemas that only contain LSP settings? I'm sure most server authors would be more than happy to accommodate this if there was a framework in place? Obviously easier said than done and someone has to put in the work 😅.

clason:
> I'm just wondering, wouldn't it be better to try to push for client-agnostic JSON schemas that only contain LSP settings? I'm sure most server authors would be more than happy to accommodate this if there was a framework in place? Obviously easier said than done and someone has to put in the work 😅.

Go ahead and do that, and report back when all 100+ server authors have bought into your proposal.

kylo252:
> I've temporarily moved these to [this gist](https://gist.github.com/williamboman/a01c3ce1884d4b57cc93422e7eae7702) which can be used for the time being (no guarantees).

@williamboman, you don't need to hard-code that value for most of these, you can use github's API, or possibly even `git ls-remote`

```console
$ gh api repos/redhat-developer/vscode-yaml/contents/package.json --jq .download_url
https://raw.githubusercontent.com/redhat-developer/vscode-yaml/main/package.json

$ gh api repos/elixir-lsp/vscode-elixir-ls/contents/package.json --jq .download_url
https://raw.githubusercontent.com/elixir-lsp/vscode-elixir-ls/master/package.json
```

mjlbach:
> I'm just wondering, wouldn't it be better to try to push for client-agnostic JSON schemas that only contain LSP settings? 

What does pushing for amending the protocol have to do with me removing something from lspconfig? Ofc it would be "better" if a way of representing available server settings was standardized, similar to how initialization options are standardized. I'm not personally interested in lobbying microsoft to amend the protocol, because I think (as @clason) alluded to, it's extremely unlikely to materialize and the cost/benefit ratio is low (there are many things I would change in the LSP specification before this).

kylo252:
> > Would it make sense to replace this with a (user curated) link to the upstream documentation for settings, if it exists? Some of them can be quite hidden...

> That would be my ideal. 

That link is literally the `package.json` schema for the most part.

> I would also consider hand-crafted schemas ported from package.json and vendored in lspconfig. I think ultimately server specific plugins should handle validation of settings.

I feel like that is more work than it's worth. It's probably easier to just accept the ones with sane settings. 

--- 

It's enough to remove the autogen and leave the link to the settings as a general "guideline" and leave the validation part for other plugins if they so wish, imho.

mjlbach:
> That link is literally the package.json schema for the most part.

No, it's not. Most servers have documentation.

> I feel like that is more work than it's worth. It's probably easier to just accept the ones with sane settings.

Disagree, the future of lsp is server specific plugins. If these want to integrate schema and provide validation on settings, that's fine.

williamboman:
> @williamboman, you don't need to hard-code that value for most of these, you can use github's API, or possibly even `git ls-remote`

Hm, but then a list of repos with these capabilities would have to be kept instead + there's now an intermediary dependency of having to go through a gh client to fetch it. I solved it pretty easily like [this](https://github.com/williamboman/nvim-lsp-installer/blob/main/scripts/autogen_metadata.lua#L150-L152) and [this](https://github.com/williamboman/nvim-lsp-installer/blob/main/scripts/autogen_metadata.lua#L160-L163).

> What does pushing for amending the protocol have to do with me removing something from lspconfig? We don't get "leverage" by keeping package.json in the repo.

Oh it was just a reply to the conversation in this thread, where there seemed to be a consensus that this is best delegated to either individual contributors to provide auxiliary LSP plugins, or repackage them in lspconfig. I think it makes sense to remove it from lspconfig in their current shape :+1:. Or well, I'm 50/50. I do think there's a net positive in displaying these in their current shape, but it is a bit of a hack at the end of the day.

> Ofc it would be "better" if a way of representing available server settings was standardized, similar to how initialization options are standardized. I'm not personally interested in lobbying microsoft to amend the protocol, because I think (as @clason) alluded to, it's extremely unlikely to materialize and the cost/benefit ratio is low (there are many things I would change in the LSP specification before this).

I'm not sure I agree this is about amending the LSP itself, it's more about awareness and general governance I think. These configuration schemas alone are just JSONSchemas (unless I'm mistaken), so there's nothing proprietary about them. I think due to the close relation between MS and LSP, the boundaries between VSCode and LSP have in many cases become very blurred, probably because it's the route of least resistance for server authors[**.**](https://en.wikipedia.org/wiki/Embrace,_extend,_and_extinguish)

A recent example to describe what I mean by awareness - recently `opencl_ls` [was added to lspconfig](https://github.com/neovim/nvim-lspconfig/pull/1654). This wasn't really feasible a few days ago, because it was built and distributed solely as a VSCode extension. After communicating there's a demand from neovim users to make the server more widely available, the author very promptly extracted the server components to a separate [repo](https://github.com/Galarius/opencl-language-server), with a separate release pipeline. I'd like to think that this is the general mindset of server authors - they've put a lot of effort in implementing it and, unless there's COI, ought to be very interested in making it the best experience for all users, no matter the editor. Native neovim LSP is gaining a lot of traction and momentum, and with it a strong mandate to drive positive change. Anyway this is getting way OT, I don't think you should reply to this in this thread 😂. I'm very happy to have rediscovered neovim last year, thanks for all your work!

PR #1662, opened 2022-01-16
fix(golangci_lint_ls): use .golangci.yaml config

### LINKED ISSUE #1661: golangci-lint lsp defaults to `enable-all`, ignoring a project's configuration file
### Description

The default command for this language server ignores any project-level configuration in the `.golangci.yaml` file and instead explicitly runs all linters.

https://github.com/neovim/nvim-lspconfig/blob/89eda6c9ba3ded27ecd38dfcee811c11c91eae05/lua/lspconfig/server_configurations/golangci_lint_ls.lua#L8

I think this command could just be reduced to the following

```lua
command = { 'golangci-lint', 'run', '--out-format', 'json' },
```

I'm not sure if this is intended behavior, but it seems like a more sane default to use the golangci-lint defaults rather than forcing all linters.

### Neovim version

N/A

### Nvim-lspconfig version

89eda6c9ba3ded27ecd38dfcee811c11c91eae05

N/A

### Affected language servers

golangci_lint_ls

### Steps to reproduce

1. Install nvim-lspconfig
2. Enable golangci-lint lsp without overriding any default configuration
3. Open any file in Go project with a `.golangci.yaml` and see that the configuration is ignored

### Actual behavior

Linting errors from all linters appear in the editor

### Expected behavior

Only linting errors should appear in the editor that are configured in `golangci.yaml`

### Minimal config

```Lua
N/A
```

### LSP log

N/A

jzelinskie:
Fixes #1661

### DIFF
--- a/lua/lspconfig/server_configurations/golangci_lint_ls.lua
+++ b/lua/lspconfig/server_configurations/golangci_lint_ls.lua
@@ -5,10 +5,10 @@ return {
     cmd = { 'golangci-lint-langserver' },
     filetypes = { 'go', 'gomod' },
     init_options = {
-      command = { 'golangci-lint', 'run', '--enable-all', '--disable', 'lll', '--out-format', 'json' },
+      command = { 'golangci-lint', 'run', '--out-format', 'json' },
     },
     root_dir = function(fname)
-      return util.root_pattern 'go.work'(fname) or util.root_pattern('go.mod', '.git')(fname)
+      return util.root_pattern 'go.work'(fname) or util.root_pattern('go.mod', '.golangci.yaml', '.git')(fname)
     end,
   },
   docs = {
@@ -28,7 +28,7 @@ go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1
 
 ]],
     default_config = {
-      root_dir = [[root_pattern('go.mod', '.git')]],
+      root_dir = [[root_pattern('go.work') or root_pattern('go.mod', '.golangci.yaml', '.git')]],
     },
   },
 }


lua/lspconfig/server_configurations/golangci_lint_ls.lua
```diff
@@ -5,10 +5,10 @@ return {
     cmd = { 'golangci-lint-langserver' },
     filetypes = { 'go', 'gomod' },
     init_options = {
-      command = { 'golangci-lint', 'run', '--enable-all', '--disable', 'lll', '--out-format', 'json' },
+      command = { 'golangci-lint', 'run', '--out-format', 'json' },
     },
     root_dir = function(fname)
-      return util.root_pattern 'go.work'(fname) or util.root_pattern('go.mod', '.git')(fname)
+      return util.root_pattern('.golangci.yaml')(fname) or util.root_pattern('go.work')(fname) or util.root_pattern('go.mod', '.git')(fname)
```
jzelinskie, line 11: I'm not certain how to best use `util.root_pattern`.
Can't this line just be the following?

```lua
return util.root_pattern('.golangci.yaml', 'go.work', 'go.mod', '.git')(fname)
````

Comment (mjlbach) [reply lua/lspconfig/server_configurations/golangci_lint_ls.lua:11]: The difference is, in the current case, `.golanci.yaml` is first searched for (up until the filesystem boundary), then `.go.work`, then `'go.mod` or `.git` (each level of the FHS is checked for either go.mod or .git`

In the latter case, all 4 files are checked for at each level of the hierarchy.

This is mostly done for projects with submodules/similar hierarchical structure. `go.work` tends to be the most unambiguous root marker.

File: lua/lspconfig/server_configurations/golangci_lint_ls.lua [hunk 2]
```diff
@@ -28,7 +28,7 @@ go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1
 
 ]],
     default_config = {
-      root_dir = [[root_pattern('go.mod', '.git')]],
+      root_dir = [[root_pattern('go.work', 'go.mod', '.golangci.yaml', '.git')]],
```
mjlbach, line 31: This should reflect line 11

mjlbach:
Style lint

PR #1664, opened 2022-01-17
fix for tailwindcss v0.7.6

kabouzeid:
Added `tailwindCSS.files.exclude` default setting, copied from the VSCode extension. Without this the server will crash. See https://github.com/tailwindlabs/tailwindcss-intellisense/issues/459

### DIFF
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -87,6 +87,13 @@ return {
           'classList',
           'ngClass',
         },
+        files = {
+          exclude = {
+            '**/.git/**',
+            '**/node_modules/**',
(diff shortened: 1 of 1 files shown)


kabouzeid:
never mind, has been fixed upstream by now

mhumeSF:
I think this issue still persists. I'm runnin `0.7.6` and had to add the above change in order not to get the error:
```
[ERROR][2022-02-06 01:03:28]  ...lsp/handlers.lua:404	"Unhandled exception: Cannot read properties of undefined (reading 'exclude')...
 ```

kabouzeid:
Yes it's still there in 0.7.6. It's fixed one commit after that, but not released in the VSCode extension yet.
You need at least version 0.0.7 of the standalone language server.

PR #1665, opened 2022-01-17
feat: add awk-language-server

lilibyte:
without handling `workspace/workspaceFolders` the server crashes at start: `server_request: no handler found for workspace/workspaceFolders`


lua/lspconfig/server_configurations/awk_language_server.lua
```diff
@@ -0,0 +1,30 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'awk-language-server' },
+    filetypes = { 'awk' },
+    root_dir = function(fname)
+        return util.path.dirname(fname)
+    end,
```
lithammer, line 9: @mjlbach this actually seems like a valid use-case for `dirname()` 🤔 I guess we still don't have support for `root_dir = nil`, right?

lithammer:
You need to format the code with Stylua and there's some commit messege errors (e.g. you need to wrap the body to not exceed 100 chars).

lua/lspconfig/server_configurations/awk_language_server.lua
lilibyte: I believe I got that function from this example. Is there a preferred way to do it? I used `vim.loop.cwd()` originally but I figured a lspconfig/util function would be better

https://github.com/neovim/nvim-lspconfig/pull/883/commits/5cd9366a063403893481f68be0ed0b1af9a59468#diff-6481f9ca36fe09d2ca7afebce8e8584728f828457f35a6217f68ebf28335d484R18

lilibyte: Also the author of awk-language-server made these changes to `handlers` if they're worth considering?

https://github.com/Beaglefoot/awk-language-server/commit/3e854556f9eaef5b69abed055bf0f1157973cd1e

mjlbach: The handler should be implemented in core, not here. Did you try with single file support enabled? I'd rather not re introduce dirname/cwd hacks

lilibyte: @mjlbach I see the problem now. Without `handlers` the language server crashes with the error in my initial comment. But I was mistaken in thinking `root_dir` had something to do with this so I'll remove it.

kylo252: @lilibyte, did you try removing the handlers after adding `single_file_support = true`?

lilibyte: @kylo252 yes `Client 1 quit with exit code 1 and signal 0` and from `lsp.log`:
```
"server_request: no handler found for"	"workspace/workspaceFolders"
```

kylo252: hmm, don't you need to update it as per the snippet that was suggested earlier anyway? https://github.com/neovim/nvim-lspconfig/pull/1665#discussion_r788936812

mjlbach: Try this PR for core instead https://github.com/neovim/neovim/pull/17149/files

lilibyte: > hmm, don't you need to update it as per the snippet that was suggested earlier anyway? [#1665 (comment)](https://github.com/neovim/nvim-lspconfig/pull/1665#discussion_r788936812)

I just tested it according to the awk-lsp maintainers' suggestions and the server crashes: https://github.com/Beaglefoot/awk-language-server/pull/23#issuecomment-1017769823 will wait to hear back before updating anything

lua/lspconfig/server_configurations/awk_language_server.lua
```diff
@@ -0,0 +1,27 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'awk-language-server' },
+    filetypes = { 'awk' },
+    single_file_support = true,
+    handlers = {
```
mjlbach, line 8: This can be removed now that https://github.com/neovim/neovim/pull/17149#event-5930028005 is merged. I would recommend gating the configuration behind a 0.7 check (you can refer to lemminx for how to do this)

lilibyte: @mjlbach I can't find the version gating within lemminx.lua

mjlbach: Ah, sorry vdmj https://github.com/neovim/nvim-lspconfig/pull/1094/files#diff-cc4d28b8a44a7645573450364c61477dffef8cf48b0005630bfaed16ab150b2aR4-R7

lua/lspconfig/server_configurations/awk_language_server.lua
lilibyte: `root_dir` was removed in https://github.com/neovim/nvim-lspconfig/pull/1665/commits/adc7ba63bbc64caac006b61ea1a3f1e5d58e97be and `handlers` was removed in https://github.com/neovim/nvim-lspconfig/pull/1665/commits/721ae1d51d9dac7a4eef05a989d2ae31e05614ed

lua/lspconfig/server_configurations/awk_language_server.lua
lilibyte: gated version to >=0.7 using `vim.version()` and removed `handlers` in https://github.com/neovim/nvim-lspconfig/pull/1665/commits/721ae1d51d9dac7a4eef05a989d2ae31e05614ed

mjlbach:
I renamed the server to awk_ls following our naming conventions and addressed the lint issues.

PR #1667, opened 2022-01-18
XSLT file types for lemminx

muhmud:


lithammer:
You need to update the commit message according to [conventional commit](https://www.conventionalcommits.org/).

muhmud:
OK, done

PR #1669, opened 2022-01-18
Improved default config of the html serverconfiguration 

rgroli:
Also added a line of documentation.


lithammer:
You need to squash and rephrase the commit according to https://www.conventionalcommits.org/

PR #1678, opened 2022-01-22
Set ALS root dir to / and show current GPR project file

TamaMcGlinn:
This fixes the problem that when you jump into a system library following a definition, a new ALS instance is started which doesn't know which gpr file to use and hence throws an error and stops working. Instead, we keep the GPR project file that you started with.

If you combine that with [this gpr selector plugin](https://github.com/TamaMcGlinn/nvim-lsp-gpr-selector) to allow switching out the selected GPR project, then you can even use ALS from projects that have multiple gpr files.

### DIFF
--- a/doc/server_configurations.md
+++ b/doc/server_configurations.md
@@ -169,6 +169,11 @@ require'lspconfig'.als.setup{}
   Default Values:
     cmd = { "ada_language_server" }
     filetypes = { "ada" }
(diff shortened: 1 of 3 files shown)


mjlbach:
This isn't the right approach. You're free to do this in your own configuration though.

TamaMcGlinn:
So, what is the right approach to the problem, then?

PR #1687, opened 2022-01-28
feat: allow attaching to paths inside archives

stephank:
This change allows attaching to buffers with `zipfile:` or `tarfile:` virtual paths.

A function `util.strip_archive_subpath(path)` was added to extract the archive path from such a virtual path, and `root_dir` for `tsserver` is configured to use it by default. This means the `root_dir` search for TypeScript starts at the archive, instead of doing a real traversal starting from the file *inside* the archive. This is a compromise, because otherwise we'd have to provide a whole bunch of support functionality (currently provided by `vim.loop`) to do the necessary checks inside archives.

I targeted TypeScript for Yarn PnP support, which keeps project dependencies in zip files instead of `node_modules`. I imagine this may also be useful for other languages, like Java with JARs, but I'm really not familiar with those.

An alternative solution I considered was to add behaviour like `strip_archive_subpath` to existing utility functions to accomplish the same, but I'm worried about breaking existing code, and perhaps other language servers need different behaviour.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -21,11 +21,13 @@ M.default_config = {
 M.on_setup = nil
 
 function M.bufname_valid(bufname)
-  if bufname and bufname ~= '' and (bufname:match '^([a-zA-Z]:).*' or bufname:match '^/') then
-    return true
-  else
+  if not bufname then
     return false
   end
+  if bufname:match '^/' or bufname:match '^[a-zA-Z]:' or bufname:match '^zipfile://' or bufname:match '^tarfile:' then
+    return true
+  end
+  return false
 end
 
 function M.validate_bufnr(bufnr)
@@ -341,6 +343,7 @@ function M.root_pattern(...)
     end
   end
   return function(startpath)
+    startpath = M.strip_archive_subpath(startpath)
     return M.search_ancestors(startpath, matcher)
   end
 end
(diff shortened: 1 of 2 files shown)


rafaeldelboni:
Thanks for this PR, it will probably help clojure-lsp as well since we do have zipfiles and [jar:files](https://github.com/Olical/conjure/blob/2717348d1a0687327f59880914fa260e4ad9c685/fnl/conjure/client/clojure/nrepl/action.fnl#L204-L205) buffer types

justinmk:
Not sure I fully understand. Is the point to trick the root finder into thinking that this file exists on the project path? Then this should be handled in the root finder, instead of special-casing a bunch of lspconfigs. And we definitely would want a few test cases like:

```
assert('...' == fixup_filepath('zipfile://...')
assert('...' == fixup_filepath('tarfile://...')
...
```

stephank:
I rebased and added tests, plus removed the special casing.

The idea is to have the root finder start the search at the zipfile itself, for a buffer with a file inside a zipfile. I have a Yarn v3 project and when I jump to the definition of a dependency, it looks like:

```
zipfile:///path/to/project/.yarn/cache/some-dependency-1.2.3-abcde456.zip::node_modules/some-dependency/index.js
```

By starting the lookup at `some-dependency-1.2.3-abcde456.zip`, it can eventually find `/path/to/project` as the root and use the same tsserver as the rest of the project.

glepnir:
LGTM

PR #1689, opened 2022-01-28
docs: update suggested configuration readme

daangoossens22:
Shouldn't the `vim.lsp.buf.formatting()` keymap be in the `on_attach` function as it should only do something when a lsp server is attached?


mjlbach:
whoops. Want to fix lspconfig.txt too?

daangoossens22:
sure, give me a sec

daangoossens22:
I think that should do it.

daangoossens22:
maybe this fixes the errors?

PR #1692, opened 2022-01-29
feat(pasls): add lpi and lpk to root dir detection

BarrOff:
Hello,

Pascal projects using [lazbuild](https://wiki.freepascal.org/lazbuild) use [lpi/lpk](https://wiki.freepascal.org/File_extensions) file extensions.
This PR adds them to detect the project root for pasls.

### DIFF
--- a/lua/lspconfig/server_configurations/pasls.lua
+++ b/lua/lspconfig/server_configurations/pasls.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'pasls' },
     filetypes = { 'pascal' },
-    root_dir = util.find_git_ancestor,
+    root_dir = util.root_pattern('*.lpi', '*.lpk', '.git'),
     single_file_support = true,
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/pasls.lua
```diff
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'pasls' },
     filetypes = { 'pascal' },
-    root_dir = util.find_git_ancestor,
+    root_dir = root_pattern('*.lpi', '*.lpk', '.git'),
```
mjlbach, line 7: you need to bring root pattern in scope, this fails lint because you are accessing an undefined function (root_pattern)

justinmk: @BarrOff to fix this, add the appropriate require() at top of the file. see other configs for example

Issue #1683, opened 2022-01-25
ALS fails after jumping into library with multiple or zero gpr files


TamaMcGlinn:
### Description

While in my project, ALS works just fine, assuming I have a single .gpr file.
But when I jump through a reference to some library, I get a new ALS instance which fails because it is given the root directory of the library, which contains multiple or zero .gpr files.

### Neovim version

NVIM v0.6.0-dev+560-g1fdbd29df
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-11 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az32-619

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "
/home/runner/work/neovim/neovim/build/nvim.AppDir/usr/share/nvim"

### Nvim-lspconfig version

c510964

### Operating system and version

Ubuntu 21.04

### Affected language servers

als

### Steps to reproduce

1. Open Ada file (e.g. src/mkproj.adb in [mkproj](https://github.com/TamaMcGlinn/mkproj))
2. Go to definition on something in the gnat runtime, e.g. Ada.Environment_Variables on line 8 of above.

### Actual behavior

ALS spits out this error: 'LSP[als] More than one .gpr found. Note: you can configure a project through the ada.projectFile setting' and also doesn't work. For example, I can't jump to definitions / references in the library file now.

[...]

PR #1693, opened 2022-01-30
feat(als): Show GPS project file in LspInfo

TamaMcGlinn:
Add extra info to the als LspInfo to show which GPR project is being used. When there is none or more than one, LspInfo shows this. In that case, als will not work.

gpr_display

### DIFF
--- a/lua/lspconfig/server_configurations/als.lua
+++ b/lua/lspconfig/server_configurations/als.lua
@@ -10,6 +10,31 @@ return {
     cmd = { bin_name },
     filetypes = { 'ada' },
     root_dir = util.root_pattern('Makefile', '.git', '*.gpr', '*.adc'),
+    lspinfo = function(cfg)
+      local extra = {}
+      local function find_gpr_project()
+        local function split(inputstr, sep)
+          if sep == nil then
+            sep = '%s'
+          end
+          local t = {}
+          for str in string.gmatch(inputstr, '([^' .. sep .. ']+)') do
+            table.insert(t, str)
+          end
+          return t
+        end
+        local projectfiles = split(vim.fn.glob(cfg.root_dir .. '/*.gpr'))
+        if #projectfiles == 0 then
+          return 'None (error)'
+        elseif #projectfiles == 1 then
+          return projectfiles[1]
+        else
+          return 'Ambiguous (error)'
+        end
+      end
+      table.insert(extra, 'GPR project:     ' .. ((cfg.settings.ada or {}).projectFile or find_gpr_project()))
+      return extra
+    end,
   },
   docs = {
     package_json = 'https://raw.githubusercontent.com/AdaCore/ada_language_server/master/integration/vscode/ada/package.json',


TamaMcGlinn:
Any issues merging this?

TamaMcGlinn:
Can you merge this please? It makes als much more useful in neovim.

mjlbach:
I'm currently in the process of trying to remove server specific workarounds, and make lspconfig a generic, vim-lsp-settings-like minimal repository. I'm interested in pushing people to create server specific plugins that can provide more tailored experience. I also don't have much time for neovim atm, but I really do not want to encourage LspInfo becoming something beyond a basic debugging tool for servers.

TamaMcGlinn:
so, would you rather I create a separate nvim-als-lspconfig plugin, that depends on nvim-lspconfig, or includes its sources? Also, what's the motivation for this? It is obviously much easier to just click 'merge' than to think about how to split off each lsp configuration server, but I guess you have some strong motivations for wanting that.

mjlbach:
> so, would you rather I create a separate nvim-als-lspconfig plugin, that depends on nvim-lspconfig, or includes its sources? 

I'd rather an ada.nvim that provided the same functionality that lspconfig does, namely addressing your other concerns about workspaceFolders/project system modules and providing any project specific info in its own UI.

>  Also, what's the motivation for this?

Namely, I want to keep the scope of lspconfig limited. There are many features users request, but would be unsustainable to have in a catch-all plugin like lspconfig. Things like:

* python virtual environment selection
* support for rust-analyzer's inlay hints
* showing the active GPS project file
* handling system libraries as workspace folders for various language servers/platforms.

The goal is to move any useful infrastructure from lspconfig to core, and then you can build a tailored language plugin that leverages the neovim lsp API. 

It's only 300 lines of lua code you would need to copy out of lspconfig to replicate it's functionality for your plugin, which is fine so long as it's Apache-2.0 license and the source is attributed.

> It is obviously much easier to just click 'merge' 

If just clicking merge was my approach to being a maintainer, I think I'd be a rather poor one.

TamaMcGlinn:
@justinmk I would still rather this be merged.

> I want to keep the scope of lspconfig limited

While this argument sounds reasonable, my fixes were both in `[lua/lspconfig/server_configurations/als.lua](https://github.com/neovim/nvim-lspconfig/pull/1693/files#diff-45537e50531c6c969942cfd5e5fc8263f1783f117c3d2e868100acc1c426c1ca)` and did not affect any other plugins. There are plenty of examples of such features being accepted and merged for languages that are in use by the plugin's maintainers. Yet, the argument could stand equally well for disallowing any fix or update that is only to a specific language config. So you have to recognize there is a double standard here.

I understand I could create a separate als-config plugin that copies a bunch of code from nvim-lspconfig and works properly. However, there is no advantage to leaving als.lua here without the feature. It is easier and better to just maintain a fork of nvim-lspconfig with Ada support added; like I said, the only things I change are in the ada.lua file, so I still don't see any problem in merging.

On the one hand, I understand not wanting to spend any time on language support for a language you don't personally use, but if that's the consideration; why keep als.lua at all? Or any other language, for that matter? I would be happy to be the maintainer for the specific lsp config of the language I am using, but it's not very nice to tell me you want to keep the scope of your plugin limited, and thereby reject any updates to the als.lua config, especially when it is broken like this.

TamaMcGlinn:
also note that I have been using this particular closed PR, rebased onto the latest master, for almost half a year now with no modifications necessary or issues encountered. If you do merge, and someone opens an issue; of course I will jump on that and fix it. I'm not asking you to maintain als.lua - just to accept fixes / updates like you do for ccls, eslint and others.

TamaMcGlinn:
Just checking in again, as I just rebased my fork onto master again; I still feel this belongs here. Any ALS user would want this feature built-in.

Please clarify how ALS is so different from all the other language server configs, in that you want its configuration in a separate repository, instead of welcoming PRs here that improve the nvim-lspconfig supplied config. What you are proposing would leave it up to users to add yet another plugin with duplicated code from nvim-lspconfig, with no apparent benefit.

Also see [issue 1683](https://github.com/neovim/nvim-lspconfig/issues/1683), which was erroneously closed. The repro steps, I'm sure, are encountered by every new user of ALS because the issue uproots even the most basic usage of ALS. I mention this because this PR, while it doesn't fix the issue, at least it shows the user what the issue is, by showing an error message in LspInfo, stating there are 0 or multiple gpr files.

TamaMcGlinn:
If this PR needs any changes before merging, please give me specific feedback on the code. Today I had another look and trimmed out a few lines of unnecessary code, but it doesn't seem that this PR was closed for any specific reason, so I'm not sure what needs doing in order to integrate this into the master branch.

I hope everyone can see that 'keeping the scope of lspconfig limited' is not achieved by rejecting all contributed updates to the als configuration. If you wanted to split out lspconfig-core and have separate plugins for each languageserver, that would be fine, but that doesn't seem to be what you are doing. As long as you are accepting changes to any languageserver config, you should also accept changes to the ALS config. The other part of this argument, put forward by @mjlbach, was to put my proposed change in a list among very different changes, where each of the others are for specific setups using the corresponding languageserver, which not all users of that languageserver would need - in contrast, als always needs a GPR file, and searches for it automatically in the same manner as this PR's code does, so every user of ALS would find my PR useful diagnostics, especially since having the wrong gpr file, or ambiguity, is fatal for ALS. (see issue #1683 - this PR does not fix it, but at least makes it a lot easier to debug. for a fix, see [nvim-lsp-gpr-selector](https://github.com/TamaMcGlinn/nvim-lsp-gpr-selector/))

Please, if you are using ALS, check if all of this works for your purposes. If not, you should have no issue merging this - I have been using this daily for six months already, so I'm pretty confident it works.

PR #1698, opened 2022-02-05
feat(leanls): start the LSP via lake serve

Julian:
Enables workspace symbols for the Lean 4 LSP.

Older versions of Lake (and Lean 4) lack a serve command,
so on those we fallback to the old behavior by directly starting
the LSP.

Closes: Julian/lean.nvim#229
Refs: leanprover/lake#46

CC @gebner

### DIFF
--- a/lua/lspconfig/server_configurations/leanls.lua
+++ b/lua/lspconfig/server_configurations/leanls.lua
@@ -1,8 +1,12 @@
 local util = require 'lspconfig.util'
 
+-- Only Lake 3.0+ supports lake serve, so for old enough Lean 4,
+-- we fallback to the builtin LSP.
+local legacy_cmd = { 'lean', '--server' }
+
 return {
   default_config = {
-    cmd = { 'lean', '--server' },
+    cmd = { 'lake', 'serve', '--' },
     filetypes = { 'lean' },
     root_dir = function(fname)
       -- check if inside elan stdlib
@@ -20,6 +24,19 @@ return {
         or util.find_git_ancestor(fname)
     end,
     on_new_config = function(config, root_dir)
+      local lake_version = ''
+      local lake_job = vim.fn.jobstart({ 'lake', '--version' }, {
+        on_stdout = function(_, d, _)
+          lake_version = table.concat(d, '\n')
+        end,
+        stdout_buffered = true,
+      })
+      if lake_job > 0 and vim.fn.jobwait({ lake_job })[1] == 0 then
+        local major = lake_version:match 'Lake version (%d).'
+        if major and tonumber(major) < 3 then
+          config.cmd = legacy_cmd
+        end
+      end
       -- add root dir as command-line argument for `ps aux`
       table.insert(config.cmd, root_dir)
     end,


mjlbach:
I want to eventually add yielding method for asyncifying these more expensive startup checks (like RA)

Julian:
Sounds good to me, I copied what I saw other LSPs use but yeah some non-blocking way would certainly be nice.

lua/lspconfig/server_configurations/leanls.lua
```diff
@@ -20,6 +24,19 @@ return {
         or util.find_git_ancestor(fname)
     end,
     on_new_config = function(config, root_dir)
+      local lake_version = ''
```
gebner, line 27: We also need to check that a `lakefile.lean` exists.  (Otherwise `lake serve` will fail.)

lua/lspconfig/server_configurations/leanls.lua
```diff
@@ -20,6 +24,19 @@ return {
...
+      local lake_job = vim.fn.jobstart({ 'lake', '--version' }, {
+        on_stdout = function(_, d, _)
+          lake_version = table.concat(d, '\n')
+        end,
+        stdout_buffered = true,
+      })
+      if lake_job > 0 and vim.fn.jobwait({ lake_job })[1] == 0 then
+        local major = lake_version:match 'Lake version (%d).'
+        if major and tonumber(major) < 3 then
+          config.cmd = legacy_cmd
```
gebner, line 37: I wonder if we should make this overridable (e.g. by adding `legacy_cmd` as a config field)?

lua/lspconfig/server_configurations/leanls.lua
```diff
@@ -20,6 +24,19 @@ return {
...
     end,
     on_new_config = function(config, root_dir)
+      local lake_version = ''
+      local lake_job = vim.fn.jobstart({ 'lake', '--version' }, {
+        on_stdout = function(_, d, _)
+          lake_version = table.concat(d, '\n')
+        end,
+        stdout_buffered = true,
+      })
+      if lake_job > 0 and vim.fn.jobwait({ lake_job })[1] == 0 then
```
gebner, line 34: If `lake --version` fails to start, we should also fall back to the legacy command (for Lean 4 versions before lake existed).

lua/lspconfig/server_configurations/leanls.lua
Julian: It's technically already overridable by end-users who want to define `on_new_config` themselves -- but generally to make it even more easily overridable, maybe someday we want `timeout` and `memory` options like VSCode does to assemble the command line, and then at that point could add one for this too -- but I'm not sure I see why it's super useful personally yet otherwise. The vast majority of people will never even know this exists once Lean 4 has another release, no?

lua/lspconfig/server_configurations/leanls.lua
Julian: You still need to use versions that old?

lua/lspconfig/server_configurations/leanls.lua
Julian: This works as-is for me like I said [here](https://github.com/Julian/lean.nvim/issues/229#issuecomment-1032565846) but let me know if I missed which case triggers this -- unless you mean we should fix single-file mode for `lake==3.0`, in which case yes that'd be nice but I don't have much energy to do it at the minute, maybe later.

lua/lspconfig/server_configurations/leanls.lua
gebner: Yes?  We're already checking for the status code, so we might as well fall back correctly.

lua/lspconfig/server_configurations/leanls.lua
gebner: The `legacy_cmd` is still used when editing files in core lean (which doesn't use lake).  Having a field for extra command-line arguments seems like a good idea (I don't think we need `timeout` and `memory` specifically, writing `-M4000` is just as easy).

lua/lspconfig/server_configurations/leanls.lua
Julian: OK I suppose -- it's less clear to me that that's more "correct" unless it's really prevalent to deal with those versions -- the deterministic behavior seemed more valuable. E.g. if for some transient reason (out of PIDs momentarily, say) `lean --version` failed, we'll fallback "incorrectly", hence why I asked whether the likelihood someone is on something like `4.0.0-m2` is bigger. But I can change.

gebner: I think it's very unlikely that `lake --version` fails but then `lean --server` works afterwards.

PR #1704, opened 2022-02-09
feat: Add slint_lsp

hunger:
We are in the process of renaming our project, so I would love to be able to get the sixtyfps lsp in under its new name.

We will unfortunately need to support both the old and the new one for a while longer, so I would like to get the new one added and will only send a PR to remove the "old" one once no users are left -- if that is OK on your side of course.

I used slint_lsp and not the plain slint since this would be too close to the existing eslint for my taste:-)

### DIFF
--- a/lua/lspconfig/server_configurations/slint_lsp.lua
+++ b/lua/lspconfig/server_configurations/slint_lsp.lua
@@ -1,0 +1,26 @@
+return {
+  default_config = {
+    cmd = { 'slint-lsp' },
+    filetypes = { 'slint' },
+    single_file_support = true,
+  },
+  docs = {
+    description = [=[
+https://github.com/slint-ui/slint
+`Slint`'s language server
+
+You can build and install `slint-lsp` binary with `cargo`:
+```sh
+cargo install slint-lsp
+```
+
+Vim does not have built-in syntax for the `slint` filetype at this time.
+
+This can be added via an autocmd:
+
+```lua
+vim.cmd [[ autocmd BufRead,BufNewFile *.slint set filetype=slint ]]
+```
+]=],
+  },
+}


mjlbach:
The convention would be `slint_ls` for naming, should we not remove the 60fps language server?

hunger:
It would be cool if the sixtyfps one could stay a bit longer, to give people time to upgrade.

I will rename the configuration.

hunger:
I changed the name. I can create a separate PR for removing the sixtyfps language server if you like. It would be cool if it could stay around for a bit longer though:-)

hunger:
Do you want the PR to drop sixtyFPS LS first or how do we proceed?

I would really love to make progress on this PR, as I apparently need it to put the slint_ls into the lsp-installer plugin -- and I would really love to have that as that would make my live so much easier.

mjlbach:
Sorry, just busy and don't have much time to review lspconfig PR's atm. There's no holdup, one thing I realize is the `cmd` is actually slint-lsp, so I'm fine if you want to call the file `slint_lsp` as well, your call, I'll merge either way. Usually people use lsp as a misnomer for language server which is why I usually push towards `*_ls`

hunger:
Oh, perfekt! I did not mean to rush, I was just unsure whether this was blocked on something I need to do. I'll change the name back then to be in line with the cmd name as I think that is less confusing.

I'll also fix the lint!

mjlbach:
Sure, although I would recommend changing upstream to be slint_ls :P 

I'm not sure why people create binaries that include "lsp" since lsp expands to language server protocol, the protocol, not language server, the server itself.

hunger:
You are right, but I am afraid I missed the chance to rename the binary: Doing so now will break user setup.

hunger:
Thanks!

PR #1708, opened 2022-02-10
fix(ansiblels): remove Yaml from ansiblels Filetype

vincentDcmps:
to avoid ansible-lint to pop error on no ansible file (by example kubernetes deploiement)


vincentDcmps:
Capture d’écran du 2022-02-10 18-18-23

PR #1712, opened 2022-02-11
feat: add hh_client for hhvm

antonk52:
This should makes `hh_client` work out of the box

### DIFF
--- a/lua/lspconfig/server_configurations/hhvm.lua
+++ b/lua/lspconfig/server_configurations/hhvm.lua
@@ -1,25 +1,19 @@
 local util = require 'lspconfig.util'
 
-local bin_name = 'hh_client'
-local cmd = {bin_name, 'lsp'}
-
-local description = [[
+return {
+  default_config = {
+    cmd = {'hh_client', 'lsp'},
+    filetypes = { 'php', 'hack' },
+    root_dir = util.root_pattern '.hhconfig',
+  },
+  docs = {
+    description = [[
 Language server for programs written in Hack
 https://hhvm.com/
 https://github.com/facebook/hhvm
 See below for how to setup HHVM & typechecker:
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/hhvm.lua
```diff
@@ -0,0 +1,27 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'hh_client'
```
mjlbach, line 3: There's no reason for these not to be defined in the table, plus the way I did autodocgen it will insert `cmd` automatically if it is inline (using the lua debug module)

lua/lspconfig/server_configurations/hhvm.lua
```diff
@@ -0,0 +1,27 @@
...
+See below for how to setup HHVM & typechecker:
+https://docs.hhvm.com/hhvm/getting-started/getting-started
+]]
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'php', 'hack' },
+    root_dir = util.root_pattern '.hhconfig',
+    settings = {},
```
mjlbach, line 19: Does this need to be an empty table?

lua/lspconfig/server_configurations/hhvm.lua
antonk52: Thanks, inlined them in the following commit

lua/lspconfig/server_configurations/hhvm.lua
antonk52: Removed in the following commit

antonk52:
@mjlbach Thanks for the review, I've addressed the comments. Please let me know if there is anything else I need to do to get this merged.

PR #1714, opened 2022-02-11
Fix documentation for eslint autocmd

aryzing:
The previously documented autocmd to fix files on save had incorrect syntax, and fails. Has been updated to the correct use the correct syntax.


lithammer:
You need to update the commit message format according to commitlint.

PR #1718, opened 2022-02-13
feat: elmls compatibility with elm-tooling

georgesboris:
As stated in the [elm-language-server docs](https://github.com/elm-tooling/elm-language-server#server-settings) – the default value of elm binaries should be an empty string so the language server searches for binaries as local project installs and only use global ones as fallback. If we specify it as the standard "global values" like this config is currently doing, we're just going straight to the fallback, breaking compatibility with local installs.

This change is essential for compatibility with tools like [elm-tooling](https://elm-tooling.github.io/elm-tooling-cli/) which is becoming the standard way of maintaining elm related tools with different versions per project.

### DIFF
--- a/lua/lspconfig/server_configurations/elmls.lua
+++ b/lua/lspconfig/server_configurations/elmls.lua
@@ -25,9 +25,9 @@ return {
       end
     end,
     init_options = {
-      elmPath = 'elm',
-      elmFormatPath = 'elm-format',
-      elmTestPath = 'elm-test',
+      elmPath = '',
+      elmFormatPath = '',
+      elmTestPath = '',
       elmAnalyseTrigger = 'change',
     },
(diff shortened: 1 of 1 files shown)


mjlbach:
Can we just drop the keys? Does this have implications for current users or should it continue to work without changes?

georgesboris:
I'm not familiar with the innards of lspconfig but those are optional settings on the elm-language-server.

And yep! The current users should work just as usual since the fallback does exactly what the current implementation does.

georgesboris:
Yup! Will do - I'm afk now but I'll let you know! Thanks

georgesboris:
@mjlbach I tested your suggestion and it indeed works as expected without those properties. I updated the PR accordingly.

PR #1737, opened 2022-02-24
fix(tailwindcss): add 'htmldjango' to filetypes

williamboman:


lcfd:
Hi @williamboman! 
I don't know why but with the `htmldjango` filetype I can't make it works.
If I set the type to `django-html` it works but I lose everything else, like syntax highlighting, other autocompletion, etc.

Do you have any idea about why this can happen?

I know that has passed quite some time but I hope that someone can help me.

PR #1747, opened 2022-02-27
add Odin language server configuration

st4ll1:


lua/lspconfig/server_configurations/ols.lua
```diff
@@ -0,0 +1,20 @@
...
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
+           https://github.com/DanielGavin/ols
+
+           `Odin Language Server`.
+        ]],
+    default_config = {
+      root_dir = [[util.root_pattern("ols.json", ".git") or current_file_dirname]],
```
mjlbach, line 17: This is not accurate

st4ll1: sorry i copied this from the zls.lua file, 
and did not check with make lint

PR #1749, opened 2022-02-28
fix(perlpls): turn on syntax checking by default

WhoIsSethDaniel:


WhoIsSethDaniel:
Turns on a newly documented feature. Brings to parity with vscode defaults.

PR #1756, opened 2022-03-05
feat: add perlnavigator support

WhoIsSethDaniel:


WhoIsSethDaniel:
I don't think the lint error has anything to do with this PR.  Seems to be older code that is formatted incorrectly.

PR #1768, opened 2022-03-10
[julials] Let root_dir fallback to the directory of the file

fredrikekre:
If the file is not in a git repository, and no Project.toml file is found, this sets root_dir to the directory of the file. Some functionality in LanguageServer.jl is limited to files inside ["workspace folders"](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceFolder). When [rootUri](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize) (`root_dir`) is passed as null to LS.jl this means that no such folders are identified.

### DIFF
--- a/lua/lspconfig/server_configurations/julials.lua
+++ b/lua/lspconfig/server_configurations/julials.lua
@@ -45,7 +45,8 @@ return {
     cmd = cmd,
     filetypes = { 'julia' },
     root_dir = function(fname)
-      return util.root_pattern 'Project.toml'(fname) or util.find_git_ancestor(fname)
+      return util.root_pattern 'Project.toml'(fname) or util.find_git_ancestor(fname) or
+             util.path.dirname(fname)
     end,
     single_file_support = true,
   },


fredrikekre:
cc @pfitzseb

fredrikekre:
I guess this is explicitly disallowed, thats unfortunate.

lithammer:
It should already be by `single_file_support = true`.

fredrikekre:
Handled in LanguageServer.jl you mean?

lithammer:
No I mean that setting `single_file_support = true` is more or less the same as adding a `util.path.dirname(fname)` fallback to `root_dir`.

https://github.com/neovim/nvim-lspconfig/blob/096398a07c2dc08d785266f9f55ddcb462ba0749/lua/lspconfig/configs.lua#L231-L236

https://github.com/neovim/nvim-lspconfig/blob/096398a07c2dc08d785266f9f55ddcb462ba0749/lua/lspconfig/configs.lua#L107-L119

fredrikekre:
Right, except https://github.com/neovim/nvim-lspconfig/blob/096398a07c2dc08d785266f9f55ddcb462ba0749/lua/lspconfig/configs.lua#L111 which means some features in LanguageServer.jl is disabled. LS.jl doesn't lint files outside the workspace folders, in particular.

mjlbach:
> https://github.com/julia-vscode/LanguageServer.jl/blob/1bce8cdd7f01bc18ab3c6c0c1544efe4ee8eb6d3/src/utilities.jl#L101-L111. But maybe this can be configured on that end instead of here.

That would be my recommendation.

I made a decision to basically only (by default) allow sending a `null` workspaceFolders/rootURI when root patterns fail, so that there is a tangible difference between editing an isolated file, and editing a file inside a Julia project. I didn't want to add case-by-case exceptions for servers that play nice when we set large directories (like $HOME) as the workspaceFolders/rootURI.

PR #1777, opened 2022-03-14
feat(eslint): support yarn2 PnP projects

williamboman:
https://yarnpkg.com/features/pnp

Yarn's PnP feature changes the way packages are installed. Instead of building on the `node_modules` resolution, it introduces a single `.pnp.*js` file in the project. This file is responsible for orchestrating and resolving dependencies. The eslint LSP server will assume that regular `node_modules` resolution applies when locating the `eslint` package - which will not work in Yarn PnP projects.

To work around this, Yarn provides the ability to run Node programs in "PnP-compat" mode via `yarn exec` and `yarn node`. My understanding is that this simply hooks into the `require()` function to resolve modules via PnP instead Node's builtin module resolution.

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -98,6 +98,13 @@ return {
         uri = new_root_dir,
         name = vim.fn.fnamemodify(new_root_dir, ':t'),
       }
+
+      -- Support Yarn2 (PnP) projects
+      local pnp_cjs = util.path.join(new_root_dir, '.pnp.cjs')
+      local pnp_js = util.path.join(new_root_dir, '.pnp.js')
+      if util.path.exists(pnp_cjs) or util.path.exists(pnp_js) then
+        config.cmd = { 'yarn', 'exec', unpack(cmd) }
+      end
     end,
     handlers = {
       ['eslint/openDoc'] = function(_, result)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -107,6 +107,13 @@ return {
         uri = new_root_dir,
         name = vim.fn.fnamemodify(new_root_dir, ':t'),
       }
+
+      -- Support Yarn2 (PnP) projects
+      local pnp_cjs = util.path.join(new_root_dir, '.pnp.cjs')
+      local pnp_js = util.path.join(new_root_dir, '.pnp.js')
```
williamboman, line 113: Globbing might be nicer here, but wanted to avoid `vim.fn` call.

mufasa71: btw, this is not working with experimental `.pnp.loader.mjs` which is tries to load vscode-eslint-language-server esm format, and fails because of cjs module

mufasa71: Yes it would help, but  because of this [issue](https://github.com/yarnpkg/berry/issues/3900) it doesn't work, but its not related to this PR

williamboman:
> Is this ready for merge?

Yep! I just tested it again in a PNP workspace, works fine (and doesn't work without this).

PR #1779, opened 2022-03-15
Update eslint `autocmd` example

angusjf:
Fixed `autocmd` to:
a) Not use <cmd> and <cr> as not valid
b) Make the example only run the linter for Javascript / Typescript & React files (as it will error on save for non-js files)


angusjf:
@mjlbach Would be great to get this in, see this thread where people are talking about it https://www.reddit.com/r/neovim/comments/qru2lb/anyone_have_success_using_eslintfixall_with_an/

lithammer:
You need to amend the commit message according to commitlint (the failing luacheck problem is unrelated though).

PR #1782, opened 2022-03-16
feat: allow `:LspStop` and `:LspRestart` commands to apply to current buffer clients

### LINKED ISSUE #1781: Apply `:LspStop` and `:LspRestart` to current buffer only
### Language server

_No response_

### Requested feature

_No response_

rish987:
This makes the argument `0` mean to stop or restart all clients for the current buffer, which is useful since it would seem common to want to restrict these commands to a certain buffer, but you rarely know what the specific IDs of the clients are.

Closes #1781.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,10 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config. Pass `autostart = false` to your `.setup{}` call for a language server if you would like to launch clients solely with this command. Defaults to all servers matching current buffer filetype.
-* `:LspStop <client_id>` Defaults to stopping all buffer clients.
-* `:LspRestart <client_id>` Defaults to restarting all buffer clients.
+* `:LspStop <client_id>` Stops the given client(s), or those for the current buffer if none are given.
+* `:LspStop!` Stop all buffer clients.
+* `:LspRestart <client_id>` Restarts the given client(s), or those for the current buffer if none are given.
+* `:LspRestart!` Restarts all buffer clients.
 
 ## Wiki
 
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -10,6 +10,15 @@ function M.available_servers()
   return vim.tbl_keys(configs)
 end
 
+local function restart_client(client, bufnr)
+  client.stop()
+  vim.defer_fn(function()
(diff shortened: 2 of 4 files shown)


mg979:
I opened a PR to fix this at #1808. Your PR doesn't really solve the main problem (`LspStop` without arguments stops everything on all buffers, but `LspStart` cannot restart all clients on all buffers, only on current buffer). So `LspStop` shouldn't stop all clients in any circumstance.

rish987:
SGTM, if it's described as such in the readme we should probably go with that. I have no issues either way though.

mjlbach:
I don't personally care if we also change the documentation. I never use these commands, so if we want to change their meaning that is fine by me. IMO `:LspStop` completely shutting down all language servers and disabling the autocommands is probably what I would expect, and `:LspStop 0` limiting it to the current buffer is what I would prefer.

mg979:
> IMO :LspStop completely shutting down all language servers and disabling the autocommands is probably what I would expect

`LspStart` restarts the clients only in the current buffer. This means that after `LspStop` you should restart each client individually. As they are the commands are useless.

> I never use these commands

Yeah. Some people would like to use them though, and they are not usable right now. Stopping all clients and restarting them all every time would be (imho) overkill, unnecessary and potentially very taxing for no gain. Personally I use LspStop to pause the clients if it misbehaves or if diagnostics annoy me.

If there is the need for some `LspStopAll`, `LspRestartAll` they should be different commands, imho. I don't see the point in stopping all clients in all buffers and restarting them every now and then. That's not what I'd normally want these commands for, so I don't know why you think that stopping all clients on all would be better. I guess it's because, as you said, you don't use these commands.

mg979:
An easy alternative for `LspStop` would be to have
```
:LspStop                     stop all clients in current buffer
:LspStop {id}                stop client with id
:LspStop!                   stop all clients on all buffers
```

But the problem with `LspStart` would remain, stopping all clients on all buffers seems a last resort to me.

rish987:
(Sorry, I'm ramping up at a new job but will get back to this soon. This PR isn't exactly what I think we want yet, and I do like the idea above about the different `:LspStop` variants.)

rish987:
Okay, sorry for the long wait, this is ready for review now! @mg979 I made it work according to your suggestion, let me know what you think!

rish987:
cc @lithammer @justinmk

glepnir:
@rish987  this pr will conflict with #1984

justinmk:
These commands are intended as a high-level interface. If you want to have fine-grained behavior, use `vim.lsp.stop_client()`.

PR #1791, opened 2022-03-18
fix(beancount-langserver): add stdio flag to cmd

kmaasrud:
The Beancount language server requires the `--stdio` flag to work with Neovim. This should be the default, requiring less configuration by the user.

Fixes https://github.com/polarmutex/beancount-language-server/issues/49

### DIFF
--- a/lua/lspconfig/server_configurations/beancount.lua
+++ b/lua/lspconfig/server_configurations/beancount.lua
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    cmd = { 'beancount-langserver' },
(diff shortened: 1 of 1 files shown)


mjlbach:
Is there a reason this is marked draft?

lithammer:
> Is there a reason this is marked draft?

I guess it's because the pending status of the upstream ticket to make this the default behaviour. On the other hand being explicit about it doesn't hurt regardless 🤷‍♂️

kmaasrud:
> Is there a reason this is marked draft?

@mjlbach Messed up on the first PR (forgot to make a feature branch) and marked as draft to make sure everything was correct. Forgot to mark it as not a draft.

PR #1808, opened 2022-03-23
fix #1781: LspStop applies to current buffer

### LINKED ISSUE #1781: Apply `:LspStop` and `:LspRestart` to current buffer only

mg979:
As per documentation. Currently stopping all managed clients on all
buffers.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -399,7 +399,7 @@ function M.get_clients_from_cmd_args(arg)
     result[id] = vim.lsp.get_client_by_id(tonumber(id))
   end
   if vim.tbl_isempty(result) then
-    return M.get_managed_clients()
+    result = vim.lsp.buf_get_clients()
   end
   return vim.tbl_values(result)
 end


lua/lspconfig/util.lua
```diff
@@ -399,7 +399,7 @@ function M.get_clients_from_cmd_args(arg)
     result[id] = vim.lsp.get_client_by_id(tonumber(id))
   end
   if vim.tbl_isempty(result) then
-    return M.get_managed_clients()
+    result = vim.lsp.buf_get_clients()
```
rish987, line 402: I'm guessing this should be restricted to clients managed by the plugin? In case there are some linters or clients managed by other plugins that we don't want to mess with (haven't had a use for this personally but I suppose that's why we had an internal function for this).

mg979: I'm assuming here that one uses lspconfig to attach lsp clients to buffers. If one uses other clients that aren't managed by lspconfig (and I don't know why or if it can be the case), one can write wrapper functions for `LspStart` that also restart the other clients.

Alternatively, `M.get_managed_clients()` should pass the current buffer as parameter, but then clients should be managed per-buffer in that function (I don't think they are). This can be done later if someone cares. This PR is meant to fix a glaring issue that totally breaks the command as it was meant to be, so other concerns are secondary (imho).

PR #1811, opened 2022-03-24
fix(solc): update root directory pattern

pedro757:
Added package.json to root directory pattern


lithammer:
I'm not familiar with solc. I skimmed through the project briefly but couldn't really find anything suggesting package.json being relevant. Maybe you can provide some context?

PR #1813, opened 2022-03-25
Allow autostart to be a callback

will:
**Motivation:**
Unlike Solargraph (and probably most other servers), Sorbet is not a general-purpose Ruby language server. A project must have a `sorbet` directory at the root for the server to start, otherwise it will error out. It's very nice when the server autostarts on my projects that use sorbet! But it's annoying when it fails and prints error messages on other random ruby projects that do not.

So it would be nice if the server only was auto started on suitable projects. It would also be nice if checking for suitability was done only in ruby projects, and not all the time. I don’t know much about all the different language servers, but I figure that there may also be other servers that may benefit from deferring the decision to auto start based on the project. 

**This PR has two patches:**

The first adds the ability for `autostart` to be a callback in addition to the boolean it has been. The function is called with the calculated root directly. I also pass in the path section of the utils functions since it seems like it will often be needed to poke around in the root directory to see if it's a good project for the language server or not.

The second patch uses this new callback feature on the sorbet server so that it only autostarts appropriately.

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -134,11 +134,18 @@ listed in `:help vim.lsp.start_client()` with the following unique entries:
 
 - {autostart}
 
-    `bool` (default: true)
+    `bool | function(root_dir, path_utils) (default: true)`
 
     Controls if the `FileType` autocommand that launches a language server is
     created. If `false`, allows for deferring language servers until manually
     launched with `:LspStart` (|lspconfig-commands|).
+
+    If a callback function is given instead of a boolean, the `FileType`
+    autocommand is created as if the value was true. However the callback will
+    run on matching filetypes with the calculated root directory and a small
+    collection of path utilities. They can be used to determine if the
+    language server is sutible for the current project. The server will only
+    be started if this function returns `true`.
 
 - {single_file_support}
 
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -54,7 +54,7 @@ function configs.__newindex(t, config_name, config_def)
       pcall(util.on_setup, config)
     end
 
-    if config.autostart == true then
+    if config.autostart then
       local event
       local pattern
       if config.filetypes then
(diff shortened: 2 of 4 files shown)


will:
The [lint error](https://github.com/neovim/nvim-lspconfig/runs/5698733900?check_suite_focus=true) seems both incorrect and unrelated to my patch.

The [dirname checker error](https://github.com/neovim/nvim-lspconfig/runs/5698733819?check_suite_focus=true) I believe is a false positive.

lithammer:
> The [lint error](https://github.com/neovim/nvim-lspconfig/runs/5698733900?check_suite_focus=true) seems both incorrect and unrelated to my patch.

Should be fixed if you rebase.

will:
> > The [lint error](https://github.com/neovim/nvim-lspconfig/runs/5698733900?check_suite_focus=true) seems both incorrect and unrelated to my patch.
> 
> Should be fixed if you rebase.

Thanks, just rebased

mjlbach:
I don't think this is the way to go, I'll respond in more detail later so DNM

mjlbach:
I want to add support for 2, a rough implementation is here https://github.com/mjlbach/projects.nvim. Projects are registered ahead of time for directory triggers.

I don't like the semantics of the current change.

Autostart is "should I manually start the server with LspStart, or should I have LspConfig automatically try to start the server". 

A couple options (which is kinda a placeholder until I finish works on projects.nvim)
* We can add a separate, optional key that takes a callback which runs after attempting to resolve the root directory, and also takes "single_file_support"
* We overload root pattern to return true/false if the server should start on the resolved root. I think this would be messy in practice, some people want to disable single_file_support for all servers (because many servers have reduced functionality if it launches with a null root. So then we'd have to pass single_file_support into `root_dir`, which makes the function somewhat complicated.

will:
I agree 100% with 

> Autostart is "should I manually start the server with LspStart, or should I have LspConfig automatically try to start the server".

and, to me at least, this change fully keeps those semantics. It just adds the option to deferring the decision to a bit later. But it's your project and if you don’t agree that's alright I suppose.

And I figured that since everyone using this particular LSP server will have the same problem, it'd be a nice thing to solve it once for everyone. But I'm fine just solving the problem for myself with an autocommand or with projets.nvim when it's more mature. Other people if they get annoyed with the thing can probably find this issue and see the solutions and do their own manual fixes.

will:
Just in case anyone is coming from search in the future here is now how I've solved it for myself based on @davidsierradz's autocommand and (currently unreleased) neovim 0.7 lua auotcommands


```lua
lspconfig.sorbet.setup {
  cmd = { "srb", "tc", "--lsp", "--enable-all-beta-lsp-features" },
  autostart = false,
}

local au_id
au_id = vim.api.nvim_create_autocmd("BufRead,BufNewFile", {
  pattern = "/full/path/to/my/project/**/*.rb",
  callback = function()
    vim.schedule(function()
      lspconfig.sorbet.launch()
      vim.api.nvim_del_autocmd(au_id)
    end)
  end,
})
```

While it seems fine to call launch over and over again, I have the autocommand delete itself here to just avoid doing extra work all the time.

PR #1821, opened 2022-04-02
Allow for custom command arguments

zappolowski:
By default all `commands` are created with `<f-args>` which prevents proper use of e.g. `-range`. Allow for setting a customized argument list to make use of e.g. `<line1>`/`<line2>` or `<count>`.

TBD: Is using a string sufficient or should it also allow for lists?

PS: I've also removed trailing white space in the documentation ... if not desired, I can delete that commit.


mjlbach:
Sorry, I punted on this because 0.7 exposes a new user commands api via lua. Do you want to take a look/give feedback on https://github.com/neovim/nvim-lspconfig/pull/1838?

zappolowski:
No problem. When I created this PR I wasn't aware, that 0.7 is right around the corner. I will take a look at the other branch.

PR #1822, opened 2022-04-03
Do not set up nonexisting buffer

przepompownia:
It can occur after (LSP) renaming the class without opening it (e.g. on its some reference).


mjlbach:
Can you give me a minimal repro that cause this?

przepompownia:
It can be not trivial but I'll try.

przepompownia:
https://github.com/przepompownia/nvim-test-lspconfig - please look how `make nonexisting-buffer` works or simply run it. 
image

przepompownia:
It's still present on the above example.

justinmk:
on_attach is a nontrivial callback so it makes sense to check the bufnr.

przepompownia:
Thanks!

Issue #1819, opened 2022-04-02
jdtls Provides a Start Up Script


schrieveslaach:
### Language server

jdtls

### Requested feature

Since version 1.9.0 of eclipse.jdt.ls there is a [wrapper script](https://github.com/eclipse/eclipse.jdt.ls/pull/2005) that handles all the start-up configuration of the JVM. So [this code](https://github.com/neovim/nvim-lspconfig/blob/700eda91316a876757bf306fd709fde75aadb36a/lua/lspconfig/server_configurations/jdtls.lua#L97-L111) could be simplified.

There are also some package managers that install the script as a executable, for example:

- Arch Linux: https://aur.archlinux.org/packages/jdtls
- Homebrew: https://formulae.brew.sh/formula/jdtls (https://github.com/Homebrew/homebrew-core/pull/96232)

agorgl:
Current nvim-lspconfig jdtls server configuration also does not work by default, checking the lsp.log shows various `java.nio.file.AccessDeniedException` exceptions. Running the cmd shown in LspInfo by hand, I can see clearly:
```
The configuration area at '/usr/share/java/jdtls/config_linux' is not writable.  Please choose a writable location using the '-configuration' command line option.
```
Tested on Arch Linux with latest jdtls AUR package.

Using the provided wrapper script will solve these kind of problems, because instead of passing the original jdtls config_<linux|win|mac> folder in -configuration option, it will make a copy first to a user provided location:

https://github.com/eclipse/eclipse.jdt.ls#running-from-command-line-with-wrapper-script

agorgl:
Prepping a PR for this.

schrieveslaach:
@agorgl, thanks for stepping in.

agorgl:
#1833 is now merged on master, this can be closed

PR #1833, opened 2022-04-10
feat: use official jdtls wrapper script

agorgl:
Implements #1819

### DIFF
--- a/lua/lspconfig/server_configurations/jdtls.lua
+++ b/lua/lspconfig/server_configurations/jdtls.lua
@@ -4,6 +4,7 @@ local handlers = require 'vim.lsp.handlers'
 local env = {
   HOME = vim.loop.os_homedir(),
   XDG_CACHE_HOME = os.getenv 'XDG_CACHE_HOME',
+  JDTLS_JVM_ARGS = os.getenv 'JDTLS_JVM_ARGS',
 }
 
 local function get_cache_dir()
@@ -20,6 +21,15 @@ end
 
 local function get_jdtls_workspace_dir()
   return util.path.join(get_jdtls_cache_dir(), 'workspace')
+end
+
+local function get_jdtls_jvm_args()
+  local args = {}
+  for a in string.gmatch((env.JDTLS_JVM_ARGS or ''), '%S+') do
+    local arg = string.format('--jvm-arg=%s', a)
+    table.insert(args, arg)
+  end
+  return unpack(args)
 end
 
 -- TextDocument version is reported as 0, override with nil so that
@@ -86,6 +96,7 @@ return {
       get_jdtls_config_dir(),
       '-data',
       get_jdtls_workspace_dir(),
+      get_jdtls_jvm_args(),
     },
     filetypes = { 'java' },
     root_dir = function(fname)
@@ -128,6 +139,14 @@ and ensure that the `PATH` variable contains the `bin` directory of the extracte
 ```lua
   -- init.lua
   require'lspconfig'.jdtls.setup{}
+```
+
+You can also pass extra custom jvm arguments with the JDTLS_JVM_ARGS environment variable as a space separated list of arguments,
+that will be converted to multiple --jvm-arg=<param> args when passed to the jdtls script. This will allow for example tweaking
+the jvm arguments or integration with external tools like lombok:
+
+```sh
+export JDTLS_JVM_ARGS="-javaagent:$HOME/.local/share/java/lombok.jar"
 ```
 
 For automatic installation you can use the following unofficial installers/launchers under your own risk:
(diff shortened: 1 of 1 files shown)


mjlbach:
@mfussenegger do you know of any cons other than the python 3.9 dependency of using the wrapper script?

agorgl:
Fixed minor lint warning

agorgl:
Having written manually the wrappers for launching correctly jdtls for my old vim configuration in all systems (lin,win,osx) in the past, I think using the upstream solution is the way to go.

The upstream wrapper script separates the usage between the read-only shared config passed with the flag `osgi.sharedConfiguration.area` and the read-write user config passed with the `-configuration` flag by us.

agorgl:
Added also support for the `--jvm-arg` flag that the jdtls wrapper script accepts. Custom jvm arguments can be passed with the `JDTLS_JVM_ARGS` environment variable as a space separated list of arguments, that will be converted to multiple `--jvm-arg=<param>` args when passed to the jdtls script. This will allow for example tweaking the jvm arguments or integration with external tools like lombok:
```
export JDTLS_JVM_ARGS="-javaagent:$HOME/.local/share/java/lombok.jar -Xbootclasspath/a:$HOME/.local/share/java/lombok.jar"
```

lua/lspconfig/server_configurations/jdtls.lua
```diff
@@ -151,19 +135,6 @@ you can keep reading here.
 For manual installation you can download precompiled binaries from the
 [official downloads site](http://download.eclipse.org/jdtls/snapshots/?d)
```
schrieveslaach, line 136:
```suggestion
[official downloads site](http://download.eclipse.org/jdtls/snapshots/?d)
and ensure that the `PATH` variable contains the `bin` directory of the extracted archive. 
```

agorgl: Fixed!

mfussenegger:
> @mfussenegger do you know of any cons other than the python 3.9 dependency of using the wrapper script?

No real cons as far as I am aware off.

Just keep in mind that the changes here will probably break the config for all people who've been using it so far.

agorgl:
Any updates on this?

mjlbach:
I have a slight concern about introducing a python 3.9 dependency, although debian bullseye has 3.9 by default, that's the main hold-up.

agorgl:
I think that any user/dev having the whole jdk in his system will probably have python too.
Also as you said almost all mainstream distros almost always come bundled with python!

mfussenegger:
Might want to hold off until it can properly parse the java version: https://github.com/eclipse/eclipse.jdt.ls/issues/2060

agorgl:
Interesting, i'm using the latest version and it seems to be working fine

BojanStipic:
@agorgl thank you very much for this PR, I am looking forward to seeing this merged.

> Added also support for the `--jvm-arg` flag that the jdtls wrapper script accepts. Custom jvm arguments can be passed with the `JDTLS_JVM_ARGS` environment variable as a space separated list of arguments, that will be converted to multiple `--jvm-arg=<param>` args when passed to the jdtls script. This will allow for example tweaking the jvm arguments or integration with external tools like lombok:
> 
> ```
> export JDTLS_JVM_ARGS="-javaagent:$HOME/.local/share/java/lombok.jar -Xbootclasspath/a:$HOME/.local/share/java/lombok.jar"
> ```

I think it would be very useful to also include this in the documentation, to ease the setup process for newcomers as Lombok is commonly used. There is currently no way to find this out, other than digging into the source code.

schrieveslaach:
@BojanStipic, @agorgl the option `-Xbootclasspath` might not work always. On my system lombok stopped working when I used it. The [project's documentation](https://projectlombok.org/setup/ecj) states that older versions of Lombok or Java requires that option:

>  You may have to add the following VM argument, if you're using an older version of Lombok or Java: 

I assume that Java 11 does not requires it in general and jdt.ls require at least Java 11. 

So I think `-Xbootclasspath` shouldn't documented as a default option.

agorgl:
Added documentation about JDTLS_JVM_ARGS environment variable as per the above suggestions!

agorgl:
@mfussenegger My fix for jdtls java version parsing landed in master: https://github.com/eclipse/eclipse.jdt.ls/pull/2063
I think we are good to go!

agorgl:
@mfussenegger java version parsing fix is now included on latest stable builds, available both on AUR (Arch Linux) and scoop (Windows), maybe we should merge this?

mfussenegger:
I don't maintain nvim-lspconfig, not up to me.

agorgl:
@mjlbach Is there anything I can do to help merge this?

agorgl:
Fixed stylua linting error in latest force push

agorgl:
Ping @justinmk as per his request!

PR #1835, opened 2022-04-12
fix(clarity_lsp): remove .cmd suffix for Windows

williamboman:
The prebuilt executables distributed via GitHub releases (https://github.com/hirosystems/clarity-lsp/releases) are binaries - with the Windows variant being `clarity-lsp.exe`. If building from source (Rust), it will also produce a `clarity-lsp.exe` binary.

### DIFF
--- a/lua/lspconfig/server_configurations/clarity_lsp.lua
+++ b/lua/lspconfig/server_configurations/clarity_lsp.lua
@@ -1,13 +1,8 @@
 local util = require 'lspconfig.util'
 
-local bin_name = 'clarity-lsp'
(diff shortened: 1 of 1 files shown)


williamboman:
@josecanhelp 👋 - am I missing something here, was there a reason behind using `.cmd` for Windows?

josecanhelp:
@williamboman I can't remember where I got that `.cmd` from. I believe if the binaries are `.exe` on releases, then we can change that to `.exe` safely. I don't have a windows system to test though.

williamboman:
Cool :+1: I've tried this on my Windows machine and it works better

PR #1838, opened 2022-04-15
feat!: 0.7 API update

mjlbach:
* switch to lua api for autocommands
* switch to nvim_create_user_command
* move to lua plugin initialization
* rename commands -> user commands to match API and free up commands
  namespace

  This will require users to update commands -> user_commands in the
  setup {} call.


zappolowski (commented):
I think, you've missed some spots which still reference the old `.commands` field. Using these changes, I at least got the basic functionality with user defined commands to work (default commands were working OOTB).

I'm still trying to understand how e.g. ranges are passed in Lua.

mjlbach:
> I'm still trying to understand how e.g. ranges are passed in Lua.

The lua callback receives a table as a single arg, it should be args.range

zappolowski:
> > I'm still trying to understand how e.g. ranges are passed in Lua.
> 
> The lua callback receives a table as a single arg, it should be args.range

Thanks, I've figured out the issue I've had before: `opts` for `nvim_create_user_command` cannot be defined by the user, but needs to be `{ range = true }` then.

```patch
diff --git a/lua/lspconfig/configs.lua b/lua/lspconfig/configs.lua
index 245e32f..edace03 100644
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -277,7 +277,9 @@ function configs.__newindex(t, config_name, config_def)
       M.user_commands = vim.tbl_deep_extend('force', M.user_commands, client.config.user_commands)
     end
     for _, command_info in pairs(M.user_commands or {}) do
-      vim.api.nvim_create_user_command(command_info.name, command_info.command, { desc = command_info.description })
+      local opts = command_info.opts or {}
+      opts.desc = command_info.description
+      vim.api.nvim_create_user_command(command_info.name, command_info.command, opts)
       M.commands_created = true
     end
   end
```

lua/lspconfig/configs.lua
```diff
@@ -37,13 +37,13 @@ function configs.__newindex(t, config_name, config_def)
...
     }
-    if config.commands then
-      for k, v in pairs(config.commands) do
+    if config.user_commands then
+      for _, v in pairs(config.user_commands) do
         validate {
-          ['command.name'] = { k, 's' },
-          ['command.fn'] = { v[1], 'f' },
+          ['command.name'] = { v.name, 's' },
+          ['command.fn'] = { v.command, 'f' },
```
zappolowski, line 46: For `config_def.user_commands` this is called `comand.command` ... maybe using a term avoids confusion.

mjlbach: I feel like I didn't push a commit as a I did a search/replace renaming for these, I'll fix it

mjlbach:
Ok, I think we should be good now? If you want tot take another look. I made the defined user_commands match the exact function signature of create_user_command

lua/lspconfig/configs.lua
```diff
@@ -37,13 +37,13 @@ function configs.__newindex(t, config_name, config_def)
       filetypes = { config.filetype, 't', true },
       on_new_config = { config.on_new_config, 'f', true },
       on_attach = { config.on_attach, 'f', true },
-      commands = { config.commands, 't', true },
+      user_commands = { config.user_commands, 't', true },
     }
     if config.commands then
-      for k, v in pairs(config.commands) do
+      for _, v in pairs(config.commands) do
```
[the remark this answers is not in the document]
zappolowski: The last patch changed this one back to the original one, which now fails as the check above checks another field.

lua/lspconfig/configs.lua
```diff
@@ -272,18 +273,17 @@ function configs.__newindex(t, config_name, config_def)
     if client.config._on_attach then
       client.config._on_attach(client, bufnr)
     end
-    if client.config.commands and not vim.tbl_isempty(client.config.commands) then
-      M.commands = vim.tbl_deep_extend('force', M.commands, client.config.commands)
+    if client.config.user_commands and not vim.tbl_isempty(client.config._usercommands) then
```
zappolowski, line 276:
```suggestion
    if client.config.user_commands and not vim.tbl_isempty(client.config.user_commands) then
```

zappolowski (commented):
Just these two typos ... besides that, it works (for my use case) :+1:

lua/lspconfig/configs.lua
```diff
@@ -37,13 +37,13 @@ function configs.__newindex(t, config_name, config_def)
       filetypes = { config.filetype, 't', true },
       on_new_config = { config.on_new_config, 'f', true },
       on_attach = { config.on_attach, 'f', true },
-      commands = { config.commands, 't', true },
+      user_commands = { config.user_commands, 't', true },
     }
-    if config.commands then
-      for k, v in pairs(config.commands) do
+    if config.user_commands then
+      for _, v in pairs(config.commands) do
```
zappolowski, line 43: The last patch reverted this to `config.commands`, which should be `config.user_commands` ... and triggers errors as the check above doesn't work as intended.

mjlbach:
Whoops, bad force push, try now.

lua/lspconfig/configs.lua
```diff
@@ -272,18 +273,17 @@ function configs.__newindex(t, config_name, config_def)
     if client.config._on_attach then
       client.config._on_attach(client, bufnr)
     end
-    if client.config.commands and not vim.tbl_isempty(client.config.commands) then
-      M.commands = vim.tbl_deep_extend('force', M.commands, client.config.commands)
+    if client.config.user_commands and not vim.tbl_isempty(client.config._user_commands) then
```
zappolowski, line 276:
```suggestion
    if client.config.user_commands and not vim.tbl_isempty(client.config.user_commands) then
```

zappolowski:
Thanks for your efforts. :+1:

mjlbach:
Thanks for catching my typos!

I'm going to give this a week to stew/people to read my warning message about breaking 0.6 compatibility before merging. Do you prefer the new interface for user_commands? I think it's more intuitive since the list maps directly onto the arguments passed to nvim_create_user_command now. Alternatively I could lift the name so that the table of user_commands is keyed by the command name instead of being a list, but IMO having the table include the *exact* arguments is cleaner. WDYT?

zappolowski:
I also prefer to have the API resemble `nvim_create_user_command` as close as possible (less cognitive load in total).

While playing around with both approaches I ran into an issue. Using the list like form of tables means, that `vim.tbl_deep_extend` as used [here](https://github.com/neovim/nvim-lspconfig/pull/1838/files#diff-0e1bd346624fa540fa153e867fbc2036ced5c83d5cd7d79d9308df1328b7c879R277) cannot be used anymore (it's intended for map-like tables). The results in vanishing default defined commands when user defined commands are used as well.

mjlbach:
That's a good point, we can either key the table with the name of the command as before, or merge the lists manually without extending the original table

zappolowski:
Personally, I would stick to the new API with lists (as written before) and thus just merge the lists.

One thing (and that's actually how I found the issue) to consider then, is how to handle both intentional (e.g. when one wants to alter an existing command without having to come up with a new name) and unintentional (implementing a command which is then later provided by this plugin) overwriting of existing commands.

My proposal would be to use the default defined commands as the base list and add user defined commands in the end. This way both sets of commands should still work and intentional overwriting will work as well. I would guess that unintentional overwriting is rare enough to be dealt later.

mjlbach:
We might even just do away with the user_command entrypoint to setup{}... there's no reason that a user can't just use on_init or on_attach to map the command themselves now that we have the lua api.

mjlbach:
I decided to just have user_commands overwrite then trying to merge so users can disable the default command mappings, also I think if people *really* want to map their own commands they can just call the create_user_commands api directly. Still need to fix docgen and then this is g2g

ranjithshegde:
Is anyone working on this since the departure of our beloved maintainer mjlbach?
I have been riding this branch since its creation and its been working great! 
Willing to help out with tests or any other way necessary

mjlbach:
I'm still alive fwiw :) I just wanted to set clear expectations on my response time/responsibilities (indefinite, and none) while I am on break.

The main reason I didn't merge this yet is because I was giving time for 0.7 to become the new standard. I know there is going to be some blowback dropping pre 0.7 support, but I wanted to do these cleanups for awhile.

justinmk:
https://github.com/neovim/nvim-lspconfig/pull/1984

Issue #1823, opened 2022-04-04
Cannot setup custom server configuration for the LSP server `astro-ls`


smithbm2316:
### Description

I'm trying to run the [Astro language server](https://github.com/withastro/language-tools/tree/main/packages/language-server) with `.astro` files on my local machine. I have used it in the past with success (it's been a few months), and all of the sudden the custom server configuration that I have does not seem to be working. Would appreciate any help that y'all can offer 🙂. Thank you for the amazing plugin and your help!

Here is the output of `:LspInfo` when I'm *not* using the `minimal_init.lua` setup:
![Screen Shot 2022-04-04 at 3 03 48 PM](https://user-images.githubusercontent.com/37027883/161639646-a00a8f63-99af-40f5-935e-751951db74f7.png)

No errors occur with my personal setup, only the errors happen when running the `minimal_init.lua` setup. But even when running `:LspStop` and `:LspStart` to try and re-run the server, it still won't attach to the buffer. I've got a `ftdetect/astro.vim` file that has the following command in it: `au BufRead,BufNewFile *.astro setfiletype astro` (I have also tried `au BufRead, BufNewFile *.astro set filetype=astro` to no avail) so that Neovim is aware of the `astro` filetype.

### Neovim version

NVIM v0.7.0-dev+1392-gdaa8ac051
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by `redacted username and computer name`

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-daa8ac0/share/nvim"

Run :checkhealth for more info

### Nvim-lspconfig version

3d1baa811b351078e5711be1a1158e33b074be9e

### Operating system and version

macOS 11.6.2

### Affected language servers

astro-ls (custom server)

### Steps to reproduce

1. `nvim -nu minimal_init.lua`
2. `:e src/pages/index.astro`

### Actual behavior

Error received after running `nvim -nu minimal_init.lua` followed by `:e src/pages.index.astro`
```
Error executing vim.schedule lua callback: ...r/neovim/HEAD-daa8ac0/share/nvim/runtime/lua/vim/lsp.lua:955: RPC[Error] code_name = InternalError, message = "Request initializ
e failed with message: Cannot read property 'astro' of undefined"
stack traceback:
        [C]: in function 'assert'
[...]

PR #1841, opened 2022-04-17
feat: add astro-ls support

piero-vic:
Based on the discussion on #1823.

### DIFF
--- a/lua/lspconfig/server_configurations/astro.lua
+++ b/lua/lspconfig/server_configurations/astro.lua
@@ -1,0 +1,32 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'astro-ls'
+local cmd = { bin_name, '--stdio' }
+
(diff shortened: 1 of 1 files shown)


mjlbach:
Just to verify we need the empty settings/init options tables? Also is there a reason this is marked draft?

piero-vic:
I've doing some testing. The `settings` table is not required. but we do need the `init_options` and `configuration` tables. Otherwie we get the error described on #1823. It will look something like this.
```lua
  default_config = {
    cmd = cmd,
    filetypes = { 'astro' },
    root_dir = util.root_pattern('package.json', 'tsconfig.json', 'jsconfig.json', '.git'),
    init_options = {
      configuration = {},
    },
  },
```

Regarding the draft, I was just trying to follow the contribution guide. I'll change it.

mjlbach:
Cool, you can drop settings then and repush and I'll merge if the tests pass.

PR #1842, opened 2022-04-17
feat: add steep support

okuramasafumi:
Steep is a static type checker for Ruby.

### DIFF
--- a/lua/lspconfig/server_configurations/steep.lua
+++ b/lua/lspconfig/server_configurations/steep.lua
@@ -1,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'steep', 'langserver' },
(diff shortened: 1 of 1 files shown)


mjlbach:
A configuration should return a table, not register itself into configs. Where did you see this style? We deprecated it awhile ago.

okuramasafumi:
@mjlbach I copied it from https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md

mjlbach:
Thanks for the catch, fixed the contributing documentation in https://github.com/neovim/nvim-lspconfig/commit/31be570201581dd23718722d51021dc55255454c

okuramasafumi:
@mjlbach Thanks, so I force-pushed another commit which mimics `typeprof` config.

Issue #1827, opened 2022-04-09
Provide abstraction for generic default root dir


tpope:
### Requested feature

TL;DR: I am looking to improve this:

> Most of the time, this is a `.git` folder, but each server defines the root config in the lua file.

This is a good 80% solution.  In the current SCM monoculture, it may even be a 95% solution.  But it has several limitations:

* It assumes you're using Git as your SCM.
* It assumes you're using an SCM at all.
* It assumes the workspace root is the same as the repository root.

The problem is that if you bump into one of these situations, you're screwed.  Since the `.git` detection is *inside* each LSP config, you need to assess each LSP you use on a case-by-case basis, and more often than not will need to yank out its internal `root_dir` and manually tweak it.

My proposed solution would be something like the following:

* Add a new Boolean config key which specifies whether a generic fallback is desired.
* Alter `launch()` to respect this config key and call `find_git_ancestor()` if no other root was found.  This could later be tweaked (e.g., to add a non-SCM based default) or made configurable by some mechanism.
* Change all existing adapters using either `find_git_ancestor()` or `root_pattern(..., '.git')` to instead set this Boolean key.

The bulk of the work is in that last step, but that's easy enough to divide and conquer, and it can be done incrementally.  As for the first part, I could certainly take a stab at it if there's interest.  I imagine someone more familiar with the project than I could make quick work of it.

mjlbach:
> The problem is that if you bump into one of these situations, you're screwed. Since the .git detection is inside each LSP config, you need to assess each LSP you use on a case-by-case basis, and more often than not will need to yank out its internal root_dir and manually tweak it.

That's true, you can compose root directory functions with the original default, but it's a bit verbose:
```
require('lspconfig').pyright.setup {
root_dir = function(fname) 
    return require('lspconfig.server_configurations.pyright').default_config.root_dir(fname) or my_fallback_func(fname)
end
}
```

```
My proposed solution would be something like the following:

* Add a new Boolean config key which specifies whether a generic fallback is desired.
* Alter launch() to respect this config key and call find_git_ancestor() if no other root was found. This could later be tweaked (e.g., to add a non-SCM based default) or made configurable by some mechanism.
* Change all existing adapters using either find_git_ancestor() or root_pattern(..., '.git') to instead set this Boolean key.
```

[...]

PR #1847, opened 2022-04-20
Add support for Mercurial repositories

0x8000-0000:
Implementation based on the similar Git utility.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -347,6 +347,14 @@ function M.find_git_ancestor(startpath)
     end
   end)
 end
+function M.find_mercurial_ancestor(startpath)
+  return M.search_ancestors(startpath, function(path)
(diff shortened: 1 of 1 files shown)


mjlbach:
I think I would lump this in with https://github.com/neovim/nvim-lspconfig/issues/1827, would you prefer to take a crack at that? It should be pretty easy.

0x8000-0000:
> I think I would lump this in with #1827, would you prefer to take a crack at that? It should be pretty easy.

I suppose #1827 would obsolete this change but I'd rather get this in as-is since it is quite fool-proof and small.

Also I have 0 Lua experience, I just got this working by copy&paste. Somebody more experienced can get it done sooner and much cleaner than me learning Lua on the job.

justinmk:
The current family of functions is very verbose so generalization looks like an improvement. But don't need to block this meanwhile.

PR #1850, opened 2022-04-22
fix(ccls): remove .git from root_dir

abgeana:
CCLS does not use .git to identify the root of a project. In addition, it
causes issues with projects that are based on the repo tool (e.g. AOSP),
which generates `.git` symlinks in many places in the source tree. In
such cases, CCLS incorrectly identifies subdirectories as the root.

### DIFF
--- a/lua/lspconfig/server_configurations/ccls.lua
+++ b/lua/lspconfig/server_configurations/ccls.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'ccls' },
     filetypes = { 'c', 'cpp', 'objc', 'objcpp' },
-    root_dir = util.root_pattern('compile_commands.json', '.ccls', '.git'),
+    root_dir = util.root_pattern('compile_commands.json', '.ccls'),
     offset_encoding = 'utf-32',
     -- ccls does not support sending a null root directory
     single_file_support = false,
@@ -37,7 +37,7 @@ lspconfig.ccls.setup {
 
 ]],
     default_config = {
-      root_dir = [[root_pattern("compile_commands.json", ".ccls", ".git")]],
+      root_dir = [[root_pattern("compile_commands.json", ".ccls")]],
     },
   },
 }


mjlbach:
I don't think this should have been merged, it is consistent with the rest of the root patterns that we fallback to the git directory.

abgeana:
> I don't think this should have been merged, it is consistent with the rest of the root patterns that we fallback to the git directory.

I am not sure I understand what you mean, but I checked for example in [clangd.lua](https://github.com/neovim/nvim-lspconfig/blob/db8357abc0c97a5110e7adb6decfe2c13ecd43b4/lua/lspconfig/server_configurations/clangd.lua#L24-L31) (which vaguely does similar things as ccls) and it does not include `.git`.

> I haven' really used ccls myself. But does it even work without a compile_commands.json file?

Yes, one can use a custom `.ccls` file, the contents of which are used by ccls to build a cache ([more details in the ccls wiki](https://github.com/MaskRay/ccls/wiki/Project-Setup#how-ccls-obtains-sources-and-compilation-commands)).

> It seems like ccls has section in its FAQ regarding project root detection. Seems like .ccls-root should be added.
> 
> https://github.com/MaskRay/ccls/wiki/FAQ#project-root-detection

That is specifically related to the `emacs-ccls` client and does not apply to this case. To be sure, I tested this in my case (without emacs :P ) and it had no effect.

To provide some more proof that the `.git` entry in `root_dir` is the cause of the issue, I am providing snippets from my `~/.cache/nvim/lsp.log`.

With the `.git` entry present, I see the following message:
```
[...] initialize.cc:275 I initialize in directory /mnt/work/trusty/trusty/kernel with uri file:///mnt/work/trusty/trusty/kernel\n
```
where indeed there is a `.git` folder in `/mnt/work/trusty/trusty/kernel`, but not the intended project root. In this scenario actually ccls incorectly fails to identify the `.ccls-cache` folder in the real project root and is unable to find various symbols.

However, without the `.git` entry present, I see the following message:
```
[...] initialize.cc:275 I initialize in directory /mnt/work/trusty with uri file:///mnt/work/trusty\n"
```
where `/mnt/work/trusty` is the intended project root and ccls is able to find the correct symbols.

Based on the above, I still believe that the commit in this pull request was correct. Would it be possible to revert the revert from https://github.com/neovim/nvim-lspconfig/pull/1857?

mjlbach:
>  and it does not include .git.
[Yes it does](https://github.com/neovim/nvim-lspconfig/blob/db8357abc0c97a5110e7adb6decfe2c13ecd43b4/lua/lspconfig/server_configurations/clangd.lua#L51)

Based on the above, I still believe that the commit in this pull request was correct. Would it be possible to revert the revert from https://github.com/neovim/nvim-lspconfig/pull/1857?

No, but you are free to override the pattern in your own configuration.

PR #1860, opened 2022-04-25
fix(ccls): .git as fallback when detecting project root

abgeana:
Use `.git` as fallback for detecting the project root in ccls, while preferring folders with `compile_commands.json` or `.ccls`. Behavior is identical to that from `clangd.lua`.


mjlbach:
LGTM! Thanks, will merge when tests are green.

PR #1862, opened 2022-04-26
fix(vls): add cmd

williamboman:
I think the comment regarding lspconfig not making assumptions about path is moot for `vls`, I feel like it'd make more sense to actually provide a `cmd` for vls. There are servers where I think this does make sense, for example `omnisharp, bicep, rescriptls, elixirls, perlnavigator` (found by grepping `assumptions about your path`).

### DIFF
--- a/lua/lspconfig/server_configurations/vls.lua
+++ b/lua/lspconfig/server_configurations/vls.lua
@@ -2,6 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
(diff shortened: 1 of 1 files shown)


mjlbach:
so long as they provide a binary, you can go ahead and update all of the configurations.

williamboman:
> so long as they provide a binary, you can go ahead and update all of the configurations.

This one and zeta_note were the only ones I could find so far. Should I merge that branch in here instead of having them separate?

PR #1865, opened 2022-04-26
feat: allow returning new config table in on_setup hook

williamboman:
This allows on_setup hooks to not have to rely on modifying table
instances in-place.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -51,7 +51,10 @@ function configs.__newindex(t, config_name, config_def)
     config = tbl_extend('keep', config, default_config)
 
     if util.on_setup then
-      pcall(util.on_setup, config)
+      local ok, new_config = pcall(util.on_setup, config)
+      if ok and new_config then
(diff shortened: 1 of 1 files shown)


williamboman:
This is mostly just to get a place to discuss the inner workings of this hook. I figured it'd be nice for the on_setup hook to be able to return a new table instance, to allow registered hooks to not have to modify the received config tables directly. This is to avoid situations where mutations to the config table carries over to configs intended for other servers, for example in the following scenario:

```lua
local opts = { on_attach = on_attach }
lspconfig.tsserver.setup(opts)
lspconfig.sumneko_lua.setup(opts)
```

I might be missing something though, but wdyt?

williamboman:
Wait I think I spoke too soon. The `tbl_extend()` call inside the `.setup(config)` makes sure to avoid the problem of table references. I still think it'd be nice to be able return a new table in the hook and have lspconfig pick that up, this would make merging different tables easier to implement (just use the builtin `vim.tbl_extend` methods)

PR #1866, opened 2022-04-27
docs(esbonio): fix URL to option docs

yhu266:
The option docs have been moved to a new page.


yhu266:
style-lint has failed on "zeta_note". I think we need a different PR to fix that.

PR #1871, opened 2022-04-27
feat: add m68k support

grahambates:
Add server configuration for Motorola 68000 family assembly.


grahambates:
The test failure on Mac appears to be spurious - 429 error downloading msgpack so I'm guessing it hit a rate limit.

PR #1834, opened 2022-04-12
fix(beancount): remove --stdio arg

williamboman:
Seems like the `--stdio` arg is not needed:

```
[ERROR][2022-04-12 18:39:14] .../vim/lsp/rpc.lua:420	"rpc"	"beancount-langserver"	"stderr"	"error: Found argument '--stdio' which wasn't expected, or isn't valid in this context\n\n\tIf you tried to supply `--stdio` as a value rather than a flag, use `-- --stdio`\n\nUSAGE:\n    beancount-langserver\n\nFor more information try --help\n"
```

https://github.com/polarmutex/beancount-language-server#installation

PR #1879, opened 2022-05-01
fix(beancount): fix beancount start command and config

hateonion:
The start up command and default config has changed a lot since the owner has refactored the code using rust. This pr will cover the change below.

Start up command change, should be beancount-language-server now.
Some config is not used anymore. And the only config used is journal_file(Previously it's camel case).

### DIFF
--- a/lua/lspconfig/server_configurations/beancount.lua
+++ b/lua/lspconfig/server_configurations/beancount.lua
@@ -7,7 +7,7 @@ return {
     root_dir = util.find_git_ancestor,
     init_options = {
       -- this is the path to the beancout journal file
-      journel_file = '',
+      journalFile = '',
     },
   },
   docs = {
@@ -17,7 +17,7 @@ https://github.com/polarmutex/beancount-language-server#installation
 See https://github.com/polarmutex/beancount-language-server#configuration for configuration options
 ]],
     default_config = {
-      root_dir = [[root_pattern("elm.json")]],
+      root_dir = [[root_pattern(".git")]]
     },
   },
 }
--- a/lua/lspconfig/server_configurations/beancount.lua
+++ b/lua/lspconfig/server_configurations/beancount.lua
@@ -18,7 +18,7 @@ https://github.com/polarmutex/beancount-language-server#installation
 See https://github.com/polarmutex/beancount-language-server#configuration for configuration options
 ]],
     default_config = {
-      root_dir = [[root_pattern(".git")]]
+      root_dir = [[root_pattern(".git")]],
     },
   },
 }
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Could you update the `root_dir` under `docs` 🙏? I'll close my #1834 in favor of this one.

```diff
-      root_dir = [[root_pattern("elm.json")]],
+      root_dir = [[root_pattern(".git")]],
```

lua/lspconfig/server_configurations/beancount.lua
```diff
@@ -2,15 +2,12 @@ local util = require 'lspconfig.util'
...
+    cmd = { 'beancount-language-server', '--stdio' },
+    filetypes = { 'beancount', 'bean' },
     root_dir = util.find_git_ancestor,
-    single_file_support = true,
     init_options = {
       -- this is the path to the beancout journal file
-      journalFile = '',
-      -- this is the path to the python binary with beancount installed
-      pythonPath = 'python3',
+      journel_file = '',
```
williamboman, line 10: This was updated in the Rust impl to be camelcased:

```suggestion
      journalFile = '',
```

hateonion: Done!

hateonion:
> Could you update the `root_dir` under `docs` 🙏? I'll close my #1834 in favor of this one.
> 
> ```diff
> -      root_dir = [[root_pattern("elm.json")]],
> +      root_dir = [[root_pattern(".git")]],
> ```

Done, thanks for this!

hateonion:
@williamboman would u mind help merge this pr? It hanged for a while.

lua/lspconfig/server_configurations/beancount.lua
```diff
@@ -2,15 +2,12 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    cmd = { 'beancount-langserver', '--stdio' },
-    filetypes = { 'beancount' },
+    cmd = { 'beancount-language-server', '--stdio' },
+    filetypes = { 'beancount', 'bean' },
     root_dir = util.find_git_ancestor,
-    single_file_support = true,
```
lithammer, line 8: So this is no longer supported...?

hateonion: Yup. https://github.com/polarmutex/beancount-language-server as long as I can read, there is no config named like this.

lithammer: `single_file_support` isn't an attribute of the server. It's nvim-lspconfig's way of setting up a pseudo workspace root when it can't find an actual workspace root (i.e. when `root_dir` resolves to `nil`).

So unless beancount _only_ works inside a git repository, this should be added back.

PR #1881, opened 2022-05-02
fix(lspinfo): respect custom PATH when checking whether cmd is executable

williamboman:
Super hacky, but this was the easiest, least intrusive, improvement I could think of.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -124,6 +124,7 @@ function configs.__newindex(t, config_name, config_def)
     M.filetypes = config.filetypes
     M.handlers = config.handlers
     M.cmd = config.cmd
+    M.cmd_env = config.cmd_env
     M.autostart = config.autostart
 
     -- In the case of a reload, close existing things.
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -30,12 +30,28 @@ local function remove_newlines(cmd)
   return cmd
 end
 
+local function is_cmd_executable(config)
+  local cmd = config.cmd[1]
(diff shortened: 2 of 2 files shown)


lithammer:
I assume this has to do with nvim-lsp-installer. But it would be nice with some context.

williamboman:
> I assume this has to do with nvim-lsp-installer. But it would be nice with some context.

Yeah I'd wager >99% of the cases where servers are set up with a modified `PATH` would be via nvim-lsp-installer. The motivation behind this is to improve the troubleshooting experience when using nvim-lsp-installer. This error will be displayed all of the time (unless the user has the server installed on `PATH`) should a server fail to start for some reason, even if the LSP client was able to locate the executable. It has led to people (rightfully) assuming that something is broken with the server installation via nvim-lsp-installer, when that is not the case.

I think it'd be nice if the client interface were to expose any errors it encounters, currently I believe it always returns `nil` if something goes wrong, for example:
- https://github.com/williamboman/neovim/blob/fa35548343afe09c969574e58a50a31b8d14c57b/runtime/lua/vim/lsp.lua#L865-L864
- https://github.com/williamboman/neovim/blob/fa35548343afe09c969574e58a50a31b8d14c57b/runtime/lua/vim/lsp/rpc.lua#L336

williamboman:
I wouldn't mind if this is declined btw, it's really hacky. Other terrible ideas I had include:

- actually attempt to spawn (via libuv, same as rpc client) the server process and immediately kill it
- spawning an ephemeral headless nvim instance with a modified path and capture its output (`nvim --headless -c "echo executable(%q)"`)

williamboman:
I'll try to hack this into lsp installer instead

williamboman:
For reference: https://github.com/williamboman/nvim-lsp-installer/pull/691

Issue #1880, opened 2022-05-02
Improve server configuration documentation


gerazov:
### Language server

pylsp

### Requested feature

It took me while to figure out how to enable disable plugins in the `pylsp` server. One thing I don't find very useful is the McCabe complexity check. Finally I got it working via the `lspconfig` setup function using:
```lua
require('lspconfig').pylsp.setup {
  on_attach = on_attach,
  flags = {
    -- This will be the default in neovim 0.7+
    debounce_text_changes = 150,
  },
  settings = {
    pylsp = {
      plugins = {
        -- pylint = { enabled = true },
        mccabe = { enabled = false }
      }
    }
  }
}
```

This might be server specific, but it would be nice to have this in the documentation to make it easier for people to find and tweak their server settings. I found these listed via `:LspInstallInfo` from [williamboman/nvim-lsp-installer](https://github.com/williamboman/nvim-lsp-installer).

![image](https://user-images.githubusercontent.com/15214418/166215810-f15af383-9fac-466a-ab52-e4f364126267.png)

justinmk:
There is a huge amount of docs in this repo. Check carefully.

If there is a specific tweak to those docs needed, send a PR.

PR #1885, opened 2022-05-03
Added plugin enabling/disabling info for pylsp

gerazov:
It's convenient to have this info in the configurations docs.

Fixes https://github.com/neovim/nvim-lspconfig/issues/1880


tombh:
I would have found this useful too. I can try submitting this if this has fallen off the radar?

gerazov:
:+1: it's on my todo list :sweat_smile:

gerazov:
Ok done here #1920

PR #1893, opened 2022-05-06
fix(glint): detect project dir inside `on_new_config` hook rather than at module import

williamboman:
This changes the configuration to detect the project's `node_modules` dir every time lspconfig detects a new workspace directory (so in short, allows the server to attach to multiple projects in one session).

Not sure if setting the `PATH` like this is desirable. I did it to get a normalized `cmd`, I think it makes it easier for users to customize the cmd should they want to.

Also updates the root_dir pattern according to https://github.com/typed-ember/glint/blob/main/packages/config/README.md#config-specification


muziejus:
Thanks, @williamboman! This is a much better solution!

PR #1898, opened 2022-05-11
Add dotnet version of omnisharp

sassanh:
This is to add dotnet 6 version of omnisharp to address this issue: https://github.com/williamboman/nvim-lsp-installer/issues/479


justinmk:
why does it need a separate config? This repo is a collection of best-effort configs. Not meant to cover every random setup. I think it's reasonable to require dotnet6. People can always copy the old config if necessary.

sassanh:
Good point 👍I wasn't aware it is not intended to be read by machine.
We need to handle it in vim-lsp-installer then.

PR #1904, opened 2022-05-14
fix(fortls): update server and default cmd args

gnikit:
`settings` and `init_options` are not currently supported in fortls, so the arguments to the server have to be passed via the `cmd` option or through our own config files.


justinmk:
Thanks for the clear explanation as well as updating the docs. 

Nit: in the future please include the explanation in the commit message :)

LGTM if CI passes

Issue #1907, opened 2022-05-14
Windows: tsserver only spawn empty cmd and does not attach


Chanakan55991:
### Description

when I open a typescript file, the cmd window pop up but it doesn't seem to attach to the neovim buffer (checked by LspInfo), If I press enter on it 2 times it will error with missing header (might not be related but just in case). 

### Neovim version

NVIM v0.8.0-dev+203-g9e4286294
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

9ff2a06cebd4c8c3af5259d713959ab310125bec

### Operating system and version

Windows 11 21H2

### Affected language servers

tsserver, tailwindcss

### Steps to reproduce

1. open javascript/typescript file on neovim with the minimal config `nvim -u minimal.lua`

### Actual behavior

the cmd of the language server will pop up but it doesn't actually seem to do anything, it does not attach to the current buffer.

### Expected behavior

for tsserver/tailwindcss to attach to the buffer and work properly

### Minimal config

```Lua
local on_windows = vim.loop.os_uname().version:match 'Windows'

local function join_paths(...)
  local path_sep = on_windows and '\\' or '/'
  local result = table.concat({ ... }, path_sep)
  return result
end

vim.cmd [[set runtimepath=$VIMRUNTIME]]

local temp_dir = vim.loop.os_getenv 'TEMP' or '/tmp'

vim.cmd('set packpath=' .. join_paths(temp_dir, 'nvim', 'site'))

local package_root = join_paths(temp_dir, 'nvim', 'site', 'pack')
local install_path = join_paths(package_root, 'packer', 'start', 'packer.nvim')
local compile_path = join_paths(install_path, 'plugin', 'packer_compiled.lua')

local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      'neovim/nvim-lspconfig',
    },
    config = {
      package_root = package_root,
      compile_path = compile_path,
    },
  }
end

_G.load_config = function()
  vim.lsp.set_log_level 'trace'
  if vim.fn.has 'nvim-0.5.1' == 1 then
    require('vim.lsp.log').set_format_func(vim.inspect)
  end
  local nvim_lsp = require 'lspconfig'
  local on_attach = function(_, bufnr)
    local function buf_set_keymap(...)
      vim.api.nvim_buf_set_keymap(bufnr, ...)
    end
    local function buf_set_option(...)
[...]

PR #1913, opened 2022-05-18
fix: tsserver detached false in windows

tamago324:
It seemed that I needed to set detached to false on Windows, as in https://github.com/neovim/nvim-lspconfig/issues/1907.
I think I will set detached to false for all other servers as well, but I would appreciate your input.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -38,6 +38,7 @@ function configs.__newindex(t, config_name, config_def)
       on_new_config = { config.on_new_config, 'f', true },
       on_attach = { config.on_attach, 'f', true },
       commands = { config.commands, 't', true },
+      detached = { config.detached, 'b', true },
     }
     if config.commands then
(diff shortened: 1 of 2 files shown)


justinmk:
@gpanders does this look like the right approach?

ref https://github.com/neovim/neovim/pull/18477

gpanders:
> @gpanders does this look like the right approach?
> 
> ref [neovim/neovim#18477](https://github.com/neovim/neovim/pull/18477)

It looks like it's only necessary for Windows. I do not think we should change the default for non-Windows platforms however. So some kind of "is Windows" check should be used.

gpanders:
@mfussenegger mentioned an issue with Windows when detaching processes in nvim-lint too. So maybe we should make the default value of `detached` system dependent.

gpanders:
https://github.com/neovim/neovim/pull/18703

tamago324:
Thanks!

PR #1915, opened 2022-05-19
docs(sumneko_lua): fix usage of runtime_path

michaeladler:
There are two ways for sumneko to search for files, namely:

1. Lua.runtime.path

When using `require`, how to find the file based on the input name.
Setting this config to `?/init.lua` means that when you enter `require
'myfile'`, `${workspace}/myfile/init.lua` will be searched from the
loaded files.  if `runtime.pathStrict` is `false`,
`${workspace}/**/myfile/init.lua` will also be searched.  If you want to
load files outside the workspace, you need to set
`Lua.workspace.library` first.

2. Lua.workspace.library

In addition to the current workspace, which directories will load files
from. The files in these directories will be treated as externally
provided code libraries, and some features (such as renaming fields)
will not modify these files.

The crucial point is that `Lua.runtime.path` only applies to
the *current* workspace. Thus it makes no sense to add any absolute
directories here. Absolute directories must be added to
workspace.library, which is already the case. The default value provided
by sumneko is what you typically would expect, so we can just stick to it.

References:

- https://github.com/sumneko/lua-language-server/blob/076dd3e5c4e03f9cef0c5757dfa09a010c0ec6bf/locale/en-us/setting.lua#L5-L13
- https://github.com/sumneko/lua-language-server/blob/e62d964ff57cc0b37eb908315f9afe3ce6a213d7/script/config/config.lua#L151

### DIFF
--- a/lua/lspconfig/server_configurations/sumneko_lua.lua
+++ b/lua/lspconfig/server_configurations/sumneko_lua.lua
@@ -36,18 +36,12 @@ initial requests (completion, location) upon starting as well as time to first d
 Completion results will include a workspace indexing progress message until the server has finished indexing.
 
 ```lua
-local runtime_path = vim.split(package.path, ';')
-table.insert(runtime_path, "lua/?.lua")
-table.insert(runtime_path, "lua/?/init.lua")
-
 require'lspconfig'.sumneko_lua.setup {
   settings = {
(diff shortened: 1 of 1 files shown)


justinmk:
Thank you for the thorough references!

PR #1917, opened 2022-05-22
feat: add :LspLog command

williamboman:
I haven't seen any previous discussions about a command like this, so
figured I'd start one by opening this PR.

From having done user support in nvim-lsp-installer for a while now, I
feel like this log file is not being reviewed by users to the extent it
should be. Although it is mentioned in the README & `:h lspconfig`, as
well as being visible in the `:LspInfo` window, I feel like a large
portion of users don't consider reviewing it while troubleshooting
issues.

I figured by providing a user command to open the log would help make it
more accessible & discoverable.

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -21,6 +21,13 @@ function M._root._setup()
       '-nargs=0',
       description = '`:LspInfo` Displays attached, active, and configured language servers',
     },
+    LspLog = {
+      function()
+        vim.cmd(string.format('tabnew %s', vim.lsp.get_log_path()))
+      end,
+      '-nargs=0',
+      description = "`:LspLog` Opens neovim's LSP client log in a new tab window.",
+    },
     LspStart = {
       function(server_name)
         if server_name then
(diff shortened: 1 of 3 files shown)


lua/lspconfig.lua
```diff
@@ -21,6 +21,13 @@ function M._root._setup()
       '-nargs=0',
       description = '`:LspInfo` Displays attached, active, and configured language servers',
     },
+    LspLog = {
+      function()
+        vim.cmd(string.format('tabnew %s', vim.lsp.get_log_path()))
+      end,
+      '-nargs=0',
+      description = "`:LspLog` Opens neovim's LSP client log in a new tab window.",
```
justinmk, line 29:
```suggestion
      description = "`:LspLog` Opens the Nvim LSP client log.",
```

justinmk:
In general we need to stop adding features to nvim-lspconfig (it should only be a collection of configs).

But I've merged this for now because

1. looks like there was a typo in the old doc
2. we are planning a more generic logging story in Nvim but LSP logs will probably remain separate from the main logs because (1) that's idiomatic for plugins, (2) the sheer volume of the log messages

PR #1919, opened 2022-05-23
feat(solidity): add another solidity lsp

qiuxiang:


justinmk:
Thanks but I don't see an explanation of the benefits on the project's readme.

nvim-lspconfig is a collection of quickstart configs. It is not a collection of every possible server that may exist. 

If there is a LS that is generally a _better_ choice for _most_ users, we will switch to that. But there is usually no reason to add different configs, which users then need to choose amongst.

PR #1920, opened 2022-05-23
Updated PyLSP lua source

gerazov:
It's convenient to have this info in the configurations docs.

Fixes https://github.com/neovim/nvim-lspconfig/issues/1880

### DIFF
--- a/lua/lspconfig/server_configurations/pylsp.lua
+++ b/lua/lspconfig/server_configurations/pylsp.lua
@@ -26,6 +26,30 @@ The language server can be installed via `pipx install 'python-lsp-server[all]'`
 Further instructions can be found in the [project's README](https://github.com/python-lsp/python-lsp-server).
 
 Note: This is a community fork of `pyls`.
+
+**Enabling/disabling plugins:**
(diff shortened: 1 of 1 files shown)


lithammer:
Wouldn't it be better to just link to https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md? Trying to duplicate the information available upstream but in a "nvim-lspconfig way" is going to turn into a maintenance nightmare.

gerazov:
Like I mentioned in https://github.com/neovim/nvim-lspconfig/issues/1880 for me it was not apparent how to configure pylsp from the nvim `init.lua` from reading https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md where lspconfig points to.

I think having just the three examples here could help people, especially since turning off `mccabe` and `E501` is something that they might want to do right away. I didn't mean to have it be exhaustive by any means.

justinmk:
> Wouldn't it be better to just link to https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md? Trying to duplicate the information available upstream but in a "nvim-lspconfig way" is going to turn into a maintenance nightmare.

+1

PR #1921, opened 2022-05-24
feat: add RustOpenDocs command

Shatur:
Borrowed from https://github.com/simrat39/rust-tools.nvim/pull/188/
Uses [Open external documentation]( https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#open-external-documentation) extension.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -10,6 +10,17 @@ local function reload_workspace(bufnr)
   end)
 end
 
+local function open_docs(bufnr)
+  bufnr = util.validate_bufnr(bufnr)
+  vim.lsp.buf_request(bufnr, 'experimental/externalDocs', vim.lsp.util.make_position_params(), function(err, url)
+    if err then
+      error(tostring(err))
+    else
+      vim.fn['netrw#BrowseX'](url, 0)
+    end
+  end)
+end
+
 return {
   default_config = {
     cmd = { 'rust-analyzer' },
@@ -62,6 +73,12 @@ return {
         reload_workspace(0)
       end,
       description = 'Reload current cargo workspace',
+    },
+    RustOpenDocs = {
+      function()
+        open_docs(0)
+      end,
+      description = 'Open documentation for the symbol under the cursor in default browser',
     },
   },
   docs = {


Shatur:
I would rename `CargoReload` into `RustReloadWorkspace` for consistency. What do you think?

Shatur:
CI fails on unrelated file.

justinmk:
out of scope for this project, which is just a collection of configs.

Shatur:
> out of scope for this project, which is just a collection of configs. 

@justinmk But why? But we already have `CargoReload` and `ClangdSwitchSourceHeader` commands. And there is a special field in this plugin called `commands`.

Shatur:
> Those should be removed.

@justinmk Where it was discussed?
I was the one who introduced `ClangdSwitchSourceHeader` command for clangd back then: #263
And recently there was a fix which brings commands back into documentation: https://github.com/neovim/nvim-lspconfig/pull/1798
(we have plenty of them)

Shatur:
> The readme used to make this more clear, but changes since then have made it less clear.

Then why @mjlbach fixed commands display in documentation? And why we even have an interface for them?

> It is a collection of configs to lower the friction for getting started with Nvim LSP

All plugin commands (7 servers currently support them) use LSP queries from LSP extensions. And they are all a few lines of code. I don't see how it harms, but working with C++ without `ClangdSwitchSourceHeader` would make very inconvenient. I don't think users should create a separate plugin for each server extensions. This kind of removes the point of nvim-lspconfig.

Shatur:
> But users also shouldn't need to learn separate commands for each language. There needs to be a generic interface

But it's unavoidable because each language server provides its own extensions. And all commands mentioned in README, [example](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#clangd).

> No idea what ClangdSwitchSourceHeader does, but I've never needed it. This kind of special-case

It does what the command says - switches between header and source file. Most people use it, this feature available in any C++ IDE. It's not a special case. And we can't apply something like this to other languages because they don't need it.

The only interface we could have - special commands specific to language server. And we **do have** such interface currently. Is there a place where their removal was discussed?

justinmk:
> But it's unavoidable because each language server provides its own extensions. And all commands mentioned in README, [example](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#clangd).

On the contrary, extensions like `experimental/externalDocs` very well can be auto-completed via a generic mechanism, maybe something like `:LspExt <tab>`. Happy to talk about that in https://github.com/neovim/neovim

> And we can't apply something like this to other languages because they don't need it.

https://github.com/tpope/vim-projectionist provides a generic mechanism for that, useful for all kinds of languages. Which is another strong reason why this kind of command doesn't belong in a collection of LSP configurations.

> Is there a place where their removal was discussed?

nvim-lspconfig since the beginning had a limited purpose. Maintaining that limited scope isn't up for discussion.

Shatur:
> On the contrary, extensions like experimental/externalDocs very well can be auto-completed via a generic mechanism, maybe something like :LspExt <tab>.

For each extension we need a separate logic which depends on the LSP extension. But I like the idea of ​​having sub-commands instead of individual commands, although that's a separate topic.

> https://github.com/tpope/vim-projectionist provides a generic mechanism for that, useful for all kinds of languages.

I can hardcore needed LSP queries and language server configuration in my dotfiles myself, but we have nvim-lspconfig to make this features easily accessible for users. People who work with clangd usually expect to have something like `ClangdSwitchSourceHeader`, especially when they migrating from VSCode or CLion.

> nvim-lspconfig since the beginning had a limited purpose. Maintaining that limited scope isn't up for discussion.

I didn'y propose to discuss the scope. What I saying is that we always had the commands feature in this plugin. And you [suggested to remove it](https://github.com/neovim/nvim-lspconfig/pull/1921#issuecomment-1143705517). This is what I think worth to discuss.

justinmk:
> > On the contrary, extensions like experimental/externalDocs very well can be auto-completed via a generic mechanism, maybe something like :LspExt .
> 
> For each extension we need a separate logic which depends on the LSP extension

If langserver extensions aren't discoverable from their API, then we aren't going to enumerate them manually in nvim-lspconfig. That is not a maintenance cost we can carry.

> People who work with clangd usually expect to have something like `ClangdSwitchSourceHeader`, especially when they migrating from VSCode or CLion.

Feature parity with vscode or other IDEs is out of scope for nvim-lspconfig. Nvim LSP is a framework, and nvim-lspconfig is a way to help users get started quickly. We cannot and will not maintain feature parity with vscode (which itself depends on third-party extensions for each LSP) in nvim-lspconfig.

Shatur:
> Isn't that basically the definition of a special case though?

What I mean is that common for C++ workflow.

> I think the point of custom command was so that you could easily extend your own configuration with them
> That is not a maintenance cost we can carry.

Then why we already have some language server specific commands? It's confusing.

> and nvim-lspconfig is a way to help users get started quickly. We cannot and will not maintain feature parity with vscode

That's fair, but to get started with C++ it's nice to have the ability to switch between source and header too. It's just a few lines of code. I don't see how it hurts, but it could be annoying to force users to past them from other people configs.

Shatur:
Opened #1937.

TamaMcGlinn:
> This kind of special-case thing is not something we can maintain for 100s of language servers.

Isn't the whole point of open-sourcing, that you don't have to? @Shatur has provided the necessary code. If you can't manage to review code and approve it, perhaps you should give more people the ability to merge? Anyway, the open-source answer to this problem is forking. So [here is my fork](https://github.com/TamaMcGlinn/nvim-lspconfig/), where I will approve everything unless I notice it is broken. Feel free to direct people there if you don't feel like merging other features.

justinmk:
Tracked in Nvim core at https://github.com/neovim/neovim/issues/28329

justinmk:
# Update

See https://github.com/neovim/nvim-lspconfig/issues/1937#issuecomment-2819257770 and https://github.com/neovim/neovim/issues/28329 .

nvim-lspconfig now has lint checks and guidance (CONTRIBUTING.md):

- Config-defined commands must be buffer-local and prefixed with `:Lsp`
- Don't create commands that merely alias existing *code-actions* or *code-lenses*. Those commands are already auto-discoverable via the `gra` default keymap (or `vim.lsp.buf.code_action()`)
- Use `client:exec_cmd()` (instead of `request(..., 'workspace/executeCommand')`)
- CI has automated checks for the above.

PR #1929, opened 2022-05-25
Feature/attach to unnamed buffer

### LINKED ISSUE #1926: LSP should be able to attach to unnamed buffer
### Language server

all (yamlls in the example)

### Requested feature

Currently, the lsp server won't attach to an unnamed buffer because of this:
https://github.com/neovim/nvim-lspconfig/blob/932164a3cc87eafdd20c7d271037bea913b31c7b/lua/lspconfig/configs.lua#L89-L90

I want the language server to do formatting linting and diagnostics on unnamed buffers when the language server or the user doesn't care about a workspace to be passed.

jsonls:

```sh
mkdir /tmp/unnamed-buffer-test
cd /tmp/unnamed-buffer-test
curl -sLO https://gist.githubusercontent.com/mosheavni/20c60ea4fe7d63554b6fbbd21a5fd59e/raw/3e001e4607664df0fa168f80d1bae138449a579d/minimal_init.lua
curl -s https://gist.githubusercontent.com/mosheavni/ef05c2c3aef46c9a14b2a63316a5c0a1/raw/313fd484f6b94727ccf1dc3d6b469659a9bdb5ca/myfile.json | nvim -nu minimal_init.lua
```

Now set the filetype to json:

```
:setf json
```

Should output `client jsonls is attached"
does not.

```
:LspInfo
```

image

_No response_

mosheavni:
Resolves #1926

This adds a boolean option to the config: `unnamed_buffer_support` with default of `false` that allows spawning the LSP server with `vim.fn.cwd()` so that we have linting and diagnostics even though we are in an unnamed buffer.
Relevant only if `single_file_support` is true

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -144,10 +144,10 @@ listed in `:help vim.lsp.start_client()` with the following unique entries:
 
     `bool` (default: nil)
 
-  - `unnamed_buffer_support` : 
-  ```lua
-  true
-  ```
+- {unnamed_buffer_support}
+
+    `bool` (default: nil)
+
     Determines if a server is started without a matching root directory.
     See |lspconfig-single-file-support|.
 
@@ -439,6 +439,7 @@ Browsing the source of the default configurations is recommended.
 
 ==============================================================================
 SINGLE FILE SUPPORT                              *lspconfig-single-file-support*
+                                              *lspconfig-unnamed-buffer-support*
 
 Language servers require each project to have a `root` in order to provide
 features that require cross-file indexing. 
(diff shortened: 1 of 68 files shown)


justinmk:
why make this optional? If we know the filetype that's all that matters.

Fabyao:
> unnamed_buffer_support
When working with restFull API its a common practice to copy the JSON returned from the server and paste it in the editor for a quick analysis and editing if required. Having the ability to format without saving the file is useful as the file is temporary.

This use case is well supported in VSCode and having the same in neovim would avoid having to switch editors just to format a temporary file.

glepnir:
@mosheavni  ci failed because you used the `util.path.dirname` in root_dir most information check . 
https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md#adding-a-server-to-lspconfig.

mosheavni:
> @mosheavni ci failed because you used the `util.path.dirname` in root_dir most information check . https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md#adding-a-server-to-lspconfig.

I can see that :) however. that doesn't make a lot of sense.
I hate the fact that I must save a file in order to apply formatting or get diagnostics on.
This was not the case for example on coc.nvim.
It's mind-boggling why this is still unanswered.

PR #1936, opened 2022-06-01
fix(docs) vim.lsp.buf.format -> vim.lsp.buf.formatting

klnusbaum:
There doesn't appear to actually be a function called `format`.
Instead, I believe what we actually want here is a call `formatting`.


kylo252:
It does exist in nightly, whereas I had wrongly assumed it was also backported to 0.7.0, sorry about the inconvenience :pray:

PR #1940, opened 2022-06-02
feat: add apex_ls

williamboman:


lua/lspconfig/server_configurations/apex_ls.lua
```diff
@@ -0,0 +1,43 @@
...
+Language server for Apex.
+
+For manual installation, download the JAR file from the [VSCode
+extension](https://github.com/forcedotcom/salesforcedx-vscode/tree/develop/packages/salesforcedx-vscode-apex).
+
+```lua
+require'lspconfig'.apex_ls.setup {
+  apex_jar_path = '/path/to/apex-jorje-lsp.jar',
+  apex_enable_semantic_errors = false, -- Whether to allow Apex Language Server to surface semantic errors
+  apex_enable_completion_statistics = false, -- Whether to allow Apex Language Server to collect telemetry on code completion usage
```
williamboman, line 38: Not sure this approach is done elsewhere, but I figured it's a nicer way to allow users to customize the server without having to manually provide the full command themselves.

lua/lspconfig/server_configurations/apex_ls.lua
```diff
@@ -0,0 +1,43 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    filetypes = { 'apexcode' },
```
williamboman, line 5: @LeviticusNelson 👋  which filetypes would make sense to use here? I saw you mentioned `apexcode` in https://github.com/williamboman/nvim-lsp-installer/issues/701 - are there any other ones that should be included?

LeviticusNelson: I just use an autocmd to set `*.cls, *.trigger` files to `apexcode` so I think that line should be fine

pogthomas:
Thank you so much for doing this @williamboman!

lua/lspconfig/server_configurations/apex_ls.lua
dkelll: I know I'm pretty late here, but this probably shouldn't have used a personal autocmd to group the file types together as the official `filetypes` list. Anyone installing this now has to manually override Salesforce filetypes for it to work correctly instead of it just automatically working for any Salesforce dev who tries using it. @williamboman @LeviticusNelson Do you guys mind if I submit a PR to change the filetypes to valid Salesforce filetypes?

williamboman: There are offical `filetype`s for Salesforice filetypes in neovim?

dkelll: By "official" I just mean the standard Salesforce filetypes (.cls, .trigger, .apex, .soql, .cmp, .page, etc.). Wouldn't setting the lspconfig to `filetypes = { 'cls', 'trigger', 'apex' }` allow it to autodetect based on the file extension, or would neovim still require a manual override?

dkelll: Ignore me, I literally don't know what I'm talking about lol -- that route would be WAY more trouble and effort than it's worth. I'm thinking it's probably best just to create a standalone plugin for Salesforce stuff. Wish me luck!

williamboman: Ah so a file extension by itself doesn't technically mean anything to Vim/Neovim. You need something that matches a file to a `filetype`. This is usually done by simply checking the file extension (`.js` is a `javascript` file). Sometimes the contents of a file is consulted (for example checking if the first line of a file matches a certain shebang, e.g. `#!/usr/bin/env ruby` is a ruby file). Neovim does ship with a set of default filetype detections that generally only matches the most popular filetypes. For more obscure languages/filetypes you'll either need a dedicated plugin, or write your own filetype detections.

>  I realize neovim probably doesn't have those types, but I guess I assumed the lsp didn't necessarily care and just utilized the file's extension if/when ft wasn't found.

Nope to both. Last time I checked Neovim did not ship with filetype detection for Salesforce filetypes (and I feel very confident that hasn't changed). When no filetype is detected the file (or rather buffer) simply has no filetype, there's no "smartness" involved.

williamboman: > I'm thinking it's probably best just to create a standalone plugin for Salesforce stuff. Wish me luck!



Let me know if I can be of help! 

Here's some impactful language plugins, for inspiration:
- https://github.com/mfussenegger/nvim-jdtls
- https://github.com/simrat39/rust-tools.nvim
- https://github.com/jose-elias-alvarez/typescript.nvim
- https://github.com/scalameta/metals

dkelll: You're awesome man, I appreciate the explanation! I'll endeavor into the dedicated plugin route I think. Salesforce has a few other tools outside of their lsp anyways that I'd like to utilise, so I'll attempt getting them all combined into one plugin 🤞🏻. 

Cheers

LeviticusNelson: @dkelll I have no use of the salesforce tools because I was doing a summer gig with my college. I hope the best for creating a better tool ecosystem for Salefroce! Here is a jumpstart to a plugin that I wrote just to get my files deployed to my sandbox: 
- https://github.com/LeviticusNelson/sfdx.nvim

PR #1946, opened 2022-06-05
feat: add Marksman markdown LSP, retire zeta-note

artempyanykh:
I'm the author of Zeta-Note which has now been retired.

Marksman is a successor to the original project, but now instead of being a "*Zettelkasten LSP* that happened to be in Markdown" it's a "*Markdown LSP* that **also** supports Zettelkasten-like note-taking".


justinmk:
Thank you for following up on this.

PR #1948, opened 2022-06-08
fix(svlangserver): don't use custom project setting resolution

williamboman:
@henry-hsieh 👋 It seems like the server configuration for `svlangserver` included some custom `exrc`-as-json-style project settings loader, which I don't think should be part of the stock configuration provided by lspconfig, but instead provided by users themselves.

### DIFF
--- a/lua/lspconfig/server_configurations/svlangserver.lua
+++ b/lua/lspconfig/server_configurations/svlangserver.lua
@@ -26,7 +26,9 @@ return {
   default_config = {
     cmd = cmd,
     filetypes = { 'verilog', 'systemverilog' },
-    root_dir = util.find_git_ancestor,
+    root_dir = function(fname)
+      return util.root_pattern '.svlangserver'(fname) or util.find_git_ancestor(fname)
+    end,
     single_file_support = true,
     settings = {
       systemverilog = {
@@ -57,7 +59,7 @@ $ npm install -g @imc-trading/svlangserver
 ```
 ]],
     default_config = {
-      root_dir = [[root_pattern(".git")]],
+      root_dir = [[root_pattern(".svlangserver", ".git")]],
     },
   },
 }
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/svlangserver.lua
```diff
@@ -26,25 +26,13 @@ return {
   default_config = {
     cmd = cmd,
     filetypes = { 'verilog', 'systemverilog' },
-    root_dir = function(fname)
-      return util.root_pattern '.nvim'(fname) or util.find_git_ancestor(fname)
-    end,
+    root_dir = util.find_git_ancestor,
```
henry-hsieh, line 29: Add detection of `.svlangserver`:
```lua
root_dir = function(fname)
  return util.root_pattern '.svlangserver'(fname) or util.find_git_ancestor(fname)
end,
```

lua/lspconfig/server_configurations/svlangserver.lua
```diff
@@ -60,7 +48,16 @@ return {
...
+Language server for SystemVerilog.
+
+`svlangserver` can be installed via `npm`:
+
+```sh
+$ npm install -g @imc-trading/svlangserver
+```
 ]],
+    default_config = {
+      root_dir = [[root_pattern(".git")]],
```
henry-hsieh, line 60: Add detection of `.svlangserver`:
```lua
root_dir = [[root_pattern(".svlangserver", ".git")]],
```

henry-hsieh (commented):
As I mentioned in imc-trading/svlangserver#22, `.svlangserver` is used for root_dir detection.

williamboman:
> The svlangserver relies LSP client (vscode, coc.nvim, etc.) to provide project local setting.

I see. What kind of project-local settings does it need, are the default ones not sufficient? In the current `on_init` it seems like `cmd`, `filetypes` and `settings` are overridden. Since this function is invoked after the server has initialized, setting `cmd` and `filetypes` will not really achieve anything, so I assume it's the `settings` that is interesting here (btw I think there's a `client.notify("workspace/didChangeConfiguration")` missing at the end, to let the server know the settings have changed). If you want to achieve project-local settings via JSON, I'd personally recommend taking a look at https://github.com/tamago324/nlsp-settings.nvim, it'll most likely provide you with the mechanism you're looking for (if you don't want to use `exrc`).

PR #1953, opened 2022-06-10
feat(php): add phan support

DrWursterich:


lua/lspconfig/server_configurations/phan.lua
```diff
@@ -0,0 +1,49 @@
...
+        '-m',
+        'json',
+        '--no-color',
+        '--no-progress-bar',
+        '-x',
+        '-u',
+        '-S',
+        '--language-server-on-stdin',
+        '--allow-polyfill-parser',
+      },
```
justinmk, line 45: instead of copy-pasting, could use a shared variable, or a function to enforce "immutability"

DrWursterich: oh right, made it a variable

PR #1981, opened 2022-07-02
Update nimlsp docs

BarrOff:
Hello,

This PR adds formats  the `description` string of the nimlsp `docs`.
It also adds the the `root_dir` key to `docs`.

### DIFF
--- a/lua/lspconfig/server_configurations/nimls.lua
+++ b/lua/lspconfig/server_configurations/nimls.lua
@@ -12,7 +12,9 @@ return {
   docs = {
     description = [[
 https://github.com/PMunch/nimlsp
+
 `nimlsp` can be installed via the `nimble` package manager:
+
 ```sh
 nimble install nimlsp
 ```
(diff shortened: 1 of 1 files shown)


justinmk:
> also adds the the `root_dir` key to `docs`

why? I keep seeing PRs where people repeat the default config in the docs. The pattern should be to use the default config and override what you need. And check the default config for hints. Duplicating the default config in the docs is not maintainable.

lua/lspconfig/server_configurations/nimls.lua
```diff
@@ -12,10 +12,13 @@ return {
...
     description = [[
 https://github.com/PMunch/nimlsp
+
 `nimlsp` can be installed via the `nimble` package manager:
+
 ```sh
 nimble install nimlsp
 ```
     ]],
+    root_dir = [[util.root_pattern '*.nimble'(fname) or util.find_git_ancestor(fname)]],
```
justinmk, line 22:
```suggestion
```

PR #1984, opened 2022-07-04
Merge 0.7 branch (rebased to master)

ranjithshegde:
Since the 0.7 branch had no one taking initiative, this is essentially #1838  rebased to master.

I have been running the branch of the PR since it was created,  have zero issues. 
Seeing as 0.7.2 is released and master already requires nvim >=0.7, it seems like the ideal time to revisit and merge this.

All credits go to the excellent work of mjlbach, who I wont tag to respect his privacy.


justinmk:
> I have been running the branch of the PR since it was created, have zero issues.

Thank you! That is valuable to know.

Why again was `commands` renamed to `user_commands` ?

ranjithshegde:
> > I have been running the branch of the PR since it was created, have zero issues.
> 
> Thank you! That is valuable to know.
> 
> Why again was `commands` renamed to `user_commands` ?

This is straight from the comments on the main PR

"rename commands -> user commands to match API and free up commands
namespace"

On the one hand, the api resemblance is nice. But do you think its too confusing for the users? as these commands are essentially wrappers around unique `lsp/commands`?

justinmk:
"API resemblance" here means "it literally matches `nvim_create_user_command` ? This is questionable because:

1. `nvim_create_user_command` itself is probably over-specific
2. this is not a strong enough motivation to ask users to perform a migration

ranjithshegde:
> "API resemblance" here means "it literally matches `nvim_create_user_command` ? This is questionable because:
> 
> 1. `nvim_create_user_command` itself is probably over-specific and should have been named `nvim_create_command` (if ever it becomes theoretically possible to define non-user commands, why would we want a separate interface?! and, if it never becomes possible, what's the point of mentioning it over and over? Not even `:command` itself mentions this, it's literaly named `:command`.)
> 2. this is not a strong enough motivation to ask users to perform a migration

Will change it back to command

justinmk:
Looking closer, and reviewing https://github.com/neovim/nvim-lspconfig/pull/1838 ,  I see that a migration will be required anyway, because the command structure changed to nvim_create_user_command. Matching the target structure is a good choice, but I wonder if instead we should maintain the old interface. And deprecate it. Because it's not really needed, people can just call `nvim_create_user_comand` from `on_init` or `on_attach`.

ranjithshegde:
> Looking closer, and reviewing #1838 , I see that a migration will be required anyway, because the command structure changed.
> 
> I wonder if instead we should maintain the old interface. And deprecate it. Because it's not really needed, people can just call `nvim_create_user_comand` from `on_init` or `on_attach`.

Perhaps we remove the interface for user to define the commands, seeing as they could just use `nvim_create_user_command` where appropriate. But many internally defined commands (`ClangdSwitchSourceHeader` as an example) rely on this mechanism.
Not sure if any language specific plugins make use of them either.

So perhaps just remove it from user_space?

justinmk:
> Perhaps we remove the interface for user to define the commands, seeing as they could just use `nvim_create_user_command` where appropriate.

I'm in favor of that.

> EDIT: Defining the commands in the setup/config of the language_server has the advantage of being able to access the server specific functions with little code.
>
> An example from the help.txt of this PR

I don't see the advantage. The `nvim_create_user_command` call could just reference `require'lspconfig'.texlab.buf_build`.

ranjithshegde:
> I don't see the advantage. The `nvim_create_user_command` call could just reference `require'lspconfig'.texlab.buf_build`.

 Many methods from this repo (chiefly `lspconfig/utils` like `vim.fs`) are already in core, along with the edition of au-events like `LspAttach`. 
Maybe after the release of nvim 0.8 +  an appropriate "settle-in" period, this repo could be stripped down to a growing collection of lsp-configs with strong defaults?

With that in mind, this PR could  supply a deprecation warning concerning `commands`, with some examples of how to set the `user_commands` in an `on_attach` function?
Removing it now might be too radical for some users just yet!

ranjithshegde:
I have added a deprecation notice for commands as discussed. There is also an example in `lspconfig.txt` to set them up in `on_attach` function.

Still not entirely sure this is ideal, having commands in individual LSP setups seems cleaner than having them in a common `on_attach` function where the commands are filtered by `client.name`. 

@justinmk Would appreciate your review and thoughts on this matter

ranjithshegde:
@justinmk  Gentle ping
Requesting your review

ranjithshegde:
Not sure who the maintainers are now to request review from.
@williamboman?

plugin/lspconfig.lua
```diff
@@ -0,0 +1,84 @@
...
+  else
+    local buffer_filetype = vim.bo.filetype
+    for _, config in pairs(configs) do
+      for _, filetype_match in ipairs(config.filetypes or {}) do
+        if buffer_filetype == filetype_match then
+          config.launch()
+        end
+      end
+    end
+  end
```
kylo252, line 55:
```suggestion
  local server_name = info.args and info.args ~= ""
  if server_name then
    if configs[server_name] then
      configs[server_name].launch()
    end
  else
    local other_matching_configs = require('lspconfig.util').get_other_matching_providers(vim.bo.filetype)
    for _, config in ipairs(other_matching_configs) do
      config.launch()
    end
  end
```

there seems to be a bug with using `fargs` and `nargs='?'`

test/minimal_init.lua
```diff
@@ -29,7 +29,7 @@ local function load_plugins()
   }
 end
 
-_G.load_config = function()
+local load_config = function()
   vim.lsp.set_log_level 'trace'
   if vim.fn.has 'nvim-0.5.1' == 1 then
```
kylo252, line 34:
```suggestion
```

should also remove the complimentary `end`

lua/lspconfig/server_configurations/zk.lua
```diff
@@ -7,17 +7,21 @@ return {
     root_dir = util.root_pattern '.zk',
   },
   commands = {
```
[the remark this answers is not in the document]
williamboman: Feels like such a breaking removal would belong in a separate PR?

lua/lspconfig/configs.lua
```diff
@@ -88,15 +89,14 @@ function configs.__newindex(t, config_name, config_def)
...
-          )
-        )
-        -- Attach for all existing buffers in this directory.
+        local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
+        vim.api.nvim_create_autocmd('BufReadPost', {
+          pattern = vim.fn.fnameescape(root_dir) .. '/*',
+          callback = function()
+            M.manager.try_add_wrapper()
+          end,
+          group = lsp_group,
```
williamboman, line 98:
```suggestion
          group = lsp_group,
          desc = string.format("Checks whether server %s should attach to a newly opened buffer inside workspace %q.", config.name, root_dir)
```

plugin/lspconfig.lua
williamboman: Doesn't that suggestion turn ´server_name` into a boolean (in the truthy case, where it matters)?

plugin/lspconfig.lua
```diff
@@ -0,0 +1,84 @@
+local version_info = vim.version()
+if version_info.minor < 7 then
```
williamboman, line 2: You never know when 1.0 is out 🤡 

```suggestion
if version_info.major < 0 and version_info.minor < 7 then
```

scripts/docgen.lua
```diff
@@ -111,11 +111,12 @@ local function make_lsp_sections()
           end
           return '**Commands:**\n'
             .. make_section(0, '\n', {
-              sorted_map_table(template_def.commands, function(name, def)
-                if def.description then
-                  return string.format('- %s: %s', name, def.description)
+              -- TODO: fixme
```
williamboman, line 114: What is there to be fixed?

lua/lspconfig/configs.lua
```diff
@@ -279,17 +277,14 @@ function configs.__newindex(t, config_name, config_def)
...
-      M.commands = vim.tbl_deep_extend('force', M.commands, client.config.commands)
+    if client.config.commands then
+      M.commands = client.config.commands
     end
-    if not M.commands_created and not vim.tbl_isempty(M.commands) then
-      -- Create the module commands
-      util.create_module_commands(config_name, M.commands)
-      M.commands_created = true
+    for _, command_info in pairs(M.commands or {}) do
+      vim.api.nvim_create_user_command(command_info.name, command_info.command, command_info.opts or {})
```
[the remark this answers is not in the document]
williamboman: I fully agree with this. Very disruptive change that is difficult to motivate.

lua/lspconfig/configs.lua
```diff
@@ -64,14 +65,14 @@ function configs.__newindex(t, config_name, config_def)
...
-          config.name
-        )
-      )
+      local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
+      vim.api.nvim_create_autocmd(event, {
+        pattern = pattern,
+        callback = function()
+          M.manager.try_add()
+        end,
+        group = lsp_group,
```
williamboman, line 74:
```suggestion
        group = lsp_group,
        desc = "Checks whether server %s should either start a new instance or attach to an existing one."
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,84 @@
...
+        end
+      end
+    end
+  end
+end, {
+  desc = 'Manually launches a language server',
+  nargs = '?',
+  complete = function()
+    return vim.tbl_keys(configs)
+  end,
```
williamboman, line 61: This style of completion (uses `:command-completion-customlist` under the hood) has a worse autocompletion experience than the previous `:command-completion-custom`. I think we should stick with the former behaviour, and do something like:

```suggestion
  complete = "custom,v:lua.lsp_complete_configured_servers"
```

williamboman: (for this we'll need to restore the `lsp_complete_configured_servers` global)

plugin/lspconfig.lua
```diff
@@ -0,0 +1,84 @@
...
+  for _, client in ipairs(get_clients_from_cmd_args(info.args)) do
+    client.stop()
+    vim.defer_fn(function()
+      configs[client.name].launch()
+    end, 500)
+  end
+end, {
+  desc = 'Manually restart the given language client(s)',
+  nargs = '?',
+  complete = lsp_get_active_client_ids,
```
williamboman, line 73:
```suggestion
  complete = "custom,v:lua.lsp_get_active_client_ids"
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,84 @@
...
+})
+
+vim.api.nvim_create_user_command('LspStop', function(info)
+  for _, client in ipairs(get_clients_from_cmd_args(info.args)) do
+    client.stop()
+  end
+end, {
+  desc = 'Manually stops the given language client(s)',
+  nargs = '?',
+  complete = lsp_get_active_client_ids,
```
williamboman, line 83:
```suggestion
  complete = "custom,v:lua.lsp_get_active_client_ids"
```

williamboman (commented):
I'm personally really opposed to breaking the command API, if the plan is to deprecate it anyway.

lua/lspconfig/configs.lua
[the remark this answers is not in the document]
ranjithshegde: That is a long story. In the original PR, mjlbach wanted create the entire interface for defining per-language-server commands called `user_commands` with identical signature to `nvim_create_user_comman`.

When I opened this PR (still 95% his work), other maintainers first suggested it be just called `commands` instead, which I changed, then it was discussed it was best to deprecate it entirely.

Right now it reflect all of those changes, in a single PR!. What is the suggestion? Keep it as the original PR by mjlbach and then deprecate it in the next nvim release cycle? Or something else?

lua/lspconfig/server_configurations/zk.lua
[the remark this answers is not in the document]
ranjithshegde: Is the deprecation still being discussed or has it been already decided?

I am neither a maintainer nor a significant contributor  so I dont know if my opinion counts for anything, but for what its wroth:
I guess for certain servers features (like `clangd/switchSourceHeader`) these are considered as "extended" features, and certainly off 3.16 spec. 
But for servers like `texlab`, where the buffer holds more than 80% plain text and very few lines of latex commands, `texlab/build` is basically a primary feature. 

Without that command internally defined, users will have to resort to defining `params`, making the lsp call, extracting the results contextually and displaying them,  and then writing a `user_command` to call it. Which most users may not be able to do. Kind of removes the quick out of quickstart feature of the plugin. 

Unless someone writes a plugin for just `texlab/build` and `texlab/forward` commands. And such plugins for each server that has those command features.

Of course, if that is the consensus I will remove/or deprecate it!

plugin/lspconfig.lua
kylo252: > Doesn't that suggestion turn ´server_name` into a boolean (in the truthy case, where it matters)?

it's specifically that `<f-args>` will get a string when using `nargs=?`, so `fargs[1]` doesn't really do anything useful, since calling `LspStart` without any arguments will give `server_name=""`

williamboman: I meant won't the following expression

```lua
local server_name = info.args and info.args ~= ""
```

always evaluate to a boolean? It's later used to index the `configs` table: `configs[server_name]`.

I'd probably do something like the following:

```lua
if info.args ~= "" then
  local server_name = info.args
  if configs[server_name] then
    configs[server_name].launch()
  end
else
  -- ...
end
```

kylo252: dang 🤦 

I meant to write it like this

```lua
local server_name = info.args ~= "" and info.args
```

but yeah, just need to make sure we're checking that args isn't an empty string

plugin/lspconfig.lua
justinmk: `vim.fn.has('nvim-0.8')` is the correct way to check this

doc/lspconfig.txt
```diff
@@ -242,34 +242,37 @@ The `configs` module is a singleton where configs are defined. The schema for
...
+      default_config = {'t', true},
     }
 <
-`commands` is a map of `name:definition` key:value pairs, where `definition`
-is a list whose first value is a function implementing the command, and the
-rest are either array values which will be formed into flags for the command,
-or special keys like `description`. Example:
+`commands` is a list of tables where each table consists of the fields
+`name`, `command`, and `opts`. See |nvim_create_user_command| for more
+details.
```
justinmk, line 261: why was it changed from a map?

lua/lspconfig/configs.lua
```diff
@@ -40,10 +40,11 @@ function configs.__newindex(t, config_name, config_def)
       commands = { config.commands, 't', true },
     }
     if config.commands then
-      for k, v in pairs(config.commands) do
+      vim.deprecate('lspconfig.[server].setup.commands', 'vim.api.nvim_create_user_command', '0.9', 'nvim-lspconfig')
```
justinmk, line 43: - soft deprecation is correct. Only show a warning/error, but continue to function
- as others mentioned, don't change the structure. we don't want a breaking change yet.

justinmk:
> When I opened this PR (still 95% his work), other maintainers first suggested it be just called commands instead, which I changed, then it was discussed it was best to deprecate it entirely.
>
> Right now it reflect all of those changes, in a single PR!. What is the suggestion? Keep it as the original PR by mjlbach and then deprecate it in the next nvim release cycle? Or something else?

@ranjithshegde this is my fault, sorry. Let's see if we can make progress here without too much work. Here's what I suggest:

1.  soft deprecate (show warning/error but continue to work) the `commands` API without a breaking change (so don't need to change the code)
2. add the documentation examples for `nvim_create_user_command`
3. remove the examples for `commands`

doc/lspconfig.txt
ranjithshegde: As I understand it, this is to make the interface identical to `nvim_create_user_command`

doc/lspconfig.txt
```diff
@@ -242,34 +242,37 @@ The `configs` module is a singleton where configs are defined. The schema for
...
-      "-range";
-      description = "Build the current buffer";
-    };
-  };
+      end,
+      opts = {
+        desc = "Build the current buffer",
+      }
+    },
+  },
```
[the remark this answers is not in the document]
ranjithshegde: in `lspconfig/server_configurations/texlab.lua` it does not accept a `range` key

lua/lspconfig/server_configurations/zk.lua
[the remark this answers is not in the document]
williamboman: > Is the deprecation still being discussed or has it been already decided?
>
> I am neither a maintainer nor a significant contributor so I dont know if my opinion counts for anything, but for what its wroth:

Likewise. If by deprecation you mean these commands for `zk` then no, I don't believe it has been discussed before (quick GitHub search).

My take on this is: I really don't see the point of removing commands just for the sake of nudging users to install a dedicated plugin (https://github.com/mickael-menu/zk-nvim) for it (even though that plugin looks very thorough and well-made). It just adds unnecessary friction. If there are people that rely on these commands it's obviously providing value to them, and I think that weighs heavier than how to technically slice and dice things. If the commands ever stop working, or if they no longer suffice, these users have a very strong personal incentive to switch to https://github.com/mickael-menu/zk-nvim themselves.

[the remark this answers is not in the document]
ranjithshegde: My suggestion would be to keep the internals, which are defined in `server_configurations/[lsp].lua` but deprecate the interface for defining them in setup as @justinmk  recommended. 

It is very trivial to just define custom commands in `nvim_create_user_command` instead of having`lspconfig` scan the the configs and define them using the same interface.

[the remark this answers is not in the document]
williamboman: Sounds extremely reasonable to me!

lua/lspconfig/configs.lua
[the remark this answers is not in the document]
ranjithshegde: What is the consensus on this? Should I revert it back to as it was?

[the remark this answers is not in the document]
ranjithshegde: Thanks! I was in the process of writing something similar. 
Will have something ready for review by (my) today evening.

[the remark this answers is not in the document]
williamboman: I updated the following part:

```
         local option, value = unpack(vim.split(v, "=", { plain = true }))
        option = option.gsub(v, "^%-+", "")
        opts[opts_aliases[option] or option] = value or true
```

This function is worthy of a test case or two (or five).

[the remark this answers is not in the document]
ranjithshegde: The command table is already being validated in `configs.lua`
so simply this will do
```lua
function M.create_module_commands(module_name, commands)
  for command_name, def in pairs(commands) do
    local opts = {}
    if def.description and def['description'] ~= '' then
      opts.desc = def.description
    end
    api.nvim_create_user_command(command_name, function(...)
      require('lspconfig')[module_name].commands[command_name][1](...)
    end, opts)
  end
end
```
I am not even sure if the `description` key needs to be verified as its also being explicitly validated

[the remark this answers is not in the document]
williamboman: Wont that suddenly exclude options like:

```lua
MyCommand = {
  function () end,
  ["-nargs=?"],
  ["-complete=..."],
  ["-bang"] = true
}
```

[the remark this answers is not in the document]
ranjithshegde: I did check through all the default configurations, and they all only provide 
```lua
name = {
function() stuff end,
description = "whatever"
}
```
Let me see what the docs specify, or if those options have been allowed in the past

scripts/docgen.lua
ranjithshegde: I think this has to be reverted as well

lua/lspconfig/configs.lua
```diff
@@ -40,6 +40,7 @@ function configs.__newindex(t, config_name, config_def)
       commands = { config.commands, 't', true },
     }
     if config.commands then
+      vim.deprecate('lspconfig.[server].setup.commands', 'vim.api.nvim_create_user_command', '0.9', 'nvim-lspconfig')
```
williamboman, line 43: While I think this deprecation message makes a lot of sense to have, it's pretty disruptive at the moment. Especially considering that some of the provided server configurations still make use of this API (i.e. nothing a user can fix themselves). May I suggest breaking the deprecation down in steps?

1) Soft-soft deprecation (deprecation notices in documentation and release notes only). This will allow 3rd party plugins time to move away from it (rust-tools.nvim is a pretty popular one, as an example).
2) Soft deprecation via warning message.
3) Removal of the feature.

williamboman (approved):
Awesome work! Code LGTM, just one comment regarding the deprecation message that I think is worth revisiting. I'll run this branch going forward and will report back if I run into issues

lua/lspconfig/configs.lua
ranjithshegde: Removed the notification. I have added something in the commit message + in `lspconfig.txt`

Also, I have been running this branch from when it was started by mjlbach. Have had no issues. 
In fact, it improves a lot of usage options as `nvim_exec_autocmd` for `lspconfig` can now be called if you are defining/configuring a server not on `FileType` event.

plugin/lspconfig.lua
```diff
@@ -0,0 +1,85 @@
...
+  else
+    local other_matching_configs = require('lspconfig.util').get_other_matching_providers(vim.bo.filetype)
+    for _, config in ipairs(other_matching_configs) do
+      config.launch()
+    end
+  end
+end, {
+  desc = 'Manually launches a language server',
+  nargs = '?',
+  complete = 'custom,v:lua.lsp_complete_configured_servers',
```
ii14, line 62: This can be a lua function, no need to create global variables. I don't remember if this was supported right from the start, or was implement in later neovim versions though.
```suggestion
  complete = lsp_complete_configured_servers,
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,85 @@
+local version_info = vim.version()
```
ii14, line 1: Add a guard here, so it can be blocked from loading.
```suggestion
if vim.g.lspconfig ~= nil then
  return
end
vim.g.lspconfig = 1

local version_info = vim.version()
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,85 @@
...
+    '[lspconfig] requires neovim 0.7 or later. Detected neovim version: 0.%s.%s',
+    version_info.minor,
+    version_info.patch
+  )
+  vim.notify_once(warning_str)
+  return
+end
+
+local util = require 'lspconfig.util'
+local configs = require 'lspconfig.configs'
```
ii14, line 13: Inconsistent use, sometimes these local variables are used, sometimes `require` is called again. In my opinion you should inline these and always call `require`, to not unnecessarily force these modules to be loaded.

plugin/lspconfig.lua
ranjithshegde: This was the initial state of the PR. @williamboman requested the use of `custom, v:lua` method

plugin/lspconfig.lua
ii14: @williamboman worse experience? What do you mean? It should be the same thing, except that `custom` should just expect a newline-separated string, and `customlist` a list.

plugin/lspconfig.lua
ranjithshegde: `configs` is used several times in the file. Perhaps I only inline `utils`?

plugin/lspconfig.lua
ii14: Okay, I think the issue was that you didn't filter completions, like this:
```lua
function complete(arg)
  local res = {}
  for _, entry in ipairs(lsp_complete_configured_servers()) do
    if entry:sub(1, #arg) == arg then
      table.insert(res, entry)
    end
  end
  return res
end
```
This can happen inside `lsp_complete_configured_servers` of course.

plugin/lspconfig.lua
ii14: It's just three times basically.

plugin/lspconfig.lua
ranjithshegde: The original functions were these
```lua
local lsp_get_active_client_ids = function()
  return vim.tbl_map(function(client)
    return ('%d (%s)'):format(client.id, client.name)
  end, require('lspconfig.util').get_managed_clients())
end
```
and 

```lua
  desc = 'Manually launches a language server',
  nargs = '?',
  complete = function()
    return vim.tbl_keys(configs)
  end,
```

ii14: Yes, you need to filter them then.

williamboman: Ah, I wasn't aware you'd have to implement the filtering yourself, that explains it. Maybe this can be rectified in a later PR, this is starting to go a bit in circles with people chiming in haha

ranjithshegde: Yes, getting this merged and then adding improvements if necessary maybe a good idea!

plugin/lspconfig.lua
ii14: Disabling those commands and defining your own instead is a valid thing to do :)

lua/lspconfig/configs.lua
```diff
@@ -182,17 +190,15 @@ function configs.__newindex(t, config_name, config_def)
         if bufnr == api.nvim_get_current_buf() then
           M._setup_buffer(client.id, bufnr)
         else
-          if vim.api.nvim_buf_is_valid(bufnr) then
```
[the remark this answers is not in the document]
ranjithshegde: Since `on_new_config` is called internally, didnt think it was necessary. But you are absolutely right, that was a mistake. Added it back

plugin/lspconfig.lua
```diff
@@ -0,0 +1,87 @@
...
+end, {
+  desc = 'Displays attached, active, and configured language servers',
+})
+
+vim.api.nvim_create_user_command('LspStart', function(info)
+  local server_name = info.args[1] ~= '' and info.args
+  if server_name then
+    if require('lspconfig.configs')[server_name] then
+      require('lspconfig.configs')[server_name].launch()
+    end
```
ii14, line 54:
```suggestion
    local config = require('lspconfig.configs')[server_name]
    if config then
      config.launch()
    end
```

ranjithshegde:
Is this good to go then?

ranjithshegde:
Is there a way to see on which file and line codespell action fails at?

glepnir:
image

plugin/lspconfig.lua
```diff
@@ -0,0 +1,91 @@
+if vim.g.lspconfig ~= nil then
+  return
+end
+vim.g.lspconfig = 1
+
+local version_info = vim.version()
+if vim.fn.has 'nvim-0.7' ~= 1 then
```
justinmk, line 7: has(nvim-0.7) will be true for all versions 7 or greater. It will only be false for 0.6 or older. The check should be has('nvim-0.8')

ranjithshegde: I am a bit confused. The minimum required version is 0.7. it should fail if its lower than 0.7. but if this is set to 0.8, wont it fail for all versions other than nightly?

test/lspconfig_spec.lua
```diff
@@ -319,7 +319,7 @@ describe('lspconfig', function()
         local _ = lspconfig.sumneko_lua
         local _ = lspconfig.tsserver
         lspconfig.rust_analyzer.setup {}
-        return lspconfig.available_servers()
+        return require("lspconfig.util").available_servers()
```
williamboman, line 322: This seems like a breaking change to me. `lspconfig.available_servers()` should still exist

ranjithshegde: Well at the moment, the `lspconfig.lua` simply sets up the metatable for `lspconfig.configs` and nothing else. If we add this, then it will setup a metatable + a single function that can actually be in utils.

williamboman: > Then the logic is simple. The portion of plugin simply sets up LSPs, with some commands in `plugin/` and some useful functions in `/utils`.

Yeah I agree it makes sense to remove it, I'm just thinking about not breaking parity with `master`. This will for sure break some user configs and plugins out there, although easily remedied.

> Also, too many PRs are being merged that potentially clash with this. Requires a bit of reshuffling

Yeah there seems to be a lot happening with this big PR still being open, I'll eject myself from further comments, let's get this merged haha

ranjithshegde: > Yeah I agree it makes sense to remove it, I'm just thinking about not breaking parity with `master`. This will for sure break some user configs and plugins out there, although easily remedied.
> 
Well this is changing/simplifying the core mechanism of the project. We could merge this as a breaking change, with a tag for previous versions?

plugin/lspconfig.lua
justinmk: my mistake, I thought you wanted to check for 0.8

justinmk (approved):
Thanks everyone for the help here.

ranjithshegde:
Squashed all the changes into single commit!

justinmk:
Will merge this now. Can someone open a PR that _removes_ stuff for 0.8? It won't be merged until 0.8 is released, but a PR will help us figure out what changes are expected..

ranjithshegde:
> Will merge this now. Can someone open a PR that _removes_ stuff for 0.8? It won't be merged until 0.8 is released, but a PR will help us figure out what changes are expected.

Will start with an issue. Maybe deprecation is requried as the first step. Many functions in `util` are available upstream.
A strategy could be to redirect those calls to upstream with a deprecation warning

mike-lloyd03:
I almost just opened an issue for this but did a bit more digging first. I see that this commit introduces a breaking change. But I cannot tell from the conversation here what needs to be changed in my config to accommodate this change. Or if it's a problem with another plugin that depends on lspconfig and I should roll back to a previous version of this plug until other plug developers can update. Is there recommended guidance somewhere I should be looking at?

Thanks and sorry for this annoying message. I did my research first but came up short.

justinmk:
@ranjithshegde  I wasn't expecting breaking changes here per https://github.com/neovim/nvim-lspconfig/pull/1984#issuecomment-1221625288  : 

> soft deprecate (show warning/error but continue to work) the commands API without a breaking change (so don't need to change the code)

What did we break?

ii14:
@justinmk looks like `require("lspconfig").available_servers` is now `require("lspconfig.util").available_servers`.

ii14:
If this is a problem, it could just be added back:
```diff
diff --git a/lua/lspconfig.lua b/lua/lspconfig.lua
index 1178202..c47707d 100644
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -4,6 +4,10 @@ local M = {
   util = require 'lspconfig.util',
 }
 
+function M.available_servers()
+  return M.util.available_servers()
+end
+
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
```

glepnir:
@justinmk  like i said in #2078 I had search in github some configs and plugins use it .but it move to `utils.lua`

PR #1986, opened 2022-07-06
chore(selene): switch to yml based standard library format

sitiom:
As of https://github.com/Kampfkarren/selene/pull/364, using TOML for the standard library is now deprecated.


sitiom:
Hmm, didn't change anything in the config and the code though, seems to be a new check from selene:
image

justinmk:
nice, that would be a nice cleanup if it's valid

PR #1988, opened 2022-07-06
add .git to jdtls root_files

MarcThu:
I have added ".git" into the root_files from jdtls, so that projects without ant, maven or gradle also get recognized by jdtls as a project. With this change, I was able to fix it for myself, so I wanted to share that.

This is my first contribution, so please tell me if I should improve things in the future, when contributing.

### DIFF
--- a/lua/lspconfig/server_configurations/jdtls.lua
+++ b/lua/lspconfig/server_configurations/jdtls.lua
@@ -81,6 +81,7 @@ local root_files = {
   {
     'build.xml', -- Ant
     'pom.xml', -- Maven
+    '.git', -- Git
     'settings.gradle', -- Gradle
     'settings.gradle.kts', -- Gradle
   },


MarcThu:
Changed commit message, to fulfil conventional commit style.

justinmk:
how is the lsp server supposed to work without a project file?

MarcThu:
I don't know if I misunderstood something, but the change does work in my case with lunar vim.

Before the change, my plain java project did not get recognized, and I was limited to just syntax errors and not for reference checks across the whole project.

As I didn't work much with java (just at university in simple projects) I did not work with maven or something else. I don't know If adding .git breaks something if there is another root_file as well.

justinmk:
same problem would then occur if `.git` is not found. 

general question for all configs: why don't we fallback to `vim.fn.getcwd()` ?

PR #1995, opened 2022-07-10
fix(denols): ignore virtual text error

sigmaSd:
If you open a file and immediately try to go to definition this error happens:
image

It seems to be related to some race condition where the virtual text handling is not ready yet

The exact error is 
```
{["error"]={["message"]="Method not found: deno/virtualTextDocument",["code"]="-32601"}}{["error"]={["message"]="Method not found: deno/virtualTextDocument",["code"]="-32601"}}{["error"]={["code"]="-32601",["message"]="Method not found: deno/virtualTextDocument"}}{["error"]={["code"]="-32601",["message"]="Method not found: deno/virtualTextDocument"}}⏎  
```

Because this error is unhelpful and the gotodefinition works normally after a couple of seconds, this pr propose to just ignore it.

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -18,19 +18,21 @@ local function virtual_text_document_handler(uri, result)
   end
 
   for client_id, res in pairs(result) do
-    local lines = vim.split(res.result, '\n')
-    local bufnr = vim.uri_to_bufnr(uri)
+    if res.result ~= nil then
+        local lines = vim.split(res.result, '\n')
+        local bufnr = vim.uri_to_bufnr(uri)
 
-    local current_buf = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
-    if #current_buf ~= 0 then
-      return nil
+        local current_buf = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+        if #current_buf ~= 0 then
+          return nil
+        end
+
+        vim.api.nvim_buf_set_lines(bufnr, 0, -1, nil, lines)
+        vim.api.nvim_buf_set_option(bufnr, 'readonly', true)
+        vim.api.nvim_buf_set_option(bufnr, 'modified', false)
+        vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)
+        lsp.buf_attach_client(bufnr, client_id)
     end
-
-    vim.api.nvim_buf_set_lines(bufnr, 0, -1, nil, lines)
-    vim.api.nvim_buf_set_option(bufnr, 'readonly', true)
-    vim.api.nvim_buf_set_option(bufnr, 'modified', false)
-    vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)
-    lsp.buf_attach_client(bufnr, client_id)
   end
 end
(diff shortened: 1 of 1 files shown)


sigmaSd:
Also I noticed that denols have this exact same issue https://github.com/fannheyward/coc-deno/issues/139 (jumping to definition then returning breaks lsp), I think it was fixed with this commit https://github.com/fannheyward/coc-deno/commit/87da2015b33b4957e71bf56d6bc94764b0cd3084

lua/lspconfig/server_configurations/denols.lua
```diff
@@ -18,19 +18,21 @@ local function virtual_text_document_handler(uri, result)
   end
 
   for client_id, res in pairs(result) do
-    local lines = vim.split(res.result, '\n')
-    local bufnr = vim.uri_to_bufnr(uri)
+    if res.result ~= nil then
```
justinmk, line 21: - a 1-line comment is worth adding here to give a hint about the reasoning.
- should probably log at warn or info level

```suggestion
    -- May be nil because of race, deno server will eventually send a result. #1995
    if res.result == nil then
      log.warn(...)
    else
```

sigmaSd: I can't find log module, where do I get it from?

justinmk: require('vim.lsp.log').warn('foo')

you can find the result in : 

    exe 'e' stdpath('log')

sigmaSd: done

sigmaSd:
It seems like this is just a workaround to  more subtle issue because a lot of denols methods seems to fail sometimes, for example if I use denolscache it also shows an error 
image

(dayjs/mod.ts doesn't really exist I just used a non existent file so it doesn't get cached)
You can see that on the bottom that the method actually worked and returned the expected `Module not found` but for some reason it also shows an error that the method denols/cache is not found

sigmaSd:
The lsp log always shows errors like this
```
[ERROR][2022-07-14 08:34:24] .../vim/lsp/rpc.lua:420	"rpc"	"deno"	"stderr"	"Error converting specifier settings (file:///home/mrcool/dev/deno/lab/a.ts): Error converting specifier settings: invalid type: null, expected struct SpecifierSettings\n"
```
maybe its relevant

justinmk:
> Error converting specifier settings: invalid type: null, expected struct SpecifierSettings\n"

sure, maybe. Need deno users to figure that out and send a fix if relevant.

sigmaSd:
I'm a deno user xD but I would need some pointers to fix this 

SpecifierSettings is here https://github.com/denoland/deno/blob/8af81d98a6d673fde3a5c2acacf6ad00c53b9223/cli/lsp/config.rs#L134 as far as I can tell nvim lsp currently only sends WorkspaceSettings with 
```
    init_options = {
            enable = true,
            unstable = false,
```
which corresponds to https://github.com/denoland/deno/blob/8af81d98a6d673fde3a5c2acacf6ad00c53b9223/cli/lsp/config.rs#L175

Where would I send those SpecifierSettings?

sigmaSd:
though it really seems optional I'll investigate when I have more free time.

PR #1996, opened 2022-07-11
fix(sumneko_lua): Add stylua.toml

nullchilly:
https://github.com/JohnnyMorganz/StyLua#finding-the-configuration

> The CLI looks for stylua.toml or .stylua.toml in the directory where the tool was executed. If not found, the default configuration is used.

It should be included in the default option. Many plugins use `stylua.toml` such as [packer.nvim](https://github.com/wbthomason/packer.nvim) or [nvim-cmp](https://github.com/hrsh7th/nvim-cmp)

### DIFF
--- a/lua/lspconfig/server_configurations/sumneko_lua.lua
+++ b/lua/lspconfig/server_configurations/sumneko_lua.lua
@@ -4,6 +4,7 @@ local root_files = {
   '.luarc.json',
   '.luacheckrc',
   '.stylua.toml',
+  'stylua.toml',
   'selene.toml',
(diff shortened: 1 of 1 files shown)


lithammer:
As far as I know Sumneko doesn't use Stylua as formatter.

nullchilly:
Yes but stylua is used to set root dir

nullchilly:
@lithammer To clarify neovim already added `.stylua.toml` as a default option, I just added `stylua.toml` to complete the default stylua's config-path

Issue #1998, opened 2022-07-13
denols: jumping to definition then returning breaks lsp


sigmaSd:
### Description

jumping to definition then returning breaks lsp
![denovim](https://user-images.githubusercontent.com/22427111/178674021-e7624ace-22f9-4773-9eec-a76b5d74d4f2.gif)

coc-deno used to have the same issue but it was recently fixed https://github.com/fannheyward/coc-deno/issues/139

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

41a826947eefbc7b59cc80cef405cce9ea6d82b8

### Operating system and version

Linux mrcool 5.18.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 02 Jul 2022 21:03:06 +0000 x86_64 GNU/Linux

### Affected language servers

denols

### Steps to reproduce

1.  `echo "console.log(4)" > test.ts`
2. `nvim test.ts`
3. move cursor to log and press `gd` to go to definition
4. press ctrl-o to return

### Actual behavior

Notice that lsp is not working correctly any more and have errors and the console and log have type any

### Expected behavior

the lsp should continue working correctly.

### Minimal config

```Lua
local on_windows = vim.loop.os_uname().version:match 'Windows'

local function join_paths(...)
    local path_sep = on_windows and '\\' or '/'
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.cmd [[set runtimepath=$VIMRUNTIME]]

local temp_dir = vim.loop.os_getenv 'TEMP' or '/tmp'

vim.cmd('set packpath=' .. join_paths(temp_dir, 'nvim', 'site'))

local package_root = join_paths(temp_dir, 'nvim', 'site', 'pack')
local install_path = join_paths(package_root, 'packer', 'start', 'packer.nvim')
local compile_path = join_paths(install_path, 'plugin', 'packer_compiled.lua')

local function load_plugins()
    require('packer').startup {
        {
            'wbthomason/packer.nvim',
            'neovim/nvim-lspconfig',
        },
        config = {
            package_root = package_root,
            compile_path = compile_path,
        },
    }
end

_G.load_config = function()
    vim.lsp.set_log_level 'trace'
    if vim.fn.has 'nvim-0.5.1' == 1 then
        require('vim.lsp.log').set_format_func(vim.inspect)
    end
    local nvim_lsp = require 'lspconfig'
[...]

PR #2000, opened 2022-07-14
fix(denols): remove unneeded lsp attach

sigmaSd:
fix https://github.com/neovim/nvim-lspconfig/issues/1998

I don't have a good idea *how* does this fixes the issue, but so far in my testing with this pr it seems to work perfectly.

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -17,7 +17,7 @@ local function virtual_text_document_handler(uri, result)
     return nil
   end
 
-  for client_id, res in pairs(result) do
+  for _, res in pairs(result) do
     -- Error might be present because of race, deno server will eventually send a result. #1995
     if res.error ~= nil then
         require('vim.lsp.log').warn('deno/virtual_text_document handler failed (might be a temporary issue), error: '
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/denols.lua
```diff
@@ -35,7 +35,6 @@ local function virtual_text_document_handler(uri, result)
         vim.api.nvim_buf_set_option(bufnr, 'readonly', true)
         vim.api.nvim_buf_set_option(bufnr, 'modified', false)
         vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)
-        lsp.buf_attach_client(bufnr, client_id)
```
justinmk, line 38: linter noticed that `client_id` is not used now.

sigmaSd: fixed

sigmaSd:
The commit that added this https://github.com/neovim/nvim-lspconfig/pull/753

@kitagry can you confirm that this is not needed anymore?

kitagry:
I added this code in order to use lsp functions (e.x. goto definition) in virtual text file. And, is it work correctly?

I'll see it this night. Thank you.

sigmaSd:
Yes it seem to work inside virtual document without it.

justinmk:
lint CI is failing because of wrong indentation

kitagry:
Hmm, I can't reproduce this bug. Can you try with neovim develpment build?

NVIM v0.8.0-dev+1470-g0124a7bfa
Build type: Release
LuaJIT 2.1.0-beta3
nvim-lspconfig c55e830

> Yes it seem to work inside virtual document without it.

I think it works only when you specify `single_file_support=true`.

sigmaSd:
> I think it works only when you specify single_file_support=true.

I can confirm this, do you have a suggestion on how to proceed then? is there a disadvantage of making `single_file_support=true` by default?

kitagry:
> is there a disadvantage of making single_file_support=true by default?

I think there are no disadvantage. I think you should this change with `single_file_support=true` by default.

PR #2003, opened 2022-07-14
feat: add glsl-language-server support

BenFradella:
Works, but it looks like the server and/or vim lsp are doing something wrong for the unsupported methods. From what I understand, this isn't something that would be handled here, but correct me if I'm wrong:
```log
[ERROR][2022-07-14 13:55:11] .../lua/vim/lsp.lua:824    "LSP[glslls]"   "on_error"  {  code = "INVALID_SERVER_MESSAGE",  err = {    error = {      code = -32601,      message = "Method 'textDocument/completion' not supported."    },    jsonrpc = "2.0"  }}
[ERROR][2022-07-14 13:56:31] .../lua/vim/lsp.lua:824    "LSP[glslls]"   "on_error"  {  code = "INVALID_SERVER_MESSAGE",  err = {    error = {      code = -32601,      message = "Method 'textDocument/didSave' not supported."    },    jsonrpc = "2.0"  }}
[ERROR][2022-07-14 13:56:44] .../lua/vim/lsp.lua:824    "LSP[glslls]"   "on_error"  {  code = "INVALID_SERVER_MESSAGE",  err = {    error = {      code = -32601,      message = "Method 'shutdown' not supported."    },    jsonrpc = "2.0"  }}
```

### DIFF
--- a/lua/lspconfig/server_configurations/glslls.lua
+++ b/lua/lspconfig/server_configurations/glslls.lua
@@ -1,0 +1,28 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'glslls', '--stdin' },
+    filetypes = { 'glsl' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+    capabilities = {
+      textDocument = {
+        completion = {
+          editsNearCursor = true,
+        },
+      },
+      offsetEncoding = { 'utf-8', 'utf-16' },
+    },
+  },
+  docs = {
+    description = [[
+https://github.com/svenstaro/glsl-language-server
+
+Language server implementation for GLSL
+
+`glslls` can be compiled and installed manually, or, if your distribution has access to the AUR,
+via the `glsl-language-server` AUR package
+    ]],
+  },
+}


ranjithshegde:
I have been watching the lsp repo in question for almost a year now. The only feature that is implemented is diagnostics. No completion, hover, definition etc. Does not seem to have most features from Lsp-spec.

Would it be a better candidate for `null-ls` until the LSP is more mature?

As a side-note: You can configure `glslc` via `efm-langserver`. Null-ls already has `glslc` for diagnostics handling. Also, I have a working method for `glslangValidator` that I can share as a gist.

theHamsta:
I've also added `glslc` and `dxc` to https://github.com/mfussenegger/nvim-lint
my setup is
```lua
  use {
    "mfussenegger/nvim-lint",
    config = function()
      local function select_executables(executables)
        return vim.tbl_filter(function(c)
          return c ~= vim.NIL and vim.fn.executable(c) == 1
        end, executables)
      end
      require("lint").linters_by_ft = {
        markdown = select_executables { "markdownlint" },
        lua = select_executables { "luacheck" },
        glsl = select_executables { "glslc" },
        hlsl = select_executables { "dxc" },
      }
      table.insert(require("lint").linters.glslc.args, "--target-env=vulkan1.3")
      vim.cmd [[au BufEnter,BufWritePost * lua require('lint').try_lint()]]
    end,
  }
```
I hope this is not spam for anyone in this PR

ranjithshegde:
Yes, more spam (with instructions).
Being able to use functions for parsing stdout is the main reason I switched from `emf-langserver` to null-ls.
Here is a snippet on how to configure `glslangValidator`

```lua
local null_ls = require "null-ls"

local function glsl()
    return {
        method = null_ls.methods.DIAGNOSTICS,
        filetypes = { "glsl" },
        generator = null_ls.generator {
            command = "glslangValidator",
            args = { "--stdin", "-S", "$FILEEXT" },
            to_stdin = true,
            from_stderr = true,
            format = "raw",
            check_exit_code = function(code, stderr)
                local success = code <= 1
                if not success then
                    print(stderr)
                end

                return success
            end,
            on_output = function(params, done)
                if params and params.output then
                    local diagnostics = {}
                    local lines = vim.split(params.output, "\n")
                    local sever, col, row, message = string.match(lines[2], "(%u+):%s(%d+):(%d+):.*:%s+(.*)")

                    table.insert(diagnostics, {
                        row = row,
                        col = col + 1,
                        end_col = col + 2,
                        source = "GLSLang",
                        message = message,
                        severity = require("null-ls.helpers").diagnostics.severities[vim.fn.tolower(sever)],
                    })
                    done(diagnostics)
                else
                    done()
                end
            end,
        },
    }
end

null_ls.register(glsl())
```

PR #2009, opened 2022-07-17
Ignore caching error

sigmaSd:
Caching actually works but it throws an error, vscode-deno also doesn't seem to care about errors https://github.com/denoland/vscode_deno/blob/196a11cead2d95c324cb772dd568ba7c2801b68a/client/src/commands.ts#L60

cache


sigmaSd:
@justinmk The error was coming from other lsp servers that received the request (and they don't implement it) I guess this is a correct fix?

Issue #2008, opened 2022-07-17
HLS configuration doesn't work properly with multi-package projects


Nexmean:
### Description

In haskell multi-package projects there is `*.cabal` file in every package folder, so `util.root_pattern` always find root of the package but not project.

I suggest to change [this line](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/hls.lua#L7) of hls config to fix it:
```lua
      root_dir = function (filepath)
         return (util.root_pattern("hie.yaml", "cabal.project", "stack.yaml")(filepath)
            or util.root_pattern("*.cabal", "package.yaml")(filepath))
      end,
```

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled  _nixbld5

Features: +acl +iconv +tui

### Operating system and version

macOS 12.4

### Affected language servers

hls

### Steps to reproduce

1. run nvim in multi-package haskell project
2. open any haskell file

### Actual behavior

There is starting different HLS instance for every package of the project

### Expected behavior

There is starting single HLS instance for the whole project

### Minimal config

```Lua
-
```

### LSP log

-

justinmk:
send a PR

PR #2010, opened 2022-07-17
Fix HLS config to work properly with multi-package projects

Nexmean:
PR addresses [this issue](https://github.com/neovim/nvim-lspconfig/issues/2008)


Nexmean:
fixed formatting with amend commit

PR #2013, opened 2022-07-18
Dartls server config windows

erietz-werner:
I was having a terrible time getting the client to attach to the buffer on Windows. This fixes the issue for me.

### DIFF
--- a/lua/lspconfig/server_configurations/dartls.lua
+++ b/lua/lspconfig/server_configurations/dartls.lua
@@ -1,11 +1,8 @@
 local util = require 'lspconfig.util'
 
-local bin_name = 'dart'
-local cmd = { bin_name, 'language-server', '--protocol=lsp' }
-
-if vim.fn.has 'win32' == 1 then
-  cmd = { 'cmd.exe', '/C', bin_name, 'language-server', '--protocol=lsp' }
-end
+local cmd = (vim.fn.has 'win32' == 1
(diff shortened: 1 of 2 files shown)


lua/lspconfig/server_configurations/dartls.lua
```diff
@@ -1,8 +1,15 @@
 local util = require 'lspconfig.util'
 
+local bin_name = 'dart'
+local cmd = { bin_name, 'language-server', '--protocol=lsp' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, 'language-server', '--protocol=lsp' }
+end
```
justinmk, line 8: ternary is more readable, avoids nesting, etc.

```suggestion
local cmd = (vim.fn.has 'win32' == 1
    and { 'cmd.exe', '/C', bin_name, 'language-server', '--protocol=lsp' }
    or { bin_name, 'language-server', '--protocol=lsp' })
```

erietz-werner: Thanks, I didn't know about the lua ternary.

justinmk:
sure seems like vim.lsp or lspconfig should be doing this by default. I keep seeing PRs like this.

erietz-werner:
Yes, it is in a ton of the server configurations. That is where I got the idea from. `grep -R "vim.fn.has 'win32'" | wc -l`

PR #2014, opened 2022-07-19
feat(denols): add deno task

sigmaSd:


sigmaSd:
task

sigmaSd:
This probably belongs to a separate plugin.

Issue #2011, opened 2022-07-17
generate root_dir docs from source code


justinmk:
## Problem

https://github.com/neovim/nvim-lspconfig/issues/1742 reported that the generated docs for `root_dir` are sometimes invalid. The response was https://github.com/neovim/nvim-lspconfig/commit/9dc02492c4a457479f8a0ec7a65aac1852ff59c0 which just removed the generation: https://github.com/neovim/nvim-lspconfig/commit/733fe92baafb23cb413300571e31cb1fec962b1a

Besides being cumbersome, the manual documentation leads to problems like: https://github.com/neovim/nvim-lspconfig/pull/2006

## Solution

1. Restore code generation removed in https://github.com/neovim/nvim-lspconfig/commit/9dc02492c4a457479f8a0ec7a65aac1852ff59c0 and...
    1. make it _link to the source_ instead of trying to parse it (e.g. `[path/to/foo.lua](https://github.com/neovim/nvim-lspconfig/blob/…/foo.lua#L132-L149)`)
    1. or just fix the parsing.
1. Use treesitter to parse the file?

sersorrel:
this would also fix e.g. [rnix's docs](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#rnix) saying that root_dir defaults to "vim's starting directory" when in fact it's `util.find_git_ancestor(fname) or vim.loop.os_homedir()`, which is very much not the same thing

PR #2018, opened 2022-07-20
feat(omnisharp): add more settings

williamboman:
As I was reviewing the Omnisharp LSP I found these configuration items that I
figured could be ported to lspconfig as well.

I've tested this new config in some sample projects (while comparing to
VSCode's C# plugin behaviour), seems to work fine!

I don't do any C# so I'm not sure if I'm missing something, but I believe this
should be fully backwards compatible.

These have primarily been sourced from the VSCode extension [1].

[1] https://github.com/OmniSharp/omnisharp-vscode/blob/88ad237e1e904ad0f91d2de39c30132198a02dcb/src/omnisharp/server.ts#L345-L434

### DIFF
--- a/lua/lspconfig/server_configurations/omnisharp.lua
+++ b/lua/lspconfig/server_configurations/omnisharp.lua
@@ -85,27 +85,51 @@ https://github.com/omnisharp/omnisharp-roslyn
 OmniSharp server based on Roslyn workspaces
 
 `omnisharp-roslyn` can be installed by downloading and extracting a release from [here](https://github.com/OmniSharp/omnisharp-roslyn/releases).
-Omnisharp can also be built from source by following the instructions [here](https://github.com/omnisharp/omnisharp-roslyn#downloading-omnisharp).
+OmniSharp can also be built from source by following the instructions [here](https://github.com/omnisharp/omnisharp-roslyn#downloading-omnisharp).
 
-Omnisharp requires the [dotnet-sdk](https://dotnet.microsoft.com/download) to be installed.
+OmniSharp requires the [dotnet-sdk](https://dotnet.microsoft.com/download) to be installed.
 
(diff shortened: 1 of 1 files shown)


williamboman:
I realized the docs probably needs to be updated

justinmk:
> I realized the docs probably needs to be updated

see also https://github.com/neovim/nvim-lspconfig/issues/2011

lua/lspconfig/server_configurations/omnisharp.lua
```diff
@@ -2,47 +2,134 @@ local util = require 'lspconfig.util'
...
       return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname)
     end,
     on_new_config = function(new_config, new_root_dir)
-      if new_root_dir then
-        table.insert(new_config.cmd, '-s')
-        table.insert(new_config.cmd, new_root_dir)
+      table.insert(new_config.cmd, '-z') -- https://github.com/OmniSharp/omnisharp-vscode/pull/4300
+      vim.list_extend(new_config.cmd, { '-s', new_root_dir })
+      vim.list_extend(new_config.cmd, { '--hostPID', tostring(vim.fn.getpid()) })
+      table.insert(new_config.cmd, 'DotNet:enablePackageRestore=false')
```
MikaelElkiaer, line 48: @williamboman What is the reason for package restore to be disabled by default? I know it has had a turbulent history, is it still not mature enough for usage?

Alternatively, it could be turned into a config property like the others you added, but kept off by default, and possibly with a warning description.

williamboman: > @williamboman What is the reason for package restore to be disabled by default? I know it has had a turbulent history, is it still not mature enough for usage?

It's taken from how the VSCode extension launches the server, no other reason than that.

> Alternatively, it could be turned into a config property like the others you added, but kept off by default, and possibly with a warning description.

Sounds reasonable to me, I'm not a maintainer though. I'm sure if you open a PR it'd get merged

MikaelElkiaer: Thanks for clarifying, will try it out and see if it's worth it.

PR #2026, opened 2022-07-24
feat(denols): Auto complete import from deno.land registery

sigmaSd:
import

The correct way to do this would be similar to https://github.com/denoland/vscode_deno/blob/main/client/src/notification_handlers.ts#L10

- listen to registry state notification
- ask the user if they want to enable completion suggestions
- update the configuration settings

The problem is with that last step I couldn't find a way to update settings at runtime in vim lsp, doing something like `client.config.init_options... = newvalue`, doesn't seem to get noticed by the server even after using `didchangeconfiguration` method.

It seems like `init_options` is something that is not in the spec ? and specific to vim lsp ? it seems to not be update-able

This pr propose to enable suggestions unconditionally form `deno.land`, it is the official provider so it can be trusted and I don't see a disadvantage for enabling it.

This is not a new thing, we already enable some options unconditionally https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/denols.lua#L81 the official vscode plugin actually asks the user before settings those.

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -63,7 +63,7 @@ local function denols_handler(err, result, ctx)
   lsp.handlers[ctx.method](err, result, ctx)
 end
 
-return {
+local denols = {
   default_config = {
     cmd = { 'deno', 'lsp' },
     -- single file support is required for now to make the lsp work correctly, see #2000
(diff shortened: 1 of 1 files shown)


sigmaSd:
This pr workaround can be applied outside of this plugin, so I guess no need to add it here.

Example:
```lua
local denols_setup = {
    on_attach = on_attach,
    capabilities = capabilities,
    init_options = {
        unstable = true,
        suggest = { imports = { hosts = {} } },
    },
}
local deno_land_host = {}
deno_land_host['https://deno.land'] = true
denols_setup.init_options.suggest.imports.hosts = deno_land_host
lspconfig.denols.setup(denols_setup)
```

PR #2027, opened 2022-07-24
feat(denols): handle registeryState notification

sigmaSd:
Port of https://github.com/denoland/vscode_deno/blob/45d343516ab1250867a5cb254460278f0ceca2a2/client/src/notification_handlers.ts

This is a more appropriate way to handle https://github.com/neovim/nvim-lspconfig/pull/2026

import


sigmaSd:
This pr also replaces init_options with settings, they seems to be the same in case of deno lsp, with settings being updateable at runtime.

sigmaSd:
This can be done in a separate plugin

PR #2038, opened 2022-07-28
fix(lemminx): Change documented binary download location

datho7561:
We no longer publish the lemminx binaries to download.jboss.org as a part of our release process for vscode-xml, so the binaries for the latest releases of lemminx are not there. However, they are available on the [vscode-xml GitHub](https://github.com/redhat-developer/vscode-xml), under releases.

Signed-off-by: David Thompson <davthomp@redhat.com>


justinmk:
Thanks!

PR #2041, opened 2022-07-30
fix(gopls): engage gopls on `go.work` files

omertuc:
https://go.dev/ref/mod#go-work-file

Seems like this should have been done as part of 5434e285d7a1cf65870d57c88201cc6b9fb6a239
but was missed


omertuc:
@lithammer

PR #2051, opened 2022-08-05
refactor: rebuild lsp command by api

### LINKED ISSUE #2068: Add a setup option to configure border of the floating window of LspInfo and others
As per discussion in gitter/matrix, @glepnir asked to create an issue and tag him, and promised to inmpement it later :grin:

glepnir:
cc @justinmk 

Fix #2068


ranjithshegde:
@glepnir #1984 already includes some of these changes

plugin/lspconfig.lua
```diff
@@ -0,0 +1,58 @@
...
+local lsp_get_active_client_ids = function()
+  return vim.tbl_map(function(client)
+    return ('%d (%s)'):format(client.id, client.name)
+  end, require('lspconfig.util').get_managed_clients())
+end
+
+if vim.fn.exists 'g:lspconfig' == 1 then
+  return
+end
+vim.g.lspconfig = 1
```
ii14, line 13: Can you move it to the top? So defining `g:lspconfig` will disable it entirely.

plugin/lspconfig.lua
```diff
@@ -0,0 +1,58 @@
+local api = vim.api
+local lspconfig = require 'lspconfig'
```
ii14, line 2: Can `require 'lspconfig'` be inlined in all places where `lspconfig` is used? Then we don't force the module to be loaded.

glepnir: done

plugin/lspconfig.lua
```diff
@@ -0,0 +1,63 @@
...
+  local lspconfig = require 'lspconfig'
+  lspconfig.lsp_start(args.args)
+end, {
+  nargs = '?',
+  complete = function(arg)
+    local lspconfig = require 'lspconfig'
+    local list = lspconfig.available_servers()
+    return vim.tbl_filter(function(s)
+      return string.match(s, '^' .. arg)
+    end, list)
```
ii14, line 40:
```suggestion
    return vim.tbl_filter(function(s)
      return s:sub(1, #arg) == arg
    end, require 'lspconfig'.available_servers())
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,63 @@
...
+api.nvim_create_user_command('LspLog', function()
+  vim.cmd(string.format('tabnew %s', vim.lsp.get_log_path()))
+end, {
+  nargs = 0,
+  desc = '`:LspLog` Opens the Nvim LSP client log.',
+})
+
+api.nvim_create_user_command('LspStart', function(args)
+  local lspconfig = require 'lspconfig'
+  lspconfig.lsp_start(args.args)
```
ii14, line 32:
```suggestion
  require 'lspconfig'.lsp_start(args.args)
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,63 @@
...
+    end, list)
+  end,
+  desc = '`:LspStart` Manually launches a language server.',
+})
+
+api.nvim_create_user_command('LspStop', function(cmd_args)
+  local lspconfig = require 'lspconfig'
+  for _, client in ipairs(lspconfig.util.get_clients_from_cmd_args(cmd_args)) do
+    client.stop()
+  end
```
ii14, line 49:
```suggestion
  for _, client in ipairs(require 'lspconfig'.util.get_clients_from_cmd_args(cmd_args)) do
    client.stop()
  end
```

plugin/lspconfig.lua
```diff
@@ -0,0 +1,63 @@
...
+end, {
+  nargs = '?',
+  complete = lsp_get_active_client_ids,
+  desc = '`:LspStop` Manually stops the given language client(s).',
+})
+
+api.nvim_create_user_command('LspRestart', function(cmd_args)
+  local lspconfig = require 'lspconfig'
+  lspconfig.lsp_restart(cmd_args)
```
ii14, line 58:
```suggestion
  require 'lspconfig'.lsp_restart(cmd_args)
```

justinmk:
let's revisit after https://github.com/neovim/nvim-lspconfig/pull/1984 ? Though maybe this isn't needed if `nvim_create_user_command` serves the purpose well enough? 

What's the use case that isn't covered by `nvim_create_user_command` ?

plugin/lspconfig.lua
glepnir: done

glepnir:
This is different from #1984. I just replaced lspconfig.vim. and  fixed #2068  now we can use `LspInfo border=single` to set the border . why not provide a config to set the border. because I think the `lspconfig` just provide the server config. compare with `vim.cmd` use string format to generate the `autocmd` use api will be more better:) and  neovim use lua file in everywhere now :)

justinmk (approved):
yeah, this looks like good changes.

plugin/lspconfig.lua
ii14: @glepnir you've missed this one

glepnir: oh sry. fix

ranjithshegde:
This conflicts with #1984 because `lspconfig.lua` logic introduced in that PR is incompatible. All these `user_commands` are moved to `plugin/lspconfig.lua`, with `lua/lspconfig.lua` just creatin/assigning a metatable for lsp_configurations.

I would recommend please wait till #1984 is merged. then this PR is simply modifying the `border` key and make it accessible by `fargs/nargs`

glepnir:
@ranjithshegde  because you don't reply my review so I continue work on this . maybe this will be merge or not @ii14  what do you think?

ranjithshegde:
> @ranjithshegde because you don't reply my review so I continue work on this . maybe this will be merge or not @ii14 what do you think?

My apologies. For whatever reason I wasn't subscribed to this thread. 

If you could please wait until that is merged, then the job of this PR will be very very simple. All the logic will already have been introduced in #1984,
 then you will simply have to make the `border` key modifiable by user.

glepnir:
> now we can use LspInfo border=single to set the border . why not provide a config to set the border. because I think the lspconfig just provide the server config.

this project only provide the server config. not provide any other config should be better. so pass the argument into the cmdline is my way. not use a `border` key config .

ii14:
> what do you think?

I don't know, I'm only reviewing the code :)

PR #2053, opened 2022-08-05
feat: add nil support

oxalica:
https://github.com/oxalica/nil

A new language server for Nix Expression Language.

~The name `nil` needs special handling since it's a keyword in Lua.~ Field name is now `nil_ls`.


lithammer:
Maybe just append `ls` or `_ls` instead of special handling?

oxalica:
> Maybe just append ls or _ls instead of special handling?

Renamed to `nil_ls` now.

oxalica:
Enabled single file support.

PR #2066, opened 2022-08-15
fix: only consider servers that have been set up as available

williamboman:
Currently, as soon as a server configuration is indexed it will be considered
"available". I'm not entirely sure if this is indeed intended behavior, all I
know is that it creates some confusion.

I figured I'd open a PR with a suggested solution immediately, to get the ball
rolling. Another solution would be to keep "available servers" as-is, and
instead introduce a new function `.configured_servers()`, and replace usage of
`.available_servers()` where appropriate (e.g., the server list at the bottom
of the `:LspInfo` window).

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -7,7 +7,13 @@ local M = {
 M._root = {}
 
 function M.available_servers()
-  return vim.tbl_keys(configs)
+  local servers = {}
+  for server, config in pairs(configs) do
+    if config.manager ~= nil then
+      table.insert(servers, server)
+    end
+  end
+  return servers
 end
 
 -- Called from plugin/lspconfig.vim because it requires knowing that the last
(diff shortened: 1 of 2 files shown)


glepnir:
LGTM. @williamboman  does this pr finished? if no other task I will merge

williamboman:
> LGTM. @williamboman does this pr finished? if no other task I will merge

Yeah it's finished!

lua/lspconfig.lua
```diff
@@ -7,7 +7,13 @@ local M = {
 M._root = {}
 
 function M.available_servers()
-  return vim.tbl_keys(configs)
+  local servers = {}
+  for server, config in pairs(configs) do
+    if config.manager ~= nil then
+      table.insert(servers, server)
```
justinmk, line 13: nit: I think vim.tbl_filter could save some code here

williamboman: It needs to access both the key and the value of the `configs` table, I assumed `tbl_filter` only passes a single argument (`value`) to the predicate function?

justinmk: oh... that's unfortunate.

Issue #2069, opened 2022-08-17
denols: root_dir ignore util.root_pattern [reopen]


xdeafbabe:
### Description

What title says. I have both `denols` and `tsserver` in my Neovim configuration, and whenever I open a JS or TS file, `denols` starts, even if I do not have `deno.json` file in my project directory. Reopening #2015, as it was closed without properly acknowledging and addressing the issue.

### Neovim version

```
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by team+vim@tracker.debian.org

Features: +acl +iconv +tui
```

### Nvim-lspconfig version

da7461b596d70fa47b50bf3a7acfaef94c47727d

### Operating system and version

Ubuntu 20.04.4

### Affected language servers

denols, tsserver

### Steps to reproduce

1. Have both denols and tsserver used in your Neovim configuration;
2. Separate them by specifying root dir for each - `deno.json` for denols and `package.json` for tsserver;
3. Open any directory not containing `deno.json` file in Neovim. Presence of `package.json` does not matter;
4. Open a TypeScript or JavaScript file;

### Actual behavior

5. See that `denols` is running.

`denols` also runs with `tsserver` simultaneously if there is `package.json` in the directory.

### Expected behavior

5. `denols` should not run.

### Minimal config

```Lua
local lspconfig = require('lspconfig')
local capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())
local illuminate = require('illuminate')
local on_attach = function(client) illuminate.on_attach(client) end

lspconfig.denols.setup({
  root_dir = lspconfig.util.root_pattern('deno.json', 'deno.jsonc'),
  on_attach = on_attach,
  capabilities = capabilities,
})

lspconfig.tsserver.setup({
  root_dir = lspconfig.util.root_pattern('package.json'),
  on_attach = on_attach,
  capabilities = capabilities,
})
```

### LSP log

https://gist.github.com/xdeafbabe/684128f50e240247f1825a07097ca336

xdeafbabe:
@sigmaSd I'd prefer to keep this issue open til the problem is actually resolved without any workarounds from users

sigmaSd:
should be fixed in https://github.com/neovim/nvim-lspconfig/pull/2070

xdeafbabe:
Removed `single_file_support = false` from my config and updated. Works fine now, thank you!

PR #2070, opened 2022-08-17
fix(denols): revert #2000

sigmaSd:
I found the root cause of denols problems its https://github.com/neovim/neovim/pull/19797

With that pr applied (and with this revert) denols works perfectly almost identical to vscode deno, its missing some features but I imagine those are out of scope for lspconfig.

Applying the revert now means that until nvim merges a fix https://github.com/neovim/nvim-lspconfig/issues/1998 and other problems will reappear.

I think that's an ok tradeoff because using `single_file_support=true` breaks workflow for other ppl see https://github.com/neovim/nvim-lspconfig/issues/2069

So my opinion is to just merge this and the user will have to wait for the fix in neovim it self.


justinmk:
Makes sense. Please rebase then will merge.

sigmaSd:
done

PR #2074, opened 2022-08-21
the funtion formatting is deprecated

ayoubelmhamdi:
I think we should use the 
```lua
vim.lsp.buf.format { async = true }
```


ubitux:
This is re-introducing the same regression fixed in 134a291fecc5c1fa5d01e398d2d6ef03b3dc5149

justinmk:
Think it requires Nvim 0.8. So yeah should revert this until format() is in stable release.

PR #2082, opened 2022-08-24
fix: LspStop should be only stop the current buffer servers by default

### LINKED ISSUE #1781: Apply `:LspStop` and `:LspRestart` to current buffer only

glepnir:
Fix #1781


kabouzeid:
This broke stopping clients that were manually started with `:LspStart` but whose filetypes don't match the current buffer.

glepnir:
what is mean?  provide reproduce your issue step

kabouzeid:
`:LspStart foo` works, even if foo doesn't list the current buffers filetype in its config. But then `:LspStop foo` won't stop the client.

The following example doesn't make much sense, but just for you to try out:
```
vim test.md
:LspStart sumneko_lua
:LspStop 1 (sumneko_lua)
```

PR #2081, opened 2022-08-24
feat: improve LspInfo

glepnir:
- the servers list always out of window when there are many servers. it should be show in a newline

- set default border and support custom border by `LspInfo border=double or other style`

- two highlight group are added `LspInfoBorder` `LspInfoList` set the servers list highlight default link to `Visual`

- replace the string format by use api

after:
image

Fix #2068

PR #2083, opened 2022-08-24
fix: use util.available_servers() instead of manually indexing configs

williamboman:
See [#2066][2066] for more background.

[2066]: https://github.com/neovim/nvim-lspconfig/pull/2066


glepnir:
maybe I can add this in  #2081 there has conflict

williamboman:
:+1: let's do that instead

PR #2084, opened 2022-08-24
feat: sort autocompletion items alphabetically

williamboman:
I figured it'd be nice to have an alphabetical sorting here, makes it much
easier to scan the list.


plugin/lspconfig.lua
```diff
@@ -14,20 +14,25 @@ if vim.fn.has 'nvim-0.7' ~= 1 then
   return
 end
 
+local completion_sort = function (items)
```
ii14, line 17: linter
```suggestion
local completion_sort = function(items)
```

Issue #1937, opened 2022-06-01
Commands interface


Shatur:
We currently have API which allows to set language server specific commands. Example from clangd:
https://github.com/neovim/nvim-lspconfig/blob/c5dae15c0c94703a0565e8ba35a9f5cb96ca7b8a/lua/lspconfig/server_configurations/clangd.lua#L52-L59

But @justinmk [suggested](https://github.com/neovim/nvim-lspconfig/pull/1921#issuecomment-1143705517) to remove them.
I personally find them useful, especially `ClangdSwitchSourceHeader`. I don't think it's good for users to copy them from other people's configs. And having a separate plugin for a few lines of code for `ClangdSwitchSourceHeader` is also weird. So I think having them here is the right place.

So it's more like a feature keep request :)

mellery451:
I very much like and need this specific functionality, but I can understand wanting to keep the core clean. If anything, maybe we can come up with a plan to move this to a community plugin like https://github.com/p00f/clangd_extensions.nvim

Shatur:
There are extensions for other LSP servers, having a separate plugin for some additional commands doesn't sound very convenient. Many people complain about the complexity of nvim-lspconfig, forcing them to install additional plugins to get the basic functionality, like `ClangdSwitchSourceHeader`,  which isn't good.

And I don't understand how the commands feature can be a maintenance issue. It's just a few lines of code and many people maintain the language servers they use. If a command breaks - the people who use it will fix it.

justinmk:
> And I don't understand how the commands feature can be a maintenance issue. It's just a few lines of code

Multiplied by 100 or more. I would like to know if we can auto-complete server extensions instead.

Shatur:
> Multiplied by 100 or more--on top of the other activity already.

Users maintain language servers they add. Why it becomes a problem to maintain 1-2 commands (if they exists, most servers don't have it) for the language server they use? How it different from maintaining the language server configuration?

> which is, again, strictly a set of quickstart configs, not a factory for all things that could possibly be useful for each random langserver.

Commands was always here. Why break people configs? Is there any maintenance problem with this feature? I would consider commands as a part of server configuration (as it was in VSCode when you installing a plugin for specific language).

> I would like to know if we can auto-complete server extensions instead. That's worth having in core.

Not sure how it could work.

justinmk:
[...]

PR #2089, opened 2022-08-25
feat(stylelint): add `StylelintFixAll` command

bennypowers:


lua/lspconfig/server_configurations/stylelint_lsp.lua
```diff
@@ -7,6 +7,37 @@ if vim.fn.has 'win32' == 1 then
...
+  opts = opts or { sync = true, bufnr = 0 }
+  local bufnr = util.validate_bufnr(opts.bufnr or 0)
+
+  local stylelint_lsp_client = util.get_active_client_by_name(bufnr, 'stylelint_lsp')
+  if stylelint_lsp_client == nil then
+    return
+  end
+
+  local request
+  if opts.sync then
```
justinmk, line 20: This kind of thing is why in https://github.com/neovim/nvim-lspconfig/issues/1937 I'm very hesitant to add these special-case commands to each lspconfig. It's unpleasant for users to have a different interface or every random command. Here we have a function with an "async" flag. Looks like it's always true currently. But this is a lot of code, and only for one command. I don't see how we can maintain 100s of configs with this much code.

bennypowers: for the record I don't like either of these commands and I think both are workarounds.

That being said, I just want to fix my stylelint errors on save. I have no idea why `autoFixOnSave` and `autoFixOnFormat` are not respected, but this config and an aucmd will will solve the problem for users.

bennypowers:
I agree with @glepnir and @justinmk's comments about how bloated this repo is and that the goal should be to release language-specific plugins

That being said, I still think this should be merged, in order to support the simple user who just wants to format and fix their file on save, rather than telling them to wait for a big rewrite.

glepnir:
@bennypowers  if this is a necessary improve it can be :) I am not familiar the stylelint need check and test it .if anyone can give some advice very thanks then.

bennypowers:
set neovim up as described in this pr (with autocmds), then
```bash
git clone https://github.com/bennypowers/nvim-stylelint-repro
cd nvim-stylelint-repro
npm ci
vim styles.css
```

write file

expect file contents to equal

```css
body {
  background-color: red;
}
```

lua/lspconfig/server_configurations/stylelint_lsp.lua
gegoune: @bennypowers 
```lua
require('lspconfig')['stylelint_lsp'].setup({
    settings = {
      stylelintplus = {
        autoFixOnFormat = true,
      },
    },
})
```
plus lukas-reineke/lsp-format.nvim seems to be fixing stylelint errors for me.

glepnir:
but as the justin said there has vim.lsp.handler and opts. sync so I need follow the justin his advice.

lua/lspconfig/server_configurations/stylelint_lsp.lua
bennypowers: Thanks for the link, but this pr is for support in lspconfig without an additional plugin

I'm definitely going to check that out for my own config tho

justinmk:
> That being said, I still think this should be merged, in order to support the simple user who just wants to format and fix their file on save, rather than telling them to wait for a big rewrite.

That's always the justification provided for each of these new features. This is how scope-creep occurs.

Not seeing a strong justification here. 

We need a very _thin_ approach (2-3 lines of code, not 50+ as in this PR) to adding support for custom (off-spec) server features. Until that happens (perhaps never), we need to stop adding custom commands. https://github.com/neovim/nvim-lspconfig/issues/1937

PR #2092, opened 2022-08-26
fix: remove the config.commands

### LINKED ISSUE #1750: How to set per-client commands to start_client?
### Description

neovim/neovim#16101 makes it possible to use per-client command handlers.
However, there seems to be no way to actually pass this to `start_client` call.

### Neovim version

NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

c276536bcdb03f2e1c353b04abb1a719814d4197

ArchLinux 5.16.11-zen1-1-zen

### Affected language servers

all

### Steps to reproduce

> The `setup{}` function takes a table which contains a superset of the keys
> listed in `:help vim.lsp.start_client()` with the following unique entries

Per documentation, the following should work, but in fact doesn't:

```lua
require('lspconfig').ltex.setup {
    commands = {
        ['_ltex.addToDictionary'] = function() end,
    }
}
```

Because lspconfig takes the `commands` key as a different meaning, as explained in the `lspconfig-configurations`
section of the documentation.

> `commands` is a map of `name:definition` key:value pairs, where `definition`
> is a list whose first value is a function implementing the command, and the
> rest are either array values which will be formed into flags for the command,
> or special keys like `description`.

So the above code errors out as it doesn't pass the check.

I'm aware of `vim.lsp.commands` as a workaround but I think it should be possible to make use of neovim/neovim#16101 and set custom command handlers.
Or at least the documentation should be updated to mention the different meaning of the `commands` key when passed to `setup`.

### Actual behavior

Error during start up:
```
...m/site/pack/paqs/opt/lspconfig/lua/lspconfig/configs.lua:46: attempt to index local 'v' (a function value)
```

### Expected behavior

_No response_

### Minimal config

```Lua

  return result
end

local install_path = join_paths(package_root, 'packer', 'start', 'packer.nvim')
local compile_path = join_paths(install_path, 'plugin', 'packer_compiled.lua')

local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      'neovim/nvim-lspconfig',
    },
    config = {
      package_root = package_root,
      compile_path = compile_path,
    },
  }
end

_G.load_config = function()
  vim.lsp.set_log_level 'trace'
  if vim.fn.has 'nvim-0.5.1' == 1 then
  end
    local function buf_set_keymap(...)
      vim.api.nvim_buf_set_keymap(bufnr, ...)
    end
    end

    -- Mappings.
    local opts = { noremap = true, silent = true }
    buf_set_keymap('n', 'gD', '<Cmd>lua vim.lsp.buf.declaration()<CR>', opts)
    buf_set_keymap('n', 'gd', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
    buf_set_keymap('n', 'K', '<Cmd>lua vim.lsp.buf.hover()<CR>', opts)
    buf_set_keymap('n', 'gi', '<cmd>lua vim.lsp.buf.implementation()<CR>', opts)
    buf_set_keymap('n', '<C-k>', '<cmd>lua vim.lsp.buf.signature_help()<CR>', opts)
    buf_set_keymap('n', '<space>wa', '<cmd>lua vim.lsp.buf.add_workspace_folder()<CR>', opts)
    buf_set_keymap('n', '<space>wr', '<cmd>lua vim.lsp.buf.remove_workspace_folder()<CR>', opts)
    buf_set_keymap('n', '<space>wl', '<cmd>lua print(vim.inspect(vim.lsp.buf.list_workspace_folders()))<CR>', opts)
    buf_set_keymap('n', '<space>D', '<cmd>lua vim.lsp.buf.type_definition()<CR>', opts)
    buf_set_keymap('n', '<space>rn', '<cmd>lua vim.lsp.buf.rename()<CR>', opts)
    buf_set_keymap('n', 'gr', '<cmd>lua vim.lsp.buf.references()<CR>', opts)
    buf_set_keymap('n', '<space>e', '<cmd>lua vim.lsp.diagnostic.show_line_diagnostics()<CR>', opts)
    buf_set_keymap('n', '[d', '<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>', opts)
    buf_set_keymap('n', ']d', '<cmd>lua vim.lsp.diagnostic.goto_next()<CR>', opts)
    buf_set_keymap('n', '<space>q', '<cmd>lua vim.lsp.diagnostic.set_loclist()<CR>', opts)
  end

  local name = 'ltex'
  if not name then
    print 'You have not defined a server name, please edit minimal_init.lua'
  end
    print [[You have not defined a server default cmd for a server
      that requires it please edit minimal_init.lua]]
  end

  nvim_lsp[name].setup {
    cmd = cmd,
    on_attach = on_attach,
    commands = {
      ['_ltex.addToDictionary'] = function() end,
    }
  }

  print [[You can find your log at $HOME/.cache/nvim/lsp.log. Please paste in a github issue under a details tag as described in the issue template.]]
end

  vim.fn.system { 'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path }
  load_plugins()
  require('packer').sync()
  vim.cmd [[autocmd User PackerComplete ++once lua load_config()]]
else
  load_plugins()
  require('packer').sync()
  _G.load_config()
end
```

### LSP log

https://gist.github.com/Aetf/a7279f4adde759bb2060aeb773961b42

glepnir:
now we can use `vim.api.nvim_create_user_command` in `on_attach` field to create commands just work for this server. so 
there is no need this field `config.commands` .

Example:
```lua
  local function organize_imports()
    local params = {
      command = 'pyright.organizeimports',
      arguments = { vim.uri_from_bufnr(0) },
    }
    vim.lsp.buf.execute_command(params)
  end

    if client.name == "pyright" then
      vim.api.nvim_create_user_command("PyrightOrganizeImports", organize_imports, {desc = 'Organize Imports'})
    end
  end

  require("lspconfig")['pyright'].setup({
    on_attach = on_attach
  })
```

Fix #1750

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -278,29 +278,6 @@ where the structure of the docs table is as follows:
     }
 <
 
-Custom commands only work for servers
-
-Example:
->
-  local function organize_imports()
-    local params = {
-      command = 'pyright.organizeimports',
-      arguments = { vim.uri_from_bufnr(0) },
-    }
-    vim.lsp.buf.execute_command(params)
-  end
-
-  local on_attach = function(client, bufnr)
-    if client.name == "pyright" then
-      vim.api.nvim_create_user_command("PyrightOrganizeImports", organize_imports, {desc = 'Organize Imports'})
-    end
-  end
-
-  require("lspconfig")['pyright'].setup({
-    on_attach = on_attach
-  })
-<
-
 The `configs.__newindex` metamethod consumes the config definition and returns
 an object with a `setup()` method, to be invoked by users:
 >
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -267,7 +267,6 @@ validating using `vim.validate` is:
       default_config = {'t'},
       on_new_config = {'f', true},
       on_attach = {'f', true},
-      commands = {'t', true},
       docs = {'t', true},
     }
 <
@@ -278,13 +277,8 @@ where the structure of the docs table is as follows:
       default_config = {'t', true},
     }
 <
-`commands` is a map of `name:definition` key:value pairs, where `definition`
-is a list whose first value is a function implementing the command, and the
-rest are either array values which will be formed into flags for the command,
-or special keys like `description`.
 
-Warning: Commands is deprecated and will be removed in future releases.
-It is recommended to use `vim.api.nvim_create_user_command()` instead in an `on_attach` function.
+Custom commands only work for servers
 
 Example:
 >
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -62,9 +62,9 @@ end
 
 ---@deprecated
 function M.create_module_commands(module_name, commands)
-  vim.deprecate('util.create_module_commands',nil,'0.1.4','vim.api.nvim_create_user_command')
+  vim.deprecate('util.create_module_commands', nil, '0.1.4', 'vim.api.nvim_create_user_command')
   for command_name, def in pairs(commands) do
-    local _,opts = unpack(def)
+    local _, opts = unpack(def)
     api.nvim_create_user_command(command_name, function(info)
       require('lspconfig')[module_name].commands[command_name][1](unpack(info.fargs))
     end, opts)
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -60,17 +60,6 @@ function M.add_hook_after(func, new_fn)
   end
 end
 
----@deprecated
-function M.create_module_commands(module_name, commands)
-  vim.deprecate('util.create_module_commands', nil, '0.1.4', 'vim.api.nvim_create_user_command')
-  for command_name, def in pairs(commands) do
-    local opts = def[2]
-    api.nvim_create_user_command(command_name, function(info)
-      require('lspconfig')[module_name].commands[command_name][1](unpack(info.fargs))
-    end, opts)
-  end
-end
-
 -- Some path utilities
 M.path = (function()
   local is_windows = uv.os_uname().version:match 'Windows'
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -64,7 +64,7 @@ end
 function M.create_module_commands(module_name, commands)
   vim.deprecate('util.create_module_commands', nil, '0.1.4', 'vim.api.nvim_create_user_command')
   for command_name, def in pairs(commands) do
-    local _, opts = unpack(def)
+    local opts = def[2]
     api.nvim_create_user_command(command_name, function(info)
       require('lspconfig')[module_name].commands[command_name][1](unpack(info.fargs))
     end, opts)
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -60,35 +60,11 @@ function M.add_hook_after(func, new_fn)
   end
 end
 
--- Maps lspconfig-style command options to nvim_create_user_command (i.e. |command-attributes|) option names.
-local opts_aliases = {
-  ['description'] = 'desc',
-}
-
----@param command_definition table<string | integer, any>
-function M._parse_user_command_options(command_definition)
-  ---@type table<string, string | boolean | number>
-  local opts = {}
-  for k, v in pairs(command_definition) do
-    if type(k) == 'string' then
-      local attribute = k.gsub(k, '^%-+', '')
-      opts[opts_aliases[attribute] or attribute] = v
-    elseif type(k) == 'number' and type(v) == 'string' and v:match '^%-' then
-      -- Splits strings like "-nargs=* -complete=customlist,v:lua.something" into { "-nargs=*", "-complete=customlist,v:lua.something" }
-      for _, command_attribute in ipairs(vim.split(v, '%s')) do
-        -- Splits attribute into a key-value pair, like "-nargs=*" to { "-nargs", "*" }
-        local attribute, value = unpack(vim.split(command_attribute, '=', { plain = true }))
-        attribute = attribute.gsub(attribute, '^%-+', '')
-        opts[opts_aliases[attribute] or attribute] = value or true
-      end
-    end
-  end
-  return opts
-end
-
+---@deprecated
 function M.create_module_commands(module_name, commands)
+  vim.deprecate('util.create_module_commands',nil,'0.1.4','vim.api.nvim_create_user_command')
   for command_name, def in pairs(commands) do
-    local opts = M._parse_user_command_options(def)
+    local _,opts = unpack(def)
     api.nvim_create_user_command(command_name, function(info)
       require('lspconfig')[module_name].commands[command_name][1](unpack(info.fargs))
     end, opts)
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -10,18 +10,7 @@ function configs.__newindex(t, config_name, config_def)
     default_config = { config_def.default_config, 't' },
     on_new_config = { config_def.on_new_config, 'f', true },
     on_attach = { config_def.on_attach, 'f', true },
-    commands = { config_def.commands, 't', true },
   }
-  if config_def.commands then
-    for k, v in pairs(config_def.commands) do
-      validate {
-        ['command.name'] = { k, 's' },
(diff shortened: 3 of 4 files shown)


doc/lspconfig.txt
```diff
@@ -278,13 +277,8 @@ where the structure of the docs table is as follows:
...
     }
 <
-`commands` is a map of `name:definition` key:value pairs, where `definition`
-is a list whose first value is a function implementing the command, and the
-rest are either array values which will be formed into flags for the command,
-or special keys like `description`.
 
-Warning: Commands is deprecated and will be removed in future releases.
-It is recommended to use `vim.api.nvim_create_user_command()` instead in an `on_attach` function.
+Custom commands only work for servers
```
ii14, line 281:
```suggestion
Custom commands only works for servers
```
I think that entire section about custom commands is not needed anymore?

lua/lspconfig/util.lua
```diff
@@ -60,35 +60,11 @@ function M.add_hook_after(func, new_fn)
...
-  end
-  return opts
-end
-
+---@deprecated
 function M.create_module_commands(module_name, commands)
+  vim.deprecate('util.create_module_commands', nil, '0.1.4', 'vim.api.nvim_create_user_command')
   for command_name, def in pairs(commands) do
-    local opts = M._parse_user_command_options(def)
+    local _, opts = unpack(def)
```
ii14, line 67:
```suggestion
    local opts = def[2]
```

doc/lspconfig.txt
glepnir: I think it can help some people so keep it . like some new vimer this can give them a tip

ii14: I doubt new people are going to define custom commands. This is an advanced thing to do, and for that they should refer to neovim LSP documentation.

glepnir: got it. remove

lua/lspconfig/util.lua
glepnir: done

lua/lspconfig/util.lua
```diff
@@ -60,35 +60,11 @@ function M.add_hook_after(func, new_fn)
...
-        attribute = attribute.gsub(attribute, '^%-+', '')
-        opts[opts_aliases[attribute] or attribute] = value or true
-      end
-    end
-  end
-  return opts
-end
-
+---@deprecated
 function M.create_module_commands(module_name, commands)
```
ii14, line 64: As far as I can tell, this function was only used in `configs.lua` and was not documented. Can't we just get rid of it?

glepnir: yup. But I'm not sure if this function is used by anyone else. Because he is derived. so i didn't remove it

ii14: Took a quick search on github, and seems like there are only lspconfig forks. I think it's okay to get rid of it entirely

glepnir: done .

ranjithshegde:
> now we can use `vim.api.nvim_create_user_command` in `on_attach` field to create commands just work for this server. so there is no need this field `config.commands` .
> 
> Example:
> 
> ```lua
>   local function organize_imports()
>     local params = {
>       command = 'pyright.organizeimports',
>       arguments = { vim.uri_from_bufnr(0) },
>     }
>     vim.lsp.buf.execute_command(params)
>   end
> 
>   local on_attach = function(client, bufnr)
>     if client.name == "pyright" then
>       vim.api.nvim_create_user_command("PyrightOrganizeImports", organize_imports, {desc = 'Organize Imports'})
>     end
>   end
> 
>   require("lspconfig")['pyright'].setup({
>     on_attach = on_attach
>   })
> ```
> 
> Fix #1750

I am a bit too late for this. But this breaks all the internal commands like `texlab-build`. 
The above example only works if the client provides `workspace/executeCommands` that dont require `params` field. 
In the case of `clangdSwitchSourceheader` and all `texlab` commands, you need to write this manually like this

```lua
local texlab_build_status = vim.tbl_add_reverse_lookup {
  Success = 0,
  Error = 1,
  Failure = 2,
  Cancelled = 3,
}

local texlab_forward_status = vim.tbl_add_reverse_lookup {
  Success = 0,
  Error = 1,
  Failure = 2,
  Unconfigured = 3,
}

local function buf_build(bufnr)
  local texlab_client = util.get_active_client_by_name(bufnr, 'texlab')
  local params = {
    textDocument = { uri = vim.uri_from_bufnr(bufnr) },
  }
  if texlab_client then
    texlab_client.request('textDocument/build', params, function(err, result)
      if err then
        error(tostring(err))
      end
      print('Build ' .. texlab_build_status[result.status])
    end, bufnr)
  else
    print 'method textDocument/build is not supported by any servers active on the current buffer'
  end
end

vim.api.nvim_create_autocmd(--[[stuff]])
```

glepnir:
it should be rewrite I think follow this in doc. notice `lspconfig` only provide the basic lsp server settting .

```
Warning: Commands is deprecated and will be removed in future releases.
It is recommended to use `vim.api.nvim_create_user_command()` instead in an `on_attach` function.
```

ranjithshegde:
> it should be rewrite I think follow this in doc. notice `lspconfig` only provide the basic lsp server settting .
> 
> ```
> Warning: Commands is deprecated and will be removed in future releases.
> It is recommended to use `vim.api.nvim_create_user_command()` instead in an `on_attach` function.
> ```

Yes but the plan was to soft deprecate it over future releases. At least thats what was discussed during #1984 

And as far as I can tell, there are no plugins out there that do this for `texlab`. This breaks config of anyone who uses texlab, and the warning has been there for 3 days, not a release cycle. 
This breaks it with no alternatives for uses to use

ii14:
Oh, server configurations were not migrated before this? Sorry, I should've checked that. I think this has to be reverted for now then

glepnir:
future releases so future is what time. now we have api can do this.  I don't understand why this field was buggy in the past. So do you think there are any commands available. Because of this field. There are constantly some prs that are deeply bound to the server. It's weird. And jusmtin has removed part of it. I think should be provide a new api like a function to do that. not use this `commands` field

ranjithshegde:
> future releases so future is what time. now we have api can do this. I don't understand why this field was buggy in the past. So do you think there are any commands available. Because of this field, there are constantly some prs that are deeply bound to the server. It's weird. And jusmtin has removed part of it. I think should be provide a new api like a function to do that. not use this `commands` field

The plan discussed with justin was that we soft deprecate it, and no longer allow users to define it in their config, but keep the internals that are part of `server_configurations/[server].lua`

Also 3 days is by no means a sufficient warning to remove something with 0 alternattives

ii14:
But it's still used in default configurations. I don't think we can just set a default `on_attach`, because it can get overridden by the user, right?

ranjithshegde:
> But it's still used in default configurations. I don't think we can just set a default `on_attach`, because it can get overridden by the user, right?

With this PR, the commands as part of default configuration no longer work

ii14:
Yes, this is exactly what I mean.

glepnir:
maybe the design of this field `commands` was a mistake from the start.  Deep bundling brings new PRs and issues

ranjithshegde:
> maybe the design of this field `commands` was a mistake from the start. Deep bundling brings new PRs and issues

I fully agree, 100%. But as it stands, this is part of so many (at least 8 that I know of) LSPs, and  most users are accustomed to having it. It needs a bit of slow deprecation.

glepnir:
need redesign and provide a new api to do .then they can implement what they want in local not in this project.

ranjithshegde:
> need redesign and provide a new api to do .then they can implement what they want in local not in this project.

This has to be done. But is not trivial. Some servers like pyright just need `workspace/executeCommand` and then its simple. Others like texlab require `textDocument/myFancyOffspecCmd` which need params and ctx handling. 
Cant make a generalized function

glepnir:
Some issues need to be looked at carefully. what do they need. Provide a function or something to bring some of the information they need.

glepnir:
revert.   I think there is a bug in `commands` so no one uses it. Then I remove something. Sorry for the inconvenience

ranjithshegde:
> revert. I think there is a bug in `commands` so no one uses it. Then I remove something. Sorry for the inconvenience

Removal is the correct and long term idea. Maybe after `0.14` we remove this from userspace and only keep the existing defaults in `server_configurations/`

PR #2094, opened 2022-08-26
fix(ui): update default `LspInfoBorder` link to `FloatBorder`

axieax:
#2081 sets the default link for the `LspInfoBorder` highlight group to `Label`, but it makes more sense (especially with pre-defined colorschemes and themes) to use `FloatBorder` since it is a border of a floating window.


glepnir:
Note that not all notes have `Floatborder` defined. For example some old vimscript themes. Maybe `FloatBorder` is not defined. But label is defined in any theme.

axieax:
Ahh ok fair enough

PR #2095, opened 2022-08-27
chore: use `loop.os_uname()` instead of `fn.has'win32'`

numToStr:
Added `util.is_windows` and replaced all the occurrences of `vim.fn.has 'win32' == 1` with the same.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,10 +28,11 @@ When choosing a server name, convert all dashes (`-`) to underscores (`_`) If th
 Note that Windows has a limitation when it comes to directly invoking a server that's installed by `npm` or `gem`, so it requires additional handling.
 
 ```lua
+local util = require 'lspconfig.util'
 local bin_name = 'typescript-language-server'
 local cmd = { bin_name, '--stdio' }
 
-if vim.fn.has 'win32' == 1 then
+if util.is_windows then
   cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
 end
 ```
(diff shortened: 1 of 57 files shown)


numToStr:
Huh...CI is failing bcz npm can't install package 🤦🏼‍♂️

glepnir:
you format all markdown file ? please do not , Honestly it's no improvement. are almost the same .

numToStr:
> you format all markdown file ? please do not , Honestly it's no improvement. are almost the same .

prettier auto formatted it. I'll revert it.

numToStr:
CI is happy 🎉

glepnir:
`is_windows` not in util table. you can' do this. and server configuration is independent. should not be associated with `util`

numToStr:
> you can' do this. and server configuration is independent. should not be associated with util

Why not? It just a boolean, its not associating anything IMO. Configurations are still independent.

glepnir:
`is_windows` not in `util` table it's not a export variable. 

By independent I mean it doesn't apply to anything in other files. lspconfig is just a temporary plugin that provides some basic server configuration. Please don't treat it as a complete plugin. Check out the roadmap for the purpose of the project

numToStr:
We can mark it private for public use. This shouldn't prevent `util` to be used internally otherwise we'll end up repeating the common task (one that this PR solves).

And, I know what's the goal of lspconfig :)

glepnir:
In my opinion no need.

numToStr:
Well then you can close this :)

PR #2096, opened 2022-08-27
chore(rust-analyzer): use `vim.json.decode` instead of `fn.json_decode`

numToStr:


scripts/docgen.lua
```diff
@@ -159,7 +159,7 @@ local function make_lsp_sections()
                 os.exit(1)
                 return
               end
-              local data = fn.json_decode(readfile(package_json_name))
+              local data = vim.json.decode(readfile(package_json_name))
```
numToStr, line 162: this is not related to `rust-analyzer` (as title says) but the changes are similar.

PR #2101, opened 2022-08-28
fix: lspstop should close the client by given client

glepnir:
`LspStop` should be work when pass client from cmdline. if no args stop the all the clients which filetypes match current buffer filetype

@kabouzeid  can you test this patch ?


kabouzeid:
works now, thx

PR #2104, opened 2022-08-30
Added elixir filetype for tailwindcss server configuration.

GGuinea:
I've came across problem while working on Phoenix + Surface UI + TailwindCSS. Tailwindcss is not working for elixir filetypes.

### DIFF
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -21,6 +21,7 @@ return {
       'htmldjango',
       'edge',
       'eelixir', -- vim ft
+      'elixir',
       'ejs',
       'erb',
       'eruby', -- vim ft


glepnir:
your commit message is wrong . please check the there. and `eelixir` and  `elixir` are different file type?
https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md

GGuinea:
Well, I was also surprised, but .ex file was recognised as `elixir` not `eelixir`.

glepnir:
okay if you do `setlocal ft=eelixir` this. the server can works well?

GGuinea:
After changing filetype to eelixir, syntax colouring not working with treesitter, so its not an option.

glepnir:
oh check neovim code I found it works for `.eex` and `.leex`. a different filetype

PR #2105, opened 2022-08-31
fix(docs): Updating duplicated functions (formatting) in the docs

gw31415:
When using the immediate code formatting function: `vim.lsp.buf.formatting`, we get a warning notify like this
```
vim.lsp.buf.formatting is deprecated. Use vim.lsp.buf.format { async = true } instead
```

This PR replaces `vim.lsp.buf.formatting` in doc/lspconfig.txt and README.md with `function() vim.lsp.buf.format{ async = true } end`.


glepnir:
because your neovim version > 0.8 . no need change . it still work on 0.7

PR #2107, opened 2022-09-01
feat: add luau_lsp

williamboman:


glepnir:
don't worry I think they will distinguish `lua` and `luau`

PR #2114, opened 2022-09-07
feat: pass user config to the on_setup hook

williamboman:
This is to allow 3rd party plugins to discern between what is
lspconfig's vendored server config and what the user provided.
Currently, these are merged into a single table which is passed to the
on_setup hook. Passing user_config as a 2nd argument would allow 3rd
party plugins to apply a more sensible precedence of configs.


glepnir:
LGTM

PR #2115, opened 2022-09-07
feat: add sonarlint

hungpt2702:


lua/lspconfig/server_configurations/sonarlint.lua
```diff
@@ -0,0 +1,92 @@
...
+
+  '.git',
+}
+
+return {
+  default_config = {
+    cmd = { 'nc', '-l', port },
+    filetypes = { 'c', 'cpp', 'html', 'javascript', 'php', 'python', 'typescript', 'xml' },
+    root_dir = function(file, bufnr)
+      if (vim.fn.getbufvar(bufnr, '&filetype')) ~= 'python' then
```
glepnir, line 24: there can be `vim.bo.filetype`. and why there judge the filetype ~= python your filetypes not include python

hungpt2702: @glepnir There is a bug with sonarlint python analyzer that slows the sonarlint language server down a lot if `root_dir` or `workspace_folders` is configured. That's why I excluded python files from `root_dir` configuration.

lua/lspconfig/server_configurations/sonarlint.lua
```diff
@@ -0,0 +1,92 @@
+local util = require 'lspconfig.util'
+
+local tcp = vim.loop.new_tcp()
+tcp:bind('127.0.0.1', 0)
```
glepnir, line 4: port in there is 0? no default port usage ?

lua/lspconfig/server_configurations/sonarlint.lua
```diff
@@ -0,0 +1,92 @@
...
+    filetypes = { 'c', 'cpp', 'html', 'javascript', 'php', 'python', 'typescript', 'xml' },
+    root_dir = function(file, bufnr)
+      if (vim.fn.getbufvar(bufnr, '&filetype')) ~= 'python' then
+        return util.root_pattern(root_files)(file)
+      end
+    end,
+    single_file_support = true,
+    handlers = {
+      ['sonarlint/isOpenInEditor'] = function(_, file_uri)
+        return vim.fn.bufexists(vim.uri_to_fname(file_uri)) == 1
```
glepnir, line 31:
```suggestion
        local buf =  vim.uri_to_fname(file_uri)
        retrun api.nvim_buf_is_valid(buf)
```

lua/lspconfig/server_configurations/sonarlint.lua
ii14, line 4: You can't create a socket here. First of all it's going to be created by just configuring the server (even if it's never actually started), and second - now if there are multiple instances of this server, they will all try to use this socket.

hungpt2702: The whole snippet
```lua
local tcp = vim.loop.new_tcp()
tcp:bind('127.0.0.1', 0)
local port = tostring(tcp:getsockname().port
```
is to generate a random, unused port. The sonarlint language server doesn't have a default port.

glepnir: oh I ignored that code ..

lua/lspconfig/server_configurations/sonarlint.lua
```diff
@@ -0,0 +1,92 @@
...
+        vim.tbl_map(function(analyzer)
+          local analyzer_dir = util.path.join(config.sonarlint_dir, string.format('analyzers/sonar%s.jar', analyzer))
+          return analyzer_dir
+        end, { 'cfamily', 'html', 'js', 'php', 'python', 'xml' })
+      )
+
+      table.insert(sonarlint_cmd, '-extraAnalyzers')
+      table.insert(sonarlint_cmd, util.path.join(config.sonarlint_dir, 'analyzers/sonarsecrets.jar'))
+
+      vim.fn.jobstart(sonarlint_cmd)
```
ii14, line 68: You're leaking the process.

Are you sure this server can only talk through TCP? If you really need to spawn and manage some external processes then I don't think lspconfig is a place for that, you should make it a separate plugin instead.

lua/lspconfig/server_configurations/sonarlint.lua
hungpt2702: I've commited the suggested changes.

PR #2119, opened 2022-09-08
feat: solidity LSP

otavioschwanck:


glepnir:
Thanks for Pr but the ci failed in style lint. you should format by stylua

PR #2120, opened 2022-09-08
fix(clojure_lsp): Update lsp url

sigmaSd:


glepnir:
@sigmaSd  change commit msg `Update` to `update`

PR #2122, opened 2022-09-08
escape wildcards patterns when calling glob

emmanueltouzery:
fixes neovim#2111

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -99,6 +99,10 @@ end
 M.path = (function()
   local is_windows = uv.os_uname().version:match 'Windows'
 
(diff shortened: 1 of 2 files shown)


lua/lspconfig/util.lua
```diff
@@ -99,6 +99,10 @@ end
 M.path = (function()
   local is_windows = uv.os_uname().version:match 'Windows'
 
+  local function escape_wildcards(path)
+    return path:gsub("([%[%]])", "\\%1")
```
glepnir, line 103: pattern can be always expanded, this only match `[]` ?

emmanueltouzery: correct, I've now added ? and * too.

PR #2127, opened 2022-09-09
feat: filter user commands

tiagovla:
This patch allows the usage of server commands without passing them using `on_init` or `before_init` (which might be overridden by a user). This is especially useful for custom servers. Therefore this is now possible:
```lua
local function custom_command(command, ctx)
    -- do something
end

local configs = require "lspconfig.configs"
if not configs["custom"] then
    configs["custom"] = {
        -- ...
        default_config = {
            -- ...
            commands = {
                ["custom.aCustomCommand"] = custom_command,
            },
        },
    }
end

```
These commands can be consumed by e.g. [code actions](https://github.com/neovim/neovim/blob/2d6735d8cecc587eb5549f65260ee9ddeb8e1d78/runtime/lua/vim/lsp/buf.lua#L807).

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -291,7 +291,13 @@ function configs.__newindex(t, config_name, config_def)
       M.commands = vim.tbl_deep_extend('force', M.commands, client.config.commands)
     end
     if not M.commands_created and not vim.tbl_isempty(M.commands) then
-      util.create_module_commands(config_name, M.commands)
+      local user_commands = {}
+      for k, v in pairs(M.commands) do
+        if k:match '^%u' then
+          user_commands[k] = v
+        end
+      end
+      util.create_module_commands(config_name, user_commands)
     end
   end
 


glepnir:
`command` field will remove in future.  I do not recommend any customization on this field. It is deeply bound to the server configuration. This is not a good direction. if you want make some commands only work for some servers .create it on `on_attach` field

tiagovla:
> `command` field will remove in future. I do not recommend any customization on this field. It is deeply bound to the server configuration. This is not a good direction. if you want make some commands only work for some servers .create it on `on_attach` field

In my use case scenario, I have a server that I need to adapt its `server.extractMethod` command triggered by a code action. 

So, I need to change the `client.commands` anyway because that's the [only way](https://github.com/neovim/neovim/blob/2d6735d8cecc587eb5549f65260ee9ddeb8e1d78/runtime/lua/vim/lsp/buf.lua#L807) to do it. I can't do something like [this](https://github.com/neovim/nvim-lspconfig/blob/f6e369e9e7feb0d8792558dc3ecf1616b19e6b2f/lua/lspconfig/server_configurations/jdtls.lua#L119) since there's no `'textDocument/codeAction'` handler, this method is just ignored. 

And since I'm already using `on_attach`, `on_init`, and `before_init` it made more sense to put it in my custom server's default configuration instead of wrapping it on setup like [this](https://github.com/mhanberg/elixir.nvim/blob/7796fe00f49612d9b93934a3b9c14ee0b24d1604/lua/elixir/language_server/init.lua#L259). The problem is that lspconfig tries to create user commands for all commands.

I can just hack my way with it, but it would be nicer if it was supported.

glepnir:
check #1937  It would be great if we could bypass the server configuration and get some desired parameters and pass them through the api.

tiagovla:
> check #1937 It would be great if we could bypass the server configuration and get some desired parameters and pass them through the api.

I think the problem is that `UserCommands` and server commands (the ones used by code actions, e.g. python.extractMethod) have the same `client.commands` key. If I understood this correctly, removing `UserCommands` is not related to removing server commands. And as it is, I can't add those in the default configuration, because lspconfig tries to create `UserCommands` out of them, which it can't since the first letter is not uppercase.

tiagovla:
Just to be clear, if `UserCommands` were to be removed, this patch wouldn't be needed. Lspconfig is using the `client.commands` key to do something different from what neovim expects.

glepnir:
yeah. So I am planning to make a lsp plugin template. Based on this template anyone can make an lsp plugin for their language. That way it will have more possibilities.

PR #2150, opened 2022-09-23
fix: update ci lint stylua-action version

glepnir:


.github/workflows/lint.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Lint with stylua
-        uses: JohnnyMorganz/stylua-action@1.0.0
+        uses: JohnnyMorganz/stylua-action@1.1.1
```
lithammer, line 33: I think you need to prefix with `v`. Also, 1.1.2 is available.

```suggestion
        uses: JohnnyMorganz/stylua-action@v1.1.2
```

glepnir: good catch.

lithammer (approved):

PR #2153, opened 2022-09-25
fix: override another progress handler for jdtls

VVKot:
Due to non-compliant message from jdtls, users may encounter issues like https://github.com/mfussenegger/nvim-jdtls/issues/327

LMK if there is a better way to do it but it seems to be the easiest temporary fix


glepnir:
change your commit message to `fix: override another progress handler for jdtls` then force push

VVKot:
Done: @glepnir

VVKot:
@glepnir sorry, one more time?

PR #2157, opened 2022-09-27
fix: added `lua/` to sumenko root patterns

folke:
Neovim plugins always have a `lua/` directory where all the source code is. This PR adds this directory to the root pattern.

### DIFF
--- a/lua/lspconfig/server_configurations/sumneko_lua.lua
+++ b/lua/lspconfig/server_configurations/sumneko_lua.lua
@@ -6,6 +6,7 @@ local root_files = {
   '.stylua.toml',
   'stylua.toml',
   'selene.toml',
+  'lua/',
 }
 
 local bin_name = 'lua-language-server'


glepnir:
there also have `plugin/xxx.lua` if we set the  root  to `plugin-name/lua/` the `plugin-name/plugin/xxx.lua` will outside the root and open as a single file ?

folke:
@glepnir the root would be the parent directory where `lua/` resides in, so should be ok?

glepnir:
`lua/` not work in my test.

```
nvim-test/
├── lua
│   └── nvim-test
│       └── init.lua
└── plugin
    └── nvim-test.lua
```

folke:
@glepnir are you sure you're using this PR? It works as expected for me.

glepnir:
min config works . maybe there have some cache in my config so not wrok .

PR #2159, opened 2022-09-28
revert: fix: added 'lua/' to sumenko root patterns

zeertzjq:
While it is true that `lua/` directory usually indicates the root directory of a Nvim plugin, it doesn't always indicate the root of a workspace. For example, in `neovim/neovim` repository, the existence of `test/functional/lua/` is causing `test/functional/` to be recognized as the root of a workspace, which is a bit annoying.


folke:
What if instead we first do rootpattern on the files and a a fallback on the `lua` directory?

This way, for the Neovim code base, the root would be the correct root.

zeertzjq:
Oh that indeed sounds better. I'll create another PR later.

PR #2168, opened 2022-10-01
I think the filetype should set as qmljs In treesitter-qmljs, the filetype is named as qmljs, So , I think the filetype should be named the same as treesitter is better

Decodetalkers:
---
name: Pull Request
about: Submit a pull request
title: ''
---


Decodetalkers:
and qt has a official lsp for qml, but still is not completed.. but this qmllsp is also not completed though..so maybe that one can also be added?

glepnir:
if someone need that server then can add it . change your commit msg to this. then force push
`fix: update qml filetype same as treesitter`

PR #2175, opened 2022-10-05
fix: gdscript use ip address instead of localhost

A-Lamia:
`vim.lsp.rpc.connect()` is unable to find a valid IP address with `localhost` changing it to `127.0.0.1` resolve the below error.

error


glepnir:
change your commit msg  to  `fix: gdscript use ip address instead of localhost` then force push

PR #2178, opened 2022-10-07
fix: lspconfig help command typo

kyoheiu:
Now readme says `:h lspconfig` to show the help, but it seems typo (:h lsp-config works in my environment).

FWIW


glepnir:
change your commit msg to `fix: lspconfig help command typo` then force push

PR #2179, opened 2022-10-07
Update lspconfig.lua

Uwo-ua:


lua/lspconfig.lua
```diff
@@ -1,18 +1,18 @@
-local configs = require 'lspconfig.configs'
+require 'user.lspconfig.configs'
```
glepnir, line 1: I don't know what does this mean? there is no module called `user`

PR #2189, opened 2022-10-13
fix(rust-analyzer): ignore cargo errors

tweksteen:
Commit 4f3e5977 added a notification message should Cargo failed to find the workspace directory. Large projects such as Android, Fuchsia or the Linux kernel rely on rust-project.json instead of Cargo.toml. Remove the error message which is expected for such configurations.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -22,16 +22,11 @@ return {
         cmd[#cmd + 1] = util.path.join(cargo_crate_dir, 'Cargo.toml')
       end
       local cargo_metadata = ''
-      local cargo_metadata_err = ''
       local cm = vim.fn.jobstart(cmd, {
         on_stdout = function(_, d, _)
           cargo_metadata = table.concat(d, '\n')
         end,
-        on_stderr = function(_, d, _)
-          cargo_metadata_err = table.concat(d, '\n')
-        end,
         stdout_buffered = true,
-        stderr_buffered = true,
       })
       if cm > 0 then
         cm = vim.fn.jobwait({ cm })[1]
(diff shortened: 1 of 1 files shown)


glepnir:
can we use `vim.notify_once` for it ? just notify once

tweksteen:
I'm not familiar with the API, but from what I can tell, this will still trigger whenever nvim is started? If so, this is still a regression from the previous behaviour.

Is there a specific error/situation we are trying to catch from `cargo metadata`? If this is only to detect that Cargo.toml is missing, we shouldn't assume that this is an error. Maybe we can keep the error message if it is different from `error: could not find Cargo.toml[...]`?

glepnir:
yup .but it only show error message once.

tweksteen:
I've pushed a new version to fix the lint error.

It is still disruptive to always have a misleading error message when starting nvim. Also note that `rust-analyzer` itself will report the error if it fails to load the Cargo.toml (eg malformed). Please consider all users for this plugin. Thanks.

glepnir:
Does some people who write rust  need this error message ?

tweksteen:
@jonhoo (who wrote the previous commit) for input

jonhoo:
The challenge here is that there are _other_ reasons why `cargo metadata` might fail. The absence of a `Cargo.toml` isn't the only one. This would hide _all_ errors from `cargo metadata` (if I'm reading it right), which may make problems much harder to debug for users. Things will just silently not work.

Honestly, even in the "`Cargo.toml` not found" case, it feels weird to hide the error. It _could_ be that this happens for some other reason than the user being in one of the projects you listed. I think what I'd want here is to hide the error only if it is "`Cargo.toml` not found" **and** a `rust-project.json` file exists.

tweksteen:
>  Things will just silently not work.

This is not the behaviour I've observed locally. `rust-analyzer` will try to execute `cargo metadata` itself once loaded. In the case of a malformed Cargo.toml for example, an error will be surfaced to the UI.

I think the current configuration is trying to be too clever for detecting `root_dir`. It should be a "best-effort" probing, more than trying to support all the scenarios (`rust-analyzer` will do the heavy lifting later on, there is no need to do it ourselves).

As another datapoint, this server configuration is also the only configuration that will `notify` the user within the `root_dir` function.

I can compromise on the `"Cargo.toml not found" and a rust-project.json file exists` approach, but it still seems unnecessary.

jonhoo:
I think it at least _used_ to be that if the initial `cargo metadata` call failed, then the whole `rust-analyzer` connection would break down, and so nothing more was reported from `rust-analyzer` (including any messages from it about `cargo metadata` failing). That may have changed though, I'm not sure?

PR #2193, opened 2022-10-13
feat: Dafny support

Gorgeous-Patrick:
Hi! This is the support for dafny. Tested on Linux

### DIFF
--- a/lua/lspconfig/server_configurations/dafny.lua
+++ b/lua/lspconfig/server_configurations/dafny.lua
@@ -1,16 +1,13 @@
 local util = require 'lspconfig.util'
 
-local root_files = {
-    'compile_commands.json',
-    '.ccls',
-}
 return {
   default_config = {
     filetypes = { 'dfy', 'dafny' },
     cmd = { 'dotnet', '/opt/dafny/DafnyLanguageServer.dll' },
     root_dir = function(fname)
-      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
+      util.find_git_ancestor(fname)
     end,
+    single_file_support = true
   },
   docs = {
     description = [[
--- a/lua/lspconfig/server_configurations/dafny.lua
+++ b/lua/lspconfig/server_configurations/dafny.lua
@@ -6,7 +6,7 @@ local root_files = {
 }
 return {
   default_config = {
-    filetypes = { 'dfy', 'dafny'},
+    filetypes = { 'dfy', 'dafny' },
     cmd = { 'dotnet', '/opt/dafny/DafnyLanguageServer.dll' },
     root_dir = function(fname)
       return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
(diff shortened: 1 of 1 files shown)


Gorgeous-Patrick:
I am looking for a proper way to set the root dir though.

glepnir:
you should use `vim.fn.win32 == 1` to check the system is windows or not. then change the cmd .

```lua
local bin_name = 'donat'
  cmd = { 'cmd.exe', '/C', bin_name,.... }
end

```

lua/lspconfig/server_configurations/dafny.lua
```diff
@@ -0,0 +1,20 @@
+local util = require 'lspconfig.util'
+
+local root_files = {
+    'compile_commands.json',
+    '.ccls',
+}
+return {
+  default_config = {
+    filetypes = { 'dfy', 'dafny'},
```
glepnir, line 9:
```suggestion
    filetypes = { 'dfy', 'dafny' },
```

Gorgeous-Patrick: Ok I will do this a bit later today. But I currently don't have a windows machine. Probably I need help for that.

Gorgeous-Patrick:
I think that I cannot make a 100% correct path here. Because the official documentation of the language server requires the users to compile themselves and put the compiled version ( a dll file or a exe file ) somewhere in their systems. I think there is no way to automatically find the file. 

The path written there is just the default path used in ArchLinux AUR package.

I will write the documentation to teach the users so that they can probably specify the path for themselves. IDK if that sounds good to you.

lua/lspconfig/server_configurations/dafny.lua
```diff
@@ -0,0 +1,25 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'dotnet'
+local dll_path = '/opt/dafny/DafnyLanguageServer.dll'
+local cmd = {bin_name, dll_path}
+if vim.fn.has 'win32' == 1 then
+  dll_path = 'C:\\Program Files\\dafny\\DafnyLanguageServer.dll'
```
glepnir, line 7: don't hardcode the path in there.

Gorgeous-Patrick:
Resolved the issues. Probably good for another review @glepnir  Thanks.

lua/lspconfig/server_configurations/dafny.lua
```diff
@@ -0,0 +1,20 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    filetypes = { 'dfy', 'dafny' },
+    root_dir = function(fname)
+      util.find_git_ancestor(fname)
+    end,
+    single_file_support = true
```
glepnir, line 9:
```suggestion
    single_file_support = true,
```

glepnir:
need fix this commit msg .

Gorgeous-Patrick:
OK I changed the commit msg. Sorry I was too careless. Thank you! I think it is ready for another review. @glepnir

PR #2213, opened 2022-10-22
fix: cmd support vim.lsp.rpc.connect function

Decodetalkers:
when user what to set cmd to vim.rpc.connect, but it is a function

nvim has already support vim.lsp.rpc.connect, but it is a function , and in the default config, it can be just set as a function, so I think lsp-config should also handle such kind of action

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -34,7 +34,11 @@ function configs.__newindex(t, config_name, config_def)
     local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
 
     validate {
-      cmd = { user_config.cmd, 't', true },
+      cmd = {
+        user_config.cmd,
+        { 'f', 't' },
+        true,
+      },
       root_dir = { user_config.root_dir, 'f', true },
       filetypes = { user_config.filetype, 't', true },
       on_new_config = { user_config.on_new_config, 'f', true },


lua/lspconfig/configs.lua
```diff
@@ -34,7 +34,15 @@ function configs.__newindex(t, config_name, config_def)
     local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
 
     validate {
-      cmd = { user_config.cmd, 't', true },
+      cmd = (function()
```
glepnir, line 37: I think there is 

```lua
cmd = { user_config.cmd , function(a) do stuff end,  true}
```

Decodetalkers: > I think there is
> 
> ```lua
> cmd = { user_config.cmd , function(a) do stuff end,  true}
> ```

Fixed now

lua/lspconfig/configs.lua
```diff
@@ -34,7 +34,17 @@ function configs.__newindex(t, config_name, config_def)
     local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
 
     validate {
-      cmd = { user_config.cmd, 't', true },
+      cmd = {
+        user_config.cmd,
+        (function()
```
glepnir, line 39: second field can be function type no need call it. `:h vim.validate`

lua/lspconfig/configs.lua
```diff
@@ -34,7 +34,17 @@ function configs.__newindex(t, config_name, config_def)
     local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
 
     validate {
-      cmd = { user_config.cmd, 't', true },
+      cmd = {
```
glepnir, line 37:
```suggestion
      cmd = { user_config.md , {'f', 't' } , true }
```

PR #2218, opened 2022-10-25
fix: after use lspstop command remove the lspconfig augroup of server

### LINKED ISSUE #2216: LspStop should remove related autocmds
### Language server

_No response_

### Requested feature

_No response_

glepnir:
remove the lspconfig augroup after use `LspStop` command .  use `LspStart`  will register the lspconfig augroup.

Fix #2216

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -31,8 +31,6 @@ function configs.__newindex(t, config_name, config_def)
   default_config.name = config_name
 
   function M.setup(user_config)
-    local lsp_group = vim.api.nvim_create_augroup('lspconfig', { clear = false })
-
     validate {
       cmd = {
         user_config.cmd,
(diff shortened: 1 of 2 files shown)


glepnir:
@ii14  do we want remove the augroup after `LspStop`?  if answer is yes please review.

lua/lspconfig/configs.lua
```diff
@@ -56,13 +54,31 @@ function configs.__newindex(t, config_name, config_def)
...
+        group_id = api.nvim_create_augroup(group_name, { clear = true })
+        configs.lsp_groups[group_name] = group_id
+      else
+        group_id = configs.lsp_groups[group_name]
+      end
+      return group_id
+    end
+
+    if not configs.lsp_groups then
+      configs.lsp_groups = {}
```
justinmk, line 70: This is a red flag. We should not need a cache or a registry for this.

kylo252: There's `on_exit` which should be a good fit to use for this purpose.

- use a unique name (some kind of hash?) for the augroup
- or: pass `create_augroup` to every instance of `create_autocmd`

```lua
on_exit = function()
  vim.schedule(function()
    pcall(function()
      vim.api.nvim_clear_autocmds { group = name }
      -- or 
      -- vim.api.nvim_del_augroup_by_name(name)
    end)
  end)
end
```

PR #2222, opened 2022-10-29
Fix dead link in `texlab` description

ayhon:
`texlab` changed where they store their documentation. Since https://github.com/latex-lsp/texlab/commit/d03e350679dfdbbd0d25e3c28131c4b2677789cf they store that information in their GitHub wiki.


glepnir:
change your commit msg to `fix: fix the dead link in texlab description`

PR #2226, opened 2022-11-05
feat: support unnamed buffer with single_file_support

### LINKED ISSUE #1926: LSP should be able to attach to unnamed buffer

glepnir:
support unnamed buffer when `single_file_support` equals `true`

Fix #1926


lua/lspconfig/configs.lua
```diff
@@ -247,7 +260,12 @@ function configs.__newindex(t, config_name, config_def)
       if root_dir then
         id = manager.add(root_dir, false)
       elseif config.single_file_support then
-        local pseudo_root = util.path.dirname(buf_path)
+        local pseudo_root
+        if #bufname == 0 then
+          pseudo_root = uv.cwd()
+        else
+          pseudo_root = util.path.dirname(buf_path)
+        end
```
justinmk, line 268: a ternary would make this clearer

lua/lspconfig/configs.lua
```diff
@@ -124,12 +128,17 @@ function configs.__newindex(t, config_name, config_def)
...
-          return
+        local pseudo_root
+        if #bufname == 0 then
+          pseudo_root = uv.cwd()
+        else
+          if not util.bufname_valid(bufname) then
+            return
+          end
+          pseudo_root = util.path.dirname(util.path.sanitize(bufname))
         end
```
justinmk, line 139:
```suggestion
        if #bufname ~= 0 and not util.bufname_valid(bufname) then
          return
        end
        local pseudo_root = (#bufname == 0
          and uv.cwd()
          or util.path.dirname(util.path.sanitize(bufname)))
```

justinmk (commented):
do we need any docs update? is a small test possible?

glepnir:
both needed. there need fix ci and also need check the behavior of other editors.

glepnir:
maybe i can ignore this ci faield ?  it just do grep the `dirname |cwd` in file..

lua/lspconfig/util.lua
```diff
@@ -20,9 +20,6 @@ M.default_config = {
 M.on_setup = nil
 
 function M.bufname_valid(bufname)
-  if not bufname then
-    return false
-  end
   if bufname:match '^/' or bufname:match '^[a-zA-Z]:' or bufname:match '^zipfile://' or bufname:match '^tarfile:' then
```
justinmk, line 23: if `bufname` is nil this will error out, is that intentional?

doc/lspconfig.txt
```diff
@@ -442,6 +442,8 @@ mode under which cross-file features may be degraded.
   `workspaceFolders` during initialization.
 - attaching subsequent files in the parent directory to the same server
   instance, depending on filetype.
+- also support the unnamed buffer which filetype match the server filetype
```
justinmk, line 445:
```suggestion
- also supports unnamed buffer if filetype matches the server filetype
```

lua/lspconfig/util.lua
glepnir: look at the source code, I don't think it can be nil.

https://github.com/neovim/neovim/blob/master/src/nvim/api/buffer.c#L1037-L1048

PR #2240, opened 2022-11-13
Try to get sourcery auth token from config file

otahontas:
Sourcery-cli creates `auth.yaml` file to users config folder, which contains the login token. 

This PR adds check for login token in auth file. Logic shouldn't be breaking, since we're getting the token from file only if it's not defined already in lsp config.

### DIFF
--- a/lua/lspconfig/server_configurations/sourcery.lua
+++ b/lua/lspconfig/server_configurations/sourcery.lua
@@ -9,6 +9,22 @@ local root_files = {
   'pyrightconfig.json',
 }
 
+local get_token_from_auth_file = function()
+  local config_home = vim.fn.getenv 'XDG_CONFIG_HOME' or vim.fn.expand '~/.config'
+  local auth_file_path = config_home .. '/sourcery/auth.yaml'
+
+  if vim.fn.filereadable(auth_file_path) == 1 then
+    local content = vim.fn.readfile(auth_file_path)
+
+    for _, line in ipairs(content) do
+      local token = line:match 'sourcery_token: (.+)'
+      if token then
+        return token
+      end
+    end
+  end
+end
+
 return {
   default_config = {
     cmd = { 'sourcery', 'lsp' },
@@ -24,6 +40,10 @@ return {
     single_file_support = true,
   },
   on_new_config = function(new_config, _)
+    local possibly_token = get_token_from_auth_file()
+    if possibly_token then
+      new_config.init_options.token = possibly_token
+    end
     if not new_config.init_options.token then
       local notify = vim.notify_once or vim.notify
       notify('[lspconfig] The authentication token must be provided in config.init_options', vim.log.levels.ERROR)
@@ -54,6 +74,8 @@ require'lspconfig'.sourcery.setup {
     },
 }
 ```
+
+Alternatively, you can login to sourcery by running `sourcery login` with sourcery-cli and your token will be grabbed from sourcery config file.
 ]],
   },
 }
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -9,6 +9,23 @@ local root_files = {
   'pyrightconfig.json',
 }
 
+local get_token_from_auth_file = function()
+  local config_home = vim.fn.getenv 'XDG_CONFIG_HOME' or vim.fn.expand '~/.config'
+  local auth_file_path = config_home .. '/sourcery/auth.yaml'
+
+  local file_readable = vim.fn.filereadable(auth_file_path)
+  if file_readable == 1 then
```
glepnir, line 17:
```suggestion
  if vim.fn.filereadable(auth_file_path) == 1  then
```

lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -9,6 +9,23 @@ local root_files = {
   'pyrightconfig.json',
 }
 
+local get_token_from_auth_file = function()
+  local config_home = vim.fn.getenv 'XDG_CONFIG_HOME' or vim.fn.expand '~/.config'
+  local auth_file_path = config_home .. '/sourcery/auth.yaml'
```
glepnir, line 14: why hardcode the path sep in there? it only work for *nix ?

otahontas: true I'll check what it is for windows

otahontas: https://github.com/neovim/nvim-lspconfig/pull/2240/commits/3d648fdbe1e731564438381ec1eb6ecabaf660ba

lua/lspconfig/server_configurations/sourcery.lua
```diff
@@ -9,6 +9,25 @@ local root_files = {
   'pyrightconfig.json',
 }
 
+local get_token_from_auth_file = function()
+  local is_windows = vim.fn.has 'win32' == 1
+  local path_sep = is_windows and '\\' or '/'
+
+  local config_home = is_windows and vim.fn.getenv 'APPDATA'
```
glepnir, line 16: maybe we can use `stdpath` for it ?

otahontas: is the stdpath always gonna be APPDATA or XDG_CONFIG_HOME? token file will in those folders

glepnir: I know it will check `XDG` . I don't know how it works on windows.

otahontas: It looks like stdpath is resolved to APPDATALOCAL in win: https://github.com/neovim/neovim/blob/4b02291f81b76eda029d4eae1c1d59c1a946a032/src/nvim/os/stdpaths.c#L27, so I wouldn't use it here

otahontas: but I can can still fix unix ones

otahontas: actually, `vim.fn.stdpath` returns `XDG_CONFIG_HOME/nvim`, so I don't think it's that usable? Why not just use `vim.fn.getenv(XDG_CONFIG_HOME)` which resolves directly to correct config root folder?

otahontas: checking sourcery docs again, it looks like in *nix config files are always in `~/.config/sourcery` so I'm going to use that instead of resolving from `XDG_CONFIG_HOME`

otahontas:
@glepnir  It seems like we don't need to pass token to lsp in case `sourcery login` has been run. Token is fetched then automatically. I fixed this to only check for token if token not set in options.

PR #2242, opened 2022-11-15
feat: include cmd option for arduino_language_server

aaaaargZombies:
Currently the `arduino_language_server` doesn't run without passing custom configuration options to `setup`.

If an Arduino project is contains `sketch.json` file all but the `cmd` option can be omitted. As this stays the same it would be nice to move it into this repository. This also has the advantage of keeping project specific settings inside the project and not in the nvim config.

I've done the above and updated the docs with a minimal guide to setting up an Arduino project.

This is working on my machine `require("lspconfig").arduino_language_server.setup()` but it would be good if others could test it.

### DIFF
--- a/lua/lspconfig/server_configurations/arduino_language_server.lua
+++ b/lua/lspconfig/server_configurations/arduino_language_server.lua
@@ -1,12 +1,15 @@
 local util = require 'lspconfig.util'
 
 return {
-  default_config = {
-    filetypes = { 'arduino' },
-    root_dir = util.root_pattern '*.ino',
-  },
-  docs = {
-    description = [[
+	default_config = {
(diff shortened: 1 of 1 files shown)


glepnir:
first change the commit msg to `fix: format with stylua` , second if current settings works fine withe the latest `arduino server` maybe we can merge it. if someone get issue . we can fix it then ?

aaaaargZombies:
> first change the commit msg to `fix: format with stylua`

done, thanks for the patience with my formatting

glepnir:
does there have any other work to do ? if not we can merge then ?

aaaaargZombies:
I think this is ready to merge if you are happy

PR #2247, opened 2022-11-19
feat: add fennel-ls support

gh-liu:
Untested on windows.

### DIFF
--- a/lua/lspconfig/server_configurations/fennel-ls.lua
+++ b/lua/lspconfig/server_configurations/fennel-ls.lua
@@ -1,0 +1,17 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { "fennel-ls" },
+    filetypes = { "fennel" },
+    root_dir = function(dir) return util.find_git_ancestor(dir) end,
+    settings = {}
+  },
+  docs = {
+    description = [[
+https://sr.ht/~xerool/fennel-ls/
+
+A language server for fennel.
+]]   ,
+  },
+}
--- a/lua/lspconfig/server_configurations/fennel-ls.lua
+++ b/lua/lspconfig/server_configurations/fennel-ls.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { "fennel-ls" },
-    filetypes = { "fennel" },
+    filetypes = { 'fennel' },
     root_dir = function(dir)
          return util.find_git_ancestor(dir)
     end,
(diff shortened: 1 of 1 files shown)


glepnir:

lua/lspconfig/server_configurations/fennel-ls.lua
```diff
@@ -0,0 +1,17 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { "fennel-ls" },
```
glepnir, line 5:
```suggestion
    cmd = { 'fennel-ls' },
```

lua/lspconfig/server_configurations/fennel-ls.lua
```diff
@@ -0,0 +1,17 @@
...
+    filetypes = { "fennel" },
```
glepnir, line 6:
```suggestion
    filetypes = { 'fennel' },
```

lua/lspconfig/server_configurations/fennel-ls.lua
```diff
@@ -0,0 +1,17 @@
...
+    root_dir = function(dir) return util.find_git_ancestor(dir) end,
```
glepnir, line 7:
```suggestion
    root_dir = function(dir)
         return util.find_git_ancestor(dir)
    end,
```

lua/lspconfig/server_configurations/fennel-ls.lua
```diff
@@ -0,0 +1,17 @@
...
+    settings = {}
```
glepnir, line 8:
```suggestion
    settings = {},
```

gh-liu:
> does this server not support windows?



> does this server not support windows?


I guess it's possible, but I haven't tried it yet.

glepnir:
so we can simply add the server cmd in windows. if it not works fine we can fix it according the error message

```lua

local cmd = 'fennel-ls'
if vim.fn.has('win32') then
  cmd = { 'cmd.exe', '/C', 'fennel-ls'}
end

```

gh-liu:
> 

I have tried in wsl2, but it's not work. 
The error in lsp.log is  `.../vim/lsp/rpc.lua:732  "rpc" "cmd.exe" "stderr"  "'fennel-ls' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n" ` , but I do not known how to compile is to a windows EXE file.

glepnir:
hmmm okay then just need fix the commit message.

gh-liu:
> hmmm okay then just need fix the commit message.

ok

PR #2248, opened 2022-11-19
fix(lua-lsp): Append 'lua/' to root_patern if the pattern is under lua

sigmaSd:
Currently when a lua file is under `..path/lua/..path` its detected by `root_pattern 'lua'` but instead of returning the expected path `..path/lua` it returns the parent path `..path/lua`

This PR fixes that


zeertzjq:
@sigmaSd Why?

PR #2255, opened 2022-11-24
fix: exclude neovim directory from lint scope

hezhizhen:
Do not lint files in `neovim/` and lint every lua file everywhere else.


glepnir (approved):

teto:
is this still needed ? I dont see a "neovim" folder anymore ?

PR #1993, opened 2022-07-08
feat: support gleam

wesleimp:
Language server for Gleam Programming Language

https://github.com/gleam-lang/gleam

cc: @lpil

PR #2266, opened 2022-11-29
feat: add gleam

UnsolvedCypher:
Add support for the Gleam programming language. Most of the work here was done by @wesleimp but I'm reopening as work on #1993 seems to be stalled. Please let me know if this isn't the right thing to do.

### DIFF
--- a/gleam.lua
+++ b/gleam.lua
@@ -1,0 +1,25 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = {'gleam', 'lsp'},
+    filetypes = { 'gleam' },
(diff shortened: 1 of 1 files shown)


gleam.lua
```diff
@@ -0,0 +1,25 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = {'gleam', 'lsp'},
```
glepnir, line 5:
```suggestion
    cmd = { 'gleam', 'lsp' },
```

lpil:
Thank you @UnsolvedCypher and @glepnir !

PR #2272, opened 2022-11-30
fix: remove unused call to util.on_setup()

alexeRadu:
This function is set to nil in lspconfig/util.lua so there is no need to check if it exists and call it when it exists since this is *never* the case.

Signed-off-by: Radu Alexe <alexeradu2007@gmail.com>

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -56,10 +56,6 @@ function configs.__newindex(t, config_name, config_def)
 
     local config = tbl_deep_extend('keep', user_config, default_config)
 
-    if util.on_setup then
-      pcall(util.on_setup, config, user_config)
-    end
-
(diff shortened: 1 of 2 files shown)


glepnir:
ci faield.

kylo252 (changes requested):
this function is needed by external plugins, e.g. meson

https://github.com/williamboman/mason-lspconfig.nvim/blob/edf15b98cd7d7ce0f83cf7d3a968145a3f974772/lua/mason-lspconfig/lspconfig_hook.lua#L57-L91

alexeRadu:
Got it! It's a way to register a hook function that runs before the setup() starts. I'm just failing to understand why you would want this kind of implementation over the straight forward one where you just call the function you need before setup() runs.

Issue #2273, opened 2022-11-30
`root_dir` is being ignored


datwaft:
### Description

Hello!

I am trying to implement [this suggestion](https://deno.land/manual@v1.28.2/getting_started/setup_your_environment#neovim-06-using-the-built-in-language-server) from Deno's documentation but I am having a problem, when setting `root_dir` with the suggested values it is being ignored. When checking with `:LspInfo` I can see that the servers still have their default values.

This issue is similar to #2069.

### Neovim version

NVIM v0.9.0-dev-408+g98695b499
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

d346335af862ede0cece75739d789cfdc46c4454

### Operating system and version

macOS 13.0.1

### Affected language servers

tsserver, denols

### Steps to reproduce

1. Use the minimal configuration and install `lspconfig`.

2. Create a Typescript project with a `.git` folder inside (can be deno or tsserver, in both happens).

3. Open any typescript file inside that project.

### Actual behavior

Both `denols` and `tsserver` are being loaded into the buffer because they still have `.git` as part of their `root_dir` configuration.

### Expected behavior

If the project doesn't contain a `package.json` nor a `deno.json` then neither of the language servers must be loaded. If the project contains a `package.json` only `tsserver` must be loaded and if the project contains a `deno.json` only `denols` must be loaded.

### Minimal config

```Lua
local lspconfig = require("lspconfig")

lspconfig.denols.setup({
  root_dir = lspconfig.util.root_pattern("deno.json", "deno.jsonc"),
})

lspconfig.tsserver.setup({
  root_dir = lspconfig.util.root_pattern("package.json"),
})
```

### LSP log

https://gist.github.com/datwaft/22e964e1a56e481a265e1b3b33f9426d

glepnir:
can't reproduce. if are using packer  I think you need recompile your config file

datwaft:
I could reproduce it in a Docker environment (I did it to verify it before creating the issue).

I will recreate the docker environment and I will send you the Github repository link.

I could reproduce it with only lspconfig installed, without packer or any other plugin.

glepnir:
I had test it in a min config with clean neovim. it works fine. test config

```lua
vim.opt.runtimepath:append("~/Workspace/nvim-lspconfig/")
local lspconfig = require("lspconfig")
lspconfig.tsserver.setup({
[...]

PR #2274, opened 2022-12-01
fix: remove searched for message in lspinfo

glepnir:
when we override the root pattern it still show the patterns from document config.  see #2273


glepnir:
see #2283

PR #2276, opened 2022-12-01
feat: add unison support

kentookura:


lua/lspconfig/server_configurations/unison.lua
```diff
@@ -0,0 +1,17 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'nc', 'localhost', '5757' },
```
lithammer, line 5: I would probably inspect the `UNISON_LSP_PORT` environment variable here as well. Something like:

```suggestion
    cmd = { 'nc', 'localhost', os.getenv 'UNISON_LSP_PORT' or '5757' },
```

glepnir: good catch. should be inspect. if someone already used 5757 port

PR #2279, opened 2022-12-01
fix: move gleam.lua to correct folder

Massolari:
In [this commit](https://github.com/neovim/nvim-lspconfig/commit/807f3a66f342e08a4f6705bb68bfe199360ff861), `gleam.lua` was created in the root folder.
I moved to the correct one


glepnir:
oh thanks .my fault. looks like ci failed

PR #2283, opened 2022-12-02
feat: improve the root dir pattern message in LspInfo

glepnir:
we should display the correct root pattern not get from the document.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -16,6 +16,8 @@ M.default_config = {
   capabilities = lsp.protocol.make_client_capabilities(),
 }
 
+M.server_root_pattern = {}
+
 -- global on_setup hook
 M.on_setup = nil
 
@@ -335,6 +337,18 @@ function M.search_ancestors(startpath, func)
   end
 end
 
+---@private
+local function get_config_by_ft()
+  local configs = require 'lspconfig.configs'
+  local servers = M.available_servers()
+  for _, server in pairs(servers) do
+    local fts = configs[server].filetypes
+    if fts and vim.tbl_contains(fts, vim.bo.filetype) then
+      return server
+    end
+  end
+end
+
 function M.root_pattern(...)
   local patterns = vim.tbl_flatten { ... }
   local function matcher(path)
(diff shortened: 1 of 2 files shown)


williamboman (commented):
I'm not too familiar with the codebase but this seems to assume the following, no?

1) a server is only launched once
2) it assumes the `filetypes` provided by lspconfig is not overridden by the user

Also, won't the `util.server_root_pattern` table grow forever as lspconfig continuously invokes `root_dir()` on every `FileType | BufReadPost`?

glepnir:
> it assumes the filetypes provided by lspconfig is not overridden by the user

Even if the user overrides the filetypes. Does it matter? The current buffer filetype should always be in there.

No   it don't grow up.

glepnir:
oh like efm etc these servers no need filetypes.

glepnir:
About server in `BufReadPost` looks like only efm.  

`rg "\s\s\s\sfiletypes\s=\s" -N --stats --sort-files`

glepnir:
look at the lsp `languageId` and efm doc . do we have a server without the filetypes ? I can't found it . the servers which currently we support must have the filetypes field.

lua/lspconfig/util.lua
```diff
@@ -347,6 +361,10 @@ function M.root_pattern(...)
     end
   end
   return function(startpath)
+    local server_name = get_config_by_ft()
+    if server_name and not M.server_root_pattern[server_name] then
+      M.server_root_pattern[server_name] = patterns
+    end
```
williamboman, line 367: Not all server configuration's `root_dir` property makes use of this `root_pattern()` utility function. I think it'll be pretty complicated to try to capture the `root_dir` predicate logic this way. A root dir function also doesn't necessarily have to deal directly with file paths (although this is most often the case) - how would these "patterns" be registered? For example (very contrived example):

```lua
lspconfig.some_server.setup {
  root_dir = function (fname)
     return sync_spawn("server-tool find-project-root", {
         cwd = vim.fn.fnamemodify(fname, "%:h")
     })
  end
}
```

I was thinking we could still simply make use of the `.docs.default_config.root_dir` value, if available. This is also what's displayed in `server_configurations.md` document that many people refer to. This is basically what was done before, but this time we can skip displaying it if we see that the user has provided a custom `root_dir`. I opened a draft PR for this alternative approach: https://github.com/neovim/nvim-lspconfig/pull/2288.

glepnir: nope. root_dir is wrong design it It doesn't meet the specification in my opinion. I am trying remake it in core. then there need some change in lspconfig. but it's break change.

glepnir: nope. root_dir is wrong design it It doesn't meet the specification in my opinion. I am trying remake it in core. then there need some change in lspconfig. but it's break change.

glepnir: back to pr. for this custom situation can we render a message like `Searched for your custom root pattern`. if user not use the root_patterns

Issue #2285, opened 2022-12-02
go-to-definition spawns new LSP server with root at definition


jonhoo:
### Description

When you use go-to-definition on a symbol that takes you outside of the current project (such as one defined in another crate in the Rust world), nvim-lspconfig will spawn a new LSP server (at least for rust-analyzer) with the current working directory set to the directory of the project that holds the symbol definition, rather than the current nvim working directory. This, in turn, causes issues when the working directory has implications for the LSP server's configuration (or the configurations of the tools it in turn invokes).

It's not entirely clear to me whether this a bug with neovim's lsp support, with nvim-lspconfig, with rust-analyzer, or with LSP more broadly, but I _suspect_ it's a nvim-lspconfig issue as VSCode for example handles this correctly.

I also believe this is related to (but not _quite_ the same as) https://github.com/neovim/nvim-lspconfig/issues/804.

I _think_ the right solution here is to either find a way to continue using the same LSP server instance, or to spawn the new LSP server instances in `:pwd`, and _then_ point it at the project directory for the definition. That way it'll still pick up the right configuration.

### Neovim version

```
NVIM v0.7.2
Build type: Debug
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/clang -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=1

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```

### Nvim-lspconfig version

d346335af862ede0cece75739d789cfdc46c4454

### Operating system and version

Amazon Linux 2 (aarch64; 5.4 kernel)

### Affected language servers

rust-analyzer, though likely also others

[...]

PR #2287, opened 2022-12-03
fix(lsp): use the same lsp server for filetype if it exists

sigmaSd:
fix https://github.com/neovim/nvim-lspconfig/issues/2285

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -253,6 +253,14 @@ function M.server_per_root_dir_manager(make_config)
     -- Check if we have a client already or start and store it.
     if not client_id then
       local new_config = make_config(root_dir)
+      if next(clients) ~= nil then
+        for _, id in pairs(clients) do
+          local client = lsp.get_client_by_id(id)
+          if client.name == new_config.name then
+            return id
+          end
+        end
+      end
       -- do nothing if the client is not enabled
       if new_config.enabled == false then
         return


sigmaSd:
I don't know if there is a spec or something to follow, this is just guessing what vscode does.

sigmaSd:
Helix handles this, so I got curious whats the trick, turns out there is no trick, they never spawn an lsp server **in response** to a goToDefinition request, which now I think about actually make sense. Why do we have that code in the first place, are there actual use cases ?

sigmaSd:
If we take a step back, theoretically in response to a gotodefiniton, the lsp server that sent us there should be the one to continue to work

sigmaSd:
It might be only really solvable at neovim level though

sigmaSd:
Nvm, helix does the exact same thing https://github.com/helix-editor/helix/blob/bcdb475b71b0fbabce57344ac8d2575c23b1bbe0/helix-view/src/editor.rs#L944

sigmaSd:
but helix has a sanity check to compare the original lsp that sent us `id` with the one that we found https://github.com/helix-editor/helix/blob/bcdb475b71b0fbabce57344ac8d2575c23b1bbe0/helix-view/src/editor.rs#L958 and it doesn't spawn a different filetype server in any case

This can only be done inside nvim itself

lua/lspconfig/configs.lua
```diff
@@ -245,7 +245,13 @@ function configs.__newindex(t, config_name, config_def)
       end
 
       if root_dir then
-        id = manager.add(root_dir, false)
+        -- Check if there is an active server already for this same file type
+        local active_clients_list = util.get_active_clients_list_by_ft(vim.bo[bufnr].filetype)
+        if #active_clients_list ~= 0 then
+          id = active_clients_list[1].id
```
glepnir, line 251: I don't think there can use index 1 if there has multiple clients for this filetype . like efm tsserver for ts file the 1 can be efm or tsserver .

sigmaSd: We don't have the original lsp client id at this point, so we can't tell which client to choose

glepnir: I am trying to remove root_dir in lsp start client in core. maybe can solve this. there is no root_dir definition in lsp doc. use wrokspace folder would be better?

sigmaSd: I'm not sure if this is a correct approach , helix also have this concept , (emacs as well?)

I'm not sure what vscode does

sigmaSd: I asked a bit , seems like vscode sets the root workspace as root but it have more logic to add more workspaces (vscode can handle multiple workspaces at the time)

Honestly I can't answer here as I'm not familiar that much with lsp protocol

glepnir: you mean helix has root_dir ? emacs lsp-bridge, yegappan/lsp  don't have root_dir and the root_dir is not in lsp specification. it just used to check there already has an exist client and reused.

sigmaSd: That's what I understood when I asked in their matrix channel, seems they're still  designing this though https://github.com/helix-editor/helix/pull/4439

glepnir: I am not familiar with helix. But this doesn't look like a good way.

glepnir: maybe we can do it like this ?

```diff
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -253,6 +253,14 @@ function M.server_per_root_dir_manager(make_config)
     -- Check if we have a client already or start and store it.
     if not client_id then
       local new_config = make_config(root_dir)
+      if next(clients) ~= nil then
+        for _, id in pairs(clients) do
+          local client = lsp.get_client_by_id(id)
+          if client.name == new_config.name then
+            return id
+          end
+        end
+      end
       -- do nothing if the client is not enabled
       if new_config.enabled == false then
         return
```

sigmaSd: I think that works  I updated the pr with your code

glepnir: do we need add the new root dir into the workspaceFolder ?

sigmaSd: I'm not sure when I tested with jonhoo original rust problem your snippet fixes it

glepnir: maybe them belong to one workspace.

xiantang: > maybe them belong to one workspace.

no, they belong different workspace when you try to open another project file

PR #2288, opened 2022-12-04
feat(lspinfo): display root dir pattern if project dir can't be found

williamboman:


lua/lspconfig/ui/lspinfo.lua
```diff
@@ -51,6 +51,7 @@ local cmd_type = {
 
 local function make_config_info(config, bufnr)
   local config_info = {}
+  local default_config = require(string.format('lspconfig.server_configurations.%s', config.name)).default_config
```
williamboman, line 54: This will break for custom server configurations, will need to revise before merge

glepnir:
it still get the root patterns from doc. if we want display it we should display the actual patterns .

williamboman:
> it still get the root patterns from doc. if we want display it we should display the actual patterns .

Yeah I agree it'd be nice to be get the actual pattern used, but from looking into the source it seems like a pretty difficult thing to pull off nicely. I think it's a fair to assume that the docs are accurate, no? If some are not it's a pretty easy fix.

glepnir:
looks like we don't need this . we can show server document in lspinfo.

PR #2295, opened 2022-12-07
feat: remove lspconfig augroup after use LspStop

### LINKED ISSUE #2216: LspStop should remove related autocmds

glepnir:
after use the `LspStop` command remove all the autocmds which in `lspconfig` augroup. that mean if there has multples servers , need use `LspStart` for each servers .

Fix #2216


justinmk:
@glepnir Was there a problem with this approach? Did we end up fixing https://github.com/neovim/nvim-lspconfig/issues/2216 in some other way?

PR #2298, opened 2022-12-08
fix: add new root into the workspace folders

glepnir:
@andrewwillette does this can works for you ?


andrewwillette:
@glepnir it does. Thank you!!

andrewwillette (approved):

PR #2300, opened 2022-12-08
Enable `terraform-ls` for `.tf` files

SG60:
The `.tf` extension seems to be what Hashicorp suggest in most of their tutorials and documentation.


gegoune:
`tf` by default relates to some mud game files or something, don't think that pr is correct.

lithammer:
Yeah the `tf` filetype isn't Terraform, it's [TinyFugue](http://tinyfugue.sourceforge.net/).

SG60:
Ah fair enough. My bad.

PR #2304, opened 2022-12-10
feat: support multiple workspaceFolders and show correct root in Lspinfo

glepnir:
if client already in `single_file_clients`, when open a new file in project . we should check the client in `single_file_clients`. then don't spawn a new server instance.  with `single_file` mode the client not initialize the `workspace_folders` field , so 
initialize this field.  then insert the project to `workspace_folders ` . 

now we don't spawn multiple clients , just use one client instance with muliple workspace folders. check the current buffer file in the project or not. and show the correct project dir in `LspInfo`, 

remove pid in `LspInfo` . the `client.rpc.pid` had removed

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -111,18 +111,13 @@ local function make_client_info(client, fname)
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
   fname = uv.fs_realpath(fname)
-  fname = is_windows and fname:lower() or fname
   local sep = is_windows and '\\' or '/'
   local fname_parts = vim.split(fname, sep, { trimempty = true })
-  --remove the host name, don't include them before compare
-  fname_parts = { unpack(fname_parts, 3) }
   if workspace_folders then
     for _, schema in pairs(workspace_folders) do
       local matched = true
       local root = uv.fs_realpath(schema.name)
-      root = is_windows and root:lower() or root
       local root_parts = vim.split(root, sep, { trimempty = true })
-      root_parts = { unpack(root_parts, 3) }
 
       for i = 1, #root_parts do
         if root_parts[i] ~= fname_parts[i] then
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -1,4 +1,4 @@
-local api, fn = vim.api, vim.fn
+local api, fn, lsp = vim.api, vim.fn, vim.lsp
 local windows = require 'lspconfig.ui.windows'
 local util = require 'lspconfig.util'
 
@@ -103,14 +103,21 @@ local function make_config_info(config, bufnr)
   return lines
 end
 
-local function make_client_info(client)
+local function make_client_info(client, fname)
   local client_info = {}
 
   client_info.cmd = cmd_type[type(client.config.cmd)](client.config)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
+  local tail_fname = fn.fnamemodify(fname, ':t')
   if workspace_folders then
-    client_info.root_dir = workspace_folders[1].name
-  else
+    for _, schema in pairs(workspace_folders) do
+      local result = vim.fs.find(tail_fname, { path = schema.name, type = 'file' })[1]
+      if result and result == fname then
+        client_info.root_dir = schema.name
+      end
+    end
+  end
+  if not client_info.root_dir then
     client_info.root_dir = 'Running in single file mode.'
   end
   client_info.filetypes = table.concat(client.config.filetypes or {}, ', ')
@@ -123,8 +130,6 @@ local function make_client_info(client)
       .. client.name
       .. ' (id: '
       .. tostring(client.id)
-      .. ', pid: '
-      .. tostring(client.rpc.pid)
       .. ', bufnr: ['
       .. client_info.attached_buffers_list
       .. '])',
@@ -150,10 +155,11 @@ end
 return function()
   -- These options need to be cached before switching to the floating
   -- buffer.
-  local buf_clients = vim.lsp.buf_get_clients()
-  local clients = vim.lsp.get_active_clients()
-  local buffer_filetype = vim.bo.filetype
   local original_bufnr = api.nvim_get_current_buf()
+  local buf_clients = lsp.get_active_clients { bufnr = original_bufnr }
+  local clients = lsp.get_active_clients()
+  local buffer_filetype = vim.bo.filetype
+  local fname = api.nvim_buf_get_name(original_bufnr)
 
   windows.default_options.wrap = true
   windows.default_options.breakindent = true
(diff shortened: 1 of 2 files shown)


musjj:
After opening a single non-project lua file, `:LspInfo` throws this error:
```
Error executing Lua callback: ...k\AppData\Local\neovim\share\nvim\runtime/lua/vim/fs.lua:203: bad argument #1 to '(for generator)' (uv_req expected, g
stack traceback:
        [C]: in function '(for generator)'
        ...\AppData\Local\neovim\share\nvim\runtime/lua/vim/fs.lua:203: in function 'find'
        ...k\packer\opt\nvim-lspconfig/lua/lspconfig/ui/lspinfo.lua:114: in function 'make_client_info'
        ...k\packer\opt\nvim-lspconfig/lua/lspconfig/ui/lspinfo.lua:203: in function <...k\packer\opt\nvim-lspconfig/lua/lspconfig/ui/lspinfo.lua:155>
        ...site\pack\packer\opt\nvim-lspconfig\plugin\lspconfig.lua:64: in function <...site\pack\packer\opt\nvim-lspconfig\plugin\lspconfig.lua:63>
```

musjj:
Also, it seems that the client now *always* run in single file mode, even in projects.

glepnir:
which file? I test it works fine in c lua etc

musjj:
Never mind, I accidentally tested it on a non-existing file.
But the single file mode issue still persists. I tested it on the `init.lua` of my nvim config:
```
.
├── .git
├── .gitignore
├── .stylua.toml
├── init.lua
├── lua
│   ├── core
│   ├── lib
│   └── plugins
└── plugin
    └── packer_compiled.lua
```
`:LspInfo`:
```
 Client: sumneko_lua (id: 1, bufnr: [1])
 	filetypes:       lua
 	autostart:       true
 	root directory:  Running in single file mode.
 	cmd:             cmd.exe /C lua-language-server
```

glepnir:
can't reproduce 
image

musjj:
Alright, it's just a display error caused by this:
https://github.com/neovim/nvim-lspconfig/blob/ab10bfe1f3fcddf1d8b75a93dc5b7519409ef268/lua/lspconfig/ui/lspinfo.lua#L115-L117
Windows can have both forward and backslashes mixed into one path, so it's best to normalize it before comparing them.
I can think of two ways of normalizing them.
With `uv.fs_realpath`:
```lua
if result and vim.loop.fs_realpath(result) == vim.loop.fs_realpath(fname) then
  client_info.root_dir = schema.name
end
```
Or with `vim.fs.normalize` and `string.lower` (Windows paths are case-insensitive):
```lua
if result and vim.fs.normalize(result):lower() == vim.fs.normalize(fname):lower() then
  client_info.root_dir = schema.name
end
```
I think the former is the most robust solution, but the latter _might_ be a _bit_ faster because it doesn't rely on I/O operations. But note that `vim.fs.normalize` still doesn't handle certain edge cases like redundant slashes (`/path//to///file`).

glepnir:
maybe we can convert it as manual ? because we convert it in core lsp just use regex..and I don't have win env. the path in win is what ?

musjj:
What do you mean by manual?
For example, this all refer to the same file:
`~\AppData\Local\nvim-data\site\pack\packer\opt\nvim-lspconfig/lua/lspconfig/ui/lspinfo.lua`
`~/AppData/Local/nvim-data/site/pack/packer/opt/nvim-lspconfig/lua/lspconfig/ui/lspinfo.lua`
`~\AppData\Local\nvim-data\site\pack\packer\opt\nvim-lspconfig\lua\lspconfig\ui\lspinfo.lua`
So if we want to do a string comparison, we need to pass them through some kind of filter first (like the one I mentioned above)

glepnir:
oh thanks ! looks like `expand` in windows will have tilde ? honestly  I don't use win 7 years ..so don't know some information of windows..

musjj:
Sorry about that, the `~` is just a placeholder. In the actual scenario it would be something like `C:\Users\Name\...`.
> use regex to convert the path.
Why not just use one of the functions above?

glepnir:
cause you mentioned performance issues. So I'm wondering if I can simply use regex to replace. can you make modifications on this pr because you can do some tests on windows.

musjj:
Ah, I don't think it actually matters, just mentioning the subtle differences between the 2 methods just for the sake of it.
`:LspInfo` is a manually triggered command so the miniscule differences caused by a bit of extra I/O would probably be invisible to the user.
The second method should also work just fine, since I don't think the function will receive any paths with redundant slashes.
Just pick whichever you prefer :), though I think writing a new normalizing function from a scratch isn't worth it.

glepnir:
nvm ... i do this .just use your first way. can you test it ?

musjj:
Found another issue. Say we have this root directory:
```
.
├── .git
├── .gitignore
├── .stylua.toml
├── init.lua
├── lua
│   ├── core
│   ├── lib
│   └── plugins
│           └── init.lua
└── plugin
    └── packer_compiled.lua
```
When we open `./lua/plugins/init.lua`, `vim.fs.find` will try to find `init.lua` in this root path, but it will end up finding `./init.lua` instead of the file we opened, `./lua/plugins/init.lua`.
So, I think it's better to just go up current path and see if we can find the root there:
```lua
if workspace_folders then
  for _, schema in pairs(workspace_folders) do
    local name = vim.fn.has 'win32' == 1 and schema.name:lower() or schema.name
    for dir in vim.fs.parents(fname) do
      dir = vim.fn.has 'win32' == 1 and dir:lower() or dir
      if dir == name then
        client_info.root_dir = schema.name
      end
    end
  end
end
```
`vim.loop.fs_realpath` can't be used here, because `vim.fs.parents` will normalize the path with `vim.fs.normalize` anyways.
What do you think?

glepnir:
can we split the fname and schema.name by os path. convert the schema.name as first, then compare the two tables. if all the elements in dir both in the fname table. can we make sure it's the root dir of fname ?

musjj:
Good idea. It's longer and more manual, but should be more robust:
```lua
if workspace_folders then
  for _, schema in pairs(workspace_folders) do
    local root = vim.loop.fs_realpath(schema.name)
    local filename = vim.loop.fs_realpath(fname)
    local sep
    if fn.has 'win32' == 1 then
      root = root:lower()
      filename = filename:lower()
      sep = '\\'
    else
      sep = '/'
    end

    local root_parts = vim.split(root, sep, { trimempty = true })
    local filename_parts = vim.split(filename, sep, { trimempty = true })
    local matched = true
    for i = 1, #root_parts do
      if root_parts[i] ~= filename_parts[i] then
        matched = false
        break
      end
    end

    if matched then
      client_info.root_dir = schema.name
    end
  end
end
```

glepnir:
and we can remove the index 1 and  2 of these tables ? because they are host name  right ? wonder index 1 and 2 same in windows ?

musjj:
What do you mean by host names? The typical Windows path after the split would probably look like;
`{ "C:", "path", "to", "something" }`
The first key is the hard drive letter, so it's definitely important.
To be honest, I'm not sure if the `:lower()` is needed. I added it because Windows paths are case-insensitive (`C:\path\to\file` and `C:\PaTh\To\FiLe` refers to the same path)
But because we've passed the path to `vim.loop.fs_realpath` first, the casing should have been taken care of. You can probably safely remove that part.

glepnir:
you are right. windows should respect the hard drive letter . I will try to remove the lower function ..then can you give me a feedback ? thanks

musjj:
Seems to work fine for me 👍

glepnir:
Thanks for your kind help :) then merge

PR #2307, opened 2022-12-11
feat: add rockspec

MrcJkb:
nvim-lspconfig is a common dependency across Neovim plugins. Using luarocks may alleviate the need for users to specify their plugins' dependencies in their plugin manager.
(e.g., vim-plug or packer).

See also:
https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html

### DIFF
--- a/nvim_lspconfig-scm-1.rockspec
+++ b/nvim_lspconfig-scm-1.rockspec
@@ -1,0 +1,38 @@
+local _MODREV, _SPECREV = 'scm', '-1'
+
+rockspec_format = "3.0"
+package = 'nvim_lspconfig'
+version = _MODREV .. _SPECREV
+
+description = {
+  summary = "A collection of common configurations for Neovim's built-in language server client.",
+  detailed = [[
+    This plugin allows for declaratively configuring, launching, and initializing language servers you have installed on your system.
+    Language server configurations are community-maintained.
+  ]],
+  homepage = 'https://github.com/neovim/nvim-lspconfig',
+  license = 'Apache/2.0',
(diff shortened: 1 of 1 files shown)


justinmk (approved):
Happy to see this! Do we need to document how and when to bump the version? (May need to add a "Release" section to bottom of readme.)

MrcJkb:
Good point.
I have set the version to `scm`, which (afaict) tells luarocks to fetch the latest version from GitHub.

If we want to configure the rockspec to align with the release versions, I would suggest working towards a github actions workflow that automatically creates a PR with a rockspec version bump on release.

nvim_lspconfig-scm-1.rockspec
```diff
@@ -0,0 +1,38 @@
...
+package = 'nvim_lspconfig'
+version = _MODREV .. _SPECREV
+
+description = {
+  summary = "A collection of common configurations for Neovim's built-in language server client.",
+  detailed = [[
+    This plugin allows for declaratively configuring, launching, and initializing language servers you have installed on your system.
+    Language server configurations are community-maintained.
+  ]],
+  homepage = 'http://github.com/nevim/nvim-lspconfig',
```
teto, line 13: typo here

teto:
@justinmk the only issue is to deal with who is going to update the package on www.luarocks.org since it can be done from only one account. See https://github.com/luarocks/luarocks-site/issues/173. 

@MrcJkb thanks for doing this !

teto:
added to https://luarocks.org/labels/neovim

PR #2313, opened 2022-12-13
fix(gdscript): support get the port from env variable

### LINKED ISSUE #2311: Allow customizing gdscript port
### Language server

gdscript

### Requested feature

The gdscript configuration hard-codes 6008 (https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/gdscript.lua), while Godot allows for customizing the port (see screenshot).

image

Since Godot 3 and 4 uses different ports, it would be great to allow customizing the port.

Godot's VS Code plugin has an option to change the port.
image

Also it seems CoC has it, however I'm not experienced using it. https://github.com/habamax/vim-godot#setting-up-cocnvim-for-godot-322

glepnir:
support get the port number from env variable `GDSript_Port`

Fix #2311

### DIFF
--- a/lua/lspconfig/server_configurations/gdscript.lua
+++ b/lua/lspconfig/server_configurations/gdscript.lua
@@ -1,6 +1,6 @@
 local util = require 'lspconfig.util'
 
-local port = os.getenv('GDScript_Port') or '6008'
+local port = os.getenv 'GDScript_Port' or '6008'
 local cmd = { 'nc', 'localhost', port }
 
 if vim.fn.has 'nvim-0.8' == 1 then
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/gdscript.lua
```diff
@@ -1,9 +1,10 @@
 local util = require 'lspconfig.util'
 
-local cmd = { 'nc', 'localhost', '6008' }
+local port = os.getenv('GDScript_Port') or '6008'
```
glepnir, line 3:
```suggestion
local port = os.getenv 'GDScript_Port' or '6008'
```

glepnir, line 3:
```suggestion
local port = os.getenv 'GDScript_Port' or '6008'
```

PR #2314, opened 2022-12-13
fix: run all matched client in LspStart

### LINKED ISSUE #2312: Client not attached to buffer after :LspStart for the second project with autostart=false
### Description

When running :LspStart for the second project with autostart=false it doesn't attach client to the buffer.

My understanding that it happens because of the following logic in M.get_other_matching_provider (https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/util.lua#L424)
```
if not vim.tbl_contains(active_clients_list, config.name) then
      local filetypes = config.filetypes or {}
      for _, ft in pairs(filetypes) do
        if ft == filetype then
          table.insert(other_matching_configs, config)
        end
      end
    end
end
```
Because we already have 1 active client (vim.tbl_contains(active_clients_list, config.name)) this function doesn't return matching config for LspStart() call.

### Neovim version

NVIM v0.9.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

0c038493b37e67bed287ff99722a9ced5cdfe617

Alt Linux Sisyphus

### Affected language servers

clangd

### Steps to reproduce

1. Consider you have two C projects - project A and project B. Also assume that you have autostart=false in your init.lua
2. Go to the file in project A and run LspStart. Execute LspInfo and check that client is attached to the buffer.
3. Go to the file in project B and run LspStart. Execute LspInfo and check the actual behavior.

### Actual behavior

LspInfo reports that client is not attached to the buffer from project B.

### Expected behavior

Client should be attached to the buffer.

### Minimal config

```Lua
local nvim_lsp = require('lspconfig')

-- Use an on_attach function to only map the following keys
-- after the language server attaches to the current buffer
  local function buf_set_keymap(...) vim.api.nvim_buf_set_keymap(bufnr, ...) end
  local function buf_set_option(...) vim.api.nvim_buf_set_option(bufnr, ...) end

  --Enable completion triggered by <c-x><c-o>

  -- Mappings.
  local opts = { noremap=true, silent=true }

  -- See `:help vim.lsp.*` for documentation on any of the below functions
  buf_set_keymap('n', 'gD', '<Cmd>lua vim.lsp.buf.declaration()<CR>', opts)
  buf_set_keymap('n', 'gd', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
  buf_set_keymap('n', '<M-.>', '<Cmd>lua vim.lsp.buf.definition()<CR>', opts)
  buf_set_keymap('n', '<F9>', '<Cmd>lua vim.lsp.buf.hover()<CR>', opts)
  buf_set_keymap('n', 'gi', '<cmd>lua vim.lsp.buf.implementation()<CR>', opts)
  buf_set_keymap('n', '<C-k>', '<cmd>lua vim.lsp.buf.signature_help()<CR>', opts)
  buf_set_keymap('n', '<leader>wa', '<cmd>lua vim.lsp.buf.add_workspace_folder()<CR>', opts)
  buf_set_keymap('n', '<leader>wr', '<cmd>lua vim.lsp.buf.remove_workspace_folder()<CR>', opts)
  buf_set_keymap('n', '<leader>wl', '<cmd>lua print(vim.inspect(vim.lsp.buf.list_workspace_folders()))<CR>', opts)
  buf_set_keymap('n', '<leader>D', '<cmd>lua vim.lsp.buf.type_definition()<CR>', opts)
  buf_set_keymap('n', '<leader>rn', '<cmd>lua vim.lsp.buf.rename()<CR>', opts)
  buf_set_keymap('n', '<leader>ca', '<cmd>lua vim.lsp.buf.code_action()<CR>', opts)
  buf_set_keymap('n', 'gr', '<cmd>lua vim.lsp.buf.references()<CR>', opts)
  buf_set_keymap('n', '<leader>e', '<cmd>lua vim.diagnostic.open_float()<CR>', opts)
  buf_set_keymap('n', '[c', '<cmd>lua vim.diagnostic.goto_prev()<CR>', opts)
  buf_set_keymap('n', ']c', '<cmd>lua vim.diagnostic.goto_next()<CR>', opts)
  buf_set_keymap('v', '[c', '<cmd>lua vim.diagnostic.goto_prev()<CR>', opts)
  buf_set_keymap('v', ']c', '<cmd>lua vim.diagnostic.goto_next()<CR>', opts)
  buf_set_keymap('n', '<leader>q', '<cmd>lua vim.diagnostic.setloclist()<CR>', opts)
  buf_set_keymap("n", "<leader>f", "<cmd>lua vim.lsp.buf.formatting_sync()<CR>", opts)
  buf_set_keymap("v", "<leader>f", "<cmd>lua vim.lsp.buf.range_formatting()<CR>", opts)
end

-- Use a loop to conveniently call 'setup' on multiple servers and
-- map buffer local keybindings when the language server attaches
local servers = { "clangd", "gopls", "pyright", "rls" }
local capabilities = require('cmp_nvim_lsp').default_capabilities()
for _, lsp in ipairs(servers) do
  nvim_lsp[lsp].setup {
    on_attach = on_attach,
    autostart = false,
    flags = {
      debounce_text_changes = 150,
    },
    settings = {
       gopls = {
         staticcheck = true
       },
    },
    capabilities = capabilities,
  }
end
```

### LSP log

https://gist.github.com/vladimir-didenko/cbdc3b3b1fb48b5c1a8da2874a08d5cb

glepnir:
Problem:   when `LspStart` without any argument it will launch the server which not in active even the filetypes match the 
                   current buffer .
Solution:  currently we support client with multiples workspaceFolders , it don't spawn another server instance . so no need 
                 use `get_other_matching_config` function .just get all matched config then launch them. 

Fix #2312

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -438,8 +438,22 @@ function M.get_other_matching_providers(filetype)
   return other_matching_configs
 end
 
+function M.get_config_by_ft(filetype)
+  local configs = require 'lspconfig.configs'
+  local matching_configs = {}
+  for _, config in pairs(configs) do
+    local filetypes = config.filetypes or {}
+    for _, ft in pairs(filetypes) do
+      if ft == filetype then
+        table.insert(matching_configs, config)
+      end
+    end
+  end
+  return matching_configs
+end
+
 function M.get_active_client_by_name(bufnr, servername)
-  for _, client in pairs(vim.lsp.buf_get_clients(bufnr)) do
+  for _, client in pairs(vim.lsp.get_active_clients { bufnr = bufnr }) do
     if client.name == servername then
       return client
     end
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -76,8 +76,8 @@ api.nvim_create_user_command('LspStart', function(info)
     end
   end
 
-  local other_matching_configs = require('lspconfig.util').get_other_matching_providers(vim.bo.filetype)
-  for _, config in ipairs(other_matching_configs) do
+  local matching_configs = require('lspconfig.util').get_config_by_ft(vim.bo.filetype)
+  for _, config in ipairs(matching_configs) do
     config.launch()
   end
 end, {


lua/lspconfig/util.lua
```diff
@@ -438,8 +438,22 @@ function M.get_other_matching_providers(filetype)
...
+        table.insert(matching_configs, config)
+      end
+    end
+  end
+  return matching_configs
+end
+
 function M.get_active_client_by_name(bufnr, servername)
-  for _, client in pairs(vim.lsp.buf_get_clients(bufnr)) do
+  for _, client in pairs(vim.lsp.get_active_clients { bufnr = bufnr }) do
```
williamboman, line 456: `vim.lsp.get_active_clients()` was added in `0.8.0` and as I understand it nvim-lspconfig still supports `0.7.0`. Maybe it'd be best to defer these changes until `0.7.0` support is dropped?

glepnir: hmm i don't notice that it added in 0.8. use a variable with version check in there would better

PR #2316, opened 2022-12-15
fix: spawning new lsp client when new root_dir

andrewwillette:
@glepnir, the same issue that [this PR resolved](https://github.com/neovim/nvim-lspconfig/pull/2298) where a new lsp client for the same lsp type would not be started for a different project has been re-introduced.

This PR resolves the problem. I'm not sure how the above PR fixed it in first place, and why it got re-introduced. Sorry.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -192,7 +192,6 @@ M.path = (function()
         return
       end
     end
-
     return it, path, path
   end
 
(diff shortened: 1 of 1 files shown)


andrewwillette:
@glepnir , can you rerun CI build? I think the lint will pass now. Apologies.

glepnir:
this check looks a little wired. can you provide a reproduce step?

PR #2320, opened 2022-12-16
fix(lspinfo): safely expand filenames that don't exist on fs

williamboman:
The filename of a buffer may point to a node on the file system that
does not yet exist. `uv.fs_realpath()` will return nil in these cases,
causing a nil exception later on. The builtin `resolve()` function will
return the provided filename as-is if it does not yet exist.

Alternatively we could stick to libuv APIs.

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -103,6 +103,8 @@ local function make_config_info(config, bufnr)
   return lines
 end
 
+---@param client table
+---@param fname string
 local function make_client_info(client, fname)
   local client_info = {}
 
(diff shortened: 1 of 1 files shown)


lua/lspconfig/ui/lspinfo.lua
```diff
@@ -103,14 +103,16 @@ local function make_config_info(config, bufnr)
...
+---@param fname string
 local function make_client_info(client, fname)
   local client_info = {}
 
   client_info.cmd = cmd_type[type(client.config.cmd)](client.config)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = uv.fs_realpath(fname)
+  fname = vim.fn.fnamemodify(vim.fn.resolve(fname), ":p")
```
williamboman, line 115: Might be preferable to just do:

```suggestion
  fname = uv.fs_realpath(fname) or fname
```

justinmk (approved):

PR #2321, opened 2022-12-17
Docs: Add instructions for instant startup for julials

josephsdavid:
## Description

When running julials without a sysimage, it takes a significant amount of time to JIT compile the LSP (about 25 seconds on my machine). This is really annoying. @thecedarprince and @fredrikekre (and @kdheepak in julia slack) wrote [a guide to setting up julials with instant startup](https://discourse.julialang.org/t/neovim-languageserver-jl/37286/83), which I feel deserves to be shared/documented in a more visible place! This PR essentially takes their guide and documents it here, however I am unsure whether it is best to put the nontrivial bits of the guide in the documentation, or if a link in the documentation would suffice :)


glepnir:
can we add it into a wiki page ?

josephsdavid:
> can we add it into a wiki page ?

This is a far better idea!!

glepnir:
okay close then.

PR #2323, opened 2022-12-17
feat: added shacl, sparql, stardog_graphql, stardog_srs and trig language servers

ArneCJacobs:
Added most language servers developed by stardog: https://github.com/stardog-union/stardog-language-servers.
All new configs were initially copied from `turtle_ls.lua`, common code was refactored out added as a function to `lua/lspconfig/util.lua`

### DIFF
--- a/lua/lspconfig/server_configurations/sparql_ls.lua
+++ b/lua/lspconfig/server_configurations/sparql_ls.lua
@@ -1,0 +1,28 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'sparql-language-server'
+local full_path = util.find_bin_path(bin_name)
+
+local cmd = { 'node', full_path, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', 'node', full_path, '--stdio' }
+end
+
+return {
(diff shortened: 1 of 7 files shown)


lua/lspconfig/server_configurations/sparql_ls.lua
```diff
@@ -0,0 +1,28 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'sparql-language-server'
+local full_path = util.find_bin_path(bin_name)
+
+local cmd = { 'node', full_path, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', 'node', full_path, '--stdio' }
```
justinmk, line 9: probably time for this to be lifted into util as well. That at least "stops the bleeding" and makes it easier to migrate to a full solution later.

lua/lspconfig/util.lua
```diff
@@ -453,4 +453,46 @@ function M.strip_archive_subpath(path)
   return path
 end
 
+
+function M.find_bin_path(bin_name)
```
glepnir, line 457: If we want to provide a public api in util it should not have limitations.

PR #2330, opened 2022-12-20
fix: send the lsp method request after insert workspace folders

glepnir:


glepnir:
@xiantang can you test this pr ?

xiantang:
> @xiantang can you test this pr ?

yes I can, I have tried but still have issues.. may we need to also change the document using `textDocument/didChange`

xiantang:
hold on.

xiantang:
look good

xiantang:
i use `        client.notify('workspace/didChangeWorkspaceFolders', params)` have issue,
but use your code look good now.

glepnir:
thanks for your kind reply

xiantang:
let me reproduce more than 5 times, if not have issue it will be ok

xiantang:
seems still have issue.. Let me sent a screenshot.

xiantang:
Kapture 2022-12-20 at 10 22 42

lua/lspconfig/util.lua
```diff
@@ -272,6 +272,7 @@ function M.server_per_root_dir_manager(make_config)
           { { uri = vim.uri_from_fname(root_dir), name = root_dir } },
           { {} }
         )
+        vim.lsp.buf_notify(0, 'workspace/didChangeWorkspaceFolders', params)
```
xiantang, line 275: `client.notify('workspace/didChangeWorkspaceFolders', params)` and 
        `vim.lsp.buf_notify(0, 'workspace/didChangeWorkspaceFolders', params)` are both not working in gopls

glepnir: in air project open another go project still get an error right?

xiantang: yes

glepnir: hmm. there already have `return client.id`

xiantang: this return is in
function  ` function manager.add` 

`return client.id` is in outer function

glepnir: I think you need this https://github.com/golang/tools/blob/master/gopls/doc/workspace.md

glepnir: looks like recently go most suggest use the go work and in my test it works fine .

xiantang: thank you, so it means I need to put the project into the same root path and create a go.work for it.
but sadly my company's go version is 1.17 which is not supported go.work, it still has issues for me without go.work.
in the previous version which is for each workspace to start a new LSP server works well, but in the master branch the behavior is broken.

glepnir: > in the previous version which is for each workspace to start a new LSP server works well, but in the master branch the behavior is broken.

This is a mistake. When you open a file in a project it spawns a new process. These items should be parsed by a language server instance.

a useful ref  https://github.com/golang/vscode-go/issues/2112 . there has a setting maybe you can use in gopls

xiantang (approved):
LGTM

lua/lspconfig/util.lua
```diff
@@ -257,7 +257,13 @@ function M.server_per_root_dir_manager(make_config)
         return
       end
       local client = lsp.get_client_by_id(id)
-      if client and client.name == conf.name then
+      if
```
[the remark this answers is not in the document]
glepnir: I think the design about `clients` is not need ..there need a redesign and refactor. but it's in another pr.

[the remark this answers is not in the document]
xiantang: I supposed that maybe need to add test case coverage for this code to keep the behavior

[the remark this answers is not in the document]
glepnir: Test cases for client behavior do not exist in current lspconfig. Most of the testing on the client side is done in the core. Most of the tests of lspconfig are tests of some public APIs of util.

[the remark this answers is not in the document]
glepnir: Maybe copy some core test code. Then do some things on lspconfig. But there is a little trouble.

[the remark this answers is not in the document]
xiantang: yes, ic. I mean if we need to do some refactoring easily having test coverage will be better, in case the code break some logic.
but actually, I have write `lua` test before, so I don't know the effort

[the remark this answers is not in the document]
glepnir: >  if we need to do some refactoring easily having test coverage will be better, in case the code break some logic.

this is correct . but some history reason ? I don't know ,maybe one day in the future there will be a change here :)

glepnir:
hmm found a problem that there is no `client.server_capabilities.workspaceFolders` ...see #2334

AlexOperator:
I don't know whether you are aware of it in #2334, but there is an issue with the way that language server instances are being started when there are more than one workspace.

For some reason what I'm seeing is that when the second instance of the language server is started, it is run as the first was, with the same arguments, but then what appear to be the correct arguments trail after them.

For example:

```
dotnet /opt/omni/OmniSharp.dll --languageserver -z -s /Users/alex/Work/workspace1 --hostPID 63560 DotNet:enablePackageRestore=false --encoding utf-8 --languageserver FormattingOptions:EnableEditorConfigSupport=true Sdk:IncludePrereleases=true

dotnet /opt/omni/OmniSharp.dll --languageserver -z -s /Users/alex/Work/workspace1 --hostPID 63560 DotNet:enablePackageRestore=false --encoding utf-8 --languageserver FormattingOptions:EnableEditorConfigSupport=true Sdk:IncludePrereleases=true -z -s /Users/alex/Work/workspace2 --hostPID 63560 DotNet:enablePackageRestore=false --encoding utf-8 --languageserver FormattingOptions:EnableEditorConfigSupport=true Sdk:IncludePrereleases=true
```

PR #2333, opened 2022-12-20
feat: add jq support

xiantang:
what is jq: https://stedolan.github.io/jq/manual/
jqlsp: https://github.com/wader/jq-lsp

Syntax checking like below:

image


glepnir:
change it to `jqls` would be better.

xiantang:
> change it to `jqls` would be better.

done

glepnir:
Thanks for your contribution :)

PR #2334, opened 2022-12-20
refactor: remove single_clients and improve clients reuse

glepnir:
The design here is a bit wired. It stores clients in categories. It is possible to reuse the language server through the workspace. There is no need for classified storage. And the logic here is a bit long-winded. You only need to judge whether root_dir is in the clients table. If in we return the existing client. If not find all stored clients. If there is a client that needs it and supports workspace. Then return directly to this client. Then insert root into the workspace field. Trigger `didChangeWorkspaceFolders`. For non-existent we start new client.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -272,12 +272,15 @@ function M.server_per_root_dir_manager(make_config)
     local client = get_client_from_cache(new_config)
 
     if client then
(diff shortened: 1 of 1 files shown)


xiantang:
related to https://github.com/neovim/nvim-lspconfig/pull/2330#issuecomment-1362164777

PR #2335, opened 2022-12-20
feat: add ruff-lsp support

igorlfs:


igorlfs:
Hey @glepnir, I just noticed the file should be named `ruff_lsp.lua` instead of `ruff-lsp.lua`, can you fix that?

glepnir:
sure

charliermarsh:
Thanks for this.

PR #2343, opened 2022-12-23
fix(lspinfo): normalize `fname` path correctly

musjj:
`vim.fn.fnamemodify` and `vim.fn.resolve` doesn't normalize slashes in Windows paths, which can cause `:LspInfo` to falsely report that the current attached client is in single file mode.
As far as I know `uv.fs_realpath` does everything that `vim.fn.fnamemodify(..., ":p")` and `vim.fn.resolve` can do and more, so I removed them.

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -112,7 +112,7 @@ local function make_client_info(client, fname)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = vim.loop.fs_realpath(fname) or vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
+  fname = uv.fs_realpath(fname) or fn.fnamemodify(fn.resolve(fname), ':p')
   if is_windows then
     fname:gsub('%/', '%\\')
   end
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -112,7 +112,7 @@ local function make_client_info(client, fname)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
+  fname = vim.loop.fs_realpath(fname)
   local sep = is_windows and '\\' or '/'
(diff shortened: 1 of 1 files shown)


glepnir:
relate #2320

musjj:
I guess we can just normalize the slashes manually?

glepnir:
yes. I prefer manually

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -112,7 +112,7 @@ local function make_client_info(client, fname)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
+  fname = vim.loop.fs_realpath(fname)
```
williamboman, line 115:
```suggestion
  fname = vim.loop.fs_realpath(fname) or fname
```

williamboman (commented):
Ah, interesting. Maybe we should stick to libuv, but we'll need to default to `fname` as-is (see #2320)

musjj:
Re-added the previous method as a fallback + a few logic to handle funky separators. What do you think?

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -112,9 +112,14 @@ local function make_client_info(client, fname)
   local workspace_folders = fn.has 'nvim-0.9' == 1 and client.workspace_folders or client.workspaceFolders
   local uv = vim.loop
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
+  fname = vim.loop.fs_realpath(fname) or vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
```
glepnir, line 115:
```suggestion
  fname = uv.fs_realpath(fname) or fn.fnamemodify(fn.resolve(fname), ':p')
```

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -112,9 +112,14 @@ local function make_client_info(client, fname)
...
   local is_windows = uv.os_uname().version:match 'Windows'
-  fname = vim.fn.fnamemodify(vim.fn.resolve(fname), ':p')
+  fname = uv.fs_realpath(fname) or fn.fnamemodify(fn.resolve(fname), ':p')
+  if is_windows then
+    fname:gsub('%/', '%\\')
+  end
   local sep = is_windows and '\\' or '/'
-  local fname_parts = vim.split(fname, sep, { trimempty = true })
+  local fname_parts = vim.tbl_filter(function(v)
+    return v ~= ''
```
glepnir, line 121:
```suggestion
    return #v > 0
```

PR #2351, opened 2022-12-25
fix: check client exist supports_method

glepnir:
# Problem
 not all the clients have the field `supports_method`

# Solution
 add a field check


glepnir:
@gregorias @lvim-tech does this pr can make it works fine for you ?

PR #2356, opened 2022-12-25
fix: check workspace after client initial

### LINKED ISSUE #2355: b0bd429 breaks sumneko_lua
### Description

When I switch between a file in my workspace and a file outside the worksapce, sumneko diagnostics get reset on all buffers and all buffers start showing errors.

The error only starts happening after b0bd429

See also https://github.com/folke/neodev.nvim/issues/98

### Neovim version

nightly

### Nvim-lspconfig version

_No response_

linux

### Affected language servers

sumneko_lua

### Steps to reproduce

1. open a lua file in your project directory that is in .gitignore
2. open a lua file in your project

### Actual behavior

The second file will list a lot of errors

### Expected behavior

no errors

### Minimal config

```Lua
..
```

### LSP log

...

glepnir:
# Problem
 logic bug

# Solution
 rewrite

Fix #2355


folke:
Can confirm that the lastest commits here fix #2355!

glepnir:
hmmm wired . there has something I don't know , as the sumneko's wiki said  there need restart server after insert workspace. I just send the `didChangeworkspacefolders` request to it . if server need restart that mean `client.stop`  then
start new client ? this will spawn a new process with new pid, we don't have client.restart function, use new config start or use config with `attached_buffers` and `workspace_folders` data from before client ? I don't know here how the sumeko lua-language-server work..

https://github.com/sumneko/lua-language-server/wiki/Developing#multiple-workspace-support

folke:
Looking at the sumneko code, it does actually support dyanmically changing the workspace folders, but before this PR, diagnostics for the current buffer would be wrong. Haven't checked your PR here, but the incorrect diagnostics now no longer show up.

folke:
Should I re-open that linked sumneko issue?

glepnir:
> it does actually support dyanmically changing the workspace folders

don't look at the sumneko code .if it support maybe wiki is old.

> but the incorrect diagnostics now no longer show up.

Just to make sure this PR solves the actual problem. I will wait for you to use it for a while. If it doesn't have any problems anymore I will merge.

> Should I re-open that linked sumneko issue?

maybe not. we don't have any handlers for request `workspace/diagnostic? / *` in core and I also don't see them in lsp-mode  lsp-bridge vim-lsp yegappan/lsp .. as lsp spec said this is useful. don't know why there. 

 https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_diagnostic

folke:
Oh right. I actually implemented the `workspace/diagnostic/refresh` handler in my config. Still need to upstream that to core.

glepnir:
there has something to do in core diagnostic , I only have a rough idea of how `workspace/diagnostic` works, but it will take some trying when i have time .

glepnir:
now confirm right ?

folke:
I've done more testing and this PR seems to fix the issues I had with diagnostics.

glepnir:
thanks for your kind reply.

PR #2363, opened 2022-12-27
fix(ruff-lsp): update docs to reflect init_options

charliermarsh:
`ruff-lsp` currently uses `init_options` and doesn't support settings. (I'll probably fix this, but for now the docs are incorrect.)

### DIFF
--- a/lua/lspconfig/server_configurations/ruff_lsp.lua
+++ b/lua/lspconfig/server_configurations/ruff_lsp.lua
@@ -2,6 +2,7 @@ local util = require 'lspconfig.util'
 
 local root_files = {
   'pyproject.toml',
+  'ruff.toml',
 }
 
 return {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/ruff_lsp.lua
```diff
@@ -26,16 +27,16 @@ Extra CLI arguments for `ruff` can be provided via
 
 ```lua
 require'lspconfig'.ruff_lsp.setup{
-  settings = {
-    ruff_lsp = {
+  init_options = {
+    settings = {
       -- Any extra CLI arguments for `ruff` go here.
-      args = {},
+      args = { '--line-length=120' },
```
lithammer, line 33: What's the reasoning for this?

charliermarsh: (It was meant to demonstrate the argument-passing style, but perhaps that's out-of-character for these docs.)

polyzen (approved):

lua/lspconfig/server_configurations/ruff_lsp.lua
lithammer: Nah I was too quick to read and didn't notice it was for the docs (I thought you updated the server config). So I guess it would've been fine.

PR #2370, opened 2022-12-30
feat: add cabal support to hls

### LINKED ISSUE #2369: Haskell HLS v1.9.0.0 should be enabled on .cabal files
### Description

The super-recent release of HLS 1.9.0.0 allows HLS to work on `.cabal` files, but only that version (and up). If it is feasible to first detect this and then set the `filetypes` field, I would be more than happy to make a PR. Otherwise I am sure it can be solved in a custom manner.

### Neovim version

--irrelevant

### Nvim-lspconfig version

_No response_

macOS 12.6.1

### Affected language servers

hls

### Steps to reproduce

--irrelevant

### Actual behavior

https://github.com/neovim/nvim-lspconfig/blob/7d0ce54a4cf6e844d87e49d1576c0fffb6d27a6e/lua/lspconfig/server_configurations/hls.lua#L6

Only supported on haskell and lhaskell filetypes.

### Expected behavior

also cabal filetype

### Minimal config

```Lua
--irrelevant
```

### LSP log

--irrelevant

mizlan:
* add note for when to modify filetypes field
* specify cabalfmt as the formatter for cabal files

Closes #2369

### DIFF
--- a/lua/lspconfig/server_configurations/hls.lua
+++ b/lua/lspconfig/server_configurations/hls.lua
@@ -14,6 +14,7 @@ return {
     settings = {
       haskell = {
         formattingProvider = 'ormolu',
+        cabalFormattingProvider = 'cabalfmt',
       },
     },
     lspinfo = function(cfg)
@@ -39,6 +40,14 @@ return {
 https://github.com/haskell/haskell-language-server
 
 Haskell Language Server
+
+If you are using HLS 1.9.0.0, enable the language server to launch on Cabal files as well:
+
+```lua
+require('lspconfig')['hls'].setup{
+  filetypes = { 'haskell', 'lhaskell', 'cabal' },
+}
+```
     ]],
 
     default_config = {


mizlan:
tested on ghc 9.4.4, hls 1.9.0.0 ^^

glepnir:
can we add this filetypes into default filetypes ? looks like there is no reason not use the latest server version ?

mizlan:
Prebuild binaries not fully backward compatible, for example 1.9.0.0 work with

    8.10.7
    9.0.2
    9.2.5
    9.4.3
    9.4.4

1.8.0.0 for example built with these


    8.6.5
    8.8.4
    8.10.7
    9.0.2
    9.2.3
    9.2.4
    9.4.1
    9.4.2

So there is still reason to use older servers, which many people will probably be using.

Also putting cabal in default will be a breaking change. Since for 1.8.0.0 it will get an error in a cabal file.

mizlan:
However, what we can do is put `cabal` in default and ask users to specify `filetypes = { 'haskell', 'lhaskell' }` for their own configuration in the docs.

glepnir:
thanks for your explain about this server.

Issue #2015, opened 2022-07-19
denols: root_dir ignore util.root_pattern


anasrar:
### Description

deno recommend installation using their script which install deno on local user `~/.deno/bin` and we need to add `~/.deno/bin` to `PATH` in user shell profile `~/.bashrc`
we can run deno, but not in neovim, it give warning every time open typescript file
```
Spawning language server with cmd: `deno` failed. The language server is either not installed, missing from PATH, or not executable.
```

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

### Nvim-lspconfig version

347947355ba0a15d803cbd61b18f8bb8f401c793

### Operating system and version

EndeavourOS - Kernel: Linux 5.18.10-arch1-1 - Architecture: x86-64

### Affected language servers

denols

### Steps to reproduce

- install deno with shell script they provide https://deno.land/#installation and add deno binary path to `PATH` in `~/.bashrc`
- restart terminal or `source` the shell profile
- open neovim and then open typescript file

### Actual behavior

when you open typescript file you will see warning
```
Spawning language server with cmd: `deno` failed. The language server is either not installed, missing from PATH, or not executable.
```
`LspInfo` output
```
Config: denols
Refer to :h lspconfig-root-detection for help.
filetypes:         javascript, javascriptreact, javascript.jsx, typescript, typescriptreact, typescript.tsx
root directory:    Not found. Searched for: root_pattern("deno.json", "deno.jsonc", ".git").
cmd:               deno lsp
cmd is executable: Unable to find executable. Please check your path and ensure the server is installed
autostart:         true
custom handlers:   textDocument/definition, textDocument/references
```

### Expected behavior

before I update `nvim-lspconfig` it respect with `util.root_pattern('deno.json')` and not showing warning if the project is not deno project, I think it does not using environment shell that open up neovim.

when I change to absolute path it works
```lua
nvim_lsp.denols.setup {
  cmd = { "/home/user/.deno/bin/deno", "lsp" },
  on_attach = on_attach,
}
```

### Minimal config

```Lua
local on_windows = vim.loop.os_uname().version:match 'Windows'

local function join_paths(...)
  local path_sep = on_windows and '\\' or '/'
  local result = table.concat({ ... }, path_sep)
  return result
end

vim.cmd [[set runtimepath=$VIMRUNTIME]]

local temp_dir = vim.loop.os_getenv 'TEMP' or '/tmp'

[...]

PR #2375, opened 2023-01-01
feat(tsserver): set `single_file_support` to true

rjooske:
Set `single_file_support` for tsserver to true by default, as typescript-language-server supports single-file mode.

### DIFF
--- a/lua/lspconfig/server_configurations/tsserver.lua
+++ b/lua/lspconfig/server_configurations/tsserver.lua
@@ -23,6 +23,7 @@ return {
       return util.root_pattern 'tsconfig.json'(fname)
         or util.root_pattern('package.json', 'jsconfig.json', '.git')(fname)
     end,
+    single_file_support = true,
   },
   docs = {
     description = [[


glepnir:
can you post the relate doc ? I don't find the typescipt-language-server support single file.

rjooske:
Actually, I also have tried to find something that mentions typescript-language-server's single file support, but couldn't find any.
My original idea was that because VSCode uses typescript-language-server and you can edit a single JS, TS, and such files with LSP features on VSCode, there's no reason that the same can't be done on Neovim.
I've been using Neovim + typescript-language-server with `single_file_support` = true, and so far it seems to work exactly like how VSCode works when editing a single JS, TS, etc files.
I'm sorry but I can't provide any other related (and possibly useful) information :(
Hope this helps.

glepnir:
vscode use tsserver not this server.

rjooske:
Oh okay, I think I was very confused about the terminology here.
Let me see if I understand it correctly now:
- tsserver
  - is part of the TypeScript ecosystem
  - is used by VSCode for completion, hover, etc
  - is **the name** of the server config for JS/TS in nvim-lspconfig, but NOT actually used by nvim-lspconfig
- typescript-language-server
  - is a language server that can speak LSP
  - is used by nvim-lspconfig to implement LSP features, under the name "tsserver"

And my assumption is that:
- tsserver
  - supports single file mode because VSCode's behavior suggests so
- typescript-language-server
  - supports single file mode because after setting `single_file_support` to true, completion, hover, etc are working as expected on Neovim

So to summarize, my original reason as to why I thought setting `single_file_support` to true would work turns out to be very much wrong, but the outcome (being able to use LSP features) happens to match my expectation by chance.
Although it doesn't answer the question of whether there is documentation about typescript-language-server's single file support...
All I can do now is say "it works on my machine" ¯\_(ツ)_/¯

glepnir:
nope . the typescript-language-server is wrapped tsserver. tsserver only work in vscode.  about `single_file_mode ` in typescript-language-server.  it's wrapped tsserver so I think it should support `single_file_mode`. but I don't know why does not add this option in default. maybe there has something wrong or something else. maybe you can test all lsp functions with  `single_file_mode` , if both works fine. we can add it into default .

rjooske:
> the typescript-language-server is wrapped tsserver

That makes a lot of sense!

>  you can test all lsp functions with `single_file_mode`

Sorry, I'm not entirely sure about what you mean by "test"...

glepnir:
oh I mean the functions `vim.lsp.buf.*`

rjooske:
Okay, so... are we going to make sure that functions in `vim.lsp.buf.*` don't do anything out of the ordinary with `single_file_support = true` or something? I still don't understand what you want to guarantee by "testing" those functions...

Also, I found [this discussion](https://github.com/typescript-language-server/typescript-language-server/issues/357) where you can find when single-file mode was introduced to typescript-language-server, down to the exact commit! Hopefully this will eliminate the need for whatever we're trying to ensure ourselves.

glepnir:
oh this is useful issue. so should support it .

nickhealy:
I think this breaks the `denols`, or at least requires some change in documentation. According to the [Deno documention](https://deno.land/manual@v1.29.2/getting_started/setup_your_environment#neovim-06-using-the-built-in-language-server), we should specify separate root directories for our ts and deno projects, so that the tsserver and denols do not both try to attach to the same buffer. With this change introduced, this workaround no longer works, and tsserver still attaches to the buffer in single file mode. When I set `single_file_support` to be `false` in my `lspconfig`, however, the behavior is as expected. 

I agree, though, that we should be able to edit a single typescript file, so I am not sure what the best work around here is, though.

anasrar:
`single_file_support=true` will ignore `root_dir`, this is considered bad for most people that work with `denols` and `tsserver` that use `root_dir` to filter which LSP should be active.

- #2015 
- #2069 
- #2070

glepnir:
```lua
lspconfig.util.on_setup = lspconfig.util.add_hook_before(lspconfig.util.on_setup, function(config)
  local cwd = vim.loop.cwd()
  if config.name == 'tsserver' and vim.fn.filereadable(cwd .. '/deno.json') then
    config.single_file_support = false
  end
end)
```

PR #2378, opened 2023-01-04
Avoid starting tailwindcss unless config is found

frangio:
The tailwindcss server is registered for a lot of file types. Since the config file will always be there if TailwindCSS is in use (as far as I can tell from experience and [the docs](https://tailwindcss.com/docs/configuration)), I would suggest only using these files for the root detection. In this way the server isn't spawned when it isn't necessary.

### DIFF
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -105,10 +105,6 @@ return {
     end,
     root_dir = function(fname)
       return util.root_pattern('tailwind.config.js', 'tailwind.config.ts')(fname)
-        or util.root_pattern('postcss.config.js', 'postcss.config.ts')(fname)
-        or util.find_package_json_ancestor(fname)
-        or util.find_node_modules_ancestor(fname)
-        or util.find_git_ancestor(fname)
     end,
   },
   docs = {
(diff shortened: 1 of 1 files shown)


lithammer:
It says the file is optional though?

> By default, Tailwind will look for an <ins>optional</ins> `tailwind.config.js` file at the root of your project where you can define any customizations.

frangio:
Sigh. Apologies, I should've read better.

I still plan to apply this on my personal configuration, but it might be too opinionated for the shared configs so I will close this PR.

polyzen:
```
In order for the extension to activate you must have [tailwindcss installed](https://tailwindcss.com/docs/installation) and a [Tailwind config file](https://tailwindcss.com/docs/installation#create-your-configuration-file) named tailwind.config.js or tailwind.config.cjs in your workspace.
```
-- https://github.com/tailwindlabs/tailwindcss-intellisense#installation

Though it seems like more variations of the filename are allowed too.

frangio:
It seems fairly complex actually. There is a [`CONFIG_GLOB`](https://github.com/tailwindlabs/tailwindcss-intellisense/blob/55d2b9e8da44faa55971ef006c0750fed5f0da2d/packages/tailwindcss-language-server/src/lib/constants.ts#L1) variable, but the config can also be found in other ways that involve [reading file contents](https://github.com/tailwindlabs/tailwindcss-intellisense/blob/07ad87e8da8bbefb824609857ec7f1fbf3fbf628/packages/tailwindcss-language-server/src/server.ts#L1499).

polyzen:
If I understand correctly the server will exit if an actual tailwind config file is not found.

PR #2379, opened 2023-01-05
fix(terraformls): update root dir

mosheavni:
According to terraform-ls' maintainer:
https://github.com/hashicorp/terraform-ls/issues/1139#issuecomment-1371987186

.terraform dir or .git are not a requirement for terraform-ls


lua/lspconfig/server_configurations/terraformls.lua
```diff
@@ -1,10 +1,19 @@
 local util = require 'lspconfig.util'
 
+local language_id_mapping = {
+  tf = 'terraform',
+  tfvars = 'terraform-vars',
```
radeksimko, line 5: What is not clear to me is how/where does `*.tf` get mapped to `tf` and `*.tfvars` to `tfvars`, i.e. pattern to filetype. Is this logic built into NeoVim somewhere? If not, then we'd need to do mapping based on patterns I think?

mosheavni: filetype is detected by neovim internals
image

image

mosheavni: ok i realized we don't need it as it's the default behaviour:

> {get_language_id} function(bufnr, filetype) -> language ID as string. Defaults to the filetype.
https://neovim.io/doc/user/lsp.html

lua/lspconfig/server_configurations/terraformls.lua
```diff
@@ -4,7 +4,8 @@ return {
   default_config = {
     cmd = { 'terraform-ls', 'serve' },
     filetypes = { 'terraform' },
-    root_dir = util.root_pattern('.terraform', '.git'),
+    root_dir = util.root_pattern('*.tf', '*.tfvars'),
```
gegoune, line 7: It isn't necessary to have any of those files in root of the project. This would break terraforms for all my projects.

radeksimko: @gegoune 👋🏻 I work on the Terraform LS and I am curious how the Terraform language server would be useful in a directory which does _not_ have any `*.tf` or `*.tfvars` files? Could you expand on this, please?

lua/lspconfig/server_configurations/terraformls.lua
radeksimko: Thank you for the context!

The only potential concern then is that "TF Mud files" will be sent to the Terraform LS too, just because these use the same `tf` filetype AFAICT: https://github.com/neovim/neovim/blob/5b9980f01eb021b0d84f540a6618f94ad2727153/runtime/filetype.vim#L1944-L1945
but I guess that's not a _new_ problem, esp. if the default mapping was in place before? 🤷🏻

lua/lspconfig/server_configurations/terraformls.lua
lithammer: This doesn't answer your question really. But another problem with the above pattern is that you want the root to resolve to the top level .tf file(s). Otherwise in tree of modules you'll start a new instance for every level.

So with this structure you'll start 4 instances of the language server if you have all files open.

```
/main.tf
/foo/main.tf
/bar/main.tf
/baz/main.tf
```

gegoune: Thanks for all your fork on tfls! My assumption was that terraformls is started for root marker's workspace, ie found by .git, and works for all nested modules, perhaps that's wrong assumption.

radeksimko: @lithammer Multiple instances are to be expected, unless NeoVim supports LSP "workspace folders" which is the ultimate way to solve this problem AFAICT.

Such a support implies that as part of `initialize` request that **single** server process receives _all_ directories as part of `workspaceFolders`. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize within the body of the request.

I am not sure if NeoVim supports this today, but if not, it probably should.

mosheavni: so would all 4 be ok?

lithammer: @radeksimko Neovim has support for workspace folders. As far as I understand it's not necessary to include all directories in `workspaceFolders` as part of the `initialize` request since you also have the [`DidChangeWorkspaceFolders`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWorkspaceFolders) notification. Which you can you use to add workspaces as you open new files.

radeksimko: Correct, given a structure such as
```
/path/main.tf
/path/foo/main.tf
/path/bar/main.tf
/path/baz/main.tf
```
the LS is provided with just `/path` as root - where the assumption is that `/path` itself may or may not contain matching files, but somewhere under `/path` there is at least one matching file.

The Terraform LS follows what I believe is implied expectation of language servers, where we walk through the FS hierarchy and index all `*.tf` and `*.tfvars` files under the given root and/or `workspaceFolders` and track each as a module. This in turn enables cross-module IntelliSense within the whole workspace. Terraform modules in this case are treated similarly to how e.g. packages are treated in Go and how gopls would provide completion from other packages.

TL;DR whether you're running the LS as single or multiple processes, you shouldn't need to worry about this.

There remains a question of relevancy of IntelliSense and understanding the hierarchy - e.g. which module is only ever used as submodule and never `init`ed directly, but this question does not have a perfect answer and I don't think it will ever have. `.terraform` directory is a good positive indicator, but bad negative one. i.e. we cannot assume a module is _not_ root just because it wasn't initialized. `.git` has a same problem, but additionally any indications are tainted by the fact that people use monorepos. There is https://github.com/hashicorp/terraform-ls/issues/179 where we track some ideas.

Either way this looks like a problem which would IMO be easier to solve on the server side, not client side.

mosheavni: @radeksimko  meaning a PR on terraform-ls?

radeksimko: I'm not sure what you're asking, I'm just trying to explain that the problems that were described above are problems which `terraform-ls` _can_ deal with and the change, as proposed in this PR is 👍🏻

PR #2381, opened 2023-01-06
ci: add luarocks upload release workflow

MrcJkb:
The recently added rockspec (see #2307) has to be prepared and uploaded to luarocks for each release.

This adds an automated workflow for it, which has been tested with plenary.nvim.

To be able to upload to luarocks, the owner of the luarocks account will have to add an API key named `LUAROCKS_API_KEY` to this repo's GitHub Actions secrets.

github-add-luarocks-api-key

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: "release"
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  luarocks-upload:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: leafo/gh-actions-lua@v9
+        with:
+          luaVersion: "luajit-2.1.0-beta3"
+      - uses: leafo/gh-actions-luarocks@v4
+      - name: Install dkjson
+        run: luarocks install dkjson
+      - name: Luarocks Upload
+        env: 
+          LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
+        run: make luarocks_upload
+      - name: Install release
+        run: make test_luarocks_install 
--- a/Makefile
+++ b/Makefile
@@ -10,3 +10,9 @@ lint:
 	stylua --check .
 
 .PHONY: test lint
+
+luarocks_upload:
+	bash ./scripts/luarocks-upload.sh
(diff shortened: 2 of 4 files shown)


glepnir:
@teto

teto:
thanks for that, I will merge once we agree with the team what LUAROCKS_API_KEY to provide. I have no experience with creating github actions but considering the work you've already done to propose this workflow in various plugins, maybe creating a Github action could be a good idea (and once it works even move it to the https://github.com/nvim-lua/ org ?)

MrcJkb:
> maybe creating a Github action could be a good idea (and once it works even move it to the https://github.com/nvim-lua/ org ?)
I have been considering that, too :)

justinmk:
> I've added a LUAROCKS_API_KEY let's see

would be a good time to add a "Release" section to the readme.md and mention this

teto:
I pushed a tag manually to trigger the pipeline. It failed when trying to install nvim-lspconfig 
https://github.com/neovim/nvim-lspconfig/actions/runs/3868300639/jobs/6593652025
maybe because the package was not present in luarocks' "root" manifest . Also there was an existing lspconfig package I had uploaded under my account. I removed the latter and retriggered the job which now fails because the version already exists:
https://github.com/neovim/nvim-lspconfig/actions/runs/3868300639/jobs/6593683541
The annoying bit is that it prevents the install job from running so I think this should become a warning instead of an error in our pipeline.
I've tested the install locally and it worked splendidly

```
➜ luarocks install --local nvim-lspconfig
Installing https://luarocks.org/nvim-lspconfig-0.1.5-1.src.rock

nvim-lspconfig 0.1.5-1 depends on lua 5.1 (5.1-1 provided by VM)
nvim-lspconfig 0.1.5-1 is now installed in /home/teto/.luarocks (license: Apache/2.0)

Checking stability of dependencies in the absence of
nvim-lspconfig 0.1.4-2...

Removing nvim-lspconfig 0.1.4-2...
Removal successful.
```

MrcJkb:
> The annoying bit is that it prevents the install job from running so I think this should become a warning instead of an error in our pipeline.

I guess releases failing because a version already exists shouldn't happen too often.
But I will look into that when creating the workflow next week.

PR #2383, opened 2023-01-07
fix: restart server when server need restart

### LINKED ISSUE #2366: `sumneko_lua` doesn't work across multiple projects
### Description

`sumneko_lua` simply doesn't work across multiple projects. The LSP is only working for the first project you opened, but it will stop for working for any subsequent projects.

### Neovim version

NVIM v0.9.0-dev-484+gb5edea655
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az368-865

### Nvim-lspconfig version

432935081337dbbbc80f4d0531868509624505df

Windows 10

### Affected language servers

sumneko_lua

### Steps to reproduce

1. `nvim -nu minimal_init.lua`
2. Open a Lua project.
3. Open a different Lua project from a different root directory, without closing the current nvim session. 

### Actual behavior

The first opened project works fine. But the subsequently opened projects are completely non-functional.
Trying to open the hover menu with `K` will give you `Workspace loading: 0/0` permanently.
Even the `require` statement is considered to be `undefined`.

I'm guessing this is because of the new workspace folders feature added by the recent commits. But I did remember testing the workspace functionality when it was first introduced to `lspconfig` and it worked _perfectly_ fine.
But after some time, it just stopped working. I suppose there was a recent commit that broke the feature, but I can't quite narrow it down.

### Expected behavior

The subsequently opened projects should work just as fine as the first one.

### Minimal config

```Lua

  return result
end

local lspconfig_path = join_paths(package_root, 'test', 'start', 'nvim-lspconfig')

  vim.fn.system { 'git', 'clone', 'https://github.com/neovim/nvim-lspconfig', lspconfig_path }
end

vim.lsp.set_log_level 'trace'
  end

  -- Mappings.
  local opts = { buffer = bufnr, noremap = true, silent = true }
  vim.keymap.set('n', 'gD', vim.lsp.buf.declaration, opts)
  vim.keymap.set('n', 'gd', vim.lsp.buf.definition, opts)
  vim.keymap.set('n', 'K', vim.lsp.buf.hover, opts)
  vim.keymap.set('n', 'gi', vim.lsp.buf.implementation, opts)
  vim.keymap.set('n', '<C-k>', vim.lsp.buf.signature_help, opts)
  vim.keymap.set('n', '<space>wa', vim.lsp.buf.add_workspace_folder, opts)
  vim.keymap.set('n', '<space>wr', vim.lsp.buf.remove_workspace_folder, opts)
  vim.keymap.set('n', '<space>wl', function()
  end, opts)
  vim.keymap.set('n', '<space>D', vim.lsp.buf.type_definition, opts)
  vim.keymap.set('n', '<space>rn', vim.lsp.buf.rename, opts)
  vim.keymap.set('n', 'gr', vim.lsp.buf.references, opts)
  vim.keymap.set('n', '<space>e', vim.diagnostic.open_float, opts)
  vim.keymap.set('n', '[d', vim.diagnostic.goto_prev, opts)
  vim.keymap.set('n', ']d', vim.diagnostic.goto_next, opts)
  vim.keymap.set('n', '<space>q', vim.diagnostic.setloclist, opts)
end

local name = 'sumneko_lua'
local cmd = {'/path/to/sumneko_lua', '--stdio' }
if not name then
  print 'You have not defined a server name, please edit minimal_init.lua'
end
  print [[You have not defined a server default cmd for a server
    that requires it please edit minimal_init.lua]]
end

nvim_lsp[name].setup {
  cmd = cmd,
  on_attach = on_attach,
}

print [[You can find your log at $HOME/.cache/nvim/lsp.log. Please paste in a github issue under a details tag as described in the issue template.]]
```

### LSP log

https://gist.github.com/musjj/1bf79a2c43957ff4c11cd92e74815145

glepnir:
# Problem
if server like sumneko lua does not support dynamically add or remove workspace Folders restart server when new add or remove workspaceFolders.

# Solution 
 add a option in server config `new_folder_restart` , when this option is `true` restart server and clean cache

Fix #2366

### DIFF
--- a/lua/lspconfig/server_configurations/sumneko_lua.lua
+++ b/lua/lspconfig/server_configurations/sumneko_lua.lua
@@ -33,6 +33,7 @@ return {
       return util.find_git_ancestor(fname)
     end,
     single_file_support = true,
+    new_folder_restart = true,
     log_level = vim.lsp.protocol.MessageType.Warning,
     settings = { Lua = { telemetry = { enable = false } } },
   },
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -279,9 +279,12 @@ function M.server_per_root_dir_manager(make_config)
     local new_config = make_config(root_dir)
     local client = get_client_from_cache(new_config)
 
-    --TODO(glepnir): do we need check language server support the workspaceFOlders?
-    --some server support it but the it not show in server_capabilities
     local register_workspace_folders = function(client_instance)
+      if single_file then
+        register_to_clients(client_instance.id)
+        return
+      end
+
       local params = {
         event = {
           added = { { uri = vim.uri_from_fname(root_dir), name = root_dir } },
@@ -298,6 +301,27 @@ function M.server_per_root_dir_manager(make_config)
         client_instance.workspace_folders = {}
       end
       table.insert(client_instance.workspace_folders, params.event.added[1])
+      if client_instance.config.new_folder_restart then
+        client_instance.stop()
+        local attached_buffers = client_instance.attached_buffers
+        new_config.workspace_folders = client_instance.workspace_folders
+        local id = lsp.start_client(new_config)
+        for buf in pairs(attached_buffers) do
+          lsp.buf_attach_client(buf, id)
+        end
+        --remove id from cache
+        for _, client_ids in pairs(clients) do
+          for k, v in pairs(client_ids) do
+            if v == client_instance.id then
+              table.remove(client_ids, k)
+              table.insert(client_ids, id)
+            end
+          end
+        end
+        register_to_clients(id)
+        return
+      end
+      register_to_clients(client_instance.id)
     end
 
     local attach_after_client_initialized = function(client_instance)
@@ -308,10 +332,7 @@ function M.server_per_root_dir_manager(make_config)
         vim.schedule_wrap(function()
           if client_instance.initialized and not timer:is_closing() then
             lsp.buf_attach_client(bufnr, client_instance.id)
-            if not single_file then
-              register_workspace_folders(client_instance)
-            end
-            register_to_clients(client_instance.id)
+            register_workspace_folders(client_instance)
             timer:stop()
             timer:close()
           end
@@ -322,10 +343,7 @@ function M.server_per_root_dir_manager(make_config)
     if client then
       if client.initialized then
         lsp.buf_attach_client(bufnr, client.id)
-        if not single_file then
-          register_workspace_folders(client)
-        end
-        register_to_clients(client.id)
+        register_workspace_folders(client)
       else
         attach_after_client_initialized(client)
       end


lua/lspconfig/server_configurations/sumneko_lua.lua
```diff
@@ -33,6 +33,7 @@ return {
       return util.find_git_ancestor(fname)
     end,
     single_file_support = true,
+    new_folder_restart = true,
```
justinmk, line 36: why is it optional? also need docs update

glepnir: because if server support dynamically add or remove workspace Folders . no need restart server. just send notify `didChangeWorkspaceFolders`.

justinmk: if only sumneko needs this then it's a sumneko bug, not something we should add code to workaround.

where is this `return` structure documented? adding random fields to it in random lang servers, will lead to a fragile codebase.

glepnir: I agree with you. But it's hard to pin the problem down to a bug in sumnkeolua. The lsp document just stipulates the rules of the communication protocol. Does not specify any details of the implementation. So you can see that each server works differently in some regard. And the author has indicated in the wiki that dynamic addition or deletion of working folders is not supported. The pr just does a behavior similar to vscode.

musjj: I've opened an issue at the repository, but there's no response yet.
What about hardcoding the restart behavior for sumneko in a way that is easy to remove later on, in case there's a fix?
So far only sumneko behaves like this, so it would be weird to expose it as a config. I also can't imagine a user who would prefer the alternative option, since it behaves incorrectly.

glepnir: I don't recommend classifying this as a language server bug. This is somewhat rude to the language server authors. The lsp does not specify how language server authors are to implement it. Depending on the author's code level, the language server behaves differently. Sumneko has described the current workspacefolder behavior in the wiki. This is a must if we want to support this server at the moment. It would be great if sumneko could be dynamically added or removed in the future.

about  this option there are many language servers. I didn't check their docs fully and maybe there has other servers that needs to be restarted. currently as far as I know is sumneko , but be careful. not every language server pays attention to `workspacefolders`, for example clang doesn't actually depend on `workspacefolders`.

lua/lspconfig/util.lua
```diff
@@ -298,6 +301,27 @@ function M.server_per_root_dir_manager(make_config)
...
+        local id = lsp.start_client(new_config)
+        for buf in pairs(attached_buffers) do
+          lsp.buf_attach_client(buf, id)
+        end
+        --remove id from cache
+        for _, client_ids in pairs(clients) do
+          for k, v in pairs(client_ids) do
+            if v == client_instance.id then
+              table.remove(client_ids, k)
+              table.insert(client_ids, id)
```
justinmk, line 317: this is a lot of new code. should live in a separate private function, e.g. `do_restart` that is called from here.

lua/lspconfig/util.lua
```diff
@@ -298,6 +301,27 @@ function M.server_per_root_dir_manager(make_config)
         client_instance.workspace_folders = {}
       end
       table.insert(client_instance.workspace_folders, params.event.added[1])
+      if client_instance.config.new_folder_restart then
```
justinmk, line 304: this is a hidden property, used only by `lua/lspconfig/server_configurations/sumneko_lua.lua`. that is hard for other contributors to follow. 

where is the structure and valid fields of `client_instance.config` documented? if nvim-lspconfig is adding fields, then nvim-lspconfig should document this somewhere.

glepnir: this is a trouble thins about workspaceFolders . currently some very popular language server not support this. so if we want support workspaceFolders there has some work to do .. no idea currently maybe i can find a good way to do this. 
about lua server restart problem, I most suggest fixed in server not client ... so I will ask  sumneko for some help or create a pr for this in lua server repo.

musjj: The issue has been fixed in https://github.com/sumneko/lua-language-server/commit/bddeb50920da81a6cecd91100aed018769bcc5b7

There's one more thing I'm confused about: why do we need to have a custom handler for `workspace/diagnostic/refresh` if you want to get rid of undefined global errors?

Can this also be fixed on the sumneko side instead?

glepnir: notice we don't have `workspace/diagnostic` handlers currently .  it's different `textDocument/diagnostic` .  

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_diagnostic

musjj: True, I guess we should we have a handler for that? It's currently not causing any noticeable problems for other LSPs though.
I've also tried the handler snippet you posted, but what happens is that you get a flash of diagnostic errors before they're "fixed".
It feels ugly and incorrect, so I wonder whether there's a more correct way of handling it on lspconfig's side, or should a fix be made on sumneko's side?

glepnir: since the lua-language-server 3.6.7 released we don't need pr right ?

musjj: Yes, we still need to do something about the handler though.

justinmk:
@glepnir @musjj  thank you for driving towards a more maintainable solution here!

PR #2384, opened 2023-01-08
feat: add openscad_lsp

williamboman:
This server is a fork of sorts of the `openscad_ls` server that is
already supported by nvim-lspconfig. I have no insight into either
projects, but the `openscad-lsp` Cargo package seems to be more
actively maintained, see:

- `openscad_ls`: https://github.com/dzhu/openscad-language-server
- `openscad_lsp (this pr)`: https://github.com/Leathong/openscad-LSP

### DIFF
--- a/lua/lspconfig/server_configurations/openscad_lsp.lua
+++ b/lua/lspconfig/server_configurations/openscad_lsp.lua
@@ -1,0 +1,32 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'openscad-lsp' },
+    filetypes = { 'openscad' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+  },
+  docs = {
+    description = [=[
+https://github.com/Leathong/openscad-LSP
+
+A Language Server Protocol server for OpenSCAD
+
+You can build and install `openscad-lsp` binary with `cargo`:
+```sh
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/openscad_lsp.lua
```diff
@@ -0,0 +1,32 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'openscad-lsp' },
```
williamboman, line 5: One sec, need to verify this cmd (the binary name is correct but it seems to default to TCP transport).

williamboman: Yep, missing `--stdio` flag.

```suggestion
    cmd = { 'openscad-lsp', '--stdio' },
```

calexanderberg:
I believe as of [April 2022](https://github.com/neovim/neovim/commit/6e6f5a783333d1bf9d6c719c896e72ac82e1ae54) openscad files are recognized, but I could be wrong.

williamboman:
> I believe as of [April 2022](https://github.com/neovim/neovim/commit/6e6f5a783333d1bf9d6c719c896e72ac82e1ae54) openscad files are recognized and have a filetype pattern, however maybe I just misunderstood the issue here.

Ah, cool! Removed that section from the docs.

PR #2387, opened 2023-01-09
feat(angularls): add Nx monorepo support

triglav:
Adds support for detecting `root_dir` in Nx monorepo projects.


glepnir:
sorry for late review. does there have some relate document ?

PR #2401, opened 2023-01-17
fix(rust_analyzer): remove stub settings from default

p-hueber:
Non-empty default settings trigger a workspace/didChangeConfiguration, potentially overwriting the initial settings.
Move this example content into the description instead.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -55,9 +55,6 @@ return {
         or util.root_pattern 'rust-project.json'(fname)
         or util.find_git_ancestor(fname)
     end,
-    settings = {
-      ['rust-analyzer'] = {},
-    },
   },
   commands = {
     CargoReload = {
@@ -73,7 +70,19 @@ https://github.com/rust-analyzer/rust-analyzer
 
 rust-analyzer (aka rls 2.0), a language server for Rust
 
-See [docs](https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user#settings) for extra settings.
+
+See [docs](https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user#settings) for extra settings. The settings can be used like this:
+```lua
+require'lspconfig'.rust_analyzer.setup{
+  settings = {
+    ['rust-analyzer'] = {
+      diagnostics = {
+        enable = false;
+      }
+    }
+  }
+}
+```
     ]],
     default_config = {
       root_dir = [[root_pattern("Cargo.toml", "rust-project.json")]],


lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -73,7 +71,19 @@ https://github.com/rust-analyzer/rust-analyzer
...
 
-See [docs](https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user#settings) for extra settings.
+
+See [docs](https://github.com/rust-analyzer/rust-analyzer/tree/master/docs/user#settings) for extra settings. The settings can be used like this:
+```lua
+require'lspconfig'.rust_analyzer.setup{
+  settings = {
+    ['rust-analyzer'] = {
+      diagnostics = {
+        enable = true;
```
polyzen, line 81: Perhaps this should be `false`, otherwise it might confuse people into thinking `true` is not the default.

glepnir:
what's the problem is ?

> Non-empty default settings trigger a workspace/didChangeConfiguration

when change the workspace it will trigger this notify. currently there has a bug in lspconfig it should check server support worksapceFolders or not . btw the `rust-analyzer` not support workspaceFolder

p-hueber:
> what's the problem is ?
> 

rust-analyzer allows passing all its configuration options as `settings` as well as `init_options`. Take this example:
```lua
{
  init_options = {
    rustfmt = {
      rangeFormatting = {
        enable = true,
      }
    }
  }
}
```
Now `rangeFormatting` must be passed as an `init_option` because it affects the capabilities of the server. And it must remain set when doing the actual formatting.

But due to the current default being a non-emtpy settings table, lspconfig will revert `rangeFormatting` to `false` after launch. So with the above example, RA will signal the range formatting capability, but reject any attempt to use it.

Even an explicit `settings = {}` in my config won't keep lspconfig from sending its own default. So currently, only this works:

```lua
{
  init_options = {
    rustfmt = {
      rangeFormatting = {
        enable = true,
      }
    }
  },
  settings = {
    ['rust-analyzer'] = {
      rustfmt = {
        rangeFormatting = {
          enable = true,
        }
      }
    }
  }
}
```

lspconfig always clearing the settings is quite surprising and not a good default for rust-analyzer.

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -55,9 +55,7 @@ return {
         or util.root_pattern 'rust-project.json'(fname)
         or util.find_git_ancestor(fname)
     end,
-    settings = {
-      ['rust-analyzer'] = {},
-    },
+    settings = {},
```
glepnir, line 58: this table can be removed by default.

p-hueber: Thanks for the suggestion, I removed the whole `settings` table and rebased onto the current master.

PR #2402, opened 2023-01-17
Add default log file location to README

WillNilges:
Useful for if someone needs to grab the log file to show to someone (eg: on an issue)


glepnir:
this is  default path of *nix system not in windows.

teto:
`:LspLog` is all you need really, it will show you the path towards the buffer.

PR #2165, opened 2022-09-30
feat: add erg-language-server support

pysan3:
Repost from #2164 

[Erg Programming Language](https://erg-lang.org/)

A generic statically typed language that can deeply improve the Python ecosystem.

Please add support for the Erg language with its language server [erg-language-server](https://github.com/erg-lang/erg-language-server).

Thanks in advance,
@pysan3

PR #2404, opened 2023-01-18
fix(erg-language-server): els has been integrated into the main compiler

pysan3:
Hi, author of #2165 here.

As noted in [ELS README](https://github.com/erg-lang/erg-language-server#els-erg-language-server), the language server has been integrated into the main compiler.

Therefore, I have changed the server configuration of els to call `erg --language-server` instead of the old method `els`.

It would be nice if you could merge this change.

Thanks in advance,
@pysan3


lua/lspconfig/server_configurations/erg_language_server.lua
```diff
@@ -2,21 +2,22 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    cmd = { 'els' },
+    cmd = { 'erg', '--language-server' },
     filetypes = { 'erg' },
     root_dir = function(fname)
-      return util.root_pattern 'package.er'(fname) or util.find_git_ancestor(fname)
+      return util.root_pattern 'package.er' (fname) or util.find_git_ancestor(fname)
```
glepnir:
```suggestion
      return util.root_pattern 'package.er'(fname) or util.find_git_ancestor(fname)
```

pysan3: Thanks, updated in the latest commit!

PR #2407, opened 2023-01-18
Add editorconfig language server

Lilja:
Hi!

I have made a WIP language server for editorconfig.

Can this server configuration be added upsteam?

### DIFF
--- a/lua/lspconfig/server_configurations/editorconfig.lua
+++ b/lua/lspconfig/server_configurations/editorconfig.lua
@@ -1,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+local root_files = { '.editorconfig', '.git' }
+return {
+  default_config = {
+    cmd = { 'ecls', '--stdio' },
+    filetypes = { 'editorconfig' },
+    root_dir = function(fname)
+      return util.root_pattern(unpack(root_files))(fname)
+    end,
+    init_options = {
+      buildDirectory = 'build',
+    },
+  },
+  docs = {
+    description = [[
+https://github.com/Lilja/ecls
+Editorconfig language server
+]],
+    default_config = {
+      root_dir = [[root_pattern('.editorconfig')]],
+    },
+  },
+}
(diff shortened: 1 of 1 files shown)


glepnir:
it only works for `.editorconfig` file right?

Lilja:
@glepnir Correct

glepnir:
so why we need `.git` ? and does it not support single file mode ?

Lilja:
> so why we need .git

A copy paste artefact. I'll remove it. 

> does it not support single file mode ?

What do you mean with single file mode?

glepnir:
`single_file_support` option that mean this server can work with a single file without root dir.  in rpc protocol is the `rootUri` is current dir `workspaceFolders` nil . send to server

Lilja:
I tried to read what the `:h lspconfig-single-file-support` said about this and I couldn't fully understand it. It's a regular LS with workspace/root dir support. Editorconfig files can be nested, so I would argue that it would need to know its workspace directory.

glepnir:
add `single_file_support` . it will use  cwd as  root send to your server. does your server can accept this ? 
like open nvim in home dir the cwd is home. and open `~/project/.editorconfig` .  the root actual is project but we send cwd.

lua/lspconfig/server_configurations/editorconfig.lua
```diff
@@ -0,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+local root_files = { '.editorconfig' }
```
glepnir, line 3:
```suggestion
```

lua/lspconfig/server_configurations/editorconfig.lua
```diff
@@ -0,0 +1,24 @@
...
+return {
+  default_config = {
+    cmd = { 'ecls', '--stdio' },
+    filetypes = { 'editorconfig' },
```
glepnir, line 7:
```suggestion
    filetypes = { 'editorconfig' },
    root_files = { '.editorconfig' },
```

Lilja:
I tried your suggestions locally and it didn't work for my server. Nothing in `:LspLog` either: img

glepnir:
well. I think it would be better support that. because it only work for `.editorconfig` it just a config file

justinmk:
> I have made a WIP language server for editorconfig.

no thanks. configs should not be added here unless they are actually used by many people.

also why does editorconfig need a lsp server?

PR #2408, opened 2023-01-19
feat(terraformls): add terraform-vars filetype

lourenci:


glepnir:
don't edit server_configruation file

PR #2411, opened 2023-01-19
fix(drools-lsp): move improper client config location and improve config processing

errantepiphany:
fix(drools-lsp) move improper client config location and improve config processing

According to [CONTRIBUTING.md](https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md), both `init_options` and `settings` are sent from the client to the server (just at different times). The drools-lsp server itself does not need anything sent to it (at least, not at this point), so it makes no sense to have client configuration under settings. Thus, we (the Drools LSP team) want to correct that by moving the drools section up one configuration level, and also move the java section under drools too, to keep everything encapsulated. Since the drools_lsp configuration in this repo is still so new (and only on master and not tagged or released), plus the fact that the server itself is currently experimental, we think the only people this would disrupt is ourselves. So now is the time to correct this. Thank you!

Signed-off-by: David Ward <dward@redhat.com>

### DIFF
--- a/lua/lspconfig/server_configurations/drools_lsp.lua
+++ b/lua/lspconfig/server_configurations/drools_lsp.lua
@@ -1,15 +1,37 @@
 local util = require 'lspconfig.util'
 
-local get = function(config, section, key)
-  return vim.tbl_get(config.settings, section, key) and config.settings[section][key] or nil
-end
-
-local get_java_bin = function()
-  local java_bin = vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java'
-  if vim.fn.has 'win32' == 1 then
-    java_bin = java_bin .. '.exe'
+local function get_java_bin(config)
+  local java_bin = vim.tbl_get(config, 'drools', 'java', 'bin')
+  if not java_bin then
+    java_bin = vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java'
+    if vim.fn.has 'win32' == 1 then
+      java_bin = java_bin .. '.exe'
+    end
   end
   return java_bin
+end
+
+local function get_java_opts(config)
+  local java_opts = vim.tbl_get(config, 'drools', 'java', 'opts')
+  return java_opts and java_opts or {}
+end
+
+local function get_jar(config)
+  local jar = vim.tbl_get(config, 'drools', 'jar')
+  return jar and jar or 'drools-lsp-server-jar-with-dependencies.jar'
+end
+
+local function get_cmd(config)
+  local cmd = vim.tbl_get(config, 'cmd')
+  if not cmd then
+    cmd = { get_java_bin(config) }
+    for _, o in ipairs(get_java_opts(config)) do
+      table.insert(cmd, o)
+    end
+    ---@diagnostic disable-next-line:missing-parameter
+    vim.list_extend(cmd, { '-jar', get_jar(config) })
+  end
+  return cmd
 end
 
 return {
@@ -17,19 +39,8 @@ return {
     filetypes = { 'drools' },
     root_dir = util.find_git_ancestor(),
     single_file_support = true,
-    on_new_config = function(config)
-      if not config.cmd then
-        local java_bin = get(config, 'java', 'bin') or get_java_bin()
-        local java_opts = get(config, 'java', 'opts') or {}
-        local drools_jar = get(config, 'drools', 'jar') or 'drools-lsp-server-jar-with-dependencies.jar'
-
-        config.cmd = { java_bin }
-        for _, o in ipairs(java_opts) do
-          table.insert(config.cmd, o)
-        end
-        --- @diagnostic disable-next-line:missing-parameter
-        vim.list_extend(config.cmd, { '-cp', drools_jar, 'org.drools.lsp.server.Main' })
-      end
+    on_new_config = function(new_config)
+      new_config.cmd = get_cmd(new_config)
     end,
   },
   docs = {
@@ -44,33 +55,19 @@ Configuration information:
 ```lua
 -- Option 1) Specify the entire command:
 require('lspconfig').drools_lsp.setup {
-  cmd = {
-    '/path/to/java',
-    '-cp',
-    '/path/to/drools-lsp-server-jar-with-dependencies.jar',
-    'org.drools.lsp.server.Main',
-  },
+  cmd = { '/path/to/java', '-jar', '/path/to/drools-lsp-server-jar-with-dependencies.jar' },
 }
 
--- Option 2) Specify just the jar location (the JAVA_HOME environment variable will be respected):
+-- Option 2) Specify just the jar path (the JAVA_HOME environment variable will be respected if present):
 require('lspconfig').drools_lsp.setup {
-  settings = {
-    drools = {
-      jar = '/path/to/drools-lsp-server-jar-with-dependencies.jar',
-    },
-  },
+  drools = { jar = '/path/to/drools-lsp-server-jar-with-dependencies.jar' },
 }
 
--- Option 3) Specify the jar location plus the java bin path and/or java opts:
+-- Option 3) Specify the java bin and/or java opts in addition to the jar path:
 require('lspconfig').drools_lsp.setup {
-  settings = {
-    java = {
-      bin = '/path/to/java',
-      opts = { '-Xmx500m' },
-    },
-    drools = {
-      jar = '/path/to/drools-lsp-server-jar-with-dependencies.jar',
-    },
+  drools = {
+    java { bin = '/path/to/java', opts = { '-Xmx100m' } },
+    jar = '/path/to/drools-lsp-server-jar-with-dependencies.jar',
   },
 }
 ```


errantepiphany:
Hi, sorry, but I don't know why commitlint is complaining of [subject-empty] and [type-empty]. It has both of those things... What am I doing wrong, please?

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,60 @@
 local util = require 'lspconfig.util'
 
-local get = function(config, section, key)
-  return vim.tbl_get(config.settings, section, key) and config.settings[section][key] or nil
+local function get(config, segments)
```
glepnir, line 3:
```lua
local function get(config, segments)
 local cfg = { }
  if vim.tbl_get(config, unpack(segments)) then
    vim.list_extend(cfg, segments)
  end
  return cfg
end
```

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,60 @@
...
+  return nil
 end
 
-local get_java_bin = function()
-  local java_bin = vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java'
-  if vim.fn.has 'win32' == 1 then
-    java_bin = java_bin .. '.exe'
+local function get_java_bin(config)
+  local java_bin = get(config, { 'drools', 'java', 'bin' })
+  if java_bin == nil then
```
glepnir, line 16:
```suggestion
  if vim.tbl_isempty(java_bin) then
```

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,60 @@
...
+    java_bin = vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java'
+    if vim.fn.has 'win32' == 1 then
+      java_bin = java_bin .. '.exe'
+    end
   end
   return java_bin
 end
 
+local function get_java_opts(config)
```
glepnir, line 25: then this no need

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,60 @@
...
+    drools_jar = 'drools-lsp-server-jar-with-dependencies.jar'
+  end
+  return drools_jar
+end
+
+local function get_cmd(config)
+  local cmd = get(config, { 'cmd' })
+  if cmd == nil then
+    cmd = { get_java_bin(config) }
+    for _, o in ipairs(get_java_opts(config)) do
```
glepnir, line 42:
```suggestion
    for _, o in ipairs(get(config, { 'drools', 'java', 'opts' })) do
```

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,60 @@
...
+  local drools_jar = get(config, { 'drools', 'jar' })
+  if drools_jar == nil then
+    drools_jar = 'drools-lsp-server-jar-with-dependencies.jar'
+  end
+  return drools_jar
+end
+
+local function get_cmd(config)
+  local cmd = get(config, { 'cmd' })
+  if cmd == nil then
```
glepnir, line 40:
```suggestion
  if vim.tbl_isempty(cmd) then
```

glepnir: how do you think this change :)

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: @glepnir, I'm not tring to add those segments (which are node "path" keys) into a new table. I'm traversing down into the config table, and extracting the value (if it exists) of a node that matches the segment "path". So, `config` is a table of nested tables, and each value in `segments` is really just a key into each subsequent nested table. I don't want to add `segments` itself.

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: @glebnir, java_in is not a table, it's a string value that I'm looking for

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: @glebnir, juts like java_bin, `cmd` is also a string, not a table

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: @glepnir My opinion is that it's more readable and maintainable to have a separate function for each config value we're looking for, and simply aggregate them into the resulting `cmd` in `get_cmd`. Is this a show-stopper?

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: @glepnir Oops, my last comment on this line was wrong, so changing it. :) Sure, I can make this change.

lua/lspconfig/server_configurations/drools_lsp.lua
glepnir: the get function your write return is `table|nil`

errantepiphany: The get function iterates over all the unpacked arguments sent to it, traversing down each node of the config until it gets to the last node that is being requested of it. For java_bin, it's not going to be a table value. It's going to be a string value.

glepnir: like i said you returned it with two type maybe one type is enough

errantepiphany: I'm sorry I do not understand what you're trying to say. We might be hitting a language barrier here.

The get function first checks if there is a value defined in the config that matches the segments "path" (using tbl_get and unpack). If not, it returns nil. If it does, then it can safely traverse down to get the value, which could be a string or a table. So the return of the get function is either nil (user didn't configure that node), a string (like java_bin), or a table (like cmd).

I don't know how to explain it better than that. If you don't like the implementation of the get function, perhaps you could suggest a complete get function that would replace it? Because otherwise I'm having trouble understanding what you are suggesting that I should do.

Very respectfully, I'm also confused as to why you consider the function is somehow "wrong" the way it is written? It works well, so these suggestions feel more like personal style preferences vs. fixing any actual bugs. I do very much appreciate the time you have spent on this, though. Thank you!

glepnir: >  I'm also confused as to why you consider the function is somehow "wrong" the way it is written?

 ```lua
local function get(config, segments)
  if vim.tbl_get(config, unpack(segments)) then
   -- cfg type is table
    local cfg = config
    for _, seg in ipairs(segments) do
      cfg = cfg[seg]
    end
    return cfg
  end
 --- return nil in there
  return nil
end
```
this function signature help return type will be `table or nil`. I say can we just return one type `table` my implment is one type

Comment (errantepiphany) [reply lua/lspconfig/server_configurations/drools_lsp.lua:16]: The return type will be a table, a string, or nil. The value of `cfg` _starts out_ as a table, but the last seg in the segments loop could be the key to a table OR a string.

So, when looping over `{ 'drools', 'java', 'opts' }`, yes, the return type is a table. However, when looping over `{ 'drools', 'java', 'bin' }`, the return type is a string. So "returning just on type `table`" will not work.

File: lua/lspconfig/server_configurations/drools_lsp.lua (hunk 3 above)
Comment (errantepiphany) [reply lua/lspconfig/server_configurations/drools_lsp.lua:25]: That function exists because if `{ 'drools', 'java', 'opts' }` is not configured by the user, then I want to return an empty table instead of nil.

File: lua/lspconfig/server_configurations/drools_lsp.lua (hunk 2 above)
Comment (glepnir) [reply lua/lspconfig/server_configurations/drools_lsp.lua:16]: `{ 'drools', 'java', 'opts' }` in `vim.tbl_get` mean `config['drools']['java']['opts']` what does this code mean ?

```lua
   local cfg = config
   for _, seg in ipairs(segments) do
     cfg = cfg[seg]
   end
```
the cfg actually equals cofig['opts']  ? you set value to it in a for iterator the value will be the. last index of segments.

errantepiphany: If you take this invocation: `get(config, { 'drools', 'java', 'opts' })`, this is what happens:

`local cfg = config`
-- ^^ the value of `cfg` _starts out as_ the root of the config (`config`)

`for` loop, where `seg` currently equals `'drools'`
-- ^^ 1st iteration in loop,  `cfg` is a table

`for` loop, where `seg` currently equals `'java'`
-- ^^ 2nd iteration in loop,  `cfg` is a table

`for` loop, where `seg` currently equals `'opts'`
-- ^^ 3rd iteration in loop,  `cfg` _ends up as_ a table, _and we return that table_.

Now if you take this invocation: `get(config, { 'drools', 'java', 'bin' })`, this is what happens:

`local cfg = config`
-- ^^ the value of `cfg` _starts out as_ the root of the config (`config`)

`for` loop, where `seg` currently equals `'drools'`
-- ^^ 1st iteration in loop,  `cfg` is a table

`for` loop, where `seg` currently equals `'java'`
-- ^^ 2nd iteration in loop,  `cfg` is a table

`for` loop, where `seg` currently equals `'bin'`
-- ^^ 3rd iteration in loop,  `cfg` _ends up as_ a string, _and we return that string_.

I have tested this, and it is the behavior I expect and want.

errantepiphany: The value of cfg keeps getting reassigned to a deeper level, and returns when it is at the level that is being requested, whether the value at the level is a table or a string.

errantepiphany: The `get` function is so that I don't have to write this ugly code for every configuration value I'm looking for:
`local java_bin = vim.tbl_get(config, 'drools', 'java', 'bin') and config['drool']['java']['bin'] or nil`

glepnir: .oh you change the cfg table pointer. my mistake.  then we can simply use `vim.tbl_get` it will return this value . in doc exmaple it return true . so I think it just check . but I look at the source code .it will return the value .

`:lua print(vim.tbl_get({ a = { b = { c= 10}}}, 'a', 'b','c'))`   it will print 10

errantepiphany: It looks like using `vim.tbl_get` is also safe if `b` is nil, and returns nil instead of an error, which is awesome. So... If I get rid of the `get` function, and use `vim.tbl_get`, will that be satisfactory?

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: I prefer to check for nil but not check for empty table. It will only be empty if the user set it and made it empty. At which point, it's their fault they did that.

lua/lspconfig/server_configurations/drools_lsp.lua
glepnir: yea. if key not exist it only return nil not trigger debug stack so just check nil is enought.

lua/lspconfig/server_configurations/drools_lsp.lua
```diff
@@ -1,35 +1,46 @@
 local util = require 'lspconfig.util'
 
-local get = function(config, section, key)
-  return vim.tbl_get(config.settings, section, key) and config.settings[section][key] or nil
+local function get_java_bin(config)
+  local java_bin = vim.tbl_get(config, 'drools', 'java', 'bin')
+  if java_bin == nil then
```
glepnir, line 5:
```suggestion
  if not java_bin then
```

lua/lspconfig/server_configurations/drools_lsp.lua
glepnir: the nil check can be simply like `if somethint then` `if not something then`.

errantepiphany: Okay. Will change.

lua/lspconfig/server_configurations/drools_lsp.lua
errantepiphany: Okay. Will change.

errantepiphany:
@glepnir All good now (please)?

glepnir:
LGTM

errantepiphany:
Thank you. Sorry to take up so much of your time.

glepnir:
it' okay :)

PR #2416, opened 2023-01-21
docs(bahsls): mention server specific config

f1rstlady:
Configuration specific to Bashls is accessible via the `bashIde` key. This is not obvious and hence is hereby documented.


glepnir:
add this in server doc would be fine.

PR #2419, opened 2023-01-24
feat: add vtsls

williamboman:


lua/lspconfig/server_configurations/vtsls.lua
```diff
@@ -0,0 +1,45 @@
...
+    end,
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
+https://github.com/yioneko/vtsls
+
+`vtsls` can be installed with npm:
+```sh
+npm install -g @vtsls/language-server
```
glepnir, line 32:
```suggestion
npm install -g @vtsls/language-server
```

glepnir (changes requested):

lua/lspconfig/server_configurations/vtsls.lua
```diff
@@ -0,0 +1,45 @@
...
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = {
+      'javascript',
+      'javascriptreact',
+      'javascript.jsx',
+      'typescript',
+      'typescriptreact',
+      'typescript.tsx',
```
williamboman, line 18: I just noticed these composite filetypes (`javascript.jsx` and `typescript.tsx`). I just copied them from `tsserver` as this server is essentially a fork of that server. 

I don't believe these filetypes are valid as-is, per my understanding they should translate to:
- `typescript.tsx`: `typescript` & `tsx`
- `javascript.jsx`: `javascript` & `jsx`

@glepnir Do you happen to know if the above filetypes are indeed valid within lspconfig, or should I look into it further?

PR #2421, opened 2023-01-25
docs(ltext): Add `settings` example

briandipalma:


justinmk:
i see that other docs call out `settings` like this, so i guess this makes sense. can you fix the lint error?

briandipalma:
Thanks. Looks like it's done/ready to merge.

briandipalma:
OK, trying again with lower case "Add"...

PR #2434, opened 2023-02-04
docs: arduino - mention upstream keywords bug

doronbehar:
BTW, perhaps the original contributor of this language server (@mjlbach) knows why does Arduino officially says that [`sketch.json` is not used by `arduino-cli`](https://arduino.github.io/arduino-cli/0.29/sketch-specification/#metadata) and how is this file related to the language server. Although I don't know the answer to this question, I'm pretty sure this documentation suggestion is still relevant.

### DIFF
--- a/lua/lspconfig/server_configurations/arduino_language_server.lua
+++ b/lua/lspconfig/server_configurations/arduino_language_server.lua
@@ -79,6 +79,8 @@ Your folder structure should look like this:
 ```
 
 For further instruction about configuration options, run `arduino-language-server --help`.
+
+Note that an [upstream bug](https://github.com/arduino/arduino-ide/issues/159) makes keywords in some cases become undefined by the language server.
 ]],
   },
 }


lua/lspconfig/server_configurations/arduino_language_server.lua
```diff
@@ -79,6 +79,8 @@ Your folder structure should look like this:
 ```
 
 For further instruction about configuration options, run `arduino-language-server --help`.
+
+Note that an [upstream bug](https://github.com/arduino/arduino-ide/issues/159) makes keywords in some cases become undefined by the language server.
```
glepnir, line 83: if you want people click or copy this link there is no need use markdown link syntax.

glepnir (changes requested):
like i said no need markdown syntax for link

doronbehar:
> no need markdown syntax for link

I don't understand, the same text includes many links such as I wrote and they get compiled as expected to https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#arduino_language_server .

glepnir:
if people use the treesitter markdown parse this link will hide.

PR #2435, opened 2023-02-05
Add Cadence Language Server

janezpodhostnik:
Add the Cadence Language Server.

If `cdc` file detection isn't working the following config fixes that:

```lua
vim.filetype.add({
  extension = {
    cdc = "cdc",
  }
})
```

Or switch to a neovim version which includes https://github.com/vim/vim/pull/11951.

### DIFF
--- a/lua/lspconfig/server_configurations/cadence.lua
+++ b/lua/lspconfig/server_configurations/cadence.lua
@@ -1,0 +1,40 @@
+local util = require 'lspconfig.util'
+
+local config_name = 'flow.json'
+local bin_name = 'flow'
+local cmd = { 'flow', 'cadence', 'language-server' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, 'cadence', 'language-server' }
+end
+
+return {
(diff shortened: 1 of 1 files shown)


glepnir (commented):
does this server not work on Windows?

janezpodhostnik:
It should, but I haven't tested it on windows. There are special installation instructions for windows for the `flow` command to be available here: https://developers.flow.com/tools/flow-cli/install#windows. I assume that if the flow command is available everything else is the same.

glepnir:
it just a install ? the start the server command should be same right ?

janezpodhostnik:
Yes, after the install the usage to start the language server (or for any other thing for that matter) should be the same. I can't find any usage differences in the [flow-cli](https://github.com/onflow/flow-cli) repo.

glepnir:
so we can assume they are same then you need add `if vim.fn.win32 then cmd.exe ....command`

glepnir:
open windows shell need cmd.exe.

janezpodhostnik:
Ah, I understand now. I added a check like I saw that others have.

PR #2438, opened 2023-02-06
feat: add fennel_language_server

Massolari:


glepnir:
does this server not work on windows?

Massolari:
@glepnir I don't know. I only tested it on MacOS

glepnir:
nope but it would be better add the windows env.

lua/lspconfig/server_configurations/fennel_language_server.lua
```diff
@@ -0,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+local cmd = { 'fennel-language-server' }
+
+if vim.fn.has 'win32' == 1 then
```
Massolari, line 5: @glepnir Done!
Is it correct?

glepnir: yes

PR #2439, opened 2023-02-06
feat: rename sumneko_lua -> lua_ls

figsoda:
@sumneko has moved lua-language-server to a new GitHub organization (@luals), https://github.com/sumneko/lua-language-server now redirects to https://github.com/LuaLS/lua-language-server

In terms of popularity, there are no competing language servers named `lua-language-server`, and the package is named `lua-language-server` on [repology](https://repology.org/project/lua-language-server/versions)

lspconfig currently doesn't have a rename policy for servers, `require("lspconfig").sumneko_lua.setup` will still work as before, just with a warning saying it has been renamed to `lua_language_server`. When this warning should be changed to a hard error is up for discussion.


lua/lspconfig/server_configurations/sumneko_lua.lua
```diff
@@ -1,91 +1,2 @@
...
-* [Lua.runtime.path](https://github.com/sumneko/lua-language-server/blob/076dd3e5c4e03f9cef0c5757dfa09a010c0ec6bf/locale/en-us/setting.lua#L5-L13)
-* [Lua.workspace.library](https://github.com/sumneko/lua-language-server/blob/076dd3e5c4e03f9cef0c5757dfa09a010c0ec6bf/locale/en-us/setting.lua#L77-L78)
-
-]],
-    default_config = {
-      root_dir = [[root_pattern(".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml", "stylua.toml", "selene.toml", "selene.yml", ".git")]],
-    },
-  },
-}
+vim.notify('`sumneko_lua` has been renamed to `lua_language_server`', vim.log.levels.WARN)
```
dundargoc, line 1: I think nvim has a specific deprecate module for this: `vim.deprecate()`

figsoda: switched to `vim.deprecate`, used `0.2.0` for version since this is the next breaking release

glepnir (approved):

scripts/docgen.lua
```diff
@@ -85,7 +85,11 @@ local function require_all_configs()
   -- Configs are lazy-loaded, tickle them to populate the `configs` singleton.
   for _, v in ipairs(vim.fn.glob('lua/lspconfig/server_configurations/*.lua', 1, 1)) do
     local module_name = v:gsub('.*/', ''):gsub('%.lua$', '')
-    configs[module_name] = require('lspconfig.server_configurations.' .. module_name)
+    -- sumneko_lua has been renamed to lua_language_server
+    -- this is a deprecated alias
+    if module_name ~= 'sumneko_lua' then
```
glepnir, line 90: does this needed ? doc can be use `lua-language-server` directly.

figsoda: without this, the script creates sections for sumneo_lua (in addition to the lua-language-server section) in server_configurations, which could be misleading to users

glepnir: the doc of `sumneko_lua` also no need ? a special condition looks wired. but it should do in next pr or after release

lua/lspconfig/server_configurations/sumneko_lua.lua
```diff
@@ -1,91 +1,2 @@
...
-* [Lua.runtime.path](https://github.com/sumneko/lua-language-server/blob/076dd3e5c4e03f9cef0c5757dfa09a010c0ec6bf/locale/en-us/setting.lua#L5-L13)
-* [Lua.workspace.library](https://github.com/sumneko/lua-language-server/blob/076dd3e5c4e03f9cef0c5757dfa09a010c0ec6bf/locale/en-us/setting.lua#L77-L78)
-
-]],
-    default_config = {
-      root_dir = [[root_pattern(".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml", "stylua.toml", "selene.toml", "selene.yml", ".git")]],
-    },
-  },
-}
+vim.deprecate('sumneko_lua', 'lua_language_server', '0.2.0', 'lspconfig')
```
glepnir, line 1: this mean it deprecate in `0.2.0` version . currently  version not match . `version` set to nil.

glepnir, line 1: this mean it deprecate in `0.2.0` version . currently  version not match . `version` should be set to nil

figsoda: this is the message when this line is run
```console
sumneko_lua is deprecated, use lua_language_server instead. See :h deprecated
This function will be removed in lspconfig version 0.2.0
```

changing `version` to `nil` gives me this error message
```console
E5108: Error executing lua vim/_editor.lua:0: attempt to concatenate a nil value
stack traceback:
	vim/_editor.lua: in function 'deprecate'
	[string ":lua"]:1: in main chunk
```

```console
$ nvim --version
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by nixbld

Features: +acl +iconv +tui
[...]
```

scripts/docgen.lua
figsoda: I agree this is a little ugly, but without this both `lua_language_server` and `sumneko_lua` are shown in `doc/server_configurations.{md,txt}` with no suggestion that the latter has been deprecated, and the diff would look like a new language server named `lua_language_server` is added.

Maybe I can do something to this instead? that will keep the aliases in one place:
https://github.com/neovim/nvim-lspconfig/blob/d3c82d2f9a6fd91ec1ffee645664d2cc57e706d9/lua/lspconfig.lua#L13

> the doc of sumneko_lua also no need?

I'm not entirely sure what that means

figsoda:
Moved the deprecation to `lua/lspconfig.lua` so it works better with `scripts/docgen.sh`, it will also make adding and removing more aliases easier in the future. The diff also looks much cleaner on github now

glepnir:
LGTM

lua/lspconfig.lua
```diff
@@ -9,9 +9,26 @@ function M.available_servers()
   return M.util.available_servers()
 end
 
+---@class Alias
+---@field to string The new name of the server
+---@field version string The version that the alias will be removed in
+---@type table<string, Alias>
+local server_aliases = {
```
glepnir, line 16: this table can add a deprecate tag ? it should be remove when after 0.2.0 release right?

figsoda: All aliases in this table are deprecated, the version field is the version it will get removed, which for sumneko_lua is 0.2.0.

glepnir: I know what's the table mean. I mean like this. this table only for sumneko lua.

```lua
---@deprecate
local server_aliases
```

justinmk (approved):

teto (approved):

glepnir:
could you rename it to `lua_ls` ? `lua_language_server` a bit long.

fkm3:
Hi. ` vim.deprecate` is a neovim 0.8 feature. Is it intentional that nvim-lspconfig requires at least 0.8 now? The main README still says 0.7

figsoda:
cc @glepnir since https://github.com/neovim/nvim-lspconfig/pull/2080 seems like the first usage of `vim.deprecate` in the code base

glepnir:
yes.but that is function can be removed I think

fkm3:
oh, i see, i can just do `s/sumneko_lua/lua_ls` to fix it. still, the error is a bit arcane looking. maybe worth a fallback that just prints an error since more people are likely to get confused. thanks!

PR #2441, opened 2023-02-07
docs: make sure username doesn't leak into generated document

figsoda:
The documents previously contained `runner`, the username of GitHub Actions runners, which would be overwritten by the user's name when run locally and generate unwanted diff.

### DIFF
--- a/scripts/docgen.lua
+++ b/scripts/docgen.lua
@@ -82,11 +82,21 @@ require'lspconfig'.{{template_name}}.setup{}
 ]]
 
 local function require_all_configs()
+  -- Make sure username doesn't leak into the generated document
+  local old_home = vim.env.HOME
+  local old_cache_home = vim.env.XDG_CACHE_HOME
+  vim.env.HOME = '/home/username'
+  vim.env.XDG_CACHE_HOME = '/home/username/.cache'
+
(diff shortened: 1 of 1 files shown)


justinmk:
Nice catch. I see examples in https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md

Though, maybe the real problem is that those example should have `$HOME` literally. Seems like some code is resolving those env vars, but shouldn't.

Meanwhile, this is a reasonable workaround.

scripts/docgen.lua
```diff
@@ -82,11 +82,21 @@ require'lspconfig'.{{template_name}}.setup{}
 ]]
 
 local function require_all_configs()
+  -- Make sure username doesn't leak into the generated document
+  local old_home = vim.env.HOME
+  local old_cache_home = vim.env.XDG_CACHE_HOME
+  vim.env.HOME = '/home/username'
+  vim.env.XDG_CACHE_HOME = '/home/username/.cache'
```
justinmk, line 89:
```suggestion
  vim.env.HOME = '/home/user'
  vim.env.XDG_CACHE_HOME = '/home/user/.cache'
```

justinmk (approved):

PR #2451, opened 2023-02-12
feat: rename servers with hyphens (`-`)

figsoda:
- fennel-ls -> fennel_ls
- starlark-rust -> starlark_rust

The generated configurations `setup`s were [wrong](https://github.com/neovim/nvim-lspconfig/blob/501cfd09b9138e51fd2fb356e2d253ad76784565/doc/server_configurations.md#fennel-ls):
```lua
require'lspconfig'.fennel-ls.setup{}
```


glepnir:
squash one commit ?

figsoda:
I split them into two commits to make it more clear what exactly is being renamed, but you can squash merge if you would prefer that way

glepnir:
you can make more details in commit message. yep I always use squash merge.

PR #2455, opened 2023-02-13
feat: add docker-compose-language-service support

marcdeop:
Potentially [related](https://github.com/williamboman/mason.nvim/pull/986)

### DIFF
--- a/lua/lspconfig/server_configurations/docker_compose_language_service.lua
+++ b/lua/lspconfig/server_configurations/docker_compose_language_service.lua
@@ -1,0 +1,32 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'docker-compose-language-service'
+local cmd = { bin_name, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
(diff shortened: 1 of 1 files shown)


williamboman (commented):
It's worth noting that the npm package itself doesn't provide an executable (it would probably be a good idea to request they do so, if someone feels like creating an issue with them). The executable used in this PR is currently only provided by mason.nvim - maybe it'd be good to omit the `cmd` until an official one is provided upstream?

marcdeop:
Request for the `cmd` [here](https://github.com/microsoft/compose-language-service/issues/114)

gegoune:
Thanks for that! Seems like that executable is called `docker-compose-langserver` on MacOS.

marcdeop:
> Thanks for that! Seems like that executable is called `docker-compose-langserver` on MacOS.

Yes, we coordinated yesterday with upstream.

I will submit a PR shortly

marcdeop:
[PR](https://github.com/neovim/nvim-lspconfig/pull/2458)

Issue #2449, opened 2023-02-10
getting root dir  not correct


LollipopFt:
### Description

file structure:
```
aoc <-- path that is :cd into
|-Cargo.toml <-- rust-analyzer should use this Cargo.toml: true project root
|-src
  |-lib.rs
|-two022 <-- rust-analyzer uses this Cargo.toml instead
  |-Cargo.toml
  |-src
    |-main.rs <-- :e two022\src\main.rs
```

rust-analyzer if run from true project root:
```diff
+ C:\Users\..\aoc>rust-analyzer diagnostics .
processing crate: 2022, module: C:\Users\..\aoc\two022\src\main.rs
-
Diagnostic { code: DiagnosticCode("inactive-code"), message: "code is inactive due to #[cfg] directives: feature = \"day2\" is disabled", range: 131..188, severity: WeakWarning, unused: true, experimental: false, fixes: None }
Diagnostic { code: DiagnosticCode("unresolved-macro-call"), message: "unresolved macro `print!`", range: 112..117, severity: Error, unused: false, experimental: true, fixes: None }
processing crate: aoc, module: C:\Users\..\aoc\src\lib.rs
Diagnostic { code: DiagnosticCode("unresolved-import"), message: "unresolved import", range: 15..17, severity: Error, unused: false, experimental: true, fixes: None }
Diagnostic { code: DiagnosticCode("unresolved-import"), message: "unresolved import", range: 28..32, severity: Error, unused: false, experimental: true, fixes: None }
Diagnostic { code: DiagnosticCode("unresolved-import"), message: "unresolved import", range: 34..38, severity: Error, unused: false, experimental: true, fixes: None }

diagnostic scan complete

[ERROR rust_analyzer] Unexpected error: diagnostic error detected
diagnostic error detected
```
rust-analyzer now:
```diff
+ C:\Users\..\aoc>rust-analyzer diagnostics two022
processing crate: two022, module: C:\Users\..\aoc\two022\src\main.rs
- Diagnostic { code: DiagnosticCode("inactive-code"), message: "code is inactive due to #[cfg] directives: feature = \"day1\" is disabled", range: 72..129, severity: WeakWarning, unused: true, experimental: false, fixes: None }
Diagnostic { code: DiagnosticCode("inactive-code"), message: "code is inactive due to #[cfg] directives: feature = \"day2\" is disabled", range: 131..188, severity: WeakWarning, unused: true, experimental: false, fixes: None }
[...]

Issue #2400, opened 2023-01-17
LSP: ESLint finds .eslintrc.js in current directory but does not traverse higher to root directory


mshirlaw-alteryx:
### Description

I have a project with a structure something like the following:

```
project
|__ src
       |__ models
           |__ Model.ts
           |__ .eslintrc.js
        |__ components
           |__ Component.tsx
.eslintrc.js
.git
```
For all of the exports that I have defined in `Model.ts` I receive a warning `Exported declaration XYZ is not used within other modules` even though these are imported in `src/components/Component.tsx`. 

When I do `:LspInfo` in this `Model.ts`  file I can see that the root directory is being set as `/Users/<username>/Documents/project/src/models` - i.e. it is finding the first `eslintrc.js`

```
 Client: eslint (id: 5, pid: 61525, bufnr: [59])
 	filetypes:       javascript, javascriptreact, javascript.jsx, typescript, typescriptreact, typescript.tsx, vue, svelte, astro
 	autostart:       true
 	root directory:  /Users/<username>/Documents/project/src/models
 	cmd:             vscode-eslint-language-server --stdio
 
 Client: tsserver (id: 3, pid: 61523, bufnr: [59])
 	filetypes:       javascript, javascriptreact, javascript.jsx, typescript, typescriptreact, typescript.tsx
 	autostart:       true
 	root directory:  /Users/<username>/Documents/project
 	cmd:             typescript-language-server --stdio
 
 Configured servers list: html, jsonls, tsserver, bashls, cssls, eslint
```

In my lsp config I am setting `root_dir` for all language servers using `util.root_pattern('.git')` which seems to work for `tsserver` (see above) but not for `eslint`. 

```
for _, lsp in pairs(servers) do
  require('lspconfig')[lsp].setup {
    root_dir = util.root_pattern('.git'),
    on_attach = on_attach,
    capabilities = capabilities,
    flags = lsp_flags, 
  }
end
```

Is there something else I need to configure to allow eslint to traverse up to the top level directory? This problem does NOT exist in vscode. The same configuration works correctly.

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

### Operating system and version

macOS 12.6

### Affected language servers

eslint

### Steps to reproduce

1. Create project with nvim-lspconfig
2. Install eslint 
3. Use directory structure above

### Actual behavior

[...]

Issue #1908, opened 2022-05-16
find_node_modules_ancestor does not always return parent


mikehaertl:
### Description

The documentation says:

> - `util.find_node_modules_ancestor`: a function that locates the first parent
>   directory containing a `node_modules` directory.

This is not quite correct: It does *not* return the **parent** directory, if there's already a `node_modules` dir in the passed directory.

So either the documentation is wrong or the behaviour should be fixed.

### Neovim version

NVIM v0.7.0
Build type: Debug
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wsuggest-attribute=pure -Wsuggest-attribute=const -Wsuggest-attribute=malloc -Wsuggest-attribute=cold -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=1 -I/home/repos/github/neovim/build/config -I/home/repos/github/neovim/src -I/home/repos/github/neovim/.deps/usr/include -I/usr/include -I/home/repos/github/neovim/build/src/nvim/auto -I/home/repos/github/neovim/build/include
Übersetzt von mike@Nilar

Features: +acl +iconv +tui
See ":help feature-compile"

          System-vimrc-Datei: "$VIM/sysinit.vim"
     Voreinstellung für $VIM: "/usr/local/share/nvim"

### Nvim-lspconfig version

0f06f7b

### Operating system and version

Linux Mint 20.3 Una

### Affected language servers

volar

### Steps to reproduce

```
mkdir -p /home/mike/test/a/node_modules
mkdir /home/mike/test/node_modules
```

Then in neovim:
```
:lua print( require'lspconfig'.util.find_node_modules_ancestor('/home/mike/test/a') )
```

### Actual behavior

It echoes `/home/mike/test/a`

### Expected behavior

[...]

PR #2459, opened 2023-02-14
refactor: rewrite root dir by using vim.fs module

glepnir:
a break change . the `vim.fs` module only works for nvim 0.8+ . so idk what time is can be merge.

relate issue: #2449  #2400 #1908

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -421,22 +421,18 @@ function M.server_per_root_dir_manager(make_config)
   return manager
 end
 
-function M.search_ancestors(startpath, func)
-  validate { func = { func, 'f' } }
-  local guard = 100
-  for path in M.path.iterate_parents(startpath) do
-    -- Prevent infinite recursion if our algorithm breaks
-    guard = guard - 1
-    if guard == 0 then
-      return
-    end
-
-    if func(path) then
-      return path
+local function searcher(startpath, patterns, tail)
+  tail = tail or false
+  for _, pattern in pairs(patterns) do
+    local res = vim.fs.find(pattern, { path = startpath, upward = true, stop = vim.env.HOME, type = 'directories' })
+    if #res ~= 0 then
+      --return latest item as root
+      return res[(tail and #res or 1)]
     end
   end
 end
 
+--TODO:rewrite there
 function M.root_pattern(...)
   local patterns = vim.tbl_flatten { ... }
   local function matcher(path)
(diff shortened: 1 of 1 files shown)


glepnir:
I want use `vim.fs` module to rewrite root dir section. but the `vim.fs` module only support 0.8+ .

@mikehaertl  can you test this ?

mikehaertl:
@glepnir Actually I don't have a use case for #1908 anymore. It was related to setting up the LSP for volar but since I migrated to mason I no longer have to deal with it.

Anyway, as described in the issue testing it is easy:

```
mkdir -p /home/mike/test/a/node_modules
mkdir /home/mike/test/node_modules
```
Then in neovim:

```
:lua print( require'lspconfig'.util.find_node_modules_ancestor('/home/mike/test/a') )
```

Output should be `/home/mike/test` not `/home/mike/test/a`.

glepnir:
i want rewrite root search in this pr. but ikd  what time this pr can be merge lol

mikehaertl:
I can't help much. But the method is called from many places so the change might break things.

glepnir:
I had rewrite it and some change is breakchange. now should works .parent will be root in some situation.

mikehaertl:
Note that `search_ancestors()` is also used in some code snippets in the docs, e.g.:

https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#volar

lua/lspconfig/util.lua
```diff
@@ -453,38 +446,45 @@ function M.root_pattern(...)
...
 end
+
 function M.find_node_modules_ancestor(startpath)
-  return M.search_ancestors(startpath, function(path)
-    if M.path.is_dir(M.path.join(path, 'node_modules')) then
-      return path
-    end
-  end)
+  return function()
+    return searcher(startpath, { '.hg' }, true)
```
mikehaertl, line 480:
```suggestion
    return searcher(startpath, { 'node_modules' }, true)
```

lua/lspconfig/util.lua
```diff
@@ -453,38 +446,45 @@ function M.root_pattern(...)
...
-    if M.path.is_dir(M.path.join(path, '.git')) or M.path.is_file(M.path.join(path, '.git')) then
-      return path
-    end
-  end)
+  return function()
+    return searcher(startpath, { '.git' })
+  end
 end
+
 function M.find_mercurial_ancestor(startpath)
```
mikehaertl, line 472: Maybe also add a `tail` param here and to the other helpers that defaults to `false`? This would not break bc.

glepnir: yep . just for test i set it to true. just search on github. found lots of function in utils are used in people config ... can't do break change ..

glepnir: Needs more consideration and design here. take compatibility issues into consideration.  🧐

PR #2486, opened 2023-03-01
fix(commands): Strange `LspStop` arguments, not matching doc

Rational-Curiosity:
`LspStop` only working like:
```
:LspStop (1)
```
putting `id` between parenthesis.

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -117,18 +117,17 @@ end, {
 
 api.nvim_create_user_command('LspStop', function(info)
   local current_buf = vim.api.nvim_get_current_buf()
-  local server_name, force
+  local server_id, force
   local arguments = vim.split(info.args, '%s')
   for _, v in pairs(arguments) do
     if v == '++force' then
       force = true
-    end
-    if v:find '%(' then
-      server_name = v
+    elseif v:find '^[0-9]+$' then
+      server_id = v
     end
   end
 
-  if not server_name then
+  if not server_id then
     local servers_on_buffer = lsp.get_active_clients { bufnr = current_buf }
     for _, client in ipairs(servers_on_buffer) do
       if client.attached_buffers[current_buf] then
@@ -136,7 +135,7 @@ api.nvim_create_user_command('LspStop', function(info)
       end
     end
   else
-    for _, client in ipairs(get_clients_from_cmd_args(server_name)) do
+    for _, client in ipairs(get_clients_from_cmd_args(server_id)) do
       client.stop(force)
     end
   end


glepnir:
in my test `LspStop 1` or `LspStop lua_ls` both works.

Rational-Curiosity:
> in my test `LspStop 1` or `LspStop lua_ls` both works.

Maybe stopping all servers.

glepnir:
it also works fine. so what's your problem ? did you try the min config ?

Rational-Curiosity:
> it also works fine. so what's your problem ? did you try the min config ?

Reading the old code only server name with parent is accepted:

```
    if v:find '%(' then
      server_name = v
    end
```

and `get_clients_from_cmd_args` only accepts ids:

```
    for _, client in ipairs(get_clients_from_cmd_args(server_name)) do
      client.stop(force)
    end
```

then only `:LspStop (1)` works

> in my test `LspStop 1` or `LspStop lua_ls` both works.

stop all servers because of

```
  if vim.tbl_isempty(result) then
    return require('lspconfig.util').get_managed_clients()
  end
```
code inside `get_clients_from_cmd_args` function

plugin/lspconfig.lua
```diff
@@ -117,26 +117,25 @@ end, {
...
-  local server_name, force
+  local server_id, force
   local arguments = vim.split(info.args, '%s')
   for _, v in pairs(arguments) do
     if v == '++force' then
       force = true
-    end
-    if v:find '%(' then
-      server_name = v
+    elseif v:find '^[0-9]$' then
```
glepnir, line 125: there can be simply `if tonumber(v)` ? if it no number it will return nil.

glepnir:
right i got your point.

plugin/lspconfig.lua
Rational-Curiosity: Thank you, I'm not a `lua` programmer.

Rational-Curiosity: Uhmm... `find` is faster than `tonumber`

```lua
local now = os.clock()
tonumber('10')
tonumber('10a')
print(os.clock() - now) -- 1.2e-05
local v = '10'
local v1 = '10a'
now = os.clock()
v:find '^[0-9]$'
v1:find '^[0-9]$'
print(os.clock() - now) -- 3e-06
```

I prefer `find`

glepnir: make sense looks like tonumber will do some memory alloc.

glepnir:
squash commit and change the commit message like title then force push will better.

Rational-Curiosity:
Done :-)

PR #2493, opened 2023-03-05
feat: add azure pipelines lsp

treatybreaker:
### Summary
This adds [microsoft's fork of yamlls](https://github.com/microsoft/azure-pipelines-language-server) which supports their custom parsing of some data values in their json specs providing better support for azure pipelines.

### Bullet Points
- Added `azure-pipelines-language-server`
- Ensured lint style followed via running `stylua **/*.lua && luacheck **/*.lua` in the repository

### Notes
This accompanies a PR to mason.nvim: [williamboman/mason.nvim#1604](https://github.com/williamboman/mason.nvim/pull/1064)

The full name of the lsp server is quite bulky, being `azure_pipelines_language_server`, and may need a renaming for users, maybe `azure_pipelines_lsp`?

### System and Environment Details 
I have tested and verified the server works on Linux with Neovim nightly, `NVIM v0.9.0 dev-1135+g419819b62`.

Config as passed to `lspconfig`:
```lua
local lsp_server_bin_dir = vim.fn.stdpath("data") .. "/mason/bin/"

require("lspconfig").azure_pipelines_ls.setup({
    cmd = { lsp_server_bin_dir .. "azure-pipelines-language-server", "--stdio" },
})
```
The server was installed using my PR branch to [`mason.nvim`](https://github.com/williamboman/mason.nvim/pull/1064), it can be installed otherwise with `npm install -g azure-pipelines-language-server` and then targeting its path with `node /PATH_TO_INSTALL/out/server.js --stdio`

### DIFF
--- a/lua/lspconfig/server_configurations/azure_pipelines_language_server.lua
+++ b/lua/lspconfig/server_configurations/azure_pipelines_language_server.lua
@@ -1,0 +1,53 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = {},
+    filetypes = { 'yaml' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+    settings = {},
+  },
+  docs = {
+    description = [[
+https://github.com/microsoft/azure-pipelines-language-server
+
+An Azure Pipelines language server
+
+**By default, `azure-pipelines-language-server` doesn't have a `cmd` set.** This is because nvim-lspconfig does not make assumptions about your path.
+You have to install the language server manually.
+
+`azure-pipelines-language-server` can be installed via `npm`:
+
+```sh
+npm install -g azure-pipelines-language-server
+```
+
+Once installed, point `cmd` to `server.js` inside the `server/out` directory:
+
+```lua
+cmd = {'node', '<path_to_repo>/server/out/server.js', '--stdio'}
+```
+
+By default `azure-pipelines-language-server` will only work in files named `azure-pipelines.yml`, this can be changed by providing additional settings like so:
+```lua
+require("lspconfig").azure_pipelines_language_server.setup {
+  ... -- other configuration for setup {}
+  settings = {
+      yaml = {
+          schemas = {
+              ["https://raw.githubusercontent.com/microsoft/azure-pipelines-vscode/master/service-schema.json"] = {
+                  "/azure-pipeline*.y*l",
+                  "/*.azure*",
+                  "Azure-Pipelines/**/*.y*l",
+                  "Pipelines/*.y*l",
+              },
+          },
+      },
+  },
+}
+```
+The Azure Pipelines LSP is a fork of `yaml-language-server` and as such the same settings can be passed to it as `yaml-language-server`.
+]],
+  },
+}
--- a/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
+++ b/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
@@ -23,10 +23,10 @@ You have to install the language server manually.
 npm install -g azure-pipelines-language-server
 ```
 
-Once installed, point `cmd` to `server.js` inside the `server/out` directory:
+Once installed, point `cmd` to `server.js` inside the `out` directory:
 
 ```lua
-cmd = {'node', '<path_to_repo>/server/out/server.js', '--stdio'}
+cmd = {'node', '<path_to_azure_pipelines_install>/out/server.js', '--stdio'}
 ```
 
 By default `azure-pipelines-ls` will only work in files named `azure-pipelines.yml`, this can be changed by providing additional settings like so:
--- a/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
+++ b/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
@@ -14,10 +14,10 @@ https://github.com/microsoft/azure-pipelines-language-server
 
 An Azure Pipelines language server
 
-**By default, `azure-pipelines-language-server` doesn't have a `cmd` set.** This is because nvim-lspconfig does not make assumptions about your path.
+**By default, `azure-pipelines-ls` doesn't have a `cmd` set.** This is because nvim-lspconfig does not make assumptions about your path.
 You have to install the language server manually.
 
-`azure-pipelines-language-server` can be installed via `npm`:
+`azure-pipelines-ls` can be installed via `npm`:
 
 ```sh
 npm install -g azure-pipelines-language-server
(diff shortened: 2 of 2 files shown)


treatybreaker:
I may need to revert [998cd1d](https://github.com/neovim/nvim-lspconfig/pull/2493/commits/998cd1dbd92535890cdd30615dd486be3f64a2ec), I did not fully grok Github Actions running the docgen.

glepnir:
it's better change the name to `azure_piplines_ls.lua` .  about the `path_to_repo` . does the repo have some special folder can detect the root of this repo ? like `node_modules` folder or something else like `tsconfig.js packages.js` etc. then we can find the root of this repo. then generate a cmd ?

treatybreaker:
I'll get the name changed over in a moment.



We could tee off of quite a few files, `LICENSE`, `package.json`, `thirdpartynotice.txt, or even the `out` directory created when installed via `npm`. The `out` directory contains `server.js`.

image

Should be doable to look for one of those to generate a `cmd`, the only issue is that it won't end up on the `PATH`, users will still have to provide the path to where it's installed.

Which file would be best for detection?

glepnir:
`LICENSE` not make sense. the `package.json thirdpartynotice.txt` and `out` dir can both used to detect the root .

treatybreaker:
I've been without internet for a bit, fun times.

Root dir cannot be determined trivially for azure pipelines, that was my bad. I was assuming you meant for detecting the the actual `server.js` to kick off the azure pipelines lsp.

An azure pipeline can live in *any* `yaml` file, the only guarantee for a root dir is that an azure pipeline is useless without being attached in some way to version control, so it's extremely likely a git directory will exist somewhere within the parent path where azure pipelines are concerned.

Most typically though, the pipelines are named `azure-pipelines.yml` and, by default, the language server won't work unless the `yaml` file it is attached to is named `azure-pipelines.yml`

So perhaps the root dir should be searching for `azure-pipelines.yml` by default. That's the only root pattern that may be common for azure pipelines.

treatybreaker:
That latest commit does root detection against `azure-pipelines.yml`

treatybreaker:
Gone ahead and ran stylua against the `azure_pipelines_ls.lua` file. Sorry for being a minor PIA.

glepnir:
what i mean is `cmd` support function type. then can generate a cmd by find some detect things.

treatybreaker:
I'm basing this off of the other configurations in the repository, such as [`visualforce_ls.lua`](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/visualforce_ls.lua). I lack the understanding of what you want me to do here. I'm not familiar enough with `cmd` usage in context of lspconfig to properly generate out the `cmd`.

Several other servers within lspconfig likewise do not provide a `cmd` or provide an empty `cmd`, with the expectation that the user provides the full `cmd`. For instance, [`raku_navigator`](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/raku_navigator.lua).

If I can get a pointer, I can implement what you're asking, but I'm not quite sure what is desired of `cmd` beyond requiring a user override.

glepnir:
```lua
cmd = function()
do some detection
return {  something you want get , '--stdin'}
end
```

treatybreaker:
The `azure-pipelines-language-server` does not end up in the user's path. The only way to do autodetection then is to attempt to find where their global `node_modules` are installed. On Linux that should be `/usr/lib/node_modules` or to the path that `NPM_CONFIG_PREFIX` is set to as an environment variable. On Windows npm is going to have the path either stored in the `NPM_CONFIG_PREFIX` variable or installed to `%appdata\npm\node_modules`.The other way, which requires some other tools to be installed (for grepping output), is to get the install prefix via `npm config ls -l` and grep for `^prefix = ".*"`.

This would be the first and only node-based non-binary producing server to do that for lspconfig from what I can tell.

You're sure you want this done?

treatybreaker:
That path (`PATH_TO_INSTALL`) was meant to be filled by wherever a user installed it to. The path is entirely determined within the scope of `npm install -g azure-pipelines-language-server` which installs it to one of the directories listed above which won't be on the environment PATH.

The chunk of code I'm using to automatically pick it up has to do with `mason.nvim` providing standard pathing (if I'm seeing where you're getting the `path_to_repo` from) from some redirects it generates. That won't be available to non-`mason.nvim` users.

Or am I misunderstanding you?

glepnir:
we don't care about the install node_module path . it's the user side. if he want use this server in neovim he will installed it right ? we care about the `cmd` field `cmd = {'node', '<path_to_repo>/server/out/server.js', '--stdio'}`

it use the `repo/server/outserver.js` as a part of cmd. we can get this repo path by some detect files or dir then generate a cmd . not find global `node_modules` folder right ?  

```lua
cmd = function()
  local cwd = vim.loop.cwd()
  if vim.fn.filereadable(cwd..'/server/out/server.js') == 1 then
    return { 'node', '/Users/xxx/test/server/out/server.js', '--stdio'}
  end
end
```

treatybreaker:
`<path_to_repo>` is probably a poor name on my part, more accurately it's `<path_to_azure_pipelines_ls_install>` which is handled by npm. Secondly, it seems I have left a mistake in that string, there is not subdir `server`, `out` is top level so I'll fix that. So it's not a path to any given repository, it's the full install path of azure pipelines ls that doesn't get registered within the `PATH` variable.

We can detect it, but it is going to require locating the global `node_modules` directory to pick it up. These non-binary producing language servers are installed via `npm install -g` which pops them under the global `node_modules` directory typically in a `lib` subdirectory within somewhere on the system.

The `server.js` file won't be readable from the current directory, it'll be somewhere else (which may be nonstandard). The easiest way to grab this is to do that grepping I mentioned earlier on `npm config ls -l` and that seems to work on `npm` version `8.19.2`, but I have no clue if it works on earlier versions at all much less if the `prefix` we grep from it is available in earlier versions.

That leads to checking common directory paths for the `server.js` for a non-binary producing npm-installed lsp server which may or may not exist in the locations we look for based on each system. Generally we can depend on the common paths though.

No other npm installed non-binary producing language servers within lspconfig seem to do detection for the `server.js` either. I think that type of detection may be within the scope of another PR for some refactoring for a better user experience, might be beyond the scope of this particular PR.

glepnir:
right. if it is no the current project dir then no issue here.  maybe we should provide a function like `global_mode_module ` to get the global node_module path . it's useful for some servers which write by ts/js .  then these server will have a complete `cmd` . user no need config it . should in another pr.

glepnir:
notice to myself (@glepnir):  improve this server cmd with `global_node_modules`

treatybreaker:
Thanks for merging and sorry for carrying you into a 18 deep discussion thread over 50 lines of code.

If I get time and I don't see `global_node_modules` implemented I may come back around to it later this week with another PR. No promises though.

glepnir:
just a idea of global node_modules . i will implement this function. no worry:)

PR #2494, opened 2023-03-05
feat: support Godot 4 port by default

pierpo:
Godot 4 changed its default port. Since Godot 4 reached stable a few days ago, we should propose the stable version's port by default.


pierpo:
Thanks for the quick merge 😊

PR #2497, opened 2023-03-07
feat: add ecsact support

zaucy:
Hi! I've developed and maintain a project called Ecsact. We've made an LSP server for our language and I'd love to include it in nvim-lspconfig as I am a big neovim user myself!

### DIFF
--- a/lua/lspconfig/server_configurations/ecsact.lua
+++ b/lua/lspconfig/server_configurations/ecsact.lua
@@ -1,0 +1,28 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'ecsact_lsp_server'
+local cmd = { bin_name, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
+  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
+end
+
+return {
+  default_config = {
+    cmd = cmd,
+    filetypes = { 'ecsact' },
+    root_dir = util.root_pattern '.git',
+    single_file_support = true,
+  },
+
+  docs = {
+    description = [[
+https://github.com/ecsact-dev/ecsact_lsp_server
+
+Language server for Ecsact.
+
+The default cmd assumes `ecsact_lsp_server` is in your PATH. Typically from the
+Ecsact SDK: https://ecsact.dev/start
+]],
+  },
+}


zaucy:
I missed my `make lint` in my previous push. I've re-ran it and force pushed now.

glepnir:

lua/lspconfig/server_configurations/ecsact.lua
```diff
@@ -0,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'ecsact_lsp_server'
+local cmd = { bin_name, '--stdio' }
+
```
glepnir, line 5:
```suggestion
if vim.fn.has 'win32' == 1 then
  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
end

```

glepnir: what i mean is you should add windows support

glepnir (changes requested):

lua/lspconfig/server_configurations/ecsact.lua
zaucy: I understand. I was saying it wasn't required because it was not installed via gem or npm which is what the contributing doc said the reason it was required.

However I was mistaken in thinking that it didn't work with the `.appinstaller` version of the Ecsact LSP. I swear I tested it with and it didn't work, but it seemed to work now. ¯\\\_(ツ)_/¯

Sorry for the confusion!

glepnir:
please squash and use correct commit format then force push.

lua/lspconfig/server_configurations/ecsact.lua
```diff
@@ -0,0 +1,28 @@
...
+  docs = {
+    description = [[
+https://github.com/ecsact-dev/ecsact_lsp_server
+
+Language server for Ecsact.
+
+The default cmd assumes `ecsact_lsp_server` is in your PATH. Typically from the
+Ecsact SDK: https://ecsact.dev/start
+		]],
```
glepnir, line 26:
```suggestion
]],
```

zaucy:
Thanks for the review and the patience :)

PR #2504, opened 2023-03-15
feat: add vhdl support

fmaggi:


glepnir:

fmaggi:
> does this server not support windows?

vhdl_ls provides both a linux and windows version, so I think it does. I haven't tried it though

glepnir:
what i mean is if it support windows then you need these codes.

```lua
  cmd = { 'cmd.exe', '/C', 'vhdl_ls}
end

```

fmaggi:
done!

PR #2505, opened 2023-03-15
Feat/vhdl support

fmaggi:
Filetypes were wrong. Fixed them


glepnir:
sync master then use new branch don't fix in prev branch.

fmaggi:
ok

Issue #2507, opened 2023-03-16
tsserver opening in deno project when root_dir doesn't match


dfontenot:
### Description

I work on both Node.JS and Deno projects. I have both language servers configured as:
```
-- from: https://deno.land/manual@v1.28.2/getting_started/setup_your_environment#neovim-06-using-the-built-in-language-server
nvim_lsp.denols.setup {
  on_attach = on_attach,
  capabilities=capabilities,
  root_dir = nvim_lsp.util.root_pattern("deno.json", "deno.jsonc"),
  init_options={ enable = true, unstable = false },
}

nvim_lsp.tsserver.setup {
  on_attach = on_attach,
  capabilities=capabilities,
  root_dir = nvim_lsp.util.root_pattern("package.json"),
}
```

If I go to a Deno project that only contains a deno.json file and no package.json file `:LspInfo` shows that both denols and tsserver clients are attached.

Please let me know if I'm misunderstanding how the root_dir setting and root_pattern is supposed to work.

### Neovim version

NVIM v0.8.0-1210-gd367ed9b2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

63f4c0082f007d2356df4bc335f55e6d414da89c

### Operating system and version

Ubuntu 22.04.2 LTS

### Affected language servers

denols, tsserver

### Steps to reproduce

1. `cd` to directory that contains a deno.json file but no package.json file
2. open any .ts or .tsx file in nvim
3. run `:LspInfo`

### Actual behavior

Both language servers are attached

### Expected behavior

Only denols should be attached

### Minimal config

```Lua
local on_windows = vim.loop.os_uname().version:match 'Windows'

local function join_paths(...)
  local path_sep = on_windows and '\\' or '/'
  local result = table.concat({ ... }, path_sep)
  return result
end

vim.cmd [[set runtimepath=$VIMRUNTIME]]

local temp_dir = vim.loop.os_getenv 'TEMP' or '/tmp'

vim.cmd('set packpath=' .. join_paths(temp_dir, 'nvim', 'site'))

local package_root = join_paths(temp_dir, 'nvim', 'site', 'pack')
local lspconfig_path = join_paths(package_root, 'test', 'start', 'nvim-lspconfig')

if vim.fn.isdirectory(lspconfig_path) ~= 1 then
  vim.fn.system { 'git', 'clone', 'https://github.com/neovim/nvim-lspconfig', lspconfig_path }
end

[...]

PR #2509, opened 2023-03-16
feat: single_file_support can be a function type

### LINKED ISSUE #2508: Add ability to control whether the server starts or not even when single_file_support=true
### Language server

N/A

### Requested feature

I have `denols` and `vtsls` for typescript.

I'm implemented the following root_dir function.

```lua
  local function deno_or_node(prefer)
    return function(fname)
      local type = {
        deno = require('lspconfig').util.root_pattern("deno.json")(fname),
        node = require('lspconfig').util.root_pattern("package.json")(fname),
      }
      local other = prefer == 'deno' and 'node' or 'deno'
      if type[prefer] and type[other] then
        local is_other = #vim.split(type[other], '/', { plain = true }) > #vim.split(type[prefer], '/', { plain = true })
        if is_other then
          return
        end
      end
      return type[prefer]
    end
  end
```

But I have to specify `single_file_support=false` for this.

It's useful if lspconfig provided a function to determine `whether to attach or not`.

_No response_

glepnir:
Problem:  if server set the `single_file_support` to true , this may cause a conflict when a filetype has two or above servers config which set `single_file_support` to true.  this filetype should be only start the server which match the root dir of current project.  the current behavior is other servers that set the `single_file_support` will also start and attach this buffer.

Solution:  make field `single_file_support` type support `function` then they can do some logic in this function to determine whether to attach or not in `single_file` mode.

After this pr:  can do it like this to avoid start tsserver in deno project.

```lua
lspconfig.denols.setup({
  root_dir = lspconfig.util.root_pattern("deno.json", "deno.jsonc"),
})

lspconfig.tsserver.setup({
  root_dir = lspconfig.util.root_pattern('package.json'),
  single_file_support = function()
    local cwd = vim.loop.cwd()
    if vim.fn.filereadable(cwd ..'/deno.json') == 0 then
      return true
    end
    return false
  end
})
```

Fix #2508 
Also Relate #2507

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -103,7 +103,7 @@ listed in `:help vim.lsp.start_client()` with the following unique entries:
 
 - {single_file_support}
 
-    `bool` (default: nil)
+    `bool` | `function` (default: nil)
 
     Determines if a server is started without a matching root directory.
     See |lspconfig-single-file-support|.
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -85,12 +85,19 @@ function configs.__newindex(t, config_name, config_def)
 
     local get_root_dir = config.root_dir
 
+    local function run_in_single(single)
+      vim.validate {
+        single = { single, { 'nil', 'b', 'f' } },
+      }
+      return (type(single) == 'boolean' or single == nil) and single or single()
+    end
+
     function M.launch()
       local root_dir
       if get_root_dir then
         local bufnr = api.nvim_get_current_buf()
         local bufname = api.nvim_buf_get_name(bufnr)
-        if #bufname == 0 and not config.single_file_support then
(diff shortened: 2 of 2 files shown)


glepnir:
@justinmk  any suggestion ?

hrsh7th:
If single_file_support=true, nvim-lspconfig use `cwd` as `root_dir` by default.

NReilingh:
@justinmk My example: I want to support running denols in single-file mode for typescript files outside of js/ts projects, and in root_dir mode for regular deno projects, but also run tsserver in single-file mode for javascript files and all other non-deno projects.

Context: one of Deno's unique qualities is its import behavior -- modules can be referenced by URL in the `import` statement so you don't need a package.json. Makes it very convenient to just script something in a `.ts` file somewhere -- one of ryan's initial intentions for the project IIRC from the original announcement talk.

PR #2512, opened 2023-03-17
feat: add openedge support

gquerret:
Language Server support for [OpenEdge](https://www.progress.com/openedge)
Based on the VSCode plugin, everything should be ok in this first version.


gquerret:
I'll change the PR so that it doesn't include the "docs" commit.

PR #2519, opened 2023-03-21
feat: add PyrightSetPythonPath command to pyright config

ianliu:
The new command accepts a single path for a python executable that will be used to reconfigure the pyright language server. This is useful to change which python will be used by pyright without having to activate the python environment beforehand. The act of finding which python executable to use isn't handled here.

### DIFF
--- a/doc/server_configurations.md
+++ b/doc/server_configurations.md
@@ -6228,6 +6228,7 @@ require'lspconfig'.pyright.setup{}
 ```
 **Commands:**
 - PyrightOrganizeImports: Organize Imports
+- PyrightSetPythonPath: Reconfigure pyright with the provided python path
 
 **Default values:**
   - `cmd` : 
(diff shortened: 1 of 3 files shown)


glepnir:
please don't edit the doc. it auto generate from lua file.

ianliu:
@glepnir I did a force push to revert doc changes

justinmk:
See discussion in https://github.com/neovim/nvim-lspconfig/issues/1937

I'm thinking we might want some sort of "sub-command" entrypoint so that langserver-specific commands can be discovered. E.g. `:Lsp <tab>` would complete all available commands provided. And `:LspCapabilties <tab>` would show... something.

ianliu:
@justinmk maybe this could be done in this function: https://github.com/neovim/nvim-lspconfig/blob/c6105c449683b944b5d2138fcf82f18c657249e9/lua/lspconfig/util.lua#L88-L95 and keep the declarative definitions of commands intact.

PR #2522, opened 2023-03-23
fix(pyright): set `useLibraryCodeForTypes` to false

Fau818:
According to the official setting document, the default value of `python.analysis.useLibraryCodeForTypes` should be `false`. See [pyright/docs/settings.md](https://github.com/microsoft/pyright/blob/main/docs/settings.md)

When this option's value is true, it will significantly affect the completion speed of the code, and I hope to merge it as soon as possible.

### DIFF
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -45,7 +45,7 @@ return {
       python = {
         analysis = {
           autoSearchPaths = true,
-          useLibraryCodeForTypes = true,
+          useLibraryCodeForTypes = false,
           diagnosticMode = 'workspace',
         },
       },
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -45,7 +45,6 @@ return {
       python = {
         analysis = {
           autoSearchPaths = true,
-          useLibraryCodeForTypes = false,
           diagnosticMode = 'workspace',
         },
       },


glepnir:
> When this option's value is true, it will significantly affect the completion speed of the code, and I hope to merge it as soon as possible.

you can set it to false in your local right ? why you so hurry ?

Fau818:
> > When this option's value is true, it will significantly affect the completion speed of the code, and I hope to merge it as soon as possible.
> 
> you can set it to false in your local right ? why you so hurry ?

Sorry I didn't express myself well; my problem.

Today at noon, I used only `Pyright` as the language server in Neovim and VSCode, and Neovim's code completion was much slower.
So I started looking for the problem and finally found that it was the `useLibraryCodeForTypes` option. 

I was concerned that this would bother the others as well.

glepnir:
commit lint failed . can you change it ?

Fau818:
> commit lint failed . can you change it ?

Of course, now I've finished.

Thank you for your positive response!

glepnir:
if you set it to default value. i think should be  remove this field.

Fau818:
> if you set it to default value. i think should be remove this field.

Yes, you are right!
I have removed this field now.

mjlbach:
Did you test this before merging? This is not a fix but a regression, and it should be reverted. I say this both as the [person who added the language server](https://github.com/neovim/nvim-lspconfig/commit/3170421b4c733cea426d7654f2f70400b68cd4fb) and former neovim/lspconfig maintainer.

There's a reason this was enabled, it matches the upstream vscode default: https://github.com/microsoft/pylance-release/blob/main/README.md used in their language server (which forwards the settings the pyright).

Without this, if libraries don't ship typestubs pyright will have degraded behavior, and users will be confused why neovim is "worse" than vscode. Here's a clear test case that illustrates this.

https://github.com/microsoft/pyright/issues/4827

Fau818:
> Did you test this before merging? This is not a `fix` but a regression (and that's not what the conventional commit would be anyways), and it should be reverted. I say this both as the [person who added the language server](https://github.com/neovim/nvim-lspconfig/commit/3170421b4c733cea426d7654f2f70400b68cd4fb) and former neovim/lspconfig maintainer.
> 
> There's a reason this was enabled, it matches the upstream vscode default: https://github.com/microsoft/pylance-release/blob/main/README.md used in their language server (which forwards the settings to pyright, thus overriding the pyright "defaults").
> 
> Without this, if libraries don't ship typestubs pyright will have degraded behavior, and users will be confused why neovim is "worse" than vscode. Here's a clear test case that illustrates this.
> 
> [microsoft/pyright#4827](https://github.com/microsoft/pyright/issues/4827)
> 
> This isn't a unique case, a cursory search reveals tons of people get directed to turn on `useLibraryCodeForTypes`:
> 
> https://github.com/microsoft/pyright/search?q=uselibrarycodefortypes&type=issues https://github.com/microsoft/pylance-release/search?q=uselibrarycodefortypes&type=issues
> 
> The correct fix for this is to file an upstream bug with pylance/pyright (replicated in vscode) showing the degraded behavior as this benefits everyone, not disabling functionality for neovim users. If you want this as the default, you should change it in your config.

My mistake; You make a good point; blame me too shallow.

Fau818:
I'm sorry for all of this.

mjlbach:
It's not your responsibility to know these things, you just filed a PR. My message is directed at glepnir as maintainer.

Fau818:
I'm sorry to bother you all again.

Just now, I updated my neovim from ` HEAD-bad218c` to `HEAD-8b7fb66`.

I found that even if I set `python.analysis.useLibraryCodeForTypes = true`, the huge completion delay I had before disappeared. Now it really works as it does in VSCode.

It seems to be an upstream issue.

mjlbach:
From looking at the commit reords there is nothing that could be relevant (I wrote a signifiant chunk of the performance critical part of the LSP client and none of that has been touched). It's more likely you updated pyright.

Fau818:
> From looking at the commit records there is nothing that could be relevant (I wrote a signifiant chunk of the performance critical part of the LSP client and none of that has been touched). It's more likely you updated pyright.

Thank you very much for your reply! Yes, the problem is with Pyright! I just rolled back my version of Pyright, and it reproduced my problem.

PR #2523, opened 2023-03-23
chore: remove redundant config for Lua

hezhizhen:
There are no different configs for Lua, so I removed its section.

I also fix some typos.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,10 +8,6 @@ end_of_line = lf
 insert_final_newline = true
 charset = utf-8
 
-[*.lua]
-indent_style = space
-indent_size = 2
-
 [{Makefile,**/Makefile,runtime/doc/*.txt}]
 indent_style = tab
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -8,10 +8,6 @@ end_of_line = lf
 insert_final_newline = true
 charset = utf-8
 
-[*.lua]
```
glepnir, line 11: why remove this from editorconfig.  this is useful for who use neovim builtin editorconfig.

hezhizhen: I think there are no different configs for Lua, so I removed its section. Does the settings in [*] section not work for neovim builtin editorconfig?

polyzen: Because it's covered already in the `[*]` block

glepnir: right . it already in `*` block.

glepnir:
LGTM

PR #2526, opened 2023-03-24
feat: add typst-lsp support

kaarmu:
Is it possible to add support for language servers in development? [This one](https://github.com/nvarner/typst-lsp) is for [Typst](https://typst.app).


lua/lspconfig/server_configurations/typst_lsp.lua
```diff
@@ -0,0 +1,18 @@
+local util = require 'lspconfig.util'
+
```
glepnir, line 2: add windows support in there.

PR #2528, opened 2023-03-26
feat: add solidity lsp built by NomicFoundation

wp-lai:
This PR adds the solidity language server built by NomicFoundation to the lspconfig server list.

More about this lsp:
- https://github.com/NomicFoundation/hardhat-vscode/blob/development/server/README.md
- https://github.com/NomicFoundation/hardhat-vscode/issues/341


lua/lspconfig/server_configurations/solidity_ls_nomicfoundation.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
```
glepnir, line 2: windows support in there.

PR #2539, opened 2023-04-01
Preserve gdscript version 1 and add gdscript version 2

julkip:
With the 4.0 release of the Godot Game Engine the new version 2 of the scripting language GDScript was introduced. The 3.x branch of the engine and with it GDScript 1 (sometimes wrongly called gdscript3, because it is engine version 3) is still widely used and should not be dumped. Because of this I introduced a configuration for version 2 and reverted the configuration for version 1 back to the defaults used in Godot 3.x


glepnir:
please read contribute.md and don't edit the doc file.

PR #2551, opened 2023-04-08
fix: fix wrong qmlls cmd path

denstiny:


lua/lspconfig/server_configurations/qmlls.lua
```diff
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    cmd = { 'qmlls' },
+    cmd = { '/usr/bin/qmlls6' },
```
glepnir, line 5: this is not correct you can't hardcode this path.

denstiny: Under arch, /usr/bin/qmlls6 exists

glepnir: I don't think other people store this executable in this path.  and you break it on windows. so you should config it in your local i think.

denstiny: You are right, it should be replaced by qmlls6

glepnir: checked it Cmakelist 6 is qt quick version . it looks like generate executable with a version bump. so can we auto generate this executable name ? then it can adapt the version 7 8 9 etc

dundargoc:
@denstiny is this still required? If so, mind fixing the conflicts?

PR #2560, opened 2023-04-14
fix: bad helm_ls cmd

mgalliou:
Loading helm files currently fails with this error:
```
Error executing lua callback: /usr/share/nvim/runtime/lua/vim/lsp.lua:239: cmd argument: expected string, got table
```

This is caused by `cmd` being set to `{ { 'helm_ls' }, 'serve' } }` instead of `{ 'helm_ls', 'serve' }`.


glepnir:
Thanks.

PR #2565, opened 2023-04-18
feat: wrap root_dir in coroutine.

glepnir:
1. make get `root_dir` can be async wrap in coroutine
2. change rust_analyzer get command output async.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -245,8 +245,6 @@ function configs.__newindex(t, config_name, config_def)
         return
       end
 
-      local root_dir
-
       local bufname = api.nvim_buf_get_name(bufnr)
       if #bufname == 0 and not config.single_file_support then
         return
@@ -257,16 +255,25 @@ function configs.__newindex(t, config_name, config_def)
       end
       local buf_path = util.path.sanitize(bufname)
 
-      if get_root_dir then
-        root_dir = get_root_dir(buf_path, bufnr)
-      end
+      local co = coroutine.create(function()
+        local root_dir
+        if get_root_dir then
+          root_dir = get_root_dir(buf_path, bufnr)
+        end
 
-      if root_dir then
-        manager.add(root_dir, false, bufnr)
-      elseif config.single_file_support then
-        local pseudo_root = #bufname == 0 and uv.cwd() or util.path.dirname(buf_path)
-        manager.add(pseudo_root, true, bufnr)
-      end
+        if root_dir then
+          vim.schedule(function()
+            manager.add(root_dir, false, bufnr)
+          end)
+        elseif config.single_file_support then
+          local pseudo_root = #bufname == 0 and uv.cwd() or util.path.dirname(buf_path)
+          vim.schedule(function()
+            manager.add(pseudo_root, true, bufnr)
+          end)
+        end
+      end)
+
+      coroutine.resume(co)
     end
 
     -- Check that the buffer `bufnr` has a valid filetype according to
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -105,7 +105,7 @@ function configs.__newindex(t, config_name, config_def)
 
       local pwd = uv.cwd()
 
-      coroutine.resume(coroutine.create(vim.schedule_wrap(function()
+      coroutine.resume(coroutine.create(function()
         local root_dir
         if get_root_dir then
           root_dir = get_root_dir(util.path.sanitize(bufname), bufnr)
@@ -145,7 +145,7 @@ function configs.__newindex(t, config_name, config_def)
           local pseudo_root = #bufname == 0 and pwd or util.path.dirname(util.path.sanitize(bufname))
           M.manager.add(pseudo_root, true, bufnr)
         end
-      end)))
+      end))
     end
 
     -- Used by :LspInfo
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -288,6 +288,7 @@ function configs.__newindex(t, config_name, config_def)
         local root_dir
         if get_root_dir then
           root_dir = get_root_dir(buf_path, bufnr)
+          reenter()
           if not api.nvim_buf_is_valid(bufnr) then
             return
           end
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -257,6 +257,16 @@ function configs.__newindex(t, config_name, config_def)
       end
       local buf_path = util.path.sanitize(bufname)
 
+      local function check_fast(root_dir, single_mode)
+        if not vim.in_fast_event() then
+          manager.add(root_dir, single_mode, bufnr)
+          return
+        end
+        vim.schedule(function()
+          manager.add(root_dir, single_mode, bufnr)
+        end)
+      end
+
       coroutine.resume(coroutine.create(function()
         local root_dir
         if get_root_dir then
@@ -264,14 +274,10 @@ function configs.__newindex(t, config_name, config_def)
         end
 
         if root_dir then
-          vim.schedule(function()
-            manager.add(root_dir, false, bufnr)
-          end)
+          check_fast(root_dir, false)
         elseif config.single_file_support then
           local pseudo_root = #bufname == 0 and pwd or util.path.dirname(buf_path)
-          vim.schedule(function()
-            manager.add(pseudo_root, true, bufnr)
-          end)
+          check_fast(pseudo_root, true)
         end
       end))
     end
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -270,6 +270,9 @@ function configs.__newindex(t, config_name, config_def)
         local root_dir
         if get_root_dir then
           root_dir = get_root_dir(buf_path, bufnr)
+          if not api.nvim_buf_is_valid(bufnr) then
+            return
+          end
         end
 
         if root_dir then
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -290,9 +290,8 @@ function configs.__newindex(t, config_name, config_def)
     -- Check that the buffer `bufnr` has a valid filetype according to
     -- `config.filetypes`, then do `manager.try_add(bufnr)`.
     function manager.try_add_wrapper(bufnr, project_root)
+      -- `config.filetypes = nil` means all filetypes are valid.
       if not config.filetypes or vim.tbl_contains(config.filetypes, vim.bo[bufnr].filetype) then
-        manager.try_add(bufnr, project_root)
-      else
         manager.try_add(bufnr, project_root)
       end
     end
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -97,54 +97,54 @@ function configs.__newindex(t, config_name, config_def)
     local get_root_dir = config.root_dir
 
     function M.launch()
-      local root_dir
-      if get_root_dir then
-        local bufnr = api.nvim_get_current_buf()
-        local bufname = api.nvim_buf_get_name(bufnr)
-        if #bufname == 0 and not config.single_file_support then
-          return
-        elseif #bufname ~= 0 then
-          if not util.bufname_valid(bufname) then
-            return
-          end
-          root_dir = get_root_dir(util.path.sanitize(bufname), bufnr)
-        end
+      local bufnr = api.nvim_get_current_buf()
+      local bufname = api.nvim_buf_get_name(bufnr)
+      if (#bufname == 0 and not config.single_file_support) or (#bufname ~= 0 and not util.bufname_valid(bufname)) then
+        return
       end
 
-      if root_dir then
-        api.nvim_create_autocmd('BufReadPost', {
-          pattern = fn.fnameescape(root_dir) .. '/*',
-          callback = function()
-            M.manager.try_add_wrapper()
-          end,
-          group = lsp_group,
-          desc = string.format(
-            'Checks whether server %s should attach to a newly opened buffer inside workspace %q.',
-            config.name,
-            root_dir
-          ),
-        })
-        for _, bufnr in ipairs(api.nvim_list_bufs()) do
-          local bufname = api.nvim_buf_get_name(bufnr)
-          if util.bufname_valid(bufname) then
-            local buf_dir = util.path.sanitize(bufname)
-            if buf_dir:sub(1, root_dir:len()) == root_dir then
-              M.manager.try_add_wrapper(bufnr)
+      coroutine.resume(coroutine.create(vim.schedule_wrap(function()
+        local root_dir
+        if get_root_dir then
+          root_dir = get_root_dir(util.path.sanitize(bufname), bufnr)
+        end
+
+        if root_dir then
+          api.nvim_create_autocmd('BufReadPost', {
+            pattern = fn.fnameescape(root_dir) .. '/*',
+            callback = function()
+              M.manager.try_add_wrapper()
+            end,
+            group = lsp_group,
+            desc = string.format(
+              'Checks whether server %s should attach to a newly opened buffer inside workspace %q.',
+              config.name,
+              root_dir
+            ),
+          })
+
+          for _, buf in ipairs(api.nvim_list_bufs()) do
+            local buf_name = api.nvim_buf_get_name(buf)
+            if util.bufname_valid(buf_name) then
+              local buf_dir = util.path.sanitize(buf_name)
+              if buf_dir:sub(1, root_dir:len()) == root_dir then
+                M.manager.try_add_wrapper(buf)
+              end
             end
           end
+        elseif config.single_file_support then
+          -- This allows on_new_config to use the parent directory of the file
+          -- Effectively this is the root from lspconfig's perspective, as we use
+          -- this to attach additional files in the same parent folder to the same server.
+          -- We just no longer send rootDirectory or workspaceFolders during initialization.
+          local buf_name = api.nvim_buf_get_name(0)
+          if #buf_name ~= 0 and not util.bufname_valid(buf_name) then
+            return
+          end
+          local pseudo_root = #buf_name == 0 and uv.cwd() or util.path.dirname(util.path.sanitize(buf_name))
+          M.manager.add(pseudo_root, true, api.nvim_get_current_buf())
         end
-      elseif config.single_file_support then
-        -- This allows on_new_config to use the parent directory of the file
-        -- Effectively this is the root from lspconfig's perspective, as we use
-        -- this to attach additional files in the same parent folder to the same server.
-        -- We just no longer send rootDirectory or workspaceFolders during initialization.
-        local bufname = api.nvim_buf_get_name(0)
-        if #bufname ~= 0 and not util.bufname_valid(bufname) then
-          return
-        end
-        local pseudo_root = #bufname == 0 and uv.cwd() or util.path.dirname(util.path.sanitize(bufname))
-        M.manager.add(pseudo_root, true, api.nvim_get_current_buf())
-      end
+      end)))
     end
 
     -- Used by :LspInfo
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -72,7 +72,7 @@ function configs.__newindex(t, config_name, config_def)
     end
 
     if config.autostart == true then
-      local event_conf = config.filetypes and { event = 'FileType', pattern = table.concat(config.filetypes, ',') }
+      local event_conf = config.filetypes and { event = 'FileType', pattern = config.filetypes }
         or { event = 'BufReadPost' }
       api.nvim_create_autocmd(event_conf.event, {
         pattern = event_conf.pattern or '*',
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -103,6 +103,8 @@ function configs.__newindex(t, config_name, config_def)
         return
       end
 
+      local pwd = uv.cwd()
+
       coroutine.resume(coroutine.create(vim.schedule_wrap(function()
         local root_dir
         if get_root_dir then
@@ -137,12 +139,11 @@ function configs.__newindex(t, config_name, config_def)
           -- Effectively this is the root from lspconfig's perspective, as we use
           -- this to attach additional files in the same parent folder to the same server.
           -- We just no longer send rootDirectory or workspaceFolders during initialization.
-          local buf_name = api.nvim_buf_get_name(0)
-          if #buf_name ~= 0 and not util.bufname_valid(buf_name) then
+          if not api.nvim_buf_is_valid(bufnr) or (#bufname ~= 0 and not util.bufname_valid(bufname)) then
             return
           end
-          local pseudo_root = #buf_name == 0 and uv.cwd() or util.path.dirname(util.path.sanitize(buf_name))
-          M.manager.add(pseudo_root, true, api.nvim_get_current_buf())
+          local pseudo_root = #bufname == 0 and pwd or util.path.dirname(util.path.sanitize(bufname))
+          M.manager.add(pseudo_root, true, bufnr)
         end
       end)))
     end
@@ -244,6 +245,7 @@ function configs.__newindex(t, config_name, config_def)
       if api.nvim_buf_get_option(bufnr, 'buftype') == 'nofile' then
         return
       end
+      local pwd = uv.cwd()
 
       local bufname = api.nvim_buf_get_name(bufnr)
       if #bufname == 0 and not config.single_file_support then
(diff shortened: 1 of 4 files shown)


lua/lspconfig/configs.lua
```diff
@@ -257,16 +255,25 @@ function configs.__newindex(t, config_name, config_def)
       end
       local buf_path = util.path.sanitize(bufname)
 
-      if get_root_dir then
-        root_dir = get_root_dir(buf_path, bufnr)
-      end
+      local co = coroutine.create(function()
```
ii14, line 258: `coroutine.resume(coroutine.create(...))`? `co` won't be used anywhere else

ii14 (commented):
`root_dir` is also called in `M.launch`, that has to be wrapped in a coroutine too.

Not sure what to do about `:LspInfo`, `root_dir` function is also called there

glepnir:
`M.launch` is also invoked `manager.add` . so it's fine. in test `LspRestart` and`LspStart` works fine.

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,47 +11,77 @@ local function reload_workspace(bufnr)
...
+    if not uv.is_closing(handle) then
+      uv.close(handle)
+    end
+  end
+
+  local stdout = uv.new_pipe(false)
+  local stderr = uv.new_pipe(false)
+  local chunks = {}
+  local handle
+  handle = uv.spawn('cargo', {
```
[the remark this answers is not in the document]
glepnir: jobstart is like a syntactic sugar based on uv.spawn. But something is done internally to make it easy to use. The current use case is simple though. So using uv.spawn directly can simplify the execution of some wrapping syntax. Although the time is negligible.

ii14:
`launch` function is used directly by `:LspStart` and `:LspRestart`. `get_workspace_dir` simply won't work when using those commands.

glepnir:
hmmm I test in a helloword rust project. `LspStop` then `LspStart` `get_workspace_dir` is run twice i get two print.  add print before` coroutin.resume(co, workspace_root)` line:46

lua/lspconfig/server_configurations/rust_analyzer.lua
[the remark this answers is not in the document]
ii14: `uv.spawn` is more verbose and harder to use correctly, and `jobstart` fixes some of its pitfalls. This code will also be an example of how to use it, so I'd just use the simpler alternative.

Also, failure has to be handled here. Looks like it falls back to other ways of determining the root directory. And if starting the process fails, the coroutine will be suspended and never resumed, not giving other code a chance to run.

ii14:
It will abort the function here:
```lua
  local co = coroutine.running()
  if not co then
    return
  end
```
`coroutine.running()` returns nil when its called outside of coroutines

glepnir:
get root coroutine in `manager.add` this function in `M.launch`

ii14:
Okay, but `root_dir` function is also called in configs.lua on line 110. That part won't be correct. We want to give a guarantee that `root_dir` functions are *always* called inside a coroutine. Try adding `assert(coroutine.running())` at the beginning of the `get_workspace_dir` function.

ii14:
Actually I guess such guarantee wouldn't be really necessary, but in that case you need to handle both sync and async context inside `root_dir` functions.

glepnir:
okay . i need think about that carefully.

ii14:
Don't forget about `:LspInfo`, it calls `root_dir` too, and everything is synchronous there. The way I see it the solutions are:

1. Make the entire `:LspInfo` asynchronous. You call `:LspInfo`, window shows up once all results are collected.
2. Show the window instantly with anything that's already available, update the buffer as you get back results asynchronously.
3. Don't guarantee that `root_dir` is always a coroutine, handle both sync/async it `root_dir` functions and leave `:LspInfo` as it is - blocking.

glepnir:
root dir in `LspInfo`  we can get it from `client.config.root_dir` this is more simply we don't need call root_dir function.

ii14:
That's what I initially thought too, but it's for the `Other clients that match the filetype` section, where it says what clients can be started, but are not necessarily actually started yet.

glepnir:
range all server name keys in configs. get the all the client that filetypes match current filetype. if client can get from active clients then root dir is `client.config.root_dir` . if not this server not start. then root dir is `not found`.

glepnir:
maybe need abstract this `coroutine.resume(coroutine.create)` like `if get_root_dir func then ` invoke a wrap function and pass get_root_dir into with a callback.   lspinfo and launch had changed. how do you think ?

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -64,9 +64,12 @@ local function make_config_info(config, bufnr)
...
   end)
-  local root_dir = config.get_root_dir and config.get_root_dir(buffer_dir)
-  if root_dir then
-    config_info.root_dir = root_dir
+
+  if config.get_root_dir then
+    config_info.root_dir = coroutine.resume(coroutine.create(vim.schedul_wrap(function()
+      local root_dir = config.get_root_dir(buffer_dir)
+      coroutine.yield(root_dir)
+    end)))
```
ii14, line 72: If `get_root_dir` calls `coroutine.yield` first, the result of that will be returned from this `coroutine.resume` (and as the second value, the first value is the status of the coroutine) and assigned to `root_dir`, which wouldn't be correct. We specifically want the result of `config.get_root_dir`.

Also there is a typo in `schedul_wrap`

lua/lspconfig/configs.lua
```diff
@@ -97,54 +97,54 @@ function configs.__newindex(t, config_name, config_def)
...
-          if not util.bufname_valid(bufname) then
-            return
-          end
+      local bufnr = api.nvim_get_current_buf()
+      local bufname = api.nvim_buf_get_name(bufnr)
+      if (#bufname == 0 and not config.single_file_support) or (#bufname ~= 0 and not util.bufname_valid(bufname)) then
+        return
+      end
+
+      coroutine.resume(coroutine.create(vim.schedule_wrap(function()
```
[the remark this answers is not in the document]
glepnir: because if don't run it on main thread it just start a server but not attach buffer.  maybe something wrong on lsp.buf_attach_client. I need check.

lua/lspconfig/ui/lspinfo.lua
glepnir: yep first value is status. a mistake. but i really don't like this in lspinfo ..

glepnir: how about now

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -64,9 +64,13 @@ local function make_config_info(config, bufnr)
...
-  local root_dir = config.get_root_dir and config.get_root_dir(buffer_dir)
-  if root_dir then
-    config_info.root_dir = root_dir
+
+  if config.get_root_dir then
+    local status, root_dir = coroutine.resume(coroutine.create(vim.schedule_wrap(function()
+      local root_dir = config.get_root_dir(buffer_dir)
+      coroutine.yield(root_dir)
+    end)))
+    config_info.root_dir = status and root_dir or error_messages.root_dir_not_found
```
ii14, line 73: Let's not use `coroutine.yield` outside of the "user code" and pass the result through a reference/upvalue. If for some reason the initial suspend is something like `coroutine.yield(42)`, `root_dir` will be assigned to 42. Not that anyone should do that, but a reference will be more robust.
```suggestion
    local root_dir
    local status = coroutine.resume(coroutine.create(vim.schedule_wrap(function()
      root_dir = config.get_root_dir(buffer_dir)
    end)))
    config_info.root_dir = status and root_dir or error_messages.root_dir_not_found
```
But doing this alone will make it stop working for anything has an async `root_dir` function. Preferably the buffer could be updated with this information once we actually get the result. Maybe checking if `coroutine.status` is `'suspended'`, and if so then save the line number and update that line? But this will make the code more complicated.

glepnir: > Maybe checking if coroutine.status is 'suspended', and if so then save the line number and update that line? But this will make the code more complicated.

honestly yes .. I don't have good idea on lspinfo .  It has some importance. But not that important.

lua/lspconfig/configs.lua
```diff
@@ -257,16 +257,23 @@ function configs.__newindex(t, config_name, config_def)
...
-      if root_dir then
-        manager.add(root_dir, false, bufnr)
-      elseif config.single_file_support then
-        local pseudo_root = #bufname == 0 and uv.cwd() or util.path.dirname(buf_path)
-        manager.add(pseudo_root, true, bufnr)
-      end
+        if root_dir then
+          vim.schedule(function()
+            manager.add(root_dir, false, bufnr)
+          end)
```
ii14, line 269: Avoid `vim.schedule` when it's not necessary? Same on the branch below
```suggestion
          if not vim.in_fast_event() then
            manager.add(root_dir, false, bufnr)
          else
            vim.schedule(function()
              manager.add(root_dir, false, bufnr)
            end)
          end
```
Or it could be even something like this, maybe even a new utility function that does this:
```lua
local function cb()
  manager.add(root_dir, false, bufnr)
end
if vim.in_fast_event() then
  vim.schedule(cb)
else
  cb()
end
```

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,44 @@ local function reload_workspace(bufnr)
   end)
 end
 
+local function get_workspace_dir(cmd)
+  assert(coroutine.running())
+  local co = coroutine.running()
+  if not co then
+    return
+  end
```
ii14, line 18:
```suggestion
  local co = assert(coroutine.running())
```

lua/lspconfig/configs.lua
glepnir: make sense

lua/lspconfig/configs.lua
```diff
@@ -257,16 +257,29 @@ function configs.__newindex(t, config_name, config_def)
       end
       local buf_path = util.path.sanitize(bufname)
 
-      if get_root_dir then
-        root_dir = get_root_dir(buf_path, bufnr)
+      local function check_fast(root_dir)
+        if not vim.in_fast_event() then
+          manager.add(root_dir, false, bufnr)
```
ii14, line 262: Second argument can't be false all the time, it used to be true on the `config.single_file_support` branch

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -64,9 +64,13 @@ local function make_config_info(config, bufnr)
   local buffer_dir = api.nvim_buf_call(bufnr, function()
     return vim.fn.expand '%:p:h'
   end)
-  local root_dir = config.get_root_dir and config.get_root_dir(buffer_dir)
-  if root_dir then
-    config_info.root_dir = root_dir
+
+  if config.get_root_dir then
+    local root_dir
+    local status = coroutine.resume(coroutine.create(vim.schedule_wrap(function()
```
ii14, line 70: I still don't get why `vim.schedule_wrap` is here, I think it's unnecessary and it only forces `get_root_dir` to be deferred. If I'm not mistaken this makes it so `root_dir` is always going to be `nil`, because it's assigned on the next event loop iteration once all of this code is done

glepnir: > I think it's unnecessary and it only forces get_root_dir to be deferred

that's right ..looks like before i don't get it work then i add it into schedule wrap . let me try again

ii14:
What's really important with coroutines is error handling. `coroutine.resume` works kinda like `pcall`, it returns the status and error message. If we don't handle errors they are just going to be silent and if something goes wrong we won't have any way of debugging it.

To make sure we are catching all unhandled errors (except in callbacks before they resume the coroutine, but that should be handled by nvim already), the coroutine body should be executed inside `pcall` like this:
```lua
local function func()
  error('test')
end

coroutine.resume(coroutine.create(function()
  local status, err = pcall(function()
    func()
  end)
  if not status then
    print('lspconfig: unhandled error: ' .. tostring(err))
  end
end))
```

glepnir:
yep use `print` can works well in coroutine thread. if use notify need yield error . or use a wrap try catch

```lua
function tryCatch(tryFunc, catchFunc)
    local success, result = pcall(tryFunc)
    if not success then
        catchFunc(result)
    end
end

tryCatch(function()
    error("xxxx")
end, function(err)
    print("Lspconfig Error: ", err)
end)
```

lua/lspconfig/configs.lua
```diff
@@ -72,17 +72,10 @@ function configs.__newindex(t, config_name, config_def)
...
-      if config.filetypes then
-        event = 'FileType'
-        pattern = table.concat(config.filetypes, ',')
-      else
-        event = 'BufReadPost'
-        pattern = '*'
-      end
-      api.nvim_create_autocmd(event, {
-        pattern = pattern,
+      local event_conf = config.filetypes and { event = 'FileType', pattern = table.concat(config.filetypes, ',') }
```
ii14, line 75: `pattern` can be a table
```suggestion
      local event_conf = config.filetypes and { event = 'FileType', pattern = config.filetypes }
```

lua/lspconfig/configs.lua
```diff
@@ -97,54 +90,57 @@ function configs.__newindex(t, config_name, config_def)
...
-        -- We just no longer send rootDirectory or workspaceFolders during initialization.
-        local bufname = api.nvim_buf_get_name(0)
-        if #bufname ~= 0 and not util.bufname_valid(bufname) then
-          return
-        end
-        local pseudo_root = #bufname == 0 and uv.cwd() or util.path.dirname(util.path.sanitize(bufname))
-        M.manager.add(pseudo_root, true, api.nvim_get_current_buf())
       end
+
+      util.async_run(try_launch)
```
ii14, line 143: Can't `try_launch` be inlined here?
```lua
util.async_run(function()
  -- ...
end)
```

lua/lspconfig/configs.lua
```diff
@@ -255,35 +250,48 @@ function configs.__newindex(t, config_name, config_def)
...
+        end
+
+        if root_dir then
+          check_fast(root_dir, false)
+        elseif config.single_file_support then
+          local pseudo_root = #bufname == 0 and pwd or util.path.dirname(buf_path)
+          check_fast(pseudo_root, true)
+        end
       end
+      util.async_run(try_start)
```
ii14, line 284: Same here, `try_start` can be inlined

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,40 @@ local function reload_workspace(bufnr)
...
+      local data = table.concat(chunks, '')
+      chunks = vim.json.decode(data)
+      local workspace_root = chunks and chunks['workspace_root'] or nil
+      coroutine.resume(co, workspace_root)
+    end,
+    stdout_buffered = true,
+    stderr_buffered = true,
+  })
+
+  if jobid < 0 then
```
ii14, line 39: Zero is failure too
```suggestion
  if jobid <= 0 then
```

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,40 @@ local function reload_workspace(bufnr)
   end)
 end
 
+local function get_workspace_dir(cmd)
+  local co = assert(coroutine.running())
+
+  local chunks = {}
+  local jobid = vim.fn.jobstart(cmd, {
+    on_stdout = function(_, data, _)
+      chunks[#chunks + 1] = unpack(data)
```
ii14, line 19: `var = unpack(data)` is equivalent to `var = data[1]`, other values are discarded. `data` array is split on newlines, so I think that should work:
```suggestion
      chunks[#chunks + 1] = table.concat(data, '\n')
```

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,40 @@ local function reload_workspace(bufnr)
...
+    end,
+    on_exit = function()
+      if next(chunks) == nil then
+        coroutine.resume(co, nil)
+        return
+      end
+
+      local data = table.concat(chunks, '')
+      chunks = vim.json.decode(data)
+      local workspace_root = chunks and chunks['workspace_root'] or nil
```
ii14, line 32: Can you move all of this out of the callback function, after `coroutine.yield()` and just call `coroutine.resume(co)` here?

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -64,9 +64,18 @@ local function make_config_info(config, bufnr)
...
+    local root_dir
+    local status = coroutine.resume(coroutine.create(function()
+      local status, err = pcall(function()
+        root_dir = config.get_root_dir(buffer_dir)
+      end)
+      if not status then
+        print('lspconfig: unhandled error: ' .. tostring(err))
+      end
+    end))
+    config_info.root_dir = status and root_dir or error_messages.root_dir_not_found
```
ii14, line 78: If we don't support async root_dir functions right now then there should be at least an error message that says that
```suggestion
    local co = coroutine.create(function()
      local status, err = pcall(function()
        root_dir = config.get_root_dir(buffer_dir)
      end)
      if not status then
        print('lspconfig: unhandled error: ' .. tostring(err))
      end
    end)
    coroutine.resume(co)
    if root_dir then
      config_info.root_dir = root_dir
    elseif coroutine.status(co) == 'suspended' then
      config_info.root_dir = error_messages.async_root_dir_function
    else
      config_info.root_dir = error_messages.root_dir_not_found
    end
```
And add to `error_messages`:
```lua
  async_root_dir_function = 'Asynchronous root_dir functions are not supported in :LspInfo',
```

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,35 @@ local function reload_workspace(bufnr)
...
+  })
+
+  if jobid <= 0 then
+    vim.api.nvim_err_writeln('Failed to start cargo metadata job id:' .. jobid)
+    return
+  end
+
+  coroutine.yield()
+  local data = table.concat(chunks, '')
+  chunks = vim.json.decode(data)
```
ii14, line 38:
```suggestion
  chunks = vim.json.decode(table.concat(chunks, ''))
```

lua/lspconfig/configs.lua
```diff
@@ -255,35 +248,47 @@ function configs.__newindex(t, config_name, config_def)
...
+        end
+        vim.schedule(function()
+          manager.add(root_dir, single_mode, bufnr)
+        end)
       end
+
+      util.async_run(function()
+        local root_dir
+        if get_root_dir then
+          root_dir = get_root_dir(buf_path, bufnr)
```
ii14, line 272: Because `get_root_dir` can be asynchronous, buffer could be deleted in the meantime
```suggestion
          root_dir = get_root_dir(buf_path, bufnr)
          if not api.nvim_buf_is_valid(bufnr) then
            return
          end
```

lua/lspconfig/configs.lua
```diff
@@ -97,54 +90,55 @@ function configs.__newindex(t, config_name, config_def)
...
+      if (#bufname == 0 and not config.single_file_support) or (#bufname ~= 0 and not util.bufname_valid(bufname)) then
+        return
+      end
+
+      local pwd = uv.cwd()
+
+      util.async_run(function()
+        local root_dir
+        if get_root_dir then
           root_dir = get_root_dir(util.path.sanitize(bufname), bufnr)
```
ii14, line 104: Same here
```suggestion
          root_dir = get_root_dir(util.path.sanitize(bufname), bufnr)
          if not api.nvim_buf_is_valid(bufnr) then
            return
          end
```

lua/lspconfig/ui/lspinfo.lua
ii14: Actually, `root_dir` in rust looks so simple that it should be very easy to support both synchronous and asynchronous versions. If `coroutine.running()` is `nil` then you just call `vim.fn.jobwait`. Maybe for now `root_dir` functions should be required to support synchronous version just to keep `:LspInfo` working

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -64,9 +65,25 @@ local function make_config_info(config, bufnr)
...
-    config_info.root_dir = root_dir
+
+  if config.get_root_dir then
+    local root_dir
+    local co = coroutine.create(function()
+      local status, err = pcall(function()
+        root_dir = config.get_root_dir(buffer_dir)
+      end)
+      if not status then
+        print('lspconfig: unhandled error: ' .. tostring(err))
```
ii14, line 76:
```suggestion
        vim.notify(('[lspconfig] unhandled error: %s'):format(tostring(err), vim.log.levels.WARN)
```

lua/lspconfig/util.lua
```diff
@@ -588,4 +588,18 @@ function M.strip_archive_subpath(path)
...
 
+function M.async_run(fn)
+  vim.validate {
+    fn = { fn, 'f' },
+  }
+  local co = coroutine.create(function()
+    local status, err = pcall(fn)
+
+    if not status then
+      print('lspconfig: unhandled error: ' .. tostring(err))
```
ii14, line 599:
```suggestion
      vim.notify(('[lspconfig] unhandled error: %s'):format(tostring(err), vim.log.levels.WARN)
```

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -10,6 +10,34 @@ local function reload_workspace(bufnr)
...
+  })
+
+  if jobid <= 0 then
+    vim.api.nvim_err_writeln('Failed to start cargo metadata job id:' .. jobid)
+    return
+  end
+
+  coroutine.yield()
+  chunks = vim.json.decode(table.concat(chunks, ''))
+  return chunks and chunks['workspace_root'] or nil
```
ii14, line 38:
```suggestion
  local stdout = {}
  local stderr = {}
  local jobid = vim.fn.jobstart(cmd, {
    on_stdout = function(_, data, _)
      stdout[#stdout + 1] = table.concat(data, '\n')
    end,
    on_stderr = function(_, data, _)
      stderr[#stderr + 1] = table.concat(data, '\n')
    end,
    on_exit = function()
      coroutine.resume(co)
    end,
    stdout_buffered = true,
    stderr_buffered = true,
  })

  if jobid <= 0 then
    vim.notify(('[lspconfig] cmd (%q) failed:\n%s'):format(table.concat(cmd, ' '), stderr), vim.log.levels.WARN)
    return
  end

  coroutine.yield()
  stdout = vim.json.decode(table.concat(stdout, ''))
  return stdout and stdout['workspace_root'] or nil
```

lua/lspconfig/configs.lua
```diff
@@ -255,35 +276,38 @@ function configs.__newindex(t, config_name, config_def)
...
-      bufnr = bufnr or api.nvim_get_current_buf()
-      local buf_filetype = api.nvim_buf_get_option(bufnr, 'filetype')
-      if config.filetypes then
-        for _, filetype in ipairs(config.filetypes) do
-          if buf_filetype == filetype then
-            manager.try_add(bufnr)
+      async_run(function()
+        local root_dir
+        if get_root_dir then
+          root_dir = get_root_dir(buf_path, bufnr)
```
ii14, line 290:
```suggestion
          root_dir = get_root_dir(buf_path, bufnr)
          reenter()
```

lua/lspconfig/configs.lua
```diff
@@ -4,6 +4,30 @@ local tbl_deep_extend = vim.tbl_deep_extend
...
+    fn = { func, 'f' },
+  }
+  local co = coroutine.create(function()
+    local status, err = pcall(func)
+
+    if not status then
+      vim.notify(('[lspconfig] unhandled error: %s'):format(tostring(err)), vim.log.levels.WARN)
+    end
+  end)
+  coroutine.resume(co)
```
ii14, line 28:
```suggestion
  coroutine.resume(coroutine.create(function()
    local status, err = pcall(func)
    if not status then
      vim.notify(('[lspconfig] unhandled error: %s'):format(tostring(err)), vim.log.levels.WARN)
    end
  end))
```

d-e-s-o:
f01429048950195bcab6c430b430fc44b2d9e05a broke `rust-analyzer` usage for me.

justinmk:
What problem does this PR solve? It is always helpful to state the problem being solved, in the PR description.

glepnir:
before rust get `workspace_root` is sync . now support async and also make gopls get cache async . in future can support `root_pattern` async by using luv fs module. 

it works well for me for a while for rust anlyzer and other people can confirm it works.  please @d-e-s-o   create an issue if you with error output. before check your env first.

d-e-s-o:
> please @d-e-s-o create an issue if you with error output. before check your env first.

What do you mean by "check your env first"? What should I be looking for, please? The problem is there is no output and I am not sure how to debug this. To the degree I can tell, all that changes is that `rust-analyzer` is never started (it's certainly not a running process as it was before the change).

glepnir:
Issue template have a guide for how to use minconfig. And i can confrim it works well for many people?   No output mean not found? Again can't reproduce with a min config

PR #2569, opened 2023-04-23
feat: make deno/cache async

sigmaSd:
This make it that it doesn't block input while its caching in the background

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -5,9 +5,10 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request('deno/cache', params, function(err)
+  client.request('deno/cache', params, function(err, _result, ctx)
     if err then
-      vim.api.nvim_err_writeln 'cache command failed'
+      local uri = ctx.params.referrer.uri
+      vim.api.nvim_err_writeln('cache command failed for ' .. uri)
     end
   end, bufnr)
 end
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -5,7 +5,11 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request_sync('deno/cache', params)
+  client.request('deno/cache', params, function(err)
+    if err then
+      vim.api.nvim_err_writeln 'cache command failed'
+    end
+  end, bufnr)
 end
 
 local function virtual_text_document_handler(uri, res, client)
@@ -59,6 +63,7 @@ end
 return {
   default_config = {
     cmd = { 'deno', 'lsp' },
+    cmd_env = { NO_COLOR = true },
     filetypes = {
       'javascript',
       'javascriptreact',
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -8,7 +8,7 @@ local function buf_cache(bufnr, client)
   client.request('deno/cache', params, function(err, _result, ctx)
     if err then
       local uri = ctx.params.referrer.uri
-      vim.api.nvim_err_writeln('cache command failed for ' .. uri)
+      vim.api.nvim_err_writeln('cache command failed for ' .. vim.uri_to_fname(uri))
     end
   end, bufnr)
 end


lua/lspconfig/server_configurations/denols.lua
```diff
@@ -5,7 +5,7 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request_sync('deno/cache', params)
+  client.request('deno/cache', params, function() end, bufnr)
```
glepnir, line 8: does there need handle error in callback when request failed?

sigmaSd: I don't think its needed, since if it fails, it will just print an error like:
```
LSP[denols][Warning] Module not found "https://deno.land/x/oak@0.1.0/mod.ts".
    at file:///home/mrcool/dev/deno/lab/tmp/a2Z/a.ts:1:16
```
I don't think we need more then that

Also I added another commit https://github.com/neovim/nvim-lspconfig/pull/2569/commits/4d1e30c1e38d4aa8597fd03abd9a27cbaee89fb0 that removes weird looking ansi escape sequences from the server messages

glepnir: ref https://github.com/mattn/vim-lsp-settings/blob/de060d969b405296a17bb9eea7c98b1bb1b15999/settings/deno.vim#L271

sigmaSd: I didn't find a way to get the result of the request, I tried 
```lua
  client.request('deno/cache', params, function(...)
    vim.print(...)
  end, bufnr)
```
but I don't see the result there,

Also I think its not important, like I said a warning is already printed

sigmaSd: Ok I see https://github.com/neovim/neovim/blob/bf0ac4f241cd22f53e026767ed5df89bf63bcf37/runtime/lua/vim/lsp.lua#L1430  what I did is supposed to show the result as a second argument but I get `true` instead

glepnir: this is sync ? callback param is`function(err, result, ctx) do err handle end`

sigmaSd: I used function(err,result,ctx) but this is what I get for those arguments
```
nil
true
{
  bufnr = 1,
  client_id = 1,
  method = "deno/cache",
  params = {
    referrer = {
      uri = "file:///home/mrcool/dev/deno/lab/tmp/Hse/a.ts"
    },
    uris = {}
  }
}
```

so I get result = true, instead of result = {} (empty object)

sigmaSd: I tried with request_sync and its the same thing I don't get an error 
image

Also note that deno lsp doesn't report the error back to the client 
```
[DEBUG][2023-04-23 13:56:22] .../vim/lsp/rpc.lua:387	"rpc.receive"	{  id = 6222,  jsonrpc = "2.0"}
```

It just prints the warning I showed initially which I think is enough

glepnir: just assert(err) or other error handle notify or print. true mean success ? important is err :)

sigmaSd: I mean even in error case it will return true, because deno server doesn't report the error back

glepnir: your print is `nil, true ,{...}` is `err, result, ctx` . in callback we only want handle err if not error is success right ? why you care about result is true ?

sigmaSd: and this https://github.com/mattn/vim-lsp-settings/blob/de060d969b405296a17bb9eea7c98b1bb1b15999/settings/deno.vim#L271 check for error in a very old version of deno, I don't think this is relevant anymore

glepnir: I don't think this relate any version . if callback return an error why you don't handle it ?  and same as
```vim
 if a:data['response']['result'] == v:true
        " Cache file would be enabled after reload buffer.
        let l:curpos = getcurpos()
        execute 'edit ' . expand('%:p')
        call setpos('.', l:curpos)
        echo 'Cache file installed'
    endif
```

in callback handler 

```lua
  if err then
       handle error
       return
  end

  if result == true then
     ....
     print('Cache file installed')
  end

Comment (sigmaSd) [reply lua/lspconfig/server_configurations/denols.lua]: I added https://github.com/neovim/nvim-lspconfig/pull/2569/commits/2d9a7f685bfc986c4f4c19bb61c65e43f7d5867a

(But what I'm arguing is that the error will never happen, deno server will always respond with result = true even if it can't cache a file)

Comment (glepnir) [reply lua/lspconfig/server_configurations/denols.lua]: https://github.com/denoland/deno/blob/bdffcb409fd1e257db280ab73e07cc319711256c/cli/lsp/language_server.rs#L175-L255

File: lua/lspconfig/server_configurations/denols.lua [hunk 2]
```diff
@@ -5,7 +5,11 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request_sync('deno/cache', params)
+  client.request('deno/cache', params, function(err, result)
+    if err or result ~= true then
```
glepnir, line 9: just `if err` is enough

lua/lspconfig/server_configurations/denols.lua
sigmaSd: The only way to hit the error branches in the match arms , is if the client sends a malformed message (which we do not)

In the case where it can't cache the file it shows a warning and return true https://github.com/denoland/deno/blob/bdffcb409fd1e257db280ab73e07cc319711256c/cli/lsp/language_server.rs#L224

Anyhow I see your point

lua/lspconfig/server_configurations/denols.lua
sigmaSd: done (I also squashed back the commits)

lua/lspconfig/server_configurations/denols.lua
sigmaSd: actually nvm you're right it probably can fail, though for some reason I can't reproduce it in practice

sigmaSd: ah I think the code is hitting this branch https://github.com/denoland/deno/blob/bdffcb409fd1e257db280ab73e07cc319711256c/cli/lsp/language_server.rs#L242 it just shows a warning and doesn't early return with an error

glepnir: also if there has `params` has some error then we can know it from response err.  the error handle not only for cache failed.

sigmaSd:
Anything else missing or it's ok to merge?

lua/lspconfig/server_configurations/denols.lua
```diff
@@ -5,7 +5,11 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request_sync('deno/cache', params)
+  client.request('deno/cache', params, function(err)
+    if err then
+      vim.api.nvim_err_writeln 'cache command failed'
```
glepnir, line 10: Is it necessary to add ctx relate information in there? because  you make it async then the buffer may changed.

sigmaSd: The context doesn't give real relevant information , for example

caching
```
import {} from "https://aze";
```
the context looks like this 
```
{
  bufnr = 1,
  client_id = 2,
  method = "deno/cache",
  params = {
    referrer = {
      uri = "file:///home/mrcool/dev/deno/lab/tmp/SZZ/a.ts"
    },
    uris = {}
  }
}
```
It doesn't give me the url I tried to cache, which is the only think I can see being useful to show (for example "errror caching `some-url`"

glepnir: if failed `local bufnr = vim.uri_to_bufnr(ctx.params.referrer.uri)` then prompt `cache for bufnr failed` .

glepnir: or `vim.uri_to_fname` prompt with a file name.

sigmaSd: ok added

lua/lspconfig/server_configurations/denols.lua
```diff
@@ -5,7 +5,12 @@ local function buf_cache(bufnr, client)
   local params = {}
   params['referrer'] = { uri = vim.uri_from_bufnr(bufnr) }
   params['uris'] = {}
-  client.request_sync('deno/cache', params)
+  client.request('deno/cache', params, function(err, _result, ctx)
+    if err then
+      local uri = ctx.params.referrer.uri
```
glepnir, line 10: `uri` not make sense  convert to file name  by using `vim.uri_to_fname`

sigmaSd: ah oops

sigmaSd: changed

PR #2576, opened 2023-04-25
fix(azure_pipelines_ls): add cmd

williamboman:
This npm package now provides an executable.

### DIFF
--- a/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
+++ b/lua/lspconfig/server_configurations/azure_pipelines_ls.lua
@@ -1,8 +1,15 @@
 local util = require 'lspconfig.util'
 
+local bin_name = 'azure-pipelines-language-server'
+local cmd = { bin_name, '--stdio' }
+
+if vim.fn.has 'win32' == 1 then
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/azure_pipelines_ls.lua
```diff
@@ -1,8 +1,15 @@
 local util = require 'lspconfig.util'
 
+local bin_name = 'azure-pipelines-language-server'
```
HieuTranTH, line 3: I don't think the npm package provide this executable. It is rather a mapping from your registry https://github.com/mason-org/mason-registry/blob/main/packages/azure-pipelines-language-server/package.yaml

williamboman: https://github.com/microsoft/azure-pipelines-language-server/pull/147, but it seems like they haven't made a release yet

HieuTranTH: oh, it's cool then

PR #2585, opened 2023-05-06
feat: add 'buck2' LSP

pbar1:
feat: add 'buck2' LSP


lua/lspconfig/server_configurations/buck2.lua
```diff
@@ -0,0 +1,27 @@
+local util = require 'lspconfig.util'
+
```
glepnir, line 2: missing window support here . check contribute.md or see what is the other server config do .

PR #2595, opened 2023-05-08
fix: use exepath on server command

### LINKED ISSUE #2130: append ".cmd" to command arg0 if it fails on Windows
## Problem

Commonly on Windows, some commands require the `.cmd` suffix. This is often missed, leading to issues like https://github.com/neovim/nvim-lspconfig/issues/2032 , and the solution is copy-pasted into configs: https://github.com/neovim/nvim-lspconfig/pull/2039

## Proposal

@williamboman https://github.com/neovim/nvim-lspconfig/issues/2032#issuecomment-1200413852 :

> `exepath()` seems to solve this pretty nicely, as Neovim [does respect `PATHEXT`](https://github.com/neovim/neovim/blob/9511faa819e8260aa7ae2c2ff140070bbc96efa9/src/nvim/os/fs.c#L321-L324). [null-ls uses it](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/47c04991af80b6acdf08a5db057908b52f4d0699/lua/null-ls/loop.lua#L147) and [mason-lspconfig.nvim successfully patches cmds with it](https://github.com/williamboman/mason-lspconfig.nvim/blob/f87c5796603aa3436d9cb1d36dbe5b2e579e4034/lua/mason-lspconfig/lspconfig_hook.lua#L66-L69). I'd be happy to open a PR that patches commands with `exepath()` on Windows + normalizes all commands in server configs.

Or is there a reason why we don't use `vim.fn.system()` (which should do this automatically) instead of `uv_spawn` (`vim.loop.spawn`))? (Probably the reason is: that doesn't work in Lua "fast" callbacks)

glepnir:
Fix #2130

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,12 +28,7 @@ When choosing a server name, convert all dashes (`-`) to underscores (`_`) If th
 Note that Windows has a limitation when it comes to directly invoking a server that's installed by `npm` or `gem`, so it requires additional handling.
 
 ```lua
-local bin_name = 'typescript-language-server'
-local cmd = { bin_name, '--stdio' }
-
-if vim.fn.has 'win32' == 1 then
-  cmd = { 'cmd.exe', '/C', bin_name, '--stdio' }
-end
+cmd = { 'typescript-language-server', '--stdio' }
 ```
 
 * `filetypes`: a list for filetypes a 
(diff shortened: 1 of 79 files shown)


justinmk:
This is a nice cleanup, thanks @glepnir . 

This also shows the cost of allowing special-case features in the configs.

PR #2596, opened 2023-05-08
feat: auto detect the `eslint.config.js` file to enable flat config support

sxyazi:
https://github.com/eslint/eslint/commit/7162d34df9a66c817c3bd4aafd3a03d226b58dd5

ESLint's new config system is complete, this PR offers auto-detecting of the new config file in the project root, then set the `experimental.useFlatConfig = true` to enable it automatedly.

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -113,7 +113,8 @@ return {
       }
 
       -- Support flat config
-      if vim.fn.filereadable(new_root_dir .. '/eslint.config.js') == 1 then
+      local path_sep = vim.fn.has 'win32' == 1 and '\\' or '/'
+      if vim.fn.filereadable(new_root_dir .. path_sep .. 'eslint.config.js') == 1 then
         config.settings.experimental.useFlatConfig = true
       end
 
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -112,6 +112,11 @@ return {
         name = vim.fn.fnamemodify(new_root_dir, ':t'),
       }
 
+      -- Support flat config
+      if #vim.fs.find('eslint.config.js', { path = new_root_dir, type = 'file' }) == 1 then
+        config.settings.experimental.useFlatConfig = true
+      end
+
       -- Support Yarn2 (PnP) projects
       local pnp_cjs = util.path.join(new_root_dir, '.pnp.cjs')
       local pnp_js = util.path.join(new_root_dir, '.pnp.js')
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -113,7 +113,7 @@ return {
       }
 
       -- Support flat config
-      if vim.fn.filereadable(new_root_dir .. "/eslint.config.js") == 1 then
+      if vim.fn.filereadable(new_root_dir .. '/eslint.config.js') == 1 then
         config.settings.experimental.useFlatConfig = true
       end
 
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -113,7 +113,7 @@ return {
       }
 
       -- Support flat config
-      if vim.fn.filereadable(new_root_dir .. "/eslint.config.js") then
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -112,6 +112,11 @@ return {
         name = vim.fn.fnamemodify(new_root_dir, ':t'),
       }
 
+      -- Support flat config
+      if #vim.fs.find('eslint.config.js', { path = new_root_dir, type = 'file' }) == 1 then
```
glepnir, line 116: this not work with link type . and missing stop condition like `stop = vim.env.HOME`

sxyazi: No, it only needs to find in the `new_root_dir` that’s resolved by `root_dir`, with the [default of `upward=false`](https://neovim.io/doc/user/lua.html#:~:text=upward%20(boolean%2C%20default%20false)) hence `stop` is unnecessary.

I'm not clear what the link type refers to, how should we fix it?

glepnir: system link type  `ln -s`

lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -113,7 +113,7 @@ return {
       }
 
       -- Support flat config
-      if #vim.fs.find('eslint.config.js', { path = new_root_dir, type = 'file' }) == 1 then
+      if vim.fn.filereadable(new_root_dir .. "/eslint.config.js") then
```
glepnir, line 116:
```suggestion
      if vim.fn.filereadable(new_root_dir .. "/eslint.config.js") == 1 then
```

sxyazi: Ah I got the return type wrong, sorry

glepnir:
Check out the documentation. This new configuration system defaults to false. It uses file and environment variables to control whether it is enabled. The environment variable `ESLINT_USE_FLAT_CONFIG` has higher priority. So shouldn't it be up to the user to control whether to enable or not. It should exist in the local configuration file. So I don't think it is necessary to add a check for this file by default right?

sxyazi:
The `ESLINT_USE_FLAT_CONFIG` environment variable exists only in ESLint, but it is ESLint LSP, i.e. [vscode-eslint](https://github.com/microsoft/vscode-eslint).

AFAIK, it can only be turned on by setting `experimental.useFlatConfig = true`, there is no configuration file or environment variable available.

polyzen:
ESLint itself uses `eslint.config.js` by default now if it's found.

Also for reference:
https://github.com/microsoft/vscode-eslint/issues/1644

lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -112,6 +112,11 @@ return {
         name = vim.fn.fnamemodify(new_root_dir, ':t'),
       }
 
+      -- Support flat config
+      if vim.fn.filereadable(new_root_dir .. '/eslint.config.js') == 1 then
```
glepnir, line 116: this file concat only work in *nix. so you may need support windows.

sxyazi: I checked the code of `util.path.join` that's used in this file, and it just uses "/" to join the paths:
https://github.com/neovim/nvim-lspconfig/blob/32f0c0465ac3ee53a1a9c0f72602a7b9cecb31d3/lua/lspconfig/util.lua#L157-L159

The "/" should work on Windows as well, right?

glepnir: nope .windows separator is `\\` . the path will sanitize

```lua
local function sanitize(path)
    if is_windows then
      path = path:sub(1, 1):upper() .. path:sub(2)
      path = path:gsub('\\', '/')
    end
    return path
  end
```

glepnir: `local path_sep = vim.fn.has('win32') == 1 and '\\' or '/'`

glepnir:
util.path has a santize function it will check is windows or not then convert `\\` to `/`

sxyazi:
The two lines above don't seem to call the `sanitize`, can you point out where it is?

glepnir:
long time no used windows check wiki https://en.wikipedia.org/wiki/Path_(computing) looks like windows also accept the '/' so maybe just path_join is enough

sxyazi:
Yeah, that's what I'm trying to say! I will revert the commit just made then.

PR #2597, opened 2023-05-09
fix: vim.lsp.rpc.connect is not allowed in cmd

Decodetalkers:
I think only table need to expand

only allow table to expand. this will still allow `vim.lsp.rpc.connection`  to work

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -86,7 +86,7 @@ function configs.__newindex(t, config_name, config_def)
 
     local config = tbl_deep_extend('keep', user_config, default_config)
 
-    if config.cmd then
+    if config.cmd and type(config.cmd) == 'table' then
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
I was using this feature and it's very useful to connect to an LSP over TCP.

Is it no longer supported? Is there an alternative approach?

WhyNotHugo:
Looks like a built-in LSP is also still using this feature: https://github.com/neovim/nvim-lspconfig/blob/7eed8b2150192e5ad05e1886fdf133493ddf2928/lua/lspconfig/server_configurations/gdscript.lua#L7

Can we maybe restore it? What's the rationale for dropping it?

WhyNotHugo:
Oh, sorry, I completely misread this PR and though it was the one dropping support for `vim.lsp.rpc.connect`. I see that it isn't. Sorry for the noise.

PR #2609, opened 2023-05-12
feat(LspconfigServerLifeCycle): stop/start server upon demand

hinell:
Start/stop servers when nvim session looses focus and restart them on demand.
This features is motivated by the fact that some servers may gain serious
memory footprint and may incur performance issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ See [server_configurations.md](doc/server_configurations.md) (`:help lspconfig-a
 nvim-lspconfig does not set keybindings or enable completion by default. The following example configuration provides suggested keymaps for the most commonly used language server functions, and manually triggered completion with omnifunc (\<c-x\>\<c-o\>).
 
 ```lua
+-- Uncomment to automatically start/stop servers when nvim is idle (non-focused) 
+-- See :h lspconfig-lifecycle for details
+-- vim.g.lspconfigServerLifeCycle = { timeout = 1000 * 60 * 30 }
+
 -- Setup language servers.
 local lspconfig = require('lspconfig')
 lspconfig.pyright.setup {}
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -473,6 +473,28 @@ contained in `:LspInfo`:
   will attempt to reattach to all previously attached buffers.
 
 ==============================================================================
+SERVER LIFECYCLE MANAGEMENT                                *lspconfig-lifecycle*
+
+`lspconfig` may automatically stop or restart workspace LSP servers by using 
+`:LspStart` / `:LspStop` commands. When `nvim` loses focus it will automatically
+stop buffer-bound LSP server upon clearing timeout threshold and restart
+otherwise. 
+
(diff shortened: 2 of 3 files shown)


lithammer:
This seems like a very drastic solution. Also, say I work on project and open file A to quickly edit something and then move on to file B. Now after 5 minutes the server would be stopped in middle of editing B? And what's worse is that it wouldn't start again until I move back to A (or some new file C) or start it manually.

I often work in projects where it can easily take 30s before the language server is ready. And having that happen every 5 min would be very frustrating.

sigmaSd:
1- What do other editors do (vscode in particular) ?
2- Do all lsp servers handle this stopping/starting gracefully ?

In my opinion this should not be a default behavior
This should be opt in for users who cares about this, or just even mentioned in the docs

hinell:
@lithammer This won't stop servers local to some specific language or buffer if you switch between them. I've tweaked to stop server only if `nvim` window looses focus.
@sigmaSd I wanted to make it optional, but there is no `lspconfig.setup()` function in this project.

lithammer:
> @lithammer This won't stop servers local to some specific language or buffer if you switch between them. I've tweaked to stop server only if `nvim` window looses focus.

Ah, it was `BufLeave` and `BufEnter` when I first looked.

hinell:
@lithammer Fixed it. My mistake. If we hook into bufers lifecycle it will constantly create/destroy timer. This may create problems if someone switches between buffers often.

hinell:
@lithammer Your suggestions? 10 minutes?

hinell:
@gegoune Checkout. The feature is now disabled by default and can only be enabled by seting a global option.

Default timeout is now set to 30 minutes. Compare two commits.

glepnir:
A shutdown restart caused some language server cache calculations to be lost. What performance problem? Is this something the client should deal with. Obviously not. This is an implementation issue of the server.  nothing to do on there.

hinell:
@glepnir Some servers may easily take up gigabytes of memory. Keeping cache in memory is problem of servers, not LSP clients. They should not rely on RAM in this sense. I have several editors opened at the same time and can tell you they take up 150MB on average at least.

glepnir:
what i mean client is `vim/neovim  emacs vscode` .  This is not something the neovim client should deal with. It is normal for the server to consume memory. No matter what editor you use, this will happen as long as you use lsp. A short reboot doesn't fix the problem. eg rust_analyzer uses the arena memory allocator. It does not free memory. vscode vim-lsp emacs lsp-mode helix I don't see this  either. I can't think of any reason to merge this pr. Sorry. Just my personal opinion. You can ask other people for their opinion.

hinell:
Cache is a not problem. Optimized servers keep theer cache on HDD instead of memory to make it shareable. This PR is intended for optimization as it requires lspconfig commands.

hinell:
Please, follow this repo for progress on standalone plugin:
* [hinell/lsp-timeout.nvim](https://github.com/hinell/lsp-timeout.nvim)

PR #2611, opened 2023-05-12
draft: feat: enable F# LSP for *.fsx files on any folder

adelarsq:
I have used this config for some time. I think that can be added to the default configs, since I didn't notice drawbacks:

```lua
lspconfig.fsautocomplete.setup({
    on_attach = my_on_attach,
    root_dir = util.root_pattern('*.sln', '*.fsproj', '.git', '*.fsx'),
    capabilities = my_capabilities,
    filetypes = { 'fsharp' }
})
```

**Note:** I have closed the previous PR to satisfy the lint.

### DIFF
--- a/lua/lspconfig/server_configurations/fsautocomplete.lua
+++ b/lua/lspconfig/server_configurations/fsautocomplete.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'fsautocomplete', '--adaptive-lsp-server-enabled' },
-    root_dir = util.root_pattern('*.sln', '*.fsproj', '.git'),
+    root_dir = util.root_pattern('*.sln', '*.fsproj', '.git', '*.fsx'),
     filetypes = { 'fsharp' },
     init_options = {
       AutomaticWorkspaceInit = true,


glepnir:
.fsx is fsharp script file right ?  I don't think pass it as root dir pattern is correct.

adelarsq:
@glepnir Yes. Is there any other way to provide the root dir for the LSP on this case? The only way that has worked so far was like that.

glepnir:
does this script file can be outside fsharp project?

adelarsq:
@glepnir Yes. `*.fsx` files can be used that same way that `*.sh` are used.

With .NET installed you can use something like `dotnet fsi somefile.fsx`.

glepnir:
okay. can you try set it into filetype and config dotnet support single file mode ?

adelarsq:
@glepnir Cool idea! I will try to make it to work. For now I will keep this PR as draft

adelarsq:
@glepnir  Closing this issue for performance reasons (for a huge amount of `*.fsx` files). The best approach it to use `.git` or other control file for the F#'s project root definition.

For a better approach I think that will be necessary some changes on the FsAutocomplete.

PR #2614, opened 2023-05-15
feat(texlab): send position when building

carlosala:
Since Texlab v5.5.0, `textDocument/build` accepts `position` argument, which is going to be used if forwardSearchAfter is enabled. Using Texlab <5.5.0 it's not an issue, the argument gets skipped and no error is thrown. Same happens if forwardSearchAfter is not enabled.

### DIFF
--- a/lua/lspconfig/server_configurations/texlab.lua
+++ b/lua/lspconfig/server_configurations/texlab.lua
@@ -17,8 +17,10 @@ local texlab_forward_status = vim.tbl_add_reverse_lookup {
 local function buf_build(bufnr)
   bufnr = util.validate_bufnr(bufnr)
   local texlab_client = util.get_active_client_by_name(bufnr, 'texlab')
+  local pos = vim.api.nvim_win_get_cursor(0)
   local params = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/texlab.lua
```diff
@@ -19,6 +19,7 @@ local function buf_build(bufnr)
   local texlab_client = util.get_active_client_by_name(bufnr, 'texlab')
   local params = {
     textDocument = { uri = vim.uri_from_bufnr(bufnr) },
+    position = { line = vim.fn.line '.' - 1, character = vim.fn.col '.' },
```
glepnir, line 22:
```lua
local pos = vim.api.nvim_win_get_cursor(0)
pos[1] -1 is line pos[2] is col
```

carlosala: Should I also change the other times it appears? It's like that in all the file.

carlosala: @glepnir changed :)

glepnir:
thanks . LGTM

PR #2622, opened 2023-05-17
fix: ruby lsp features should be opt-out

zhengpd:
Without the `enabledFeatures` init option, ruby lsp would enable all the supported features. This change was introduced in https://github.com/Shopify/ruby-lsp/pull/562


glepnir:
the link pr make sense. thanks

PR #2637, opened 2023-05-27
feat: implement workspace

### LINKED ISSUE #2518: rust_analyzer skip root_dirs for external crates inside `~/.cargo`
### Language server

rust_analyzer

### Requested feature

In VSCode, external crates are skipped and not added to the `workspace_folders`.

This is not the case for the root_dir function in lsp-config.

The consequence is that using `goto definition` on a symbol which takes you to source code in `~/.cargo/registry/src/...` the external crate is added and rust-analyzer will then run `cargo check` or `cargo clippy` inside that folder as well, which will cause diagnostics from those crates to show up in workspace diagnostics.

This results in both increased time for producing diagnostics and other linting when saving, as RA has to check more, as well as cluttering the results of workspace diagnostics with warnings from other crates or the standard library, which are plentiful.

Relevant solution, which doesn't work with cargo workspaces as it uses a completely different root pattern resolution than the built in one: https://users.rust-lang.org/t/neovim-lsp-shows-diagnostic-for-external-crates/89842/3

_No response_

glepnir:
The workspace of vscode does not exist in neovim. It is also not possible to simply set a workspace for each root. This is wrong behavior. So I implemented the concept of a workspace in lspconfig. The way it works is. You can create multiple worksapce. A default space is automatically created when the first project is opened. If you don't change the workspace. All open projects will belong to the default space. When you open the project after changing the workspace shown, it will check whether the server supports the workspace_folder feature. Send didchangeworkspacefolder notification if supported. Re-use the server. Otherwise a new server instance will be started.

eg: when open first file which will spawn a server then workspace is 

```
{ defaule = {  {ProjectA , client.id} --a tuple } }
```
run `LspWorkspaceAdd myspace` this will create a new workspace and change current workspace to `myspace`  then open a project it will be

```
{
  defaule = {  {ProjectA , client.id} --a tuple },
  myspace = { {ProjectB , client.id } }
}
```
- [ ] add test case
- [ ] write doc for new workspace

Ref: 
1. https://code.visualstudio.com/docs/editor/workspaces
2. https://github.com/emacs-lsp/lsp-mode/discussions/3095
3. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders

Fix #2518

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -1,4 +1,4 @@
-local api, fn, lsp = vim.api, vim.fn, vim.lsp
+local api, lsp = vim.api, vim.lsp
 local windows = require 'lspconfig.ui.windows'
 local util = require 'lspconfig.util'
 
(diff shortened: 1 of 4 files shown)


glepnir:
improve lspinfo now show workspace and root 

image

glepnir:
need add more commands ? like `WorkspaceRemove ` `WorkspaceChagne` ? not sure

PR #2638, opened 2023-05-29
fix(denols): pass config to vim.lsp.handlers

uga-rosa:
SSIA


uga-rosa:
@glepnir Please review :)

PR #2642, opened 2023-06-02
fix(clangd): make sure root_dir use `.git` pattern

Fau818:


glepnir:
I think this should be fix in new version root which use `vim.fs` module and `stop` is `vim.env.HOME` not current.

Fau818:
> I think this should be fix in new version root which use `vim.fs` module and `stop` is `vim.env.HOME` not current.

I agree with your point. The changes I've made are just a reference to [lua_ls](https://github.com/neovim/nvim-lspconfig/blob/9166622781a39a829878d1fd24c174529d996838/lua/lspconfig/server_configurations/lua_ls.lua#LL18C1-L27C9).

And if this PR is not necessary, feel free to close it.

PR #2645, opened 2023-06-03
fix(rust_analyzer): check file is library

### LINKED ISSUE #2518: rust_analyzer skip root_dirs for external crates inside `~/.cargo`

glepnir:
Problem: 
if file is library file it will also register to workspaceFolder this will cause some diagnostic for this file

Solution:
check file is a library file then use active client root dir 

Fix #2518

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -48,11 +48,28 @@ local function get_workspace_dir(cmd)
   return stdout and stdout['workspace_root'] or nil
 end
 
+local function is_library(fname)
+  local cargo_home = os.getenv 'CARGO_HOME' or util.path.join(vim.env.HOME, '.cargo')
+  local registry = util.path.join(cargo_home, 'registry', 'src')
+  local toolchains = util.path.join(vim.env.HOME, '.rustup', 'toolchains')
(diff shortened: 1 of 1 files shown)


bukzor:
I'm getting this error from this code:

```
[lspconfig] unhandled error: ...ig/lua/lspconfig/server_configurations/rust_analyzer.lua:23: attempt to index a nil value
```

I don't know how I managed to get the error but perhaps you do?

PR #2648, opened 2023-06-04
fix(lspinfo): improve lspinfo behavior

glepnir:
problem: if file is a library file it will reuse a client .  now lspinfo will it running in single mode

solution:  add `single_buffers` in client object if file is running in single file mode passed it to this table. when show lspinfo can use this field to check file  run in workspace or single file mode.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -278,6 +278,18 @@ function M.server_per_root_dir_manager(make_config)
       end
     end
 
+    local function register_client_single(client)
+      if not client.single_buffers then
+        client.single_buffers = {}
+      end
+      client.single_buffers[bufnr] = true
+      api.nvim_buf_attach(bufnr, false, {
+        on_detach = function()
+          client.single_buffers[bufnr] = nil
+        end,
+      })
+    end
+
     local function attach_and_cache(root, client_id)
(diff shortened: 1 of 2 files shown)


lua/lspconfig/util.lua
```diff
@@ -347,6 +359,7 @@ function M.server_per_root_dir_manager(make_config)
         return
       end
       attach_and_cache(root_dir, client_id)
+      register_client_single(vim.lsp.get_client_by_id(client_id))
```
justinmk, line 362: does this affect core lspconfig functionality, not only "lspinfo" ?

glepnir: It shouldn't have any side effects. When there is a file opened in single file mode. It create a  field called`single_buffers`  in client table. And insert this buffer. Clean up this field when the buffer is detach.

justinmk: but isn't it strange that LspInfo features require this kind of change? This is a hint that the `server_per_root_dir_manager` and all that "cache" code is doing too much. Why is lspconfig maintaining its own client cache...?

glepnir: we have a client cache but is simple cached the client id

justinmk:
Needs a rebase since recent changes. See also https://github.com/neovim/nvim-lspconfig/issues/3376

PR #2652, opened 2023-06-06
Fix docstring typo

masaeedu:


glepnir:
no need edit any doc file. just lua file.

masaeedu:
@glepnir Whoops, didn't realize that was maintained by CI. It's a bit weird because the file appeared to be out of sync with the lua values.

PR #2655, opened 2023-06-06
feat: consider .mjs and .cjs in tailwindcss root_pattern

phelipetls:
Also consider `.mjs` and `.cjs` extensions in Tailwind CSS language server's `root_pattern`.

### DIFF
--- a/lua/lspconfig/server_configurations/tailwindcss.lua
+++ b/lua/lspconfig/server_configurations/tailwindcss.lua
@@ -98,11 +98,13 @@ return {
       end
     end,
     root_dir = function(fname)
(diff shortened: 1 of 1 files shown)


glepnir:
any spec doc of these file ?

phelipetls:
Not sure of any spec links, but here are some links:

https://webpack.js.org/guides/ecma-script-modules/#flagging-modules-as-esm
https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions

It's a Node.js thing apparently. `.mjs` is always treated as ESM (regardless of `package.json`) and `.cjs` is always treated as CommonJS.

PR #2662, opened 2023-06-08
Add support for gjs and gts files for the ember language server

NullVoxPopuli:
As an aside, I need to add a second root_dir pattern -- is there a way to have two root_patterns?

or is that what the args are in util.root_pattern? if first not found, use second?

### DIFF
--- a/lua/lspconfig/server_configurations/ember.lua
+++ b/lua/lspconfig/server_configurations/ember.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'ember-language-server', '--stdio' },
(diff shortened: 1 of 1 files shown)


glepnir:
you can pass multiple patterns in `root_pattern` .

NullVoxPopuli:
the root_pattern is probably fine for now

NullVoxPopuli:
@glepnir are there more changes you'd like to see?

glepnir:
these filetype is special. we don't have them in builtin right ? so it would be nice to  add how to add this filetype in doc ? like use `vim.filtype.add`

NullVoxPopuli:
they're keyed off of/set by treesitter, and some LSP tools, what's vim.filetype.add?

PR #2667, opened 2023-06-12
feat: add nelua-lsp

LucasTavaresA:
it actually works really well for how scuffed it looks

### DIFF
--- a/lua/lspconfig/server_configurations/nelua_lsp.lua
+++ b/lua/lspconfig/server_configurations/nelua_lsp.lua
@@ -1,0 +1,36 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    filetypes = { 'nelua' },
+    root_dir = util.root_pattern('Makefile', '.git', '*.nelua'),
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
+https://github.com/codehz/nelua-lsp
+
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/nelua_lsp.lua
```diff
@@ -0,0 +1,39 @@
+local util = require 'lspconfig.util'
+local path = '/usr/local/lib/nelua/lib/nelua-lsp'
```
glepnir, line 2: why hardcorde this path. if you download this into other dir .what happen and this path not exist on windows.

LucasTavaresA: this is the path where nelua installs and looks for its libraries, only linux of course...
nelua recommends and supports MSYS2 on windows this looks messy

i think making like omnisharp does and letting the user set the cmd is the best i can do, i have no idea of how this would work on windows

glepnir: it's better use system check on this path

LucasTavaresA:
so now im wondering if those backslashes work correctly or are even necessary, and do i need to account for msys32 path

LucasTavaresA:
i tried to not use `-L` but nelua can't detect it, so nothing better can be done

i think this is an appropriate place to put it on windows

PR #2661, opened 2023-06-08
feat(gopls): attach previous lsp instance for GOMODCACHE

icholy:
Issue: https://github.com/neovim/nvim-lspconfig/issues/804
Reboots: https://github.com/neovim/nvim-lspconfig/pull/879

PR #2673, opened 2023-06-14
perf(gopls): make get go mod cache path async

glepnir:
problem: get go mod cache path env variable is sync now may have performance issue 
solution: make it async in root_dir coroutine

Follow #2661  
Fix #2671

### DIFF
--- a/lua/lspconfig/server_configurations/gopls.lua
+++ b/lua/lspconfig/server_configurations/gopls.lua
@@ -1,6 +1,5 @@
 local util = require 'lspconfig.util'
-
-local mod_cache = vim.trim(vim.fn.system 'go env GOMODCACHE')
+local mod_cache = nil
 
 return {
   default_config = {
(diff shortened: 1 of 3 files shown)


justinmk:
Why is this non-trivial logic in a single config? Is each config going to maintain its own cache implementation? That is too complicated.

glepnir:
There is no need to maintain a cache for each configuration. This is a special case. For example rust-analyzer  and gopls. When you navigate to a lib file by goto definition or reference. Usually installed in some cache directory. For example. cargo and gopath/cache etc. These directories will match the root pattern. This will spawn a new server process. It is necessary to determine whether the file is a lib file which need reuse already exists. Similar logic you can see in `emacs lsp-mode` and `vscode`. `lsp-mode` uses a `library-fn` function registered for each server. Similar logic exists in rust-analyzer and gopls extensions in vscode. Currently only rust-analyzer and gopls are in lspconfig.

justinmk:
> There is no need to maintain a cache for each configuration. This is a special case.

if it's needed for more than one config then it's not special. Doing this kind of non-trivial stuff per-config is too hard to maintain and makes it also hard to discover what kind of concepts should be upstreamed into Nvim core.



> Similar logic you can see in `emacs lsp-mode` and `vscode`

It doesn't matter what they do. vscode especially has teams of people writing per-language LSP integrations. That can't possibly be the strategy for nvim-lspconfig.

glepnir:
If you want to implement something like that on the core. Something like `library-fn` needs to be implemented in the opt of `vim.lsp.start`. Used to determine whether the file is a lib file. If yes, there are already started clients Then return  the client. Otherwise start a new client. `library-fn` is a function that receives `fname`.

glepnir:
going to merge as problem solution. about check file is library maybe can do something on `vim.lsp.config` in future.

PR #2678, opened 2023-06-16
Added the default cmd in openedge_ls

TechnicalDC:


lua/lspconfig/server_configurations/openedge_ls.lua
```diff
@@ -2,6 +2,15 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
+    cmd = {
```
glepnir, line 5: does there have any relate standard doc ? post it in comment or in default doc will be useful.

PR #2679, opened 2023-06-16
fix(rust_analyzer): use RUSTUP_HOME in library check

tobyvin:
When checking if file is part of a library, use `RUSTUP_HOME` env variable if it is set, in the same way as `CARGO_HOME`.


glepnir:
correct i forget this. thanks for contribute

PR #2681, opened 2023-06-19
fix(rust-analyzier): avoid "attempt to index nil" error

bukzor:
This issue was introduced in https://github.com/neovim/nvim-lspconfig/pull/2645

I've no idea if this is the right fix, but it may be.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -20,7 +20,9 @@ local function is_library(fname)
   for _, item in ipairs { toolchains, registry } do
     if fname:sub(1, #item) == item then
       local clients = vim.lsp.get_active_clients { name = 'rust_analyzer' }
-      return clients[#clients].config.root_dir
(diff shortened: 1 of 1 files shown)


glepnir:
it's wired . it must have a `config.root_dir` in rust_analyer

bukzor:
No, not since I patched it :)

glepnir:
could you provide a reproduce step ?

bukzor:
I needed this patch again today, to prevent rust-analyzer setup from crashing.
I've no idea the root cause, so I don't know a minimal reproduction.
For me, it's sufficient to 

```sh
$ cd ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/env_logger-0.10.0/src
$ nvim lib.rs
```

Perhaps a clue :LspInfo reports (after above patch is applied): `root directory: Asynchronous root_dir functions are not supported in :LspInfo`. Perhaps this `config` attribute is set asynchronously, causing this nil error?

bukzor:
Thanks!

PR #2683, opened 2023-06-20
Add more root files for lua-ls

DarkWiiPlayer:
* Busted
* LDoc
* Git

### DIFF
--- a/lua/lspconfig/server_configurations/lua_ls.lua
+++ b/lua/lspconfig/server_configurations/lua_ls.lua
@@ -1,13 +1,16 @@
 local util = require 'lspconfig.util'
 
 local root_files = {
+  '.busted',
+  '.git',
+  '.luacheckrc',
   '.luarc.json',
   '.luarc.jsonc',
-  '.luacheckrc',
   '.stylua.toml',
-  'stylua.toml',
+  'config.ld', -- ldoc
   'selene.toml',
   'selene.yml',
+  'stylua.toml',
 }
 
 return {


lua/lspconfig/server_configurations/lua_ls.lua
```diff
@@ -1,13 +1,16 @@
 local util = require 'lspconfig.util'
 
 local root_files = {
+  '.busted',
```
glepnir, line 4: `.busted` just for test case ?

lua/lspconfig/server_configurations/lua_ls.lua
```diff
@@ -1,13 +1,16 @@
...
+  '.git',
```
glepnir, line 5: why add `.git` as a file in there ? usually it's a folder?

glepnir: this is duplicate we have `util.find_git_ancestor(fname)` .

lua/lspconfig/server_configurations/lua_ls.lua
```diff
@@ -1,13 +1,16 @@
...
+  '.luacheckrc',
   '.luarc.json',
   '.luarc.jsonc',
-  '.luacheckrc',
   '.stylua.toml',
-  'stylua.toml',
+  'config.ld', -- ldoc
```
glepnir, line 10: this not make sense .

DarkWiiPlayer: `config.ld` is the project-level config file for ldoc

DarkWiiPlayer:
I just spotted the `find_git_ancestor` at the end of the function.

Does this make sense as the last option? I've repeatedly run into trouble because my Lua projects are inside a `lua` directory and even if they are a git repository, the function will first return the `lua` dir before even getting to the git check.

A workaround that would work both for git repositories inside a `lua` directory as well as larger multi-language projects with Lua code in a `lua` subdirectory would be to prefer the longer of the two paths, if both are found.

PR #2689, opened 2023-06-22
fix: nixd root_pattern

Ruixi-rebirth:


lua/lspconfig/server_configurations/nixd.lua
```diff
@@ -5,9 +5,7 @@ return {
     cmd = { 'nixd' },
     filetypes = { 'nix' },
     single_file_support = true,
-    root_dir = function(fname)
-      return util.root_pattern(unpack { 'nixd.json', 'flake.nix' })(fname) or util.find_git_ancestor(fname)
```
glepnir, line 9: why remove these patterns of root detect .  It's better add link of spec which point the patterns in pr comment and `docs.description`

Ruixi-rebirth: nixd will read .nixd.json , and if it detects a git repository as it did before, then the root_dir will be in the root of the project rather than changing to the directory with .nixd.json

glepnir: if so add nixd doc about there will better.

Ruixi-rebirth: I tested it again and it doesn't need to be removed, I didn't think it through, now it's just fixing the missing prefix in the original `nixd.json` (just a dot)

Ruixi-rebirth: After many iterations of my testing. Only needed to fix the missing dot for the file in the previous root_pattern

glepnir:
thanks for contribute

PR #2697, opened 2023-07-04
Update pyright.lua to use `cwd()` if no python config files are present.

hacker-DOM:
This behavior is more natural and the default in VS Code: https://github.com/microsoft/pyright/issues/5407#issuecomment-1619132310.

### DIFF
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -32,7 +32,9 @@ return {
   default_config = {
     cmd = { 'pyright-langserver', '--stdio' },
     filetypes = { 'python' },
-    root_dir = util.root_pattern(unpack(root_files)),
+    root_dir = function()
+      return util.root_pattern(unpack(root_files))() or vim.loop.cwd()
+    end,
     single_file_support = true,
     settings = {
       python = {


glepnir:
there has `single_file_mode = true`  it will use `cwd` .

hacker-DOM:
> there has `single_file_mode = true` it will use `cwd` .

I'm not 100% this covers my use case. My use case is:

a directory with stuff like

tests/
  f.py 
pytypes/
  a.py

In tests/f.py, I would like to do `from pytypes.a import A`

In Vs Code this works. In Nvim I currently have to create a pyproject.toml `tool.pyright.include = ['.']`. This PR changes the root dir to default to cwd if no config file is found among parents (just as in vs code).

glepnir:
if in `tests` folder open  test.py the root will be `tests` i guess.  `tests` and `pytypes` both in a project?

hacker-DOM:
> if in `tests` folder open f.py the root will be `tests` i guess. `tests` and `pytypes` both in a project?

no no, my `cwd` is still same, I am not `cd`ing anywhere.

Yeah we have [pytypes](https://ackeeblockchain.com/woke/docs/latest/testing-framework/getting-started/#generating-pytypes) as well as tests.

glepnir:
hmm. then change your commit message force push.

anthony-S93:
This is a bad idea. Like I mentioned in #2728 , this PR disables single-file mode entirely. You are forcing the root directory to always fallback to Neovim's current working directory even though some files are meant to be edited with the LSP launched in single-file mode without any root directory set.

> This behavior is more natural and the default in VS Code

Just because a behavior is the default for VS Code does not mean it is appropriate for the workflow of Neovim.  Let me give you an example of why this PR is such a bad idea.

Let's say I'm currently in my home directory when I launch neovim from the terminal. This will essentially set Neovim's cwd to my home directory. After neovim launches, I open a python file `script.py` from inside neovim `:e /path/to/script.py`. `script.py` is not a component of some grand software repository. It is a utility script that performs some routine task; as such, it will sit inside a directory without any root markers. And now because of this PR, `script.py` will never open in single-file mode. Since neovim's cwd is my home directory, `nvim-lspconfig` tells `pyright` to open `script.py` with the root directory set to my home directory. But what if my home directory is a huge directory with hundreds of files? This causes significant deterioration in terms of the LSP's performance. Since the root is set to my home directory, pyright will have to analyse the entire directory for no reason at all. `script.py` is not even part of any repo. It is meant to be edited in single-file mode. On my system, it took a few seconds for Pyright to even load. 

**Neovim is not VSCode**
How does one uses VSCode? You launch VSCode, select a working directory from the side panel, and then start editing the files inside the working directory you just selected. Here's the thing: not everyone uses neovim like this. In neovim, you  sometimes want to open files that are outside of neovim's cwd. 

@glepnir 
Do I have permission to revert the changes in this PR? I have already tested the plugin without this PR. Pyright's performance issue went away after I removed this PR (duh).

glepnir:
lsp was born based on vscode. Other editors benefit from this agreement. So you can see that most editors use vscode as a reference when implementing lsp. The problem is due to a difference in the design of the editors. There is no guarantee that the behavior will be 100% consistent with vscode. T

pr breaks the single-file mode. Because the root directory of the cwd will be generated anyway. pyright will retrieve all pyhton files under cwd. My suggestion is to leave it as it is (remove cwd). If you need to use cwd,  modify it in your local pyright. （lspconfig provides a basic server configuration. Meet the needs of most people. For individual usage scenarios. Local modifications are more appropriate）

anthony-S93:
> lsp was born based on vscode. Other editors benefit from this agreement. So you can see that most editors use vscode as a reference when implementing lsp. The problem is due to a difference in the design of the editors. There is no guarantee that the behavior will be 100% consistent with vscode.
> 
> pr breaks the single-file mode. Because the root directory of the cwd will be generated anyway. pyright will retrieve all pyhton files under cwd. My suggestion is to leave it as it is (remove cwd). If you need to use cwd, modify it in your local pyright. （lspconfig provides a basic server configuration. Meet the needs of most people. For individual usage scenarios. Local modifications are more appropriate）
@glepnir 
Huh? So are you agreeing with me that this PR should be reverted? I'm a bit confused here.

glepnir:
> My suggestion is to leave it as it is (remove cwd). 

revert or remove .

anthony-S93:
> > My suggestion is to leave it as it is (remove cwd).
> 
> revert or remove .

Creating a PR to fix this now.

PR #2708, opened 2023-07-09
add shebangs to config

sigmaSd:
This is useful for running one of scripts with the lsp

see https://github.com/sigmaSd/deno-nvim/issues/2 for prior discussion

Also for precedence helix also has this feature

example of usage
```lua
require("deno-nvim").setup {
  server = {
    on_attach = require("lvim.lsp").common_on_attach,
    on_init = require("lvim.lsp").common_on_init,
    root_dir = require "lspconfig.util".root_pattern('deno.json', 'deno.jsonc', 'denonvim.tag'),
    -- single_file_support = true,
    shebangs = { "deno" }, -- new part here
```
a.ts
```ts
#!/usr/bin/env deno
Deno
```
now the script a.ts will automatically trigger deno lsp


williamboman:
This can easily be solved by adding a filetype detection instead and not adding a bespoke filetype detection implementation to nvim-lspconfig:

```lua
vim.filetype.add {
  pattern = {
    [".*"] = {
      priority = -math.huge,
      function(path, bufnr)
        local content = vim.filetype.getlines(bufnr, 1)
        if vim.filetype.matchregex(content, [[^#!/usr/bin/env deno]]) then
          return "javascript"
        end
      end,
    },
  },
}
```

sigmaSd:
The problem is I need to hook to the internals of lspconfig in order to start the lsp server in single file mode in case of a match, or do you have an example on how to do it externally?

Also I'm not really sure of the point of the file detection because I still use the ts and js extension (so file type is detected already) as that's what this PR assumes

williamboman:
> The problem is I need to hook to the internals of lspconfig in order to start the lsp server in single file mode in case of a match, or do you have an example on how to do it externally?
> 
> Also I'm not really sure of the point of the file detection because I still use the ts and js extension (so file type is detected already) as that's what this PR assumes

Ah if the language server supports single file support the flag could be enabled for it. This could either be done manually or be part of the default server config:

```lua
lspconfig.denols.setup {
  single_file_support = true,
}
```

sigmaSd:
But the point is to enable it only when the shebang exists

johnpyp:
Hey, following up on this - like @sigmaSd mentioned, the usecase mentioned here about disambiguating when to use `tsserver` or `denols`, when filetypes aren't enough.

For example:
- Node should apply when there's a `package.json` root file
- Deno should apply when there's a `deno.json` root file
- Node should apply when the hashbang calls `node`
- Deno should apply when the hashbang calls `deno`
- In the case of a tie, the user should be able to pick which one applies

Assuming this was implemented flexibly, you could imagine something like a "buffer rule" that determines whether or not the lsp should be attached after the rest of the existing filetype/root file rules already apply. This would also enable additional better custom heuristics like:

- Deno applies if any import fetches from a url that contains `deno.land` (a common import url)
- Deno applies if the `Deno` global is used in the buffer
... etc.

I'm not sure what the best way to implement this would be that's flexible enough

justinmk:
Instead of this, would it make sense for `:LspStart foo` to force the "foo" lsp client to attach to the current buffer? Maybe `:LspStart! foo`.

PR #2715, opened 2023-07-12
feat(omnisharp): find root directory for .csx scripts

### LINKED ISSUE #2714: Support for dotnet scripts
### Language server

omnisharp

### Requested feature

Omnisharp supports [dotnet-scripts](https://github.com/dotnet-script/dotnet-script) automatically. 
Those are ".csx" files with a shebang and a special syntax for dependencies. 
But neovim doesn't because the rootdir is not defined.

Example (using Lunarvim distribution):

Command: `:LspInfo`
Screenshot 2023-07-12 at 15 17 13

 [dotnet-scripts](https://github.com/dotnet-script/dotnet-script) has a cli to scaffold a new script and this is the structure:

```
$ mkdir Hello
$ cd Hello
$ dotnet script init
$ tree
.
├── main.csx
└── omnisharp.json

0 directories, 2 files

```
Note that a .csx script can be used in Azure Function (https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-csharp?tabs=functionsv2)

So, there is 3 options to find the root directory for scripts:

1. trying to find the file `omnisharp.json`

The problem is that the `omnisharp.json` is required to have the lsp to work, but this file is not require to run the script.
Example: [Azure Function doesn't have omnisharp.json](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-csharp?tabs=functionsv2#folder-structure)

2. trying to find main.csx

The problem is the file ".csx" could be named differently.
Example: in [Azure Function](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-csharp?tabs=functionsv2#folder-structure)

3. trying to find *.csx

Problems could happen in a ./script folder for example with a lot of csx. in different folder. Script could loaded between them using the for example: `#load ../OtherScript.csx`. [Reference](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-csharp?tabs=functionsv2#reusing-csx-code)

I think the "safer" is to find for `omnisharp.json` for dotnet-script and `function.json` for Azure Function, because it has less chance to create side effects. 
 
I have linked a PR with those changes.

_No response_

titouancreach:
close #2714

### DIFF
--- a/lua/lspconfig/server_configurations/omnisharp.lua
+++ b/lua/lspconfig/server_configurations/omnisharp.lua
@@ -39,7 +39,7 @@ return {
 
     filetypes = { 'cs', 'vb' },
     root_dir = function(fname)
-      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname)
+      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname) or util.root_pattern 'omnisharp.json'(fname) or util.root_pattern 'function.json'(fname)
     end,
     on_new_config = function(new_config, _)
       -- Get the initially configured value of `cmd`
--- a/lua/lspconfig/server_configurations/omnisharp.lua
+++ b/lua/lspconfig/server_configurations/omnisharp.lua
@@ -40,12 +40,7 @@ return {
     filetypes = { 'cs', 'vb' },
     root_dir = function(fname)
       local root_patterns = { '*.sln', '*.csproj', 'omnisharp.json', 'function.json' }
-      for _, pattern in ipairs(root_patterns) do
-        local match = util.root_pattern(pattern)(fname)
-        if match then
-          return match
-        end
-      end
+      return util.root_pattern(root_patterns)(fname)
     end,
     on_new_config = function(new_config, _)
       -- Get the initially configured value of `cmd`
--- a/lua/lspconfig/server_configurations/omnisharp.lua
+++ b/lua/lspconfig/server_configurations/omnisharp.lua
@@ -39,7 +39,13 @@ return {
 
     filetypes = { 'cs', 'vb' },
     root_dir = function(fname)
-      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname) or util.root_pattern 'omnisharp.json'(fname) or util.root_pattern 'function.json'(fname)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/omnisharp.lua
```diff
@@ -39,7 +39,7 @@ return {
 
     filetypes = { 'cs', 'vb' },
     root_dir = function(fname)
-      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname)
+      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname) or util.root_pattern 'omnisharp.json'(fname) or util.root_pattern 'function.json'(fname)
```
glepnir, line 42: it's better use a root_file table

```lua
local root_patterns = { '*.sln', '*.csproj' ,'omnisharp.json', 'function.json'} 
for _, pattern  in iparis(root_patterns) do
end
```

titouancreach: I updated the PR thanks !

lua/lspconfig/server_configurations/omnisharp.lua
```diff
@@ -39,7 +39,13 @@ return {
 
     filetypes = { 'cs', 'vb' },
     root_dir = function(fname)
-      return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname)
+      local root_patterns = { '*.sln', '*.csproj', 'omnisharp.json', 'function.json' }
+      for _, pattern in ipairs(root_patterns) do
```
glepnir, line 43: no need do for range in there just

```lua
return util.root_pattern(patterns)(fname),
```

titouancreach: Thanks for the feedback, I let copilot write the code :)

cedliang:
This PR seems to have caused behavioural changes for projects with `.sln` files that reference `.csproj` files in subdirectories. The original behaviour checked for `.csproj` files only after a `.sln` file was not found.

Now, the root will not be set correctly to `src/` for the following structure if `Main.cs` is opened:

``` 
src
├── All.sln
├── Proj
    ├── Proj.csproj
    ├── Main.cs
```

Can we reimplement without the `root_file` table? Unlike other ecosystems, `.sln` and `.csproj` files are not equal in importance in determining the root directory of a project - one is subordinate to the other.

PR #2720, opened 2023-07-13
feat: add `javascript` to emmet_ls

pedrobl85:
Adds  `javascript` support to emmet_ls as it was missing and emmet_ls implemented support for it (as seen in https://github.com/aca/emmet-ls#supported-filetypes)

### DIFF
--- a/lua/lspconfig/server_configurations/emmet_ls.lua
+++ b/lua/lspconfig/server_configurations/emmet_ls.lua
@@ -9,6 +9,7 @@ return {
       'eruby',
       'html',
       'htmldjango',
+      'javascript',
       'javascriptreact',
       'less',
       'pug',


TheBlckbird:
> You could manually select filetypes during setup for the language server, like so:
> 
> `require('lspconfig').emmet_ls.setup({ filetypes = { "css", "eruby", "html", "javascriptreact", "less", "sass", "scss", "svelte", "pug", "typescriptreact", "vue" }, }) ` I'd be fine with reversing it though, since I realize most people who would want to use emmet would be working with `.jsx` most likely.

Oh, thanks, I didn't think of that. But I believe many people find this annoying. What even is the use case for this in JS?

TheBlckbird:
@pedrobl85 Could you please do this?

PR #2730, opened 2023-07-21
fix(pyright): re-enable single-file mode

### LINKED ISSUE #2728: Root detection for `pyright` is broken
### Description

Now, the plugin always sets the root directory to `neovim`'s  current working directory even though a root marker is present.

In this example, I first launch neovim from my home directory. And then, from inside Neovim, I open a file that resides in a directory called `~/Dev/testing/`, which has the following structure.

https://github.com/neovim/nvim-lspconfig/assets/69449791/6f7f8d11-799a-4b84-808b-c4ff1f089d3c

The root detection here is wrong. The correct project root should be `~/Dev/testing`.  

I am not entirely sure what is causing this, but I suspect the culprit is #2697 

The new behavior is that the project root is always assigned to Neovim's current working directory, which is highly problematic for 2 reasons:
1. It automatically disables single-file mode. If root detection automatically defaults to cwd, **I am no longer able to open python files that reside in different project roots in the same neovim session** - because the root is now forced to be whatever directory my neovim session is in.
2. It will deteriorate the lsp's performance. If neovim's cwd is in directory `D` and I open my python file in a non-root directory `C`. Instead of setting the lsp mode to `single-file mode` like it's supposed to, the root is assigned to `D`. But what if `D` is a large directory with hundreds of files? Pyright would have to look at all of them even though it is completely unrelated to the project. 

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

_No response_

Arch Linux x86_64

### Affected language servers

pyright

### Steps to reproduce

1. `cd` into your home directory. 
2. Creating a test directory `mkdir Dev/testing` 
3. Create a root marker in test directory `touch Dev/testing/pyrightconfig.json`
4. Create a test file inside test directory `touch Dev/testing/testfile.py`
5. Launch neovim from your home directory `cd ; nvim`
6. (At this point, your neovim's cwd is your home directory) Open the test file from inside nvim `:e /Dev/testing/testfile.py`
7. Check the root directory with `:LspInfo`

### Actual behavior

Case 1: At the end of the steps, you should see that the root directory is set to your home directory, not the directory containing the root marker (`pyrightconfig.json`).

Case 2: If you delete the root marker `pyrightconfig.json` and open the file again, you will see that the root directory is also set to your home directory.

### Expected behavior

For case 1, the root directory needs to be set to the directory containing the root marker, not neovim's current working directory. 

For case 2, the lsp needs to be launched in `single-file` mode without the root marker. 

### Minimal config

```Lua
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
    require('packer').startup({
        function(use)
            use 'wbthomason/packer.nvim'
            use
            {
                'neovim/nvim-lspconfig',
                config = function()
                    local opts = { noremap = true, silent = true }

                    --Keymappings on attach
                        vim.api.nvim_buf_set_option(bufnr, 'omnifunc', 'v:lua.vim.lsp.omnifunc')
                        local bufopts = { noremap = true, silent = true, buffer = bufnr }
                        vim.keymap.set('n', 'gD', vim.lsp.buf.declaration, bufopts)
                        vim.keymap.set('n', 'gd', vim.lsp.buf.definition, bufopts)
                        vim.keymap.set('n', 'K', vim.lsp.buf.hover, bufopts)
                        vim.keymap.set('n', 'gi', vim.lsp.buf.implementation, bufopts)
                        vim.keymap.set('n', '<C-k>', vim.lsp.buf.signature_help, bufopts)
                        vim.keymap.set('n', '<space>wa', vim.lsp.buf.add_workspace_folder, bufopts)
                        vim.keymap.set('n', '<space>wr', vim.lsp.buf.remove_workspace_folder, bufopts)
                        vim.keymap.set('n', '<space>wl', function()
                        end, bufopts)
                        vim.keymap.set('n', '<space>D', vim.lsp.buf.type_definition, bufopts)
                        vim.keymap.set('n', '<space>rn', vim.lsp.buf.rename, bufopts)
                        vim.keymap.set('n', '<space>ca', vim.lsp.buf.code_action, bufopts)
                        vim.keymap.set('n', 'gr', vim.lsp.buf.references, bufopts)
                    end

                    local lsp_flags = {
                        debounce_text_changes = 150,
                    }

                    --Customize diagnostics Icons
                    local signs = { Error = '󰅙', Warn = '󰀦', Hint = "󰐗", Info = "󰋼" }
                    for type, icon in pairs(signs) do
                        local hl = "DiagnosticSign" .. type
                        vim.fn.sign_define(hl, { text = icon, texthl = hl, numhl = hl })
                    end
                    vim.diagnostic.config({
                        underline = false,
                        virtual_text = false,
                    })
                    
                    --Configuration for each language server
                    lspconfig.pyright.setup({
                        on_attach = on_attach,
                        capabilities = default_capabilities,
                    })
                    --Borders for floating windows
                    local lspconfig_windows = require("lspconfig.ui.windows")
                    lspconfig_windows.default_options = {
                        border = "rounded"
                    }
                end
            }
        end,
        config = {
            package_root = package_root,
            compile_path = install_path .. '/plugin/packer_compiled.lua',
            display = { non_interactive = true },
        }
    })
end

    vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
end

load_plugins()
require('packer').sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!"]]
```

### LSP log

https://gist.github.com/anthony-S93/ba22075cd4f7732088fc3b2f7f8255f3

anthony-S93:
Reverts problematic PR #2697 

Closes #2728

### DIFF
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -36,7 +36,6 @@ return {
     root_dir = function(fname)
       return util.root_pattern(unpack(root_files))(fname)
     end,
-    root_dir = util.root_pattern(unpack(root_files)),
     single_file_support = true,
     settings = {
       python = {
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -33,6 +33,9 @@ return {
   default_config = {
     cmd = { 'pyright-langserver', '--stdio' },
     filetypes = { 'python' },
+    root_dir = function(fname)
+      return util.root_pattern(unpack(root_files))(fname)
+    end,
     root_dir = util.root_pattern(unpack(root_files)),
     single_file_support = true,
     settings = {
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -33,9 +33,7 @@ return {
   default_config = {
     cmd = { 'pyright-langserver', '--stdio' },
     filetypes = { 'python' },
-    root_dir = function(fname)
-      return util.root_pattern(unpack(root_files))(fname) or vim.loop.cwd()
-    end,
+    root_dir = util.root_pattern(unpack(root_files)),
     single_file_support = true,
     settings = {
       python = {


lua/lspconfig/server_configurations/pyright.lua
```diff
@@ -33,9 +33,7 @@ return {
   default_config = {
     cmd = { 'pyright-langserver', '--stdio' },
     filetypes = { 'python' },
-    root_dir = function(fname)
-      return util.root_pattern(unpack(root_files))(fname) or vim.loop.cwd()
-    end,
+    root_dir = util.root_pattern(unpack(root_files)),
```
glepnir, line 36: fname need keep. just remove vim.loop.cwd

anthony-S93: Oops.  I made a mistake there.

PR #2514, opened 2023-03-17
docs: recommend LspAttach in readme

ii14:
Most users want to use the same on_attach function for all language servers.

* Bump the required nvim version in the readme to 0.8 (for `LspAttach` event)
* In the suggested configuration replace `on_attach` functions with an `LspAttach` autocommand
* Remove `noremap=true` (it's the default) and `silent=true` (why was it there?) from keymap options
* Remove `{ debounce_text_changes = 150 }` flags (it's the default in 0.8)

PR #2732, opened 2023-07-23
docs: make vimdocs consistent with readme

ii14:
Add changes from #2514 to vimdoc. Better late than never. Also fix `lsp-config` links, it's `lspconfig`

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -1,4 +1,4 @@
-*lspconfig.txt*         For Nvim version 0.7+
+*lspconfig.txt*         For Nvim version 0.8+
 
 nvim-lspconfig provides user-contributed configs for the Nvim |lsp| client.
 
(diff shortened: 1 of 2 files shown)


doc/lspconfig.txt
```diff
@@ -477,70 +477,64 @@ EXAMPLE KEYBINDINGS                                      *lspconfig-keybindings*
...
-      -- Server-specific settings...
-      settings = {
-        ["rust-analyzer"] = {}
-      }
-  }
+  vim.api.nvim_create_autocmd('LspAttach', {
+    group = vim.api.nvim_create_augroup('UserLspConfig', {}),
+    callback = function(ev)
+      -- Enable completion triggered by <c-x><c-o>
+      vim.bo[ev.buf].omnifunc = 'v:lua.vim.lsp.omnifunc'
```
glepnir, line 508: in 0.9 this set by default when `omnifunc` is empty . do we need a special statement on there ？

justinmk (approved):
good to fix this, but also the direction should be to reduce the lspconfig docs as much as possible, or at least remove duplication by putting most things into the vimdoc.

ii14:
As far as the current readme goes, in my opinion the way it is right now is okay, I don't see anything that's unnecessary there. You only get the most important information to get you started and the most common pitfalls. Yes, there is some duplication, but in this case I think it's worth it to get the basic instructions on how to configure lspconfig more visible.

PR #2738, opened 2023-07-27
Fix `CargoReload` sending RPC requests to unrelated LSP clients

speelbarrow:
The configuration for `rust_analyzer` in this plugin sets up a command `CargoReload` that calls an internal function `reload_workspace`. This function had the same problem as the function of the same name in `rust-tools.nvim` (see simrat39/rust-tools.nvim#267 and simrat39/rust-tools.nvim#410). I've applied the same fix I did here as I did there.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -2,12 +2,16 @@ local util = require 'lspconfig.util'
 
 local function reload_workspace(bufnr)
   bufnr = util.validate_bufnr(bufnr)
-  vim.lsp.buf_request(bufnr, 'rust-analyzer/reloadWorkspace', nil, function(err)
-    if err then
-      error(tostring(err))
-    end
-    vim.notify 'Cargo workspace reloaded'
-  end)
+  local clients = vim.lsp.get_active_clients { name = 'rust_analyzer', bufnr = bufnr }
+  for _, client in ipairs(clients) do
+    vim.notify 'Reloading Cargo Workspace'
+    client.request('rust-analyzer/reloadWorkspace', nil, function(err)
+      if err then
+        error(tostring(err))
+      end
+      vim.notify 'Cargo workspace reloaded'
+    end, 0)
+  end
 end
 
 local function is_library(fname)


lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -1,13 +1,16 @@
...
-    if err then
-      error(tostring(err))
-    end
-    vim.notify 'Cargo workspace reloaded'
-  end)
+local function reload_workspace()
+  local client = vim.lsp.get_active_clients { name = 'rust_analyzer' }
+  if #client ~= 0 then
+    vim.notify 'Reloading Cargo Workspace'
+    client[1].request('rust-analyzer/reloadWorkspace', nil, function(err)
```
glepnir, line 7: there can't use 1 index if there has two ?   get client which attach current buffer?

speelbarrow: I’m using index 1 because I’m searching clients by name. It’s possible for multiple clients to attach to one buffer, but if there are multiple clients with the name `”rust_analyzer”` then something has clearly gone wrong.

glepnir: > but if there are multiple clients with the name ”rust_analyzer” then something has clearly gone wrong.

maybe nope. if someone rewrite root detect for spawn two servers? Please don't use hard-coded assumptions. It's not 100% established

speelbarrow: Alright, will update accordingly.

lua/lspconfig/server_configurations/rust_analyzer.lua
```diff
@@ -1,13 +1,16 @@
...
 
-local function reload_workspace(bufnr)
-  bufnr = util.validate_bufnr(bufnr)
-  vim.lsp.buf_request(bufnr, 'rust-analyzer/reloadWorkspace', nil, function(err)
-    if err then
-      error(tostring(err))
-    end
-    vim.notify 'Cargo workspace reloaded'
-  end)
+local function reload_workspace()
```
glepnir, line 3: this function still need bufnr. then you can use it to get the client which attach this buffer .and send  request.

PR #2739, opened 2023-07-29
fix(efm): set `init_options` to enable format and code action.

milanglacier:
With the release of `v0.0.46` in efm, the `init_options` must be manually set to allow the format and code action abilities to be usable in neovim. I believe the format and code action are the most frequently used features of efm and should be enabled by default. The user shouldn't be bothered to manually set those options.


llllvvuu:
Once https://github.com/mattn/efm-langserver/pull/256 is merged and versioned, this should no longer be necessary

dundargoc:
@llllvvuu @milanglacier I guess this isn't necessary then? I will close it then, lemme know if I misunderstood.

PR #2740, opened 2023-07-29
feat(marksman): enable for `markdown.mdx`

llllvvuu:
`marksman` works for MDX files. We don't want to `set ft=markdown`, since there exist tools that don't work for MDX, so we want to keep .md and .mdx as separate filetypes.


llllvvuu:
closing since there is an actual mdx language server: https://github.com/mdx-js/mdx-analyzer/pull/226

PR #2759, opened 2023-08-11
lua_ls: Update links to refer to new wiki

carsakiller:
Small PR that updates links for the [lua language server](https://github.com/luals/lua-language-server) to point to the new wiki that is replacing the old GitHub wiki.


carsakiller:
add(comment): i think a pull request template could go a long way in helping new contributors

PR #2763, opened 2023-08-14
feat: add LspReconfigure command

overhacked:
Add command `LspReconfigure` that allows overriding settings for LSP servers attached to the current buffer.

Does some command-line parsing so that:
- Bare string literals don't have to be quoted:
  `:LspReconfigure Lua.hint.paramName=Disable` ➡️ `Lua.hint.paramName="Disable"`
- Non-Lua-Name characters get properly quoted, and JSON literals (`true`/`false`/`null`) do not:
  `rust-analyzer.assist.emitMustUse=true` ➡️ `["rust-analyzer"].assist.emitMustUse=true`
- List-based settings can be assigned without using JSON syntax:
  `Lua.diagnostics.disable=cast-local-type,cast-type-mismatch` ➡️ `Lua.diagnostics.disable=["cast-local-type","cast-type-mismatch"]`
  `Lua.diagnostics.disable=cast-local-type,` ➡️ `Lua.diagnostics.disable=["cast-local-type"]`

In all of these cases, an "autotable" is generated so that arbitrarily-nested settings keys can be used, and the value can be any legal JSON expression so complex lists and mappings can be used if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ Most of the time, the reason for failure is present in the logs.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config. Pass `autostart = false` to your `.setup{}` call for a language server if you would like to launch clients solely with this command. Defaults to all servers matching current buffer filetype.
 * `:LspStop <client_id>` Defaults to stopping all buffer clients.
 * `:LspRestart <client_id>` Defaults to restarting all buffer clients.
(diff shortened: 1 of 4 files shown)


justinmk:
This is too much code and too complicated. 

This repo is a collection of configs. It is not another platform that accumulates endless features. Actual LSP features should be added in Nvim core or as third-party plugins.

overhacked:
Understand and appreciate rejecting feature creep. I'll look for a better home for this functionality.

PR #2774, opened 2023-08-22
feat(phpactor): attach language client if config found

bart-jaskulski:
Based on presence of phpactor-specific configuration files, allow client to attach, even if we are not inside a composer package or a git repository.


justinmk:
sure why not

PR #2780, opened 2023-08-23
fix(jedi): add .git to jedi_language_server root_dir

lkhphuc:
Similar to #2722 . Detect .git as project root.


justinmk:
given that https://github.com/neovim/nvim-lspconfig/pull/2722 is also for python, can't argue with this

PR #2783, opened 2023-08-24
fix(pyright): set diagnosticMode = openFilesOnly

lkhphuc:
When you go to definition of a library function, pyright will analyze the entire packages to report diagnostic information. This is extremely slow and heavy https://www.reddit.com/r/neovim/comments/135fqp9/why_is_pyright_constantly_analyzing_files_it/ Switch to the default value, similar to in VSCode.

Screenshot 2023-08-24 at 11 36 33

Changing this setting does not affect the type inference from library code, does not affect the diagnostic of the opened files from user or library code.
As far as I know, "openFilesOnly" only report diagnostic information for the file you have opened, i.e workspace diagnostics won't report for files you have not opened in this session.
This seems like an acceptable tradeoff for me.

P/s: Or considering delete everything analysis setting to use the default value. This will keep it in sync with VSCode as well.

### DIFF
--- a/lua/lspconfig/server_configurations/pyright.lua
+++ b/lua/lspconfig/server_configurations/pyright.lua
@@ -42,7 +42,7 @@ return {
         analysis = {
           autoSearchPaths = true,
           useLibraryCodeForTypes = true,
-          diagnosticMode = 'workspace',
+          diagnosticMode = 'openFilesOnly',
         },
       },
     },


justinmk:
> Or considering delete everything analysis setting to use the default value. This will keep it in sync with VSCode as well.

no objection, but we can try this meanwhile

last-partizan:
Is there a way to change this on the fly?

I like the speed improvements gained from this change, but i'd like to have project diagnostics available sometimes.

PR #2793, opened 2023-09-03
feat(denols): enable import completion from registries

sigmaSd:
same as the vscode extension https://github.com/denoland/vscode_deno/blob/abd33d2edcf752d69e0c53f244ee549a04a6eba5/package.json#L247

`init_options` fails, it needs to be `settings = {deno`

Also I removed the `unstable = false`, since that is the default


justinmk:
would you mind fixing the lint issue?

sigmaSd:
done, didn't notice it

PR #2805, opened 2023-09-12
fix(ccls): add cuda filetype support

charleschetty:
I used to use clangd as my language server, and today I changed to ccls.  However, I found that ccls did not offer support to cuda.
In the configuration of clangd, the cuda filetype is added by default. So maybe it is better for ccls to support cuda by default?


glepnir:
commit amend your commit msg `fix(ccls): add cuda filetype support`

charleschetty:
Sorry for that , I have not read the CONTRIBUTING.md, and now the commit style has been changed

glepnir:
`git commit --amend`

PR #2806, opened 2023-09-12
fix(ccls): add cuda filetype support

charleschetty:
I used to use clangd as my language server, and today I changed to ccls. However, I found that ccls did not offer support to cuda.
In the configuration of clangd, the cuda filetype is added by default. So maybe it is better for ccls to support cuda by default?


charleschetty:
Sorry for the  inconvenience I made, and now I amend my commit

glepnir:
you can change before pr commit message by using `git commit --ammend` command ...

PR #2808, opened 2023-09-13
feat(tsserver): add _typescript.rename handler

faergeek:
tsserver calls this method after extracting a variable or function via code action. It enables giving extracted function or variable a name right away without requiring a separate step. Otherwise extracted variable or function is left with a generic name like "newLocal".

### DIFF
--- a/lua/lspconfig/server_configurations/tsserver.lua
+++ b/lua/lspconfig/server_configurations/tsserver.lua
@@ -17,6 +17,33 @@ return {
         or util.root_pattern('package.json', 'jsconfig.json', '.git')(fname)
     end,
     single_file_support = true,
+    handlers = {
+      ['_typescript.rename'] = function(_, result, ctx)
+        if not result then
+          return {}
+        end
+
+        local client = vim.lsp.get_client_by_id(ctx.client_id)
+
+        if client then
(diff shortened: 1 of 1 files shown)


faergeek:
I understand that this is an off-spec handler, though despite that I saw some handlers like `eslint/openDoc` or similar. The difference might in that they won't work without those methods at all and this added method is just for convenience. At the same time this doesn't look like a complicated handler.

justinmk:
I've discouraged building up features in lspconfig beyond just the minimum needed for basic configuration. We can't maintain hundreds of special handlers for LSP servers, that should be pushed out to specialized plugins.

One thing we should do in lspconfig is establish a pattern for recommending third-party plugins for a given LSP server.

faergeek:
Fair. I just noticed that some specific handlers were here already and thought may be this one could be added as well.

But leaving it to language-specific plugins makes sense, since it's absence doesn't break anything.

justinmk:
> Fair. I just noticed that some specific handlers were here already and thought may be this one could be added as well.

Yeah. I mean we could add this one too, but it's a slippery slope because we keep getting more. As a typescript user I agree this would be a nice feature.

PR #2820, opened 2023-09-19
feat(lua_ls): add main.lua as a root file

musjj:
Some types of lua projects uses `main.lua` as the main entry-point, so it should be considered as a root file. Example: Love2D, mpv plugins, etc.


justinmk:
Some types of projects have all kinds of files. "main.lua" could easily be a false positive, i.e. it might not be the root in some projects.

You need to customize this for your project.

PR #2822, opened 2023-09-20
fix: remove usages of private (now deprecated) functions

gpanders:
Functions prefixed with _ should not be used, as they have no stability
guarantees from Nvim and can be removed at anytime with no warning or
replacement. _trim was such a function that was recently removed.


hinell:
You should have actually used `vim.lsp.util.trim_empty_lines()` for that. AFAIK it was mentioned in the `deprecated.txt`

PR #2829, opened 2023-09-23
add util.containing_dir, allow enabling typst_lsp outside of git repos

classabbyamp:
- feat: add util.containing_dir function
    - this allows LSP configurations to use the first parent directory as the root_dir or as a fallback for the root_dir.
- docs: document util.containing_dir
- fix(typst_lsp): allow enabling outside of a git repo
    - typst-lsp does not support single files, as typst files can import other typst files in the same path. However, as a typesetting engine, typst documents don't always exist in a git repo. This allows typst-lsp to be enabled outside of git repos by adding a fallback to the first parent directory.

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -375,11 +375,15 @@ below returns a function that takes as its argument the current buffer path.
 >
    root_dir = util.root_pattern('pyproject.toml', 'requirements.txt')
 <
+- `util.containing_dir`: a function that returns the first parent directory.
+>
+   root_dir = util.containing_dir
+<
 - `util.find_git_ancestor`: a function that locates the first parent directory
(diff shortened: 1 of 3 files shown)


classabbyamp:
> root_dir: a function (or function handle) which returns the root of the project used to determine if lspconfig should launch a new language server, or attach a previously launched server when you open a new buffer matching the filetype of the server. Note, lspconfig does not offer a dedicated single file mode (this is not codified in the spec). **Do not add vim.fn.cwd or util.path.dirname in root_dir. A future version of lspconfig will provide emulation of a single file mode until this is formally codified in the specification.** A good fallback is util.find_git_ancestor, see other configurations for examples.

whoops, didn't see that. good job CI

PR #2831, opened 2023-09-25
fix(texlab): don't overly prioritize .latexmkrc

tomtomjhj:
$HOME/.latexmkrc is per-user config, so it shouldn't be prioritized over closer directory containing .git.


justinmk:
@clason any objection to this?

Issue #2816, opened 2023-09-18
Allow stopping automatic stopping of client


arenevier:
### Requested feature

I sometimes have to load files from a huge cpp project. nvim-lspconfig and clangd help me navigate those files easily. But after closing all the buffers, clangd process is not killed, and continues to use a lot of memory.

Maybe lsp-config could detach the buffer from the client on BufUnload, and stop the client if it does not any active buffer left.

justinmk:
lspconfig is a collection of configs. We are not going to add more and more features to it. It's just a collection of configs.

Nvim has `vim.lsp.start()`, you can call that and customize it endlessly. There doesn't need to be special options for every random thing.

arenevier:
lspconfig is already starting the clients. It was easier for me to modify lspconfig slightly than to reimplement everything. In case someone else is wants to stop clients on unload, here is what I have been using for the last two weeks: https://github.com/arenevier/nvim-lspconfig/commit/634aeadcce747a42da21c944fe95fb8003f8c794

PR #2837, opened 2023-09-29
On BufUnload, unload clients started by lspconfig

arenevier:
See https://github.com/neovim/nvim-lspconfig/issues/2816


justinmk:
The timer stuff looks like trouble. And the counterpoint is that some servers might take time to start, then users have to wait for them again.

PR #2839, opened 2023-09-29
eslint: deprecate packageManager setting

mochaaP:
https://github.com/microsoft/vscode-eslint/pull/1646

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -64,7 +64,7 @@ return {
     -- Refer to https://github.com/Microsoft/vscode-eslint#settings-options for documentation.
     settings = {
       validate = 'on',
-      packageManager = 'npm',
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/eslint.lua
```diff
@@ -64,7 +64,7 @@ return {
     -- Refer to https://github.com/Microsoft/vscode-eslint#settings-options for documentation.
     settings = {
       validate = 'on',
-      packageManager = 'npm',
+      packageManager = nil,
```
glepnir, line 67: if nil can be remove

mochaaP: I suppose keeping this as an explicit field can help with the typing, so we can have `nil|"npm"|"yarn"|"pnpm"`?

glepnir: fair

PR #2843, opened 2023-10-01
feat: Set default border rounded

nfwyst:


glepnir:
you could set border in your local.

PR #2844, opened 2023-10-01
refactor: default_opts for border

nfwyst:
the default border config is not used, and `options.border and options.border` is not necessary


glepnir:
It's used you can enable border by using

```lua
        require('lspconfig.ui.windows').default_options.border = 'single'
```

PR #2863, opened 2023-10-15
Add htmx-lsp support

OZoneGuy:


lua/lspconfig/server_configurations/htmx.lua
```diff
@@ -0,0 +1,27 @@
+local util = require 'lspconfig.util'
+-- NOTE: the binary can be installed in custom a location. The user should
+-- change the cmd to match the installation path
+local cargo_home = os.getenv 'CARGO_INSTALL_ROOT' or os.getenv 'CARGO_HOME' or os.getenv 'HOME' .. '/.cargo'
+
+return {
+  default_config = {
+    cmd = { cargo_home .. '/bin/htmx-lsp' },
```
glepnir, line 8: This doesn't make sense. If I install the binary somewhere else. will fail here

OZoneGuy: That is correct. Pushed a change to call without the path.

lua/lspconfig/server_configurations/htmx.lua
```diff
@@ -0,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'htmx-lsp' },
+    filetypes = { 'html' },
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname) or vim.fn.expand '%:p:h' or vim.loop.os_homedir()
```
glepnir, line 8:
```suggestion
      return util.find_git_ancestor(fname)
```

lua/lspconfig/server_configurations/htmx.lua
```diff
@@ -0,0 +1,24 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'htmx-lsp' },
+    filetypes = { 'html' },
```
glepnir, line 6:
```suggestion
    filetypes = { 'html' },
    single_file_support = true,
```

glepnir (changes requested):

lua/lspconfig/server_configurations/htmx.lua
```diff
@@ -0,0 +1,25 @@
...
+    single_file_support = true,
+    root_dir = function(fname)
+      return util.find_git_ancestor(fname)
+    end,
+    settings = {},
```
glepnir, line 11: no need

OZoneGuy:
Thanks for the review and your time! Really appreciate it.

PR #2870, opened 2023-10-22
feat: add stimulus-language-server support

Gert-JanPeeters:
Hi,

I wrote a simple config file for the stimulus-language-server (https://github.com/marcoroth/stimulus-lsp).

### DIFF
--- a/lua/lspconfig/server_configurations/stimulus_ls.lua
+++ b/lua/lspconfig/server_configurations/stimulus_ls.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'stimulus-language-server', '--stdio' },
-    filetypes = { 'html', 'ruby', 'eruby', 'haml', 'slim', 'php' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/stimulus_ls.lua
```diff
@@ -0,0 +1,26 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'stimulus-language-server', '--stdio' },
+    filetypes = { 'html', 'ruby', 'eruby', 'haml', 'slim', 'php' },
```
glepnir, line 6: > This Language Server works for HTML and ERB files

I just saw this on this server README. so why there have many filetypes. related ?

marcoroth: > I just saw this on this server README. so why there have many filetypes. related ?

I think we would be fine if we just added these `filetypes` for now:

```lua
filetypes = { 'html', 'ruby', 'eruby', 'blade', 'php' },
```

PR #2873, opened 2023-10-25
Better config for root directory in typst lsp

Feel-ix-343:
---
name: Pull Request
about: Submit a pull request
title: 'Support the current working directory for root directory of typst lsp instead of only configured a configured git repository'
---


Feel-ix-343:
It is unintuitive that somebody would need to set up a git repo to get this lsp to work. I, for example, was trying to test out a .typ file and I did not realize it needed a git repo for quite a while.

glepnir:
please read contribute first

PR #2885, opened 2023-11-10
fix: util.root_pattern prioritises pattern order

### LINKED ISSUE #2612: `csharp_ls` default config finds incorrect root directory
### Description

The default config for `csharp_ls` does not identify the correct root directory. Consider the following example filestructure.

```
/MyProject
    MyProject.sln
    /MyProject.Program
        MyProject.Program.csproj
        Program.cs
    /MyProject.Library
        MyProject.Library.csproj
        Class1.cs
```

If you open `/MyProject/MyProject.Program/Program.cs` then the language server identifies the root_dir as `/MyProject/MyProject.Program`. This is not correct. Logically this is one project (`MyProject` as indicated by `MyProject.sln` solution file), with two subprojects (`MyProject.Program` and `MyProject.Library`) which each have their respecive project files (`MyProject.Program.csproj` and ` MyProject.Library.csproj`). Ideally, you would want one language server running for the entire solution.

Here is the relevant line of the default config for `csharp_ls`:

https://github.com/neovim/nvim-lspconfig/blob/df58d91c9351a9dc5be6cf8d54f49ab0d9a64e73/lua/lspconfig/server_configurations/csharp_ls.lua#L6

What is happening here is that `util.root_pattern` is visiting each ancestor directory (including the current directory), checking if *any* of the specified patterns match. The `.csproj` file is discovered before the `.sln` file, and the wrong root directory gets set.

I managed to get around this by overriding the `root_dir` field like this

```diff
lspconfig.csharp_ls.setup({
+    root_dir = function(startpath)
+        return lspconfig.util.root_pattern("*.sln")(startpath)
+            or lspconfig.util.root_pattern("*.csproj")(startpath)
+            or lspconfig.util.root_pattern("*.fsproj")(startpath)
+            or lspconfig.util.root_pattern(".git")(startpath)
+    end,
    on_attach = on_attach,
    capabilities = capabilities,
})
```

This searches ancestor directories for any `.sln` file, and if none are found then tries the same thing looking for a `.csproj` file, and so on.

Others are also experiencing this, and there is an open issue in the `csharp_ls` repo: https://github.com/razzmatazz/csharp-language-server/issues/62

I see three possible ways to remedy the situation:

1. In `util.root_pattern`, do `util.search_ancestors` once for each supplied pattern, instead of once with all patterns (as it does currently). This is less efficient, as you have to explore ancestors at most N times for N supplied patterns. This is a bit of an edge case so maybe not the best idea to negatively impact performance for the unaffected language servers.
2.  Update the `csharp_ls` default config with something like the above.
3. Keep the `csharp_ls` defaults the same, but add a note suggesting the above. 

I'm happy to make an attempt at fixing it.

### Neovim version

NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3

### Nvim-lspconfig version

df58d91c9351a9dc5be6cf8d54f49ab0d9a64e73

macOS 13.3

### Affected language servers

csharp_ls

### Steps to reproduce

```
mkdir MyProject && cd MyProject
dotnet new console -o MyProject.Program
dotnet new classlib -o MyProject.Library
dotnet new sln
dotnet sln add MyProject.Program
dotnet sln add MyProject.Library
dotnet tool install --global csharp-ls
nvim -u minimal_init.lua MyProject.Program/Program.cs
```

Run `:LspInfo` and see that root dir is set to `MyProject/MyProject.Program`

### Actual behavior

`csharp_ls` finds the `csproj` file before the `sln` file, and the incorrect `root_dir` is set for the LSP.

### Expected behavior

If a `sln` file is present, this should be preferred over any `csproj` file. Only use `csproj` file if no `sln` file exists in the directory's ancestors.

### Minimal config

```Lua
local on_windows = vim.loop.os_uname().version:match("Windows")

	local path_sep = on_windows and "\\" or "/"
	return result
end

vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local lspconfig_path = join_paths(package_root, "test", "start", "nvim-lspconfig")

	vim.fn.system({ "git", "clone", "https://github.com/neovim/nvim-lspconfig", lspconfig_path })
end

vim.lsp.set_log_level("trace")
require("vim.lsp.log").set_format_func(vim.inspect)
local nvim_lsp = require("lspconfig")
	end

	buf_set_option("omnifunc", "v:lua.vim.lsp.omnifunc")

	-- Mappings.
	local opts = { buffer = bufnr, noremap = true, silent = true }
	vim.keymap.set("n", "gD", vim.lsp.buf.declaration, opts)
	vim.keymap.set("n", "gd", vim.lsp.buf.definition, opts)
	vim.keymap.set("n", "K", vim.lsp.buf.hover, opts)
	vim.keymap.set("n", "gi", vim.lsp.buf.implementation, opts)
	vim.keymap.set("n", "<C-k>", vim.lsp.buf.signature_help, opts)
	vim.keymap.set("n", "<space>wa", vim.lsp.buf.add_workspace_folder, opts)
	vim.keymap.set("n", "<space>wr", vim.lsp.buf.remove_workspace_folder, opts)
	vim.keymap.set("n", "<space>wl", function()
	end, opts)
	vim.keymap.set("n", "<space>D", vim.lsp.buf.type_definition, opts)
	vim.keymap.set("n", "<space>rn", vim.lsp.buf.rename, opts)
	vim.keymap.set("n", "gr", vim.lsp.buf.references, opts)
	vim.keymap.set("n", "<space>e", vim.diagnostic.open_float, opts)
	vim.keymap.set("n", "[d", vim.diagnostic.goto_prev, opts)
	vim.keymap.set("n", "]d", vim.diagnostic.goto_next, opts)
	vim.keymap.set("n", "<space>q", vim.diagnostic.setloclist, opts)
end

local name = "csharp_ls"
if not name then
	print("You have not defined a server name, please edit minimal_init.lua")
end
	print([[You have not defined a server default cmd for a server
    that requires it please edit minimal_init.lua]])
end

nvim_lsp[name].setup({
	on_attach = on_attach,
})

print([[You can find your log at $HOME/.cache/nvim/lsp.log. Please paste in a github issue under a details tag as described in the issue template.]])
```

### LSP log

https://gist.github.com/emilioziniades/2bfff850f2c22f143c3864b17fe8df86

emilioziniades:
Because of the way `util.root_dir` is defined, `.csproj` files are found before `.sln` files and the incorrect root directory is set. This prevents go-to-definition between multiple projects in a single solution and causes a few other issues. See #2612 for a more in-depth explanation.

This updates the default configuration to avoid this footgun.

EDIT:

The changes now cover the above issue, but does so in a more general way by updating the `util.root_pattern` function.

closes #2612 
related to razzmatazz/csharp-language-server#57

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -250,18 +250,21 @@ end
 
 function M.root_pattern(...)
   local patterns = vim.tbl_flatten { ... }
-  local function matcher(path)
-    for _, pattern in ipairs(patterns) do
-      for _, p in ipairs(vim.fn.glob(M.path.join(M.path.escape_wildcards(path), pattern), true, true)) do
-        if M.path.exists(p) then
-          return path
-        end
-      end
-    end
-  end
   return function(startpath)
     startpath = M.strip_archive_subpath(startpath)
-    return M.search_ancestors(startpath, matcher)
+    for _, pattern in ipairs(patterns) do
+      local match = M.search_ancestors(startpath, function(path)
+        for _, p in ipairs(vim.fn.glob(M.path.join(M.path.escape_wildcards(path), pattern), true, true)) do
+          if M.path.exists(p) then
+            return path
+          end
+        end
+      end)
+
+      if match ~= nil then
+        return match
+      end
+    end
   end
 end
 
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -371,7 +371,8 @@ below returns a function that takes as its argument the current buffer path.
 - `util.root_pattern`: function which takes multiple arguments, each 
   corresponding to a different root pattern against which the contents of the
   current directory are matched using |vim.fn.glob()| while traversing up the
-  filesystem.
+  filesystem. Parent directories are traversed once per pattern, in the order
+  the patterns are specified.
 >
    root_dir = util.root_pattern('pyproject.toml', 'requirements.txt')
 <
--- a/test/lspconfig_spec.lua
+++ b/test/lspconfig_spec.lua
@@ -136,7 +136,7 @@ describe('lspconfig', function()
         -- change the working directory to test directory
         vim.api.nvim_command 'cd ./test/test_dir/a'
         local cwd = vim.fn.getcwd()
-        eq(true, cwd == lspconfig.util.root_pattern { 'root_marker.txt', 'a_marker.txt' }(cwd))
+        eq(true, cwd == lspconfig.util.root_pattern { 'a_marker.txt', 'root_marker.txt' }(cwd))
       end)
 
       it('resolves to root_marker.txt', function()


lua/lspconfig/server_configurations/csharp_ls.lua
```diff
@@ -3,7 +3,12 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'csharp-ls' },
-    root_dir = util.root_pattern('*.sln', '*.csproj', '*.fsproj', '.git'),
+    root_dir = function(startpath)
+      return util.root_pattern '*.sln'(startpath)
+        or util.root_pattern '*.csproj'(startpath)
+        or util.root_pattern '*.fsproj'(startpath)
+        or util.root_pattern '.git'(startpath)
+    end,
```
justinmk, line 11: why can't the fix be done in `util.root_pattern` instead of special-cased for csharp?

emilioziniades: Yeah there's no reason why not. I agree that having the special case is weird. Will put something together and we can take a look.

emilioziniades:
@justinmk would you mind taking another look? I've updated `util.root_pattern` as discussed.

Previously, the function did this:
```
for each parent directory:
    for each pattern:
        if match:
            return match
```

I have switched the two loops around so that it now does the following:
```
for each pattern:
    for each parent directory:
        if match:
            return match
```

There is a slight performance hit because instead of traversing parent directories once, it now traverses parent directories `n` times for `n` patterns.

On the plus side though this resolves the issue not only for `csharp_ls`, but also for a whole bunch of other language servers which use a similar pattern of doing `return util.root_pattern("pattern1")(path) or util.root_pattern("pattern2")(path) or ...`. Some examples:

- [omnisharp](https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/omnisharp.lua#L41-L49)
- [tsserver](https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/tsserver.lua#L15-L18)
- [cairo_ls](https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/cairo_ls.lua#L11-L14)
- [hls](https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/hls.lua#L8-L12)

justinmk (approved):
Thanks for the clear explanation!

- should mention "in order" in the doc for root_pattern https://github.com/neovim/nvim-lspconfig/blob/0a0682d4646a6869b85a4e4d0e30da5ef8b11f66/doc/lspconfig.txt#L371-L374
- @glepnir any objections? this might accidentally change behavior of existing configs, but they can be updated.

After this change if a config has root_pattern `{'package.json', '.git'}` then parents are traversed until `package.json` is found, a random `.git` won't incorrectly define the workspace. If someone wants `.git` to be higher precedence, then the root_pattern should be changed to `{'.git', 'package.json'}`

justinmk:
otoh , does make it impossible to define root markers with "equal precedence"? e.g. https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/tsserver.lua#L15-L18

is that important?

emilioziniades:
Force pushed to update spelling of "prioritises" in commit message, and updated `util.root_pattern` docs.


> otoh , does make it impossible to define root markers with "equal precedence"? e.g.

Yes it does.

> is that important?

Not sure. Struggling to think of a situation where equal precedence would be desirable/useful.

_Case A_
```
directory/
    marker_a
    marker_b
```
_Case B_
```
directory/
    marker_a
    subdirectory/
        marker_b
```
_Case C_
```
directory/
    marker_b
    subdirectory/
        marker_a
```

In case A, precedence doesn't matter, any order of markers will identify the same root directory.

The only situation where equal precedence would be useful is if you _always_ wanted to identify `directory/subdirectory` as the root, regardless of whether `marker_a` or `marker_b` is present. i.e. you have one project that is like case B, and another project using the same language server that is like case C. Does this actually ever happen?

lua/lspconfig/util.lua
```diff
@@ -250,19 +250,22 @@ end
...
   local patterns = vim.tbl_flatten { ... }
-  local function matcher(path)
+  return function(startpath)
+    startpath = M.strip_archive_subpath(startpath)
     for _, pattern in ipairs(patterns) do
-      for _, p in ipairs(vim.fn.glob(M.path.join(M.path.escape_wildcards(path), pattern), true, true)) do
-        if M.path.exists(p) then
-          return path
+      local match = M.search_ancestors(startpath, function(path)
+        for _, p in ipairs(vim.fn.glob(M.path.join(M.path.escape_wildcards(path), pattern), true, true)) do
```
glepnir, line 257: it's better use vim.fs module

glepnir:
thought after 0.10 we can use `vim.iter` with `find` .

justinmk:
> Force pushed to update spelling of "prioritises" in commit message

Did CI complain about that? 😆 

> after 0.10 we can use `vim.iter` with `find` .

We can save the `vim.fs` refactoring until 0.10 then?

Since we have carefully considered the implications and the behavior of this PR seems to match user intention, I think we should go ahead and try this.

In a followup PR we should cleanup the existing configs since the `or` cases are now redundant.

glepnir:
nope.  ci has been fixed on master. just need rebase :)

emilioziniades:
> Did CI complain about that? 😆

lol no it was just bothering me  :stuck_out_tongue_closed_eyes: 

> nope. ci has been fixed on master. just need rebase :)

Rebased master and CI is passing now.

tmillr:
> otoh , does make it impossible to define root markers with "equal precedence"? e.g.
> 
> https://github.com/neovim/nvim-lspconfig/blob/90a28fd7637b66e055af62387ecee06f7cbd3173/lua/lspconfig/server_configurations/tsserver.lua#L15-L18
> 
> is that important?

Yes it is important 🫤.

justinmk:
> > is that important?
> 
> Yes it is important 🫤.

@tmillr does https://github.com/neovim/neovim/pull/33485 match the behavior you want?

Issue #2733, opened 2023-07-25
9a2cc569c88662fa41d414bdb65b13ea72349f86 (#2673) Breaks golsp


EJammy:
### Description

`mod_cache` does not seem to be set correctly with the performance improvement using `async_run_command`. Reverting the changes fix this.

### Neovim version

NVIM v0.8.3
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az171-224

Features: -acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM\sysinit.vim"
  fall-back for $VIM: "C:/Program Files (x86)/nvim/share/nvim"

Run :checkhealth for more info

### Operating system and version

Windows 10

### Affected language servers

gopls

### Steps to reproduce

1. `go mod init example/hello`
2. `nvim hello.go`

### Actual behavior

```
[lspconfig] unhandled error: ...-lspconfig/lua/lspconfig/server_configurations/gopls.lua:16: attempt to get length of upval
ue 'mod_cache' (a nil value)
```

### Minimal config

```Lua
.
```

### LSP log

.

glepnir:
output of `go env GOMODCACHE`

EJammy:
```
# go env GOMODCACHE
C:\Users\connor.chang\go\pkg\mod
```

glepnir:
problem is why there does not get this value ..although we can add a nil check but it should be have a value.

EJammy:
In here:
https://github.com/neovim/nvim-lspconfig/blob/654dd9e95c49f64726b8751d75cbe147ec626875/lua/lspconfig/server_configurations/gopls.lua#L11-L14

It looks like `result` is nil. I'm not sure how `async_run_command` works, but somehow it's failing. I think we should change the code to fall back to using `vim.fn.system` for now.
```lua
        local result = util.async_run_command 'go env GOMODCACHE' 
        if result and result[1] then
          mod_cache = vim.trim(result[1])
        else
          mod_cache = vim.fn.system 'go env GOMODCACHE'
        end
```
I don't think it make sense to have a code path where mod_cache is nil anyway

glepnir:
`vim.fn.system` is sync. and problem not is `nil` check.  it's only in windows async get command output failed. in *nix it works fine !

EJammy:
[...]

PR #2896, opened 2023-11-14
fix: async.run_command: report error exit code and stderr

merrilymeredith:
I ran into an error like #2733 when opening a project where `go` was unable to start, and noticed `async.run_command` didn't report an error at all.

This adds a check for the exit code of the command, reporting that back to the user with stderr output, and changes the existing error check:  jobid -1 is returned when the command _can't start at all_, and there shouldn't be any stderr output at that point either.  Normally, you could get further detail "errno" style, but i don't see that this is exposed by the job api.

Example after fix:

It's possible merging this will add some insight into the ways people are still running into #2733 - though it doesn't mess with the actual place where mod_cache is nil and emitting _that_ message.

### DIFF
--- a/lua/lspconfig/async.lua
+++ b/lua/lspconfig/async.lua
@@ -16,6 +16,8 @@ function M.run_command(cmd)
 
   local stdout = {}
   local stderr = {}
+  local exit_code = nil
+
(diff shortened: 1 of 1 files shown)


glepnir:
make sense thanks.

Issue #2897, opened 2023-11-16
bug: jdtls $/progress does not display the messages correctly


s1n7ax:
### Description

At the moment, `1.29.0` (released on 2023-10-26 13:40) is the latest version of jdtls. Right now in lspconfig, `language/status` and `$/progress` both handled by the same `on_language_status` function.

https://github.com/neovim/nvim-lspconfig/blob/d5d7412ff267b92a11a94e6559d5507c43670a52/lua/lspconfig/server_configurations/jdtls.lua#L122-L123

However, the data sent from jdtls are different.

**language/status** 2nd parameter

```lua
{
  message = "0% Starting Java Language Server",
  type = "Starting"
}
```

**$/progress** 2nd parameter

```lua
{
  token = "38f85eca-ebcf-4c4a-8e83-7b27fd530f99",
  value = {
    kind = "report",
    message = "Initialize Workspace - 30% ",
    percentage = 30
  }
}
```

Due to this progress handler prints just `nil`.

Progress handler was added a fix to https://github.com/neovim/nvim-lspconfig/pull/2153 in Nov 21, 2022

I tested the `vim.lsp.handlers['$/progress']` handler and it works fine right now. So it's only matter of removing the handler in the config. I would like any feedback & ideas regarding removing `$/progress` handler from the jdtls config.

### Neovim version

NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1693350652

### Nvim-lspconfig version

36e4ffe

### Operating system and version

NixOS 23

### Affected language servers

jdtls

### Steps to reproduce

1. `lspconfig.jdtls.setup({})`
2. check the progress messages

### Actual behavior

`nil` being printed

### Expected behavior

progress message should be displayed correctly

### Minimal config

```Lua
-
```

### LSP log

-

justinmk:
> Progress handler was added a fix to #2153
>
> ... I tested the `vim.lsp.handlers['$/progress']` handler and it works fine right now. So it's only matter of removing the handler in the config. I would like any feedback & ideas regarding removing `$/progress` handler from the jdtls config.

Please do. This repo should avoid custom things as much as possible.

le4ker:
I don't get any formatting or diagnostics messages and progress from my LSPs in my [statusline](https://github.com/NvChad/ui/blob/v2.0/lua/nvchad/statusline/default.lua#L136). 

@s1n7ax @justinmk can this be related? 

[...]

PR #2898, opened 2023-11-16
fix: remove outdated progress message handler

s1n7ax:
[ISSUE]
previously the message type of progress event was

{
  message = "0% Starting Java Language Server",
  type = "Starting"
}

however in the latest (1.29.0) use the proper data structure like below. so the custom handler is no longer needed.

{
  token = string,
  value = {
    kind = "report",
    message = "Initialize Workspace - 30% ",
    percentage = 30
  }
}

By removing the custom handler in jdtls config, vim.lsp.handlers['$/progress'] will be used.


s1n7ax:
Issue: https://github.com/neovim/nvim-lspconfig/issues/2897

Before:

screenshot 23-11-16 22:09:33

After:

PR #2899, opened 2023-11-16
fix: remove outdated jdtls progress message handler (#2897)

s1n7ax:
progress message

**before**
screenshot 23-11-16 22:09:33

**After**
screenshot 23-11-16 22:19:08


s1n7ax:
May be that screenshots are misleading. in the bottom right, noice.nvim displays the progress status by overriding `vim.lsp.handlers['$/progress']`. Before, jdtls had defined a handler withing the config so it never reached noice.nvim. `nil` in the first screenshot was printed by the jdtls config `$/progress` handler.

PR #2905, opened 2023-11-18
fix(rust-analyzer): don't run cargo if Cargo.toml is not found

kitterion:
After commit https://github.com/neovim/nvim-lspconfig/commit/d5d7412ff267b92a11a94e6559d5507c43670a52 I started getting the message
```
[lspconfig] cmd failed with code 101: table: 0x0104f90fe0
error: could not find `Cargo.toml` in `<path>` or any parent directory
Press ENTER or type command to continue
```
every time I open a rust file in a project that doesn't use cargo. For some reason `cargo` is launched even if `Cargo.toml` is not found which seems wrong, so I moved all the cargo logic inside `if cargo_crate_dir ~= nil`.
rust-analyzer supports non-cargo projects via `rust-project.json` that nvim-lspconfig also supports, and it's what I use.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -50,19 +50,26 @@ return {
       end
 
       local cargo_crate_dir = util.root_pattern 'Cargo.toml'(fname)
-      local cmd = { 'cargo', 'metadata', '--no-deps', '--format-version', '1' }
-      if cargo_crate_dir ~= nil then
-        cmd[#cmd + 1] = '--manifest-path'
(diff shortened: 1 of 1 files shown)


glepnir:
i don't think this is correct. This command is still available when cargo_crate_dir does not exist. Try running this command `cargo metadata --no-deps --format-version=1 `

kitterion:
I get the exact same error because there's no `Cargo.toml`
```
$ cargo metadata --no-deps --format-version=1
error: could not find `Cargo.toml` in `<path>` or any parent directory
```

glepnir:
okay.. command must run with cargo.toml. thanks.

PR #2907, opened 2023-11-19
feat: add support for slangd

theHamsta:
`slangd` is the language server for [slang](https://github.com/shader-slang/slang), a shader language that extends HLSL with additional features.

The filetype `shaderslang` is a invention of myself, but I plan to also use it for a tree-sitter parser https://github.com/nvim-treesitter/nvim-treesitter/pull/5659 and a TBD PR to vim for filetype detection. `shaderslang` is used because Vim/Neovim already has support for a filetype called `slang` which I assume is for https://en.wikipedia.org/wiki/S-Lang with extension `*.sl` while `shaderslang` uses `*.slang`. The language server also works with regular `hlsl` https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl / https://de.wikipedia.org/wiki/High_Level_Shading_Language (which would also require a PR to be detected by Vim/Neovim). I'm not aware of any Vim plugins for slang.

Please let me know when this PR should remain in draft status until the Vim PR is completed.

### DIFF
--- a/lua/lspconfig/server_configurations/slangd.lua
+++ b/lua/lspconfig/server_configurations/slangd.lua
@@ -1,0 +1,44 @@
+local util = require 'lspconfig.util'
+local bin_name = 'slangd'
+
+if vim.fn.has 'win32' == 1 then
+  bin_name = 'slangd.exe'
+end
+
+return {
+  default_config = {
+    cmd = { bin_name },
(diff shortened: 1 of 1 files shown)


glepnir:
There is no order actually. There are many custom file types  lsp here. just make sure filetype register (vim.filetype.add) before lspconfig load .

PR #2911, opened 2023-11-22
fix(elixirls): smarter detection of root_dir for umbrella apps

giddie:
Should fix issues identified following merge of https://github.com/neovim/nvim-lspconfig/pull/2910

### DIFF
--- a/lua/lspconfig/server_configurations/elixirls.lua
+++ b/lua/lspconfig/server_configurations/elixirls.lua
@@ -1,11 +1,14 @@
-local util = require 'lspconfig.util'
-
 return {
   default_config = {
     filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
     root_dir = function(fname)
-      return util.root_pattern 'mix.exs'(fname) or util.find_git_ancestor(fname) or vim.loop.os_homedir()
+      local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
+      local child_or_root_path, maybe_umbrella_path = unpack(matches)
+      local root_dir = vim.fs.dirname(maybe_umbrella_path or child_or_root_path)
+
+      return root_dir
     end,
+    single_file_support = true,
   },
   docs = {
     description = [[
@@ -31,9 +34,11 @@ require'lspconfig'.elixirls.setup{
     ...
 }
 ```
+
+'root_dir' is chosen like this: if two or more directories containing `mix.exs` were found when searching directories upward, the second one (higher up) is chosen, with the assumption that it is the root of an umbrella app. Otherwise the directory containing the single mix.exs that was found is chosen.
 ]],
     default_config = {
-      root_dir = [[util.find_git_ancestor(fname) or util.root_pattern 'mix.exs'(fname) or vim.loop.os_homedir()]],
+      root_dir = '{{see description above}}',
     },
   },
 }


lua/lspconfig/server_configurations/elixirls.lua
```diff
@@ -4,7 +4,15 @@ return {
   default_config = {
     filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
     root_dir = function(fname)
-      return util.root_pattern 'mix.exs'(fname) or util.find_git_ancestor(fname) or vim.loop.os_homedir()
+      if not fname or fname == "" then
```
georgeguimaraes, line 7: I think you don't need this check at all since we can trust there's a fname coming in.

lua/lspconfig/server_configurations/elixirls.lua
```diff
@@ -4,7 +4,15 @@ return {
...
+        fname = vim.fn.getcwd()
+      end
+
+      local matches = vim.fs.find({ "mix.exs" }, { upward = true, limit = 2, path = fname })
+      local child_or_root_path, maybe_umbrella_path = unpack(matches)
+      local root_dir = vim.fs.dirname(maybe_umbrella_path or child_or_root_path)
```
georgeguimaraes, line 13: maybe this works?

`local root_dir = util.root_pattern 'mix.exs'(maybe_umbrella_path) or util.root_pattern 'mix.exs'(child_or_root_path)`

a bit redundant, but this way we're not using dirname and instead rely on the utils from lspconfig.

georgeguimaraes (commented):
Gave an idea to circumvent the dirname restriction

lua/lspconfig/server_configurations/elixirls.lua
giddie: My main issue is that I don't understand what the constraints are here. Why can't we use `dirname`? Does this mean we can't use `vim.fs.find` either? Can @glepnir offer any guidance?

giddie: Almost. The issue is that we need to return a directory path, right? So we need `dirname`. This was my attempt:

```lua
local matches = vim.fs.find({ "mix.exs" }, { upward = true, limit = 2, path = fname })
local child_or_root_path, maybe_umbrella_path = unpack(matches)
local root_dir = vim.fs.dirname(maybe_umbrella_path or child_or_root_path)
return root_dir
```

giddie: But we need to return a directory path, right? Hence the use of `dirname`. Again, for me the question is really what functions we can and can't use, and why? What is different about this "single file mode"?

glepnir: no different. when single file mode enable it will use current dir as root.

giddie: So what's with the CI complaining `dirname`? The CONTRIBUTING document says: "Do not add vim.fn.cwd or util.path.dirname in root_dir.", but doesn't offer a reason _why_, so I don't really have a way of knowing what exactly I need to avoid, or what other function or behaviour could be problematic.

glepnir: just set `single_file_mode = true` and remove `cwd`

giddie: I'm sorry, where should I be setting this? A search in `nvim-lspconfig` reveals zero instances of `single_file_mode`, so I'm all out of clues.

BTW, the code in this PR works just fine for me. As far as I'm concerned, and unless I can understand the constraints better, the problem seems to be the CI...?

lua/lspconfig/server_configurations/elixirls.lua
```diff
@@ -4,7 +4,15 @@ return {
...
+
+      return root_dir
     end,
   },
```
glepnir, line 17:
```suggestion
    end,
    single_file_support = true,
  },
```

lua/lspconfig/server_configurations/elixirls.lua
giddie: Your [suggestion below](https://github.com/neovim/nvim-lspconfig/pull/2911#discussion_r1422266681) led me to find `lspconfig-single-file-support` in the `lspconfig` help. IIUC, the idea here is that if `root_dir` returns `nil`, and `single_file_support` is on, then this mode will be used?

I don't really get how this helps with the issue of `dirname` being rejected by the CI, though :confused:

glepnir: yup dirname ci can ignore just fix lint .

giddie: Oh OK; great. Then I think we're good to go :partying_face:

lua/lspconfig/server_configurations/elixirls.lua
glepnir: emm you forget here

giddie: Done :+1:

georgeguimaraes:
Great work everyone!

Issue #2915, opened 2023-11-27
util.find_git_ancestor Should Optionally Ignore Worktrees


vantaboard:
### Requested feature

There are some cases where I am using submodules, and I don't want those to get caught up as the root of the directory especially because if I am using yarn workspaces this is incredibly misleading and leads to LSP misconfigurations.

PR #2916, opened 2023-11-27
feat(util.find_git_ancestor): added exclude flag (#2915)

vantaboard:
Adds the enhancement suggested in https://github.com/neovim/nvim-lspconfig/issues/2915.


justinmk:
Thank you for the PR, but this codebase needs to become simpler, not more complex.

If this kind of customization is needed then the user should create a custom config. We aren't going to add unlimited features here, this repo is just a collection of basic configs.

PR #2919, opened 2023-11-28
feat(hydra-lsp): add hydra-lsp

liepieshov:
Add [hydra language server](https://github.com/Retsediv/hydra-lsp) for [Hydra](https://hydra.cc/) yaml config files.

Really popular in Python research community for configuration of experiments.


glepnir:
doesn't support single file mode ?

liepieshov:
> doesn't support single file mode ?

Yes, you are right. Added single_file_support

PR #2926, opened 2023-12-09
feat: add cds-lsp

stijnme:
Basic LSP config for @sap/cds-lsp

### DIFF
--- a/lua/lspconfig/server_configurations/cds_lsp.lua
+++ b/lua/lspconfig/server_configurations/cds_lsp.lua
@@ -1,11 +1,17 @@
 local util = require 'lspconfig.util'
 
+local root_files = {
+  'package.json',
+  'db',
+  'srv',
+}
+
 return {
   default_config = {
     cmd = { 'cds-lsp', '--stdio' },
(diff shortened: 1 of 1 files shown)


nbys:
It might be due to my moderate experience with Neovim, but I struggled a bit to get this language server working. :LspInfo indicated that cds_lsp was not working for my CDS views files. I investigated the path, tried different versions of Node.js, and experimented with various configurations using lsp-config. However, the problem was much simpler: Neovim was not detecting the filetype cds.

You can check the filetype with :set ft?. If it works, it shows filetype=cds.

To fix this, I added these lines to my init.lua file. Now, cds_lsp is correctly attached:
```sh
vim.filetype.add({
    extension = {
        cds = "cds",
    },
})
```

justinmk:
> To fix this, I added these lines to my init.lua file. Now, cds_lsp is correctly attached:
> 
> ```shell
> vim.filetype.add({
>     extension = {
>         cds = "cds",
>     },
> })
> ```

Because nvim/vim doesn't have "cfs" filetype detection by default.

Issue #2758, opened 2023-08-09
Show capabilities summary for LSP in `:LspInfo`


jeanlucthumm:
### Requested feature

A common pattern of users' `on_attach` functions is to gate keymaps and functionality based on the server's capabilities, e.g.:

```lua
if client.server_capabilities.documentHighlightProvider then
  -- ...
end
```
As such, it'd be nice to have an overview of what the currently attached LSP server is capable of.

MariaSolOs:
Given the resolution of #2932, should we just close this?

Issue #2412, opened 2023-01-20
Show current LSP config in :LspInfo floating window


ellisonleao:
### Requested feature

Hey people, i was wondering if it would be nice to show the contents of the loaded lsp server configs in `:LspInfo` output. I know we already have the docs for the default config, but sometimes I need to use the output from `vim.api.get_active_clients()`  to check what's actually my current config. 

I was wondering if we could add something like a hidden accordion with the full config where you can just toggle it if needed

![screenshot-github com-2023 01 20-16_25_09](https://user-images.githubusercontent.com/178641/213790262-6696106e-0848-402e-9e35-612fe1bfd74e.png)

glepnir:
currently the `checkhealth` command show some lsp relate like server id root .   and catch some custom pattern will add some codes.  I don't think @justinmk  will allow these codes.

PR #2932, opened 2023-12-17
feat: press to show the client object

thenbe:
https://github.com/neovim/nvim-lspconfig/assets/33713262/2c3e624a-33f2-41c6-9604-0eb88db186b4

This PR adds a keymap that prints the [LSP client object](https://neovim.io/doc/user/lsp.html#vim.lsp.client) of the client under the cursor in the LspInfo window. I find that peeking under the hood of a running LSP client helps when debugging LSP issues. For new users especially, as doing this manually requires pulling together a handful of concepts (inspecting a table, printing to new buffer, etc).

- The output is printed to a new buffer for easy search/yank/diff/etc.
- While the output is not valid lua syntax, it's close enough that calling `set syntax=lua` does make it easier to do a quick eye scan.
- I have no strong opinion on what key to assign this to. I currently have it set to `K`.
- `K` doesn't need to be pressed on a specific column, anywhere on the same line will do.

### DIFF
--- a/lua/lspconfig/ui/lspinfo.lua
+++ b/lua/lspconfig/ui/lspinfo.lua
@@ -184,6 +184,33 @@ local function make_client_info(client, fname)
   return lines
 end
 
+---Returns the id of the client on the current line
+---@return number|nil
+local function get_client_id_under_cursor()
+  local line = vim.api.nvim_get_current_line()
+  local id = string.match(line, 'id:%s*(%d+)')
+  if id then
+    return tonumber(id)
+  else
+    return nil
+  end
+end
+
+---Prints the LSP client object to a new buffer
+---@param filter table
+local show_client_config = function(filter)
+  local output = vim.lsp.get_clients(filter)
+  vim.cmd [[
+    execute 'vsplit | enew'
+    setlocal buftype=nofile
+    setlocal bufhidden=hide
+    setlocal noswapfile
+    setlocal wrap
+    set syntax=lua
+]]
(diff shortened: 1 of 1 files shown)


glepnir:
thanks for pr. but lspconfig just a config collection. we don't want add more in this repo. and there also have check health interface. anyway thanks agian.

thenbe:
Now that you mention it, I don't think I've ever used `:checkhealth vim.lsp` before. My go to has always been `:LspInfo`. I agree that, ideally, we should decide on using only one and appreciate that that is what you're trying to do here. This is anecdotal, but from browsing the /r/neovim subreddit I think most users rely on `:LspInfo` at this time?

While I understand that you don't want to increase the maintenance burden of this repo, I'd argue that this PR reduces it. It invites curious new users to peek into the LSP client config. It is those types of users who then become likely to identify issues and contribute bug fixes and improvements to the configs hosted in this repo, reducing the overall maintenance effort required by any single maintainer.

It's also worth mentioning that this has been requested at least once (#2412) before this year, by a plugin author no less (as opposed to a newer user). It would also close #2758.

That was only _my_ opinion on the matter. And while I am disappointed, I respect that you might have different views because you see things differently (and you interact with users _a lot_ more than I do).

glepnir:
> That was only my opinion on the matter. And while I am disappointed

first I absolutely understand how you feel. I had the same experience before.

 but i think it can be very simply to do like `:echo v:lua.vim.lsp.get_clients() | put=execute('messages')` just wrap in a command.

cc @justinmk

PR #2936, opened 2023-12-19
feat(denols): follow the latest default value of the vscode extension

### LINKED ISSUE #2934: feat(denols): disable import completion from some registries in default
### Language server

denols

### Requested feature

same as the vscode extension:
https://github.com/denoland/vscode_deno/blob/3d20875396a779fc7047ac826c97775571d997e5/package.json#L430-L440

```diff
           "default": {
-            "https://deno.land": true,
-            "https://x.nest.land": true,
-            "https://crux.land": true
+            "https://deno.land": true
           },
```

This option has set in #2793 to according to the vscode extension.

_No response_

uga-rosa:
Close #2934

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -27,9 +27,9 @@ local function virtual_text_document_handler(uri, res, client)
   end
 
   vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, lines)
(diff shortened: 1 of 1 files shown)


uga-rosa:
I will also note why I agree with this follow-up.
In my environment, when `https://x.nest.land` is enabled, it can take several minutes for the server to start up, so I don't think it is a good idea to have this enabled by default.

PR #2944, opened 2023-12-20
fix(zls): use workspace zls.json if available

lawrence-laz:
Language server config is read from `zls.json`. By default common global and user config directories are searched. However, this omits per-project configuration, which can be available in the workspace dir.

We already look for `zls.json` to determine root dir, but we don't use it when starting the language server.

This change appends `cmd` with `--config-path "zls.json"`, which uses the file when it's available and fall-backs to default logic when file is not available.

### DIFF
--- a/lua/lspconfig/configs/zls.lua
+++ b/lua/lspconfig/configs/zls.lua
@@ -3,6 +3,11 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'zls' },
+    on_new_config = function(new_config, new_root_dir)
+      if vim.fn.filereadable(vim.fs.joinpath(new_root_dir, 'zls.json')) ~= 0 then
+        new_config.cmd = { 'zls', '--config-path', 'zls.json' }
+      end
+    end,
     filetypes = { 'zig', 'zir' },
     root_dir = util.root_pattern('zls.json', 'build.zig', '.git'),
     single_file_support = true,
@@ -14,6 +19,7 @@ https://github.com/zigtools/zls
 Zig LSP implementation + Zig Language Server
         ]],
     default_config = {
+      cmd = { 'zls' },
       root_dir = [[util.root_pattern("zls.json", "build.zig", ".git")]],
     },
   },


lua/lspconfig/server_configurations/zls.lua
```diff
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    cmd = { 'zls' },
+    cmd = { 'zls', '--config-path', '"zls.json"' },
```
glepnir, line 5: why we don't check zls.json first build a cmd

lawrence-laz: We don't really have to, since `zls` goes back to using global config if a specified config file does not exists, but we can.

Would you rather have `cmd` as a function that checks if `zls.json` exists in workspace and and then generates command based on that?

glepnir: This is fine if the server can fallback. Just wondering if it might bother some newbies.

lawrence-laz: I tried looking into how to provide a function for `default_config.cmd`, but couldn't find it, does it always has to be a table?

glepnir: nope it can be a function.

lawrence-laz: I'm struggling a bit here.
Tried providing a function instead of a table and printed the arguments, which got me:
```lua
{
  notification = <function 1>,
  on_error = <function 2>,
  on_exit = <function 3>,
  server_request = <function 4>
}
```
I then tried looking for these function names in source, but got unrelated stuff or in case of `server_request` even no results at all.
Tried looking through docs again, but they only seem to contain an example for table as well:
```
- {cmd} `list[string]`

  a list where each entry corresponds to the blankspace delimited part of
  the command that launches the server. The first entry is the binary used
  to run the language server. Additional entries are passed as arguments.

  The equivalent `cmd` for: 
 
  foo --bar baz
 
  is:
 
  {'foo', '--bar', 'baz'}
```
So not sure how to do this.
Or should I use `on_setup` hook instead?
Could you give some pointers on how this could be implemented?

glepnir:
```lua
 cmd = function() xxxxxx end
```

lua/lspconfig/server_configurations/zls.lua
```diff
@@ -14,6 +14,7 @@ https://github.com/zigtools/zls
 Zig LSP implementation + Zig Language Server
         ]],
     default_config = {
+      cmd = { 'zls', '--config-path', '"zls.json"' },
```
justinmk, line 17:
```suggestion
      cmd = { 'zls', '--config-path', 'zls.json' },
```

lua/lspconfig/server_configurations/zls.lua
justinmk, line 5:
```suggestion
    cmd = { 'zls', '--config-path', 'zls.json' },
```

justinmk:

lua/lspconfig/server_configurations/zls.lua
lawrence-laz: Added a check for `zls.json` first

lawrence-laz:
rebased

PR #2952, opened 2023-12-29
fix(ols): remove single_file_support

losinggeneration:
It doesn't seem like ols really has support for single_file_support. The ols application crashes with the following message: "ols/src/server/requests.odin(1494:28) Index 0 is out of range 0..<0\n"

This happens on workspace/didChangeConfiguration when settings are provided in the users' ols setup.

To work around this, essentially turning off single_file_support & adding *.odin as a way to detect a root_dir. This isn't very common but at least one other language server uses this method, qml_ls. However, without this, the ols doesn't really work out of the box for projects that lack a .git directory or ols.json file.

### DIFF
--- a/lua/lspconfig/server_configurations/ols.lua
+++ b/lua/lspconfig/server_configurations/ols.lua
@@ -5,7 +5,6 @@ return {
     cmd = { 'ols' },
     filetypes = { 'odin' },
     root_dir = util.root_pattern('ols.json', '.git', '*.odin'),
-    single_file_support = false,
   },
   docs = {
     description = [[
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/ols.lua
```diff
@@ -4,8 +4,8 @@ return {
   default_config = {
     cmd = { 'ols' },
     filetypes = { 'odin' },
-    root_dir = util.root_pattern('ols.json', '.git'),
-    single_file_support = true,
+    root_dir = util.root_pattern('ols.json', '.git', '*.odin'),
+    single_file_support = false,
```
glepnir, line 8: if it not support remove.

losinggeneration: I went ahead and removed `single_file_support` since it seems to default to false. :+1:

PR #2961, opened 2024-01-11
fix(docs): Add note about filetype for docker-compose

Uzaaft:
I propose adding a note about the filetype, as I just spent a few hours trying to figure out why the lsp didn't run when entering a docker-compose.yaml file. I think this would avoid anyone doing the same.


Uzaaft:
My bad, I'll fix it.

PR #2962, opened 2024-01-11
spyglassmc: Make root_dir unrequired.

amqndin:
by default, root_dir set to pack.mcmeta, and if pack.mcmeta is missing in your directory regardless whether single_file_support is true or false, the spyglass language server will throw an error and wont work. this pr fixes the problem.

this could be a wrong approach tho, ill try to find other ways to fix this later too.


amawdiin:
i see, ill try to fix that

amawdiin:
i will try to fix it

amawdiin:
to whoever requested to review - this is not done yet, i havent even started

lua/lspconfig/server_configurations/spyglassmc_language_server.lua
```diff
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'spyglassmc-language-server', '--stdio' },
     filetypes = { 'mcfunction' },
-    root_dir = util.root_pattern 'pack.mcmeta',
+    root_dir = function() return vim.loop.cwd() end,
```
glepnir, line 7: this is not correct.

amawdiin: it works though.

asmodeus812: it would be better to extend the default patterns to be more inclusive, maybe add a .git .svn or another specific to spyglass root directory marker, cwd() is a bad idea here.

PR #2969, opened 2024-01-14
chore: update dcmls config

RobertBrunhage:


glepnir:
It's better test in your local to make sure all it can works . and add more spec in default description about  why add yaml filetype .

RobertBrunhage:
> It's better test in your local to make sure all it can works.
Got feedback from creator of `DCM` that `--` was missing. Was not visible locally.

> and add more spec in default description about why add yaml filetype 
It can analyze yaml files as well for lint rules etc.

glepnir:
emm I think it's better disable yaml as  default.. because yaml has their own server ymls.

RobertBrunhage:
> emm I think it's better disable yaml as default.. because yaml has their own server ymls.

Got it, will do

PR #2977, opened 2024-01-19
feat(texlab): add cancelBuild

DimitrisDimitropoulos:
Add a command for the new workspace command of texlab for canceling the build, including the onSave ones. I think it is pretty useful. Thanks for your effort and help


lua/lspconfig/server_configurations/texlab.lua
```diff
@@ -14,6 +14,18 @@ local texlab_forward_status = vim.tbl_add_reverse_lookup {
...
 }
 
+local function buf_cancel_build()
+  local texlab_client = util.get_active_client_by_name(0, 'texlab')
+  if texlab_client then
+    texlab_client.request('workspace/executeCommand', { command = 'texlab.cancelBuild' }, function(err)
+      if err then
+        error(tostring(err))
+      end
+      print 'Build cancelled'
```
lithammer, line 24: Use `vim.notify()` instead of print.

lua/lspconfig/server_configurations/texlab.lua
```diff
@@ -14,6 +14,18 @@ local texlab_forward_status = vim.tbl_add_reverse_lookup {
   Unconfigured = 3,
 }
 
+local function buf_cancel_build()
+  local texlab_client = util.get_active_client_by_name(0, 'texlab')
+  if texlab_client then
+    texlab_client.request('workspace/executeCommand', { command = 'texlab.cancelBuild' }, function(err)
```
justinmk, line 20: isn't `workspace/executeCommand` part of the LSP spec?  should this use `:help vim.lsp.commands` ?

DimitrisDimitropoulos: Thanks for the reply and the suggestion. Reading the docs on the `vim.lsp.commands`  I too think it should be used here. However, the docs were not very helpful on how to implement it, even looking at the source code in neovim i could not get it to work. Moreover, i found another implementation with the `request`  in the following part from the es-lint [config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/eslint.lua/#L4-L33). I could really use some help regarding how to proceed, thanks in advance.

justinmk: Hoping that @glepnir can help.



> i found another implementation with the `request` in the following part from the es-lint [config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/eslint.lua/#L4-L33).

Perhaps that pattern can be lifted into `util.lua` in this repo. We need to avoid copy-pasting lots of code in all of these configs.

Issue #2663, opened 2023-06-09
Add context to LTeX filetypes


usertzuiop:
### Language server

ltex-ls

### Requested feature

LTeX supports Context by default. Vim knows the file type context and can determine it. Please add the "context" file type to the default language server settings.
![image](https://github.com/neovim/nvim-lspconfig/assets/74067428/fc5ead10-9e35-49b1-833b-607e9e220510)
https://valentjn.github.io/ltex/supported-languages.html

glepnir:
fixed by #2979

PR #2979, opened 2024-01-21
Improve LTeX support

metiulekm:
This adds ConTeXt support as requested in #2663. It also makes LTeX work correctly in rnoweb, HTML, XHTML and git commit messages, as well as in plaintext (though this is not enabled by default).


metiulekm:
Lint failed due to URLs in some commit messages. Should I remove them?

PR #2982, opened 2024-01-23
feat: add dprint lsp

WhoIsSethDaniel:


lua/lspconfig/server_configurations/dprint.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'dprint', 'lsp' },
+    filetypes = {},
```
glepnir, line 6: Can be remove

g-plane: It should contain file types that dprint officially supports: https://github.com/nvimtools/none-ls.nvim/blob/a311c7cc8f17543143a7482cdbe3a384c371d56a/lua/null-ls/builtins/formatting/dprint.lua#L17-L29

WhoIsSethDaniel: I don't think null-ls should be used as a source of truth. dprint, when installed, doesn't come with any support for anything. You can later configure it for one or many different languages. 

If placing all these filetypes in the filetypes is what it takes to add this to   nvim-lspconfig I'll do it. I don't think having the filetypes be *too* expansive is that bad.

PR #2987, opened 2024-01-27
fix: disable autostart after :LspStop

brookhong:
Remove autocmd BufReadPost on LspStop so that lsp won't be auto started on buf read after explicit stop with LspStop.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -139,6 +139,9 @@ function configs.__newindex(t, config_name, config_def)
           api.nvim_create_autocmd('BufReadPost', {
             pattern = fn.fnameescape(root_dir) .. '/*',
             callback = function(arg)
+              if #M.manager:clients() == 0 then
+                return true
+              end
               M.manager:try_add_wrapper(arg.buf, root_dir)
             end,
             group = lsp_group,


lua/lspconfig/configs.lua
```diff
@@ -136,7 +137,7 @@ function configs.__newindex(t, config_name, config_def)
         end
 
         if root_dir then
-          api.nvim_create_autocmd('BufReadPost', {
+          buf_read_post_autocmd_id = api.nvim_create_autocmd('BufReadPost', {
             pattern = fn.fnameescape(root_dir) .. '/*',
             callback = function(arg)
               M.manager:try_add_wrapper(arg.buf, root_dir)
```
justinmk, line 143: Can the callback check if the client was force stopped? Then it can return `true`, which will automatically remove the autocmd.

```suggestion
              if client.is_stopped -- psuedocode
                return true
              end
              M.manager:try_add_wrapper(arg.buf, root_dir)
```

lua/lspconfig/configs.lua
```diff
@@ -172,6 +173,12 @@ function configs.__newindex(t, config_name, config_def)
       end)
     end
 
+    function M.shutdown()
```
justinmk, line 176: why does it use a randomly different term instead of "stop"?
```suggestion
    function M.stop()
```

plugin/lspconfig.lua
```diff
@@ -144,6 +144,10 @@ api.nvim_create_user_command('LspStop', function(info)
       client.stop(force)
     end
   end
+  local matching_configs = require('lspconfig.util').get_config_by_ft(vim.bo.filetype)
```
justinmk, line 147: This seems fragile. OTOH, I guess `LspStop` itself is kind of a hack so...

lua/lspconfig/configs.lua
brookhong: Nice, with this addressed, other two comments should be also good.

Thanks, updated.

lua/lspconfig/configs.lua
```diff
@@ -139,6 +139,9 @@ function configs.__newindex(t, config_name, config_def)
           api.nvim_create_autocmd('BufReadPost', {
             pattern = fn.fnameescape(root_dir) .. '/*',
             callback = function(arg)
+              if #M.manager:clients() == 0 then
+                return true
+              end
```
justinmk, line 144: I was hoping there is a way to checking that the clients were force-stopped. Could there be an edge case where the client dies because of an error, which would then (unintentionally) remove this autocmd?

I guess we can try this.

justinmk (approved):

artkpv:
@justinmk , @glepnir  Thank you for this tool.

I wish to stop Lsp in my personal wiki. I start it with 

```
 nvim -c 'cd ~/mydir/notes' -c WikiEnable  -c 'lua vim.g.init_my_wiki_config()' "$@"
```
Where this funciton there is

```
--- Inside my Lazy config for my wiki pluging (https://github.com/lervag/wiki.vim)
    'lervag/wiki.vim',
    lazy = false, -- we don't want to lazy load VimTeX
    config = function()
      vim.g.init_my_wiki_config = function()
        --- ...
        vim.cmd 'LspStop'
       end
    end
```

It works when for my first buffer I open. But when I move to another buffer (.md) the lsp starts. And I need to enter `:LspStop`. Then the lsp doesn't start for other buffers. 

How to stop the lsp? Expected behavior: my function there in my config in init.lua stops the lsp and I don't need to type `:LspStop` each time I open another buffer.

glepnir:
which server you used for wiki.vim find it and set auto_start =false.

PR #2990, opened 2024-01-30
Fix(terraform): Add 'tf' to filetypes

AlexvZyl:
Seems like neovim sets a "*.tf" file to filetype "tf".

### DIFF
--- a/lua/lspconfig/server_configurations/terraform_lsp.lua
+++ b/lua/lspconfig/server_configurations/terraform_lsp.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'terraform-lsp' },
-    filetypes = { 'terraform', 'hcl' },
+    filetypes = { 'terraform', 'hcl', 'tf' },
     root_dir = util.root_pattern('.terraform', '.git'),
   },
   docs = {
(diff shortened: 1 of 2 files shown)


AlexvZyl:
Sorry for all the force pushes, just trying to get those checks passing :)

lua/lspconfig/server_configurations/terraform_lsp.lua
```diff
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'terraform-lsp' },
-    filetypes = { 'terraform', 'hcl' },
+    filetypes = { 'terraform', 'hcl', 'tf' },
```
justinmk, line 6: This is wrong. If filetype=tf that means the detection failed: https://github.com/neovim/neovim/blob/5e5b004da44c7075ed1e20ae9d05ab09c6f2ac58/runtime/lua/vim/filetype/detect.lua#L1524

filetype=tf is not terraform.

justinmk:
If you think there's an improvement that can be made to the detection for terraform, send a change to vim and it will be backported to https://github.com/neovim/neovim/blob/5e5b004da44c7075ed1e20ae9d05ab09c6f2ac58/runtime/lua/vim/filetype/detect.lua#L1524-L1535

AlexvZyl:
Oh, I see the filetype is `ft` when the file is empty, otherwise `terraform`.

Funny, sorry for the noise!

PR #2991, opened 2024-01-30
feat(swift_mesonls): add meson.options as a root marker

tristan957:
This is the preferred filename for defining options for a few versions. meson_options.txt is legacy, but will continue to stick around.

### DIFF
--- a/lua/lspconfig/server_configurations/swift_mesonls.lua
+++ b/lua/lspconfig/server_configurations/swift_mesonls.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'Swift-MesonLSP', '--lsp' },
     filetypes = { 'meson' },
-    root_dir = util.root_pattern('meson_options.txt', '.git'),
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/swift_mesonls.lua
```diff
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'Swift-MesonLSP', '--lsp' },
     filetypes = { 'meson' },
-    root_dir = util.root_pattern('meson_options.txt', '.git'),
+    root_dir = util.root_pattern('meson_options.txt', 'meson.options', '.git'),
```
justinmk, line 7: order matters. if the new file has higher precedence, it should be first in the list.

tristan957: You're not allowed to have both files, so the point seems moot. Happy to submit a follow-up if that's what you want.

PR #2995, opened 2024-02-02
rust-analyzer: Fix is_library

jeremija:
I'm not that well versed in Lua, but the previous check

   if fname:sub(1, #item) == item

didn't really work as expected and I was still experiencing high CPU
usage when using the `go to definition` functionality. After adding some
print statements for debugging, I noticed that only a few chars were
removed from the base dir and thus the comparison was failing. Perhaps
it might work differently in other operating systems, but my Linux
machine it did not work correctly.

I replaced this check with `util.path.is_descendant` which now works
correctly, as well as added another path to check when git repositories
are used as cargo dependencies.

### DIFF
--- a/lua/lspconfig/server_configurations/rust_analyzer.lua
+++ b/lua/lspconfig/server_configurations/rust_analyzer.lua
@@ -19,12 +19,13 @@ local function is_library(fname)
   local user_home = util.path.sanitize(vim.env.HOME)
   local cargo_home = os.getenv 'CARGO_HOME' or util.path.join(user_home, '.cargo')
   local registry = util.path.join(cargo_home, 'registry', 'src')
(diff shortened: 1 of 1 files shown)


glepnir:
LGTM thanks :)

jeremija:
Cool, thanks for the blazing fast review and merge!

jeremija:
Actually, I got confused, the `fname:sub(1, #item) == item` did work as expected, but I was running into this issue only because my deps being in a git repo. So that part could've stayed, but I suppose `util.path.is_descendant(item, fname)` is still easier to grok :slightly_frowning_face:

glepnir:
> because my deps being in a git repo.

for some special cases, the directory cannot be judged by truncation maybe 🤔

PR #3002, opened 2024-02-09
feat(fennel-ls): allow utf-8 offsetEncoding

XeroOl:
The `fennel-ls` server can support `utf-8` offsetEncoding now, and is faster when it doesn't need to compute offsets.


XeroOl:
Apologies for the force push to my branch. The only change is the commit message, which I made in order to pass the `commit_lint` review.

PR #3007, opened 2024-02-13
fix: Follow latest denols implmention

4513ECHO:
Since https://github.com/denoland/deno/pull/22182 Deno removed `deno/cache` custom command.
This PR make the plugin use `workspace/executeCommand` instead.

### DIFF
--- a/lua/lspconfig/server_configurations/denols.lua
+++ b/lua/lspconfig/server_configurations/denols.lua
@@ -2,12 +2,13 @@ local util = require 'lspconfig.util'
 local lsp = vim.lsp
 
 local function buf_cache(bufnr, client)
-  local params = {}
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/denols.lua
```diff
@@ -90,24 +91,14 @@ return {
...
     DenolsCache = {
       function()
-        local clients = vim.lsp.get_active_clients()
-        for _, client in ipairs(clients) do
-          if client.name == 'denols' then
-            buf_cache(0, client)
-            break
-          end
+        local clients = vim.lsp.get_active_clients { name = 'denols' }
+        if #clients > 0 then
```
glepnir, line 100: i think there should find out the client which current buffer attached.

PR #3010, opened 2024-02-13
fix(quick_lint_js): enable for typescript

chrisgrieser:
`quick-lint` works for typescript as well, but the config did not include typescript in the filetypes.


chrisgrieser:
A bit puzzled why commitlint is failing here? Adhering to conventional commits here

PR #3027, opened 2024-02-24
feat: add Delphi LSP

St8Razor:
Add config for Delphi LSP server from Embarcadero

### DIFF
--- a/lua/lspconfig/server_configurations/delphi_ls.lua
+++ b/lua/lspconfig/server_configurations/delphi_ls.lua
@@ -1,25 +1,25 @@
 local util = require("lspconfig.util")
 
 return {
-	default_config = {
-		cmd = { "DelphiLSP.exe" },
-		filetypes = { "pascal" },
-		root_dir = util.root_pattern("*.dpr"),
-		single_file_support = true,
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/delphi_ls.lua
```diff
@@ -0,0 +1,25 @@
...
+  default_config = {
+    cmd = { "DelphiLSP.exe" },
+    filetypes = { "pascal" },
+    root_dir = util.root_pattern("*.dpr"),
+    single_file_support = true,
+    settings = {
+      --look for *.delphilsp.json - Delphi lsp config file
+      settingsFile = vim.fs.find(function(name)
+        return name:match(".*%.delphilsp.json$")
+      end, { limit = math.huge, type = "file", upward = true })[1],
```
glepnir, line 13: it's better add stop condition

St8Razor: Yes, I agree, the limit was is slightly too big 😅 . Set it to 3.

St8Razor:
will resubmit it later

PR #3030, opened 2024-02-25
Feat: Add markdown-oxide lsp

Feel-ix-343:
Add initial support for https://github.com/Feel-ix-343/markdown-oxide

### DIFF
--- a/lua/lspconfig/server_configurations/markdown_oxide.lua
+++ b/lua/lspconfig/server_configurations/markdown_oxide.lua
@@ -1,0 +1,18 @@
+return {
+  default_config = {
+    root_dir = function()
+      return vim.fn.getcwd()
+    end, -- this is applicable because markdown vaults have no defined root files
+    filetypes = { 'markdown' },
+    cmd = { 'markdown-oxide' },
+  },
+  docs = {
+    description = [[
+https://github.com/Feel-ix-343/markdown-oxide
+
+Markdown language server with advanced linking support made to be completely compatible with Obsidian; An Obsidian Language Server
+
+Check the readme to see how to get nvim-cmp properly working.
+    ]]
+  }
+}
--- a/lua/lspconfig/server_configurations/markdown_oxide.lua
+++ b/lua/lspconfig/server_configurations/markdown_oxide.lua
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    root_dir = util.find_git_ancestor
+    root_dir = util.find_git_ancestor,
     single_file_support = true,
     filetypes = { 'markdown' },
     cmd = { 'markdown-oxide' },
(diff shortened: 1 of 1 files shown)


Session.vim
```diff
@@ -0,0 +1,60 @@
+let SessionLoad = 1
```
glepnir, line 1: please don't add any not relate file !!!!!!!!!

Feel-ix-343: Yes thank you. I am trying to clean this up right now.

lua/lspconfig/server_configurations/markdown_oxide.lua
```diff
@@ -0,0 +1,18 @@
+return {
+  default_config = {
+    root_dir = function()
```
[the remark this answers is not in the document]
Feel-ix-343: The language server uses root_dir in the initialization, so right now single_file_support does not work. 

I do think that root_dir makes sense as is because this way the context in which the file is opened is sent as the root_dir, which makes sense for vaults with folders. The same root dir will be sent if neovim is called with `nvim .`, `nvim file`, and `nvim folder/file`, which adhears to the vault structure and the LSP completion structure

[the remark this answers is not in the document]
glepnir: if you set the single_file_support the root dir is cwd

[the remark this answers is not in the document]
Feel-ix-343: As many people do track their vaults with git, I think it is best to both resolve .git as root and cwd; I think root_dir is best. 

My recent commits have this

lua/lspconfig/server_configurations/markdown_oxide.lua
```diff
@@ -0,0 +1,18 @@
+local lspconfig = require 'lspconfig'
```
glepnir, line 1:
```suggestion
local util = require 'lspconfig.util'
```

lua/lspconfig/server_configurations/markdown_oxide.lua
```diff
@@ -0,0 +1,18 @@
...
+
+return {
+  default_config = {
+    root_dir = lspconfig.util.root_pattern('.git', vim.fn.getcwd()),
```
glepnir, line 5:
```suggestion
    root_dir = util.find_git_ancestor
    single_file_support = true,
```

Feel-ix-343:
Awesome; works on my end!

Feel-ix-343:
Hey @glepnir I don't think the `single_file_support` is working. There are some issues open on my repo saying that the LS doesn't start unless there is a git repo -- in which case `root_dir` would be sent. I think `single_file_support` is supposed to send the root_dir but as of neovim 0.9.5 and nightly, `root_dir` is not being sent (unless there is a git repo). 

I believe the solution for neovim 0.9.5 is to send the cwd(), and for nightly to send the vim.uv.cwd()

Shall I open a new PR?

glepnir:
emm sure. but i may take a look code of there .forget some code on how neovim initialize client when root is emtpy. :p . most work on neovim core rencently ..

PR #3034, opened 2024-02-26
feat: add delphilsp support

St8Razor:
Add support for Delphi Language Server, which is available only for Windows.
The config file for the language server *.delphilsp.json is located in the root_dir.
Unfortunately the root_dir is not available at the configuration creation time,
so I had to look for it. Maybe, there is a better solution?

### DIFF
--- a/lua/lspconfig/server_configurations/delphi_ls.lua
+++ b/lua/lspconfig/server_configurations/delphi_ls.lua
@@ -1,29 +1,11 @@
 local util = require 'lspconfig.util'
 
-local function get_lsp_config_path()
-  -- first find *.dpr project file, as lsp config might not be present
-  local dpr_path = vim.fs.find(function(name)
-    return name:match '.*%.dpr$'
-  end, { type = 'file', upward = true })[1]
-
-  -- if found, then lsp config should be located in the same folder
-  if dpr_path then
-    return vim.fs.find(function(name)
-      return name:match '.*%.delphilsp.json$'
-    end, { type = 'file', path = vim.fs.dirname(dpr_path), upward = false })[1]
-  end
-end
-
 return {
   default_config = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/delphi_ls.lua
```diff
@@ -0,0 +1,37 @@
...
+  -- first find *.dpr project file, as lsp config might not be present
+  local dpr_path = vim.fs.find(function(name)
+    return name:match '.*%.dpr$'
+  end, { type = 'file', upward = true })[1]
+
+  -- if found, then lsp config should be located in the same folder
+  if dpr_path then
+    return vim.fs.find(function(name)
+      return name:match '.*%.delphilsp.json$'
+    end, { type = 'file', path = vim.fs.dirname(dpr_path), upward = false })[1]
```
glepnir, line 13: dirname is unnecessary.

St8Razor: Unfortunately, it is required. The dpr_path contains file name that needs to be removed for vim.fs.find to work.
It would be nice to replace the dpr_path with root_dir, as Delphi's lsp config and project file are always located in the root folder. With current implementation we have to look for the same folder twice, first time for lsp config and second time for the root_dir

glepnir: what about if find return an empty table ?

St8Razor: Setting DelphiLSP configuration on server setup was wrong, @glepnir thank you for being patient  with me.
The proper way is to send it to LS through  **didChangeConfiguration**  on **on_attach**.
At this point the **root_dir** is already set. 
I've added a configuration sample in the documentation section of delphi_ls.

PR #3056, opened 2024-03-05
Update server_configurations.md

oaleynik:
Typo in Vue plugin configuration for tsserver

PR #3060, opened 2024-03-09
docs(vue): update vue support section

RayGuo-ergou:
### Description

Updated the documentation for Vue support in tsserver, adding two essential points for configuring Vue with tsserver.


RayGuo-ergou:
This can also close https://github.com/neovim/nvim-lspconfig/pull/3056

PR #3070, opened 2024-03-12
chore: remove rnix-lsp instructions

adamcstephens:
rnix-lsp is an archived project as of January. It can no longer be securely installed due to dependence on nix 2.15, which has been marked insecure. Finally, I've merged the removal of rnix-lsp from nixpkgs. https://github.com/NixOS/nixpkgs/pull/294488


glepnir:
we don't remove a file directly should be mark as deprecated.

adamcstephens:
I guess you'll do that? rnix is no longer supported, nor in fact installable any longer with nix. It seems odd to keep directions around, but ok.

glepnir:
yup I will mark it to deprecate later and it will remove in next release version.

adamcstephens:
Great, thank you!

PR #3075, opened 2024-03-16
feat: add tinymist

Parsifa1:


lua/lspconfig/server_configurations/tinymist.lua
```diff
@@ -0,0 +1,15 @@
+return {
+  default_config = {
+    cmd = { 'tinymist' },
+    filetypes = { 'typst' },
+    root_dir = function()
```
glepnir, line 5: I aussme there can use single_file_support  and root can be use .git folder?

Parsifa1: no single_file_support,but i added the `.git` support

lua/lspconfig/server_configurations/tinymist.lua
```diff
@@ -0,0 +1,22 @@
...
+  default_config = {
+    cmd = { 'tinymist' },
+    filetypes = { 'typst' },
+    root_dir = function(fname)
+      if util.find_git_ancestor(fname) then
+        return util.find_git_ancestor(fname)
+      else
+        return vim.fn.getcwd()
+      end
+    end,
```
glepnir, line 13:
```suggestion
    root_dir = util.find_git_ancestor,
```

lua/lspconfig/server_configurations/tinymist.lua
```diff
@@ -0,0 +1,16 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'tinymist' },
+    filetypes = { 'typst' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = false,
```
glepnir, line 8:
```suggestion
    single_file_support = true,
```

Parsifa1: Theoretically, this lsp does not have single file support...

Parsifa1: Well, actually it’s not that it can’t run on a single file, it just has an incomplete experience.

glepnir: okay .but in there this also mean use cwd as root . so i wonder this not work ?

Parsifa1: If you use cwd as root, it works, but if you don't carry root initially, it may take time to initialize and cause lagging
21c683246004b17b040c3403148c564c

Parsifa1: You can revise it and make the final decision, it's my first time to do this kind of pr, please forgive me if there may be some wrong.

Myriad-Dreamin: > but in there this also mean use cwd as root . so i wonder this not work ?

I don't really understand `single_file_support`, but if neovim as a client will somehow provide a root_dir, the lsp will be not problem at all. They we can go with `single_file_support = true`, and let people report issues if any.

glepnir: some code here , you can find in configs.lua

```lua
          -- This allows on_new_config to use the parent directory of the file
          -- Effectively this is the root from lspconfig's perspective, as we use
          -- this to attach additional files in the same parent folder to the same server.
          -- We just no longer send rootDirectory or workspaceFolders during initialization.
          if not api.nvim_buf_is_valid(bufnr) or (#bufname ~= 0 and not util.bufname_valid(bufname)) then
            return
          end
          local pseudo_root = #bufname == 0 and pwd or util.path.dirname(util.path.sanitize(bufname))
          M.manager:add(pseudo_root, true, bufnr)
        end
```

Parsifa1:
Overall, I think it should be possible to merge now
`single_file_support = true` can be accepted

Andrew15-5:
Without `single_file_support = true` LSP server doesn't work.

Andrew15-5:
I'm not 100% sure if it does the same thing as adding

```lua
  root_dir = function()
    return vim.fn.getcwd()
  end,
```

Because that is what I normally do if the `root directory:    Not found` with default settings.

PR #3079, opened 2024-03-17
Correct schema for LSP server

baco:
There was a mistake on the original schema. The original project stores their settings under the [`python`](https://github.com/microsoft/pyright/blob/main/docs/settings.md) key, the fork does it under the [`basedpyright`](https://github.com/DetachHead/basedpyright/blob/main/docs/settings.md) key.


AThePeanut4:
`:PyrightSetPythonPath` still doesn't work. This simple change fixes it for me:

```diff
diff --git a/lua/lspconfig/server_configurations/basedpyright.lua b/lua/lspconfig/server_configurations/basedpyright.lua
index a360e8c..172ed23 100644
--- a/lua/lspconfig/server_configurations/basedpyright.lua
+++ b/lua/lspconfig/server_configurations/basedpyright.lua
@@ -32,7 +32,7 @@ local function set_python_path(path)
   }
   for _, client in ipairs(clients) do
     if client.settings then
-      client.settings.python = vim.tbl_deep_extend('force', client.settings.python, { pythonPath = path })
+      client.settings.python = vim.tbl_deep_extend('force', client.settings.python or {}, { pythonPath = path })
     else
       client.config.settings = vim.tbl_deep_extend('force', client.config.settings, { python = { pythonPath = path } })
     end
```

baco:
@AThePeanut4 "fixes" as in you were able to use the command **and** effectively the Python to check the code changed, or it just stopped raising an error?

AThePeanut4:
> @AThePeanut4 "fixes" as in you were able to use the command **and** effectively the Python to check the code changed, or it just stopped raising an error?

@baco The former - the LSP reloaded and I stopped getting missing import errors (because I was setting it to the python inside the project's venv)

If I understand correctly, the reason why the actual pyright doesn't need the ` or {}` is because its default settings table has a `python` key in it, so `client.settings.python` won't ever be `nil`. But since basedpyright uses `basedpyright` rather than `python` (what this PR originally fixed), the default settings table can have `client.settings.python == nil`, which causes `vim.tbl_deep_extend` to error.

baco:
> > @AThePeanut4 "fixes" as in you were able to use the command **and** effectively the Python to check the code changed, or it just stopped raising an error?
> 
> @baco The former - the LSP reloaded and I stopped getting missing import errors (because I was setting it to the python inside the project's venv)
> 
> If I understand correctly, the reason why the actual pyright doesn't need the ` or {}` is because its default settings table has a `python` key in it, so `client.settings.python` won't ever be `nil`. But since basedpyright uses `basedpyright` rather than `python` (what this PR originally fixed), the default settings table can have `client.settings.python == nil`, which causes `vim.tbl_deep_extend` to error.

All right, but then the fix, shouldn't be something like:
```diff
diff --git a/lua/lspconfig/server_configurations/basedpyright.lua b/lua/lspconfig/server_configurations/basedpyright.lua
index a360e8c..41c103a 100644
--- a/lua/lspconfig/server_configurations/basedpyright.lua
+++ b/lua/lspconfig/server_configurations/basedpyright.lua
@@ -32,9 +32,17 @@ local function set_python_path(path)
   }
   for _, client in ipairs(clients) do
     if client.settings then
-      client.settings.python = vim.tbl_deep_extend('force', client.settings.python, { pythonPath = path })
+      client.settings.basedpyright = vim.tbl_deep_extend(
+        'force',
+        client.settings.basedpyright,
+        { pythonPath = path }
+      )
     else
-      client.config.settings = vim.tbl_deep_extend('force', client.config.settings, { python = { pythonPath = path } })
+      client.config.settings = vim.tbl_deep_extend(
+        'force',
+        client.config.settings,
+        { basedpyright = { pythonPath = path } }
+      )
     end
     client.notify('workspace/didChangeConfiguration', { settings = nil })
   end
```
replacing `python` with `basedpyright` instead?

I am trying to make it work, but I am not getting de import errors to disappear when setting the path to Python inside a `venv`.

baco:
@AThePeanut4  I will be making a new PR as soon as I make it work on my local. I just don't know why your approach would be the correct one.

baco:
ok, I get it, _sigh_. The official documentation says that: https://github.com/DetachHead/basedpyright/blob/main/docs/settings.md. There are 2 options that remain under the `python` namespace in the upstream project.

> `python.pythonPath` [path]: Path to Python interpreter. This setting is being deprecated by the VS Code Python extension in favor of a setting that is stored in the Python extension’s internal configuration store. Pyright supports both mechanisms but prefers the new one if both settings are present.
> 
> `python.venvPath` [path]: Path to folder with subdirectories that contain virtual environments. The python.pythonPath setting is recommended over this mechanism for most users. For more details, refer to the [import resolution](https://github.com/DetachHead/basedpyright/blob/main/docs/import-resolution.md#configuring-your-python-environment) documentation.

Although, neither option is loaded from configuration files. I'm not confident they will persist in the future.

AThePeanut4:
> ok, I get it, _sigh_. The official documentation says that: https://github.com/DetachHead/basedpyright/blob/main/docs/settings.md. There are 2 options that remain under the `python` namespace in the upstream project.
> 
> > `python.pythonPath` [path]: Path to Python interpreter. This setting is being deprecated by the VS Code Python extension in favor of a setting that is stored in the Python extension’s internal configuration store. Pyright supports both mechanisms but prefers the new one if both settings are present.
> > `python.venvPath` [path]: Path to folder with subdirectories that contain virtual environments. The python.pythonPath setting is recommended over this mechanism for most users. For more details, refer to the [import resolution](https://github.com/DetachHead/basedpyright/blob/main/docs/import-resolution.md#configuring-your-python-environment) documentation.

Yeah my bad, I had looked at that docs page, should have added a link to it.

> Although, neither option is loaded from configuration files. I'm not confident they will persist in the future.

I do hope they stay, the LSP setting is the easiest way (at least for me) to get the LSP to use a venv python rather than the system one. I think the reason why `pythonPath` specifically isn't loaded from configuration files is because the path will generally be system-specific and thus shouldn't be shared: [here](https://github.com/DetachHead/basedpyright/blob/main/docs/configuration.md) under "venvPath" it says `If you are working on a project with other developers, it is best not to specify this setting in the config file, since this path will typically differ for each developer. Instead, it can be specified on the command line or in a per-user setting.`

PR #3086, opened 2024-03-22
feat: adds ginko_ls support

SethGower:
Adds configuration for [`ginko_ls`](https://github.com/Schottkyc137/ginko?tab=readme-ov-file#ginko_ls).

I tested this locally and it works well. I tested with the test file from `ginko`'s readme

```dts
/dts-v1/;

/ {
    pic@10000000 {
        phandle = <1>;
        interrupt-controller;
        reg = <0x10000000 0x100>;
    }
};
```

### DIFF
--- a/lua/lspconfig/server_configurations/ginko_ls.lua
+++ b/lua/lspconfig/server_configurations/ginko_ls.lua
@@ -1,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'ginko_ls' },
+    filetypes = { 'dts' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/ginko_ls.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'ginko_ls' },
+    filetypes = { 'dts' },
+    root_dir = util.find_git_ancestor,
+    settings = {},
```
glepnir, line 8: if is empty can be remove.

PR #3094, opened 2024-04-04
fix: tinymist

Parsifa1:
fix https://github.com/Myriad-Dreamin/tinymist/issues/125


Myriad-Dreamin:
> Do not add vim.fn.cwd or util.path.dirname in root_dir. A future version of lspconfig will provide emulation of a single file mode until this is formally codified in the specification. A good fallback is util.find_git_ancestor, see other configurations for examples.

We may have to close this PR according to it.

Considering that, the only solution is probably to tell users to override the default configuration provided by nvim-lspconfig manually. Example configuration from some typst-lsp user:

```lua
{
  root_dir = function(fname)
    return require('lspconfig.util').root_pattern('typst.toml', '.git')(fname)
      or vim.fn.getcwd() -- this is strongly rejected by nvim-lspconfig, so do configure it by yourself
  end,
}
```

PR #3099, opened 2024-04-09
feat(omnisharp)!: use standard `settings` table for configuration

JosefLitos:
The standard way in lspconfig is configuring through the `settings` table. Omnisharp doesn't support being passed config at runtime, so previously some common options were toggled and added to the command args. That happened for only the preconfigured options.

This PR generalizes the behaviour in an attempt to allow users to set their options in the standard way, allowing for a more seamless syntax transition from JSON format that is used in `omnisharp` global config and `vscode`.

Although I haven't experienced any issues when leaving the `settings` table in the config, I assume `nvim` then tries to feed the settings to `omnisharp`, which may be avoidable by `settings=nil` (opinion needed).

### DIFF
--- a/lua/lspconfig/server_configurations/omnisharp.lua
+++ b/lua/lspconfig/server_configurations/omnisharp.lua
@@ -2,40 +2,44 @@ local util = require 'lspconfig.util'
 
 return {
   default_config = {
-    -- Enables support for reading code style, naming convention and analyzer
-    -- settings from .editorconfig.
-    enable_editorconfig_support = true,
-
-    -- If true, MSBuild project system will only load projects for files that
-    -- were opened in the editor. This setting is useful for big C# codebases
-    -- and allows for faster initialization of code navigation features only
-    -- for projects that are relevant to code that is being edited. With this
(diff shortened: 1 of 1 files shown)


glepnir:
thanks.

lua/lspconfig/server_configurations/omnisharp.lua
```diff
@@ -51,32 +55,21 @@ return {
       table.insert(new_config.cmd, '--languageserver')
 
       -- Append configuration-dependent command arguments
-      if new_config.enable_editorconfig_support then
-        table.insert(new_config.cmd, 'FormattingOptions:EnableEditorConfigSupport=true')
+      local function flatten(tbl)
```
justinmk, line 58: why not use `vim.tbl_flatten()` ?

JosefLitos: I didn't know about such function, but after trying it, it returns an empty table, so it probably does something completely different from what we want here.

JosefLitos: I see now - that function works on lists, this is a normal table/map/dict.

PR #3100, opened 2024-04-10
feat: add agda-language-server

foxyseta:
Created as draft as per the contributing guidelines. Let me know if there are any comments I need to address.


glepnir:
just take a quick look at is okay

foxyseta:
Thanks! The server itself does not seem to implement many LSP features, but I tried hovering on symbols to get types and that works consistently.

PR #3105, opened 2024-04-12
feat: Rename ruby_ls to ruby_lsp

nithinbekal:
## Description

The Ruby LSP team calls the LSP server `ruby-lsp`, while nvim-lspconfig calls it `ruby_ls`. The ruby-lsp team have suggested that this be renamed to `ruby_lsp` so the naming is consistent everywhere ([comment](https://github.com/Shopify/ruby-lsp/pull/1877/files#r1559571577)). cc @vinistock 

As part of this, I've also added a deprecation, so that people using `ruby_ls` in their config will be notified, but can continue using `ruby-lsp`.

## Testing

Tested with [my neovim config](https://github.com/nithinbekal/dotfiles/blob/a2bbd5bf33493c8690715b57a3f2557f2b66d944/.config/nvim/init.lua#L223-L252) which uses [mason-lspconfig.nvim](https://github.com/williamboman/mason-lspconfig.nvim) to set up ruby_ls.

I've tested that ruby-lsp continues working even after the rename, but with this deprecation message on load:

ruby-ls-deprecation

## Next steps

- Update mason-lspconfig.nvim: https://github.com/williamboman/mason-lspconfig.nvim/pull/395
- Update the [neovim install instructions in ruby-lsp](https://github.com/Shopify/ruby-lsp/blob/main/EDITORS.md#neovim)


kevintraver:
Is there a way to hide or suppress the warning?

```
Warn: ruby_ls is deprecated, use ruby_lsp instead.
This feature will be removed in lspconfig version 0.2.0
```

nithinbekal:
@kevintraver How have you configured Ruby LSP? If you're using mason-lspconfig, you might need to wait till https://github.com/williamboman/mason-lspconfig.nvim/pull/395 gets merged.

kevintraver:
Yes, I am using **_mason-lspconfig_**, which is why I was just curious if there's a way to hide the notice until it's resolved. If not, it's no problem at all.

PR #3109, opened 2024-04-14
fix: gitlab-ci-ls filetype

alesbrelih:
Now it loads on yaml.gitlab. Also added yamlls dependency to the same filetype.

Else it was being loaded on every yaml file.


polyzen:
@alesbrelih, is there a plugin or something that sets the filetype to `yaml.giltab` for you?

alesbrelih:
Hey @polyzen,
I just use a simple autocmd:

``` lua
vim.api.nvim_create_autocmd({ "BufRead", "BufNewFile" }, {
  pattern = ".gitlab*",
  callback = function()
    vim.bo.filetype = "yaml.gitlab"
  end,
})
```

glepnir:
`:help vim.filetype.add` this is more easy

PR #3110, opened 2024-04-15
Minor cleanup of flake.nix

Rexcrazy804:
used ```with pkgs;``` to simplify statements referring to it in the scope of mkshell. Restructured  input declaration to follow the typical structure.

I am unsure whether I should mark this with doc: or not, sorry

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,14 +3,18 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs";
-    flake-utils.url = "github:numtide/flake-utils";
   };
 
   outputs = { self, nixpkgs, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
-      let pkgs = nixpkgs.legacyPackages.${system}; in
-      rec {
-        devShell = with pkgs; mkShell {
+    let
+      forAllSystems = function: nixpkgs.lib.genAttrs [
+        "x86_64-linux"
+        "aarch64-linux"
+        "aarch64-darwin"
+      ] (system: function nixpkgs.legacyPackages.${system});
+    in {
+      devShells = forAllSystems (pkgs: {
+        default = with pkgs; mkShell {
           buildInputs = [
             stylua
             luaPackages.luacheck
@@ -18,6 +22,6 @@
             selene
           ];
         };
-      }
-    );
+      });
+    };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -1,19 +1,21 @@
 {
   description = "Quickstart configurations for the Nvim LSP client";
 
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
 
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system:
       let pkgs = nixpkgs.legacyPackages.${system}; in
       rec {
-        devShell = pkgs.mkShell {
+        devShell = with pkgs; mkShell {
           buildInputs = [
-            pkgs.stylua
-            pkgs.luaPackages.luacheck
-            pkgs.luajitPackages.vusted
-            pkgs.selene
+            stylua
+            luaPackages.luacheck
+            luajitPackages.vusted
+            selene
           ];
         };
       }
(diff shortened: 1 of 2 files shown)


Rexcrazy804:
If permitted, I'd like to add a shell hook that executes the default shell using the ```$SHELL``` env var so running ```nix develop``` will drop you into whatever shell you are using instead of defaulting to bash. I suppose that could be another pr?

justinmk:
@teto

flake.nix
```diff
@@ -1,19 +1,21 @@
...
+        devShell = with pkgs; mkShell {
           buildInputs = [
-            pkgs.stylua
-            pkgs.luaPackages.luacheck
-            pkgs.luajitPackages.vusted
-            pkgs.selene
+            stylua
+            luaPackages.luacheck
+            luajitPackages.vusted
+            selene
```
itslychee, line 18: Instead of using `with`, which isn't a [very encouraged practice](https://nix.dev/guides/best-practices#with-scopes), use `builtin.attrValues` with `inherit`

For example
```nix
devShell = pkgs.mkShell {
  packages = builtin.attrValues {
    inherit (pkgs)
     stylua
     luaPackages.luacheck
     ...
     ;
  };
};
```

Also `packages` should be used over `buildInputs`, another nitpick of mine is that we can drop the `flake-utils` input entirely in favor of using `genAttrs` which provides a cleaner, readable Flake. Read this for more on that https://ayats.org/blog/no-flake-utils

itslychee (changes requested):

itslychee:
> If permitted, I'd like to add a shell hook that executes the default shell using the `$SHELL` env var so running `nix develop` will drop you into whatever shell you are using instead of defaulting to bash. I suppose that could be another pr?

I disagree with this, perhaps check out direnv if you haven't already?

flake.nix
Rexcrazy804: hmm I can see why it isn't encouraged but the points they mention are as follows:
> Static analysis can’t reason about the code, because it would have to actually evaluate this file to see which names are in scope.

> When more than one with used, it’s not clear anymore where the names are coming from.

> Scoping rules for with are not intuitive, see this [Nix issue for details](https://github.com/NixOS/nix/issues/490).

These don't apply for our specific use case here, simply using ```with pkgs``` is far more practical and intuitive here. That aside, I love that article regarding no-flake-utils, thank you for that :D

itslychee: ❤️ no problem, I don't disagree with the `with` part but personally i just find the syntax nicer to read, but I do think flake utils should be removed

Rexcrazy804: yep, I'll do that as early as I can 👍

Rexcrazy804:
@itslychee  I have specified the three major systems, do you reckon there's anything else to add?

itslychee:
> @itslychee I have specified the three major systems, do you reckon there's anything else to add?

Don't forget to change `buildInputs` to `packages`, and after that I'd squash the newer commits to keep the history clean.

itslychee:
Also since `eachDefaultSystem` is used currently, it's best to add `"x86_64-linux"` to the platform list to keep it backwards compatible as `eachDefaultSystem` used 4 platforms as described [here](https://github.com/numtide/flake-utils?tab=readme-ov-file#defaultsystems--system)

Rexcrazy804:
Could you clarify that a bit further? The platforms list does contain "x86_56-linux" but it has come to my attention that "x86_64-darwin" is not included in the list, I suppose that is the one I should add.

itslychee:
> Could you clarify that a bit further? The platforms list does contain "x86_64-linux" but it has come to my attention that "x86_64-darwin" is not included in the list, I suppose that is the one I should add.

Oops yeah that's the one, something possessed me to write linux instead :sob:

itslychee (approved):

flake.nix
```diff
@@ -1,21 +1,28 @@
...
+    let
+      forAllSystems = function: nixpkgs.lib.genAttrs [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ] (system: function nixpkgs.legacyPackages.${system});
+    in {
+      devShells = forAllSystems (pkgs: {
+        default = with pkgs; mkShell {
```
teto, line 18: why remove the pkgs prefix ? it's just churn . If anything it makes perf and linter work harder please revert

teto:
I get that one would prefer to remove the dependency to `flake-utils` but what's the point if you have to replace allDefaultSystems by actually typing the defaults yourself ? The handling of system is one of the controversial point of flakes and nothing in this PR is objectively better so it's just churn. I would just merge an update to the lockfile to get a more recent version of the dependencies.

justinmk:

PR #3117, opened 2024-04-23
fix(nixd): update root_dir

Ruixi-rebirth:
starting with nixd 2.0.2 the '.nixd.json' file will no longer be needed.


glepnir:
stylint error unrelate your pr

PR #3131, opened 2024-04-29
fix(command): check server exist in configs

glepnir:


glepnir:
@wojciech-kulik  could you check this.

wojciech-kulik:
There is no error now, but the LSP is not attached again.

wojciech-kulik:
Now it works for my regular LSP, but for copilot it prints tons of:
image

I have to restart nvim then

glepnir:
copilot not in lspconfig

wojciech-kulik:
Maybe copilot is some special case. Because it doesn't appear on Configured servers list:  from :LspInfo

I'm using the official plugin: `github/copilot.vim`

glepnir:
@wojciech-kulik fixed. but notice the copilot will not restart because it not exist in lspconfig. you must need do some wrap

wojciech-kulik:
Makes sense! Now I checked and everything works fine on both Neovim versions! Thank you 🍻

PR #3142, opened 2024-05-07
docs(dockerls): document ignoreMultilineInstructions setting

fho:
Document how the ignoreMultilineInstructions settings for dockerls can
be configured.

That the settings key must be "docker" instead of "dockerls" makes it
non-intuitive and harder to figure out.

Sources:
- https://github.com/rcjsuen/dockerfile-language-server?tab=readme-ov-file#language-server-settings
- https://github.com/rcjsuen/dockerfile-language-server/blob/9e3f8580f711e54a1de0351631baeac49665d6f3/src/server.ts#L330-L344

### DIFF
--- a/lua/lspconfig/server_configurations/dockerls.lua
+++ b/lua/lspconfig/server_configurations/dockerls.lua
@@ -15,6 +15,21 @@ https://github.com/rcjsuen/dockerfile-language-server-nodejs
 ```sh
 npm install -g dockerfile-language-server-nodejs
 ```
+
+Additional configuration can be applied in the following way:
+```lua
+require("lspconfig").dockerls.setup {
+    settings = {
+        docker = {
+	    languageserver = {
+	        formatter = {
+		    ignoreMultilineInstructions = true,
+		},
+	    },
+	}
+    }
+}
+```
     ]],
     default_config = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/dockerls.lua
```diff
@@ -6,6 +6,15 @@ return {
     filetypes = { 'dockerfile' },
     root_dir = util.root_pattern 'Dockerfile',
     single_file_support = true,
+    settings = {
+      docker = {
+        languageserver = {
+          formatter = {
+            ignoreMultilineInstructions = true,
```
glepnir, line 13: it's better add an introduce of this config in docs/descripition

fho: moved, thanks for the fast feedback!

fho:
I'm struggling with the commitlint check:
- It complains that the URLs in the commit message exceed the max line lenght. It does not make sense to break those URLs into multiple lines though.
- I don't understand the warning `footer must have leading blank line [footer-leading-blank]` I have a newline before "Sources:" which I would consider to be the footer

@glepnir how can I solve this?

glepnir:
commit lint.. i can force update it no worry.

fho:
@glepnir thank you!

PR #3145, opened 2024-05-09
fix: support compound file-types

baod-rate:
Vim supports compound filetypes (see [|'filetype'|](https://neovim.io/doc/user/options.html#'filetype')):

> When a dot appears in the value then this separates two filetype names.
> This works both for filetype plugins and for syntax files. More than one dot may appear.

nvim-lspconfig currently always compares the full 'filetype' option against the configured filetypes. So buffers set to multiple filetypes will not match at all.

Fix utility functions to match against each individual filetype

### DIFF
--- a/lua/lspconfig/server_configurations/elmls.lua
+++ b/lua/lspconfig/server_configurations/elmls.lua
@@ -13,7 +13,7 @@ return {
     filetypes = { 'elm' },
     root_dir = function(fname)
       local filetype = api.nvim_buf_get_option(0, 'filetype')
-      if filetype == 'elm' or (filetype == 'json' and fname:match 'elm%.json$') then
+      if util.ft_matches(filetype, 'elm') or (util.ft_matches(filetype, 'json') and fname:match 'elm%.json$') then
         return elm_root_pattern(fname)
       end
     end,
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -21,6 +21,21 @@ M.default_config = {
 -- global on_setup hook
 M.on_setup = nil
 
+---@param filetype string the filetype to check (can be a compound, dot-separated filetype; see |'filetype'|)
+---@param expected string|string[] the filetype(s) to match against
+---@return boolean
+function M.ft_matches(filetype, expected)
+  expected = type(expected) == 'table' and expected or { expected }
+  for ft in filetype:gmatch '([^.]+)' do
+    for _, expected_ft in ipairs(expected) do
+      if ft == expected_ft then
+        return true
+      end
+    end
+  end
+  return false
+end
+
 function M.bufname_valid(bufname)
   if bufname:match '^/' or bufname:match '^[a-zA-Z]:' or bufname:match '^zipfile://' or bufname:match '^tarfile:' then
     return true
@@ -337,10 +352,8 @@ function M.get_active_clients_list_by_ft(filetype)
   local clients_list = {}
   for _, client in pairs(clients) do
     local filetypes = client.config.filetypes or {}
-    for _, ft in pairs(filetypes) do
-      if ft == filetype then
-        table.insert(clients_list, client.name)
-      end
+    if M.ft_matches(filetype, filetypes) then
+      table.insert(clients_list, client.name)
     end
   end
   return clients_list
@@ -353,10 +366,8 @@ function M.get_other_matching_providers(filetype)
   for _, config in pairs(configs) do
     if not vim.tbl_contains(active_clients_list, config.name) then
       local filetypes = config.filetypes or {}
-      for _, ft in pairs(filetypes) do
-        if ft == filetype then
-          table.insert(other_matching_configs, config)
-        end
+      if M.ft_matches(filetype, filetypes) then
+        table.insert(other_matching_configs, config)
       end
     end
   end
@@ -368,10 +379,8 @@ function M.get_config_by_ft(filetype)
   local matching_configs = {}
   for _, config in pairs(configs) do
     local filetypes = config.filetypes or {}
-    for _, ft in pairs(filetypes) do
-      if ft == filetype then
-        table.insert(matching_configs, config)
-      end
+    if M.ft_matches(filetype, filetypes) then
+      table.insert(matching_configs, config)
     end
   end
   return matching_configs


lua/lspconfig/util.lua
```diff
@@ -21,6 +21,21 @@ M.default_config = {
 -- global on_setup hook
 M.on_setup = nil
 
+---@param filetype string the filetype to check (can be a compound, dot-separated filetype; see |'filetype'|)
+---@param expected string|string[] the filetype(s) to match against
+---@return boolean
+function M.ft_matches(filetype, expected)
```
glepnir, line 27: there has vim.fs module.

baod-rate: sorry, can you explain? I don't see anything in `vim.fs` that does this. vim.fs seems to deal mostly with filesystem operations. I'm not actually reading files here, I am only splitting and comparing `filetype` strings

glepnir: okay see vim.filetype moudle. vim.filetype.match will return the filetype if it already in default detect table or register by vim.filetype.add . we don't want add more utils function now

baod-rate: `vim.filetype.match` takes an existing buffer and returns the `filetype` (using the rules defined in `filetype.lua`). the result of that might be (for example) `c.doxygen`. see `|'filetype'|`:

```vimhelp
						*'filetype'* *'ft'*
'filetype' 'ft'		string	(default "")
			local to buffer  |local-noglobal|
...
	Setting this option to a different value is most useful in a modeline,
	for a file for which the file type is not automatically recognized.
	Example, for in an IDL file: >c
		/* vim: set filetype=idl : */
<	|FileType| |filetypes|
	When a dot appears in the value then this separates two filetype
	names.  Example: >c
		/* vim: set filetype=c.doxygen : */
<	This will use the "c" filetype first, then the "doxygen" filetype.
	This works both for filetype plugins and for syntax files.  More than
	one dot may appear.
	This option is not copied to another buffer, independent of the 's' or
	'S' flag in 'cpoptions'.
	Only normal file name characters can be used, `/\*?[|<>` are illegal.
```

In other words:

```lua
local ft = vim.filetype.match({buf=0})
ft == "c.doxygen"
util.ft_matches(ft, 'idl') == false
util.ft_matches(ft, 'c') == true
util.ft_matches(ft, 'doxygen') == true
```

glepnir: you can also pass filename  into match

baod-rate: yes, `vim.filetype.match` can take a filename and give us a filetype. but what we want is to take *that* filetype and check it against a list of filetypes (e.g. the ones supported by a particular lsp server). 

i.e. 

```lua
local ft = vim.filetype.match({ filename = 'main.lua' })
ft == "lua"
local filetypes_supported_by_current_client = { "lua", "sh", "c" }
util.ft_matches(ft, filetypes_supported_by_current_client) == true
```

Usually this would just be a `==` comparison but because a filetype can contain multiple filetypes (e.g. `c.doxygen`), we have to split it first.

```lua
local ft = "yaml.ansible" -- this is not a filename! it is a _compound_ filetype
util.ft_matches(ft, "ansible") == true
util.ft_matches(ft, "yaml") == true
```

notice that `vim.filetype.match` doesn't help us do this comparison! it takes a buffer/file/filepath and gives us a filetype. but we don't need that. we **have** a filetype, it is provided by the caller. We are trying to match that against a list of acceptable filetypes, also provided by the caller.

glepnir: does register a filetype  by using vim.filetype.add for a special file not make sense in user side

baod-rate: A completely unique type is not desirable because we want to use the existing behaviors defined for the primary filetype. see [mfussenegger/nvim-ansible](https://github.com/mfussenegger/nvim-ansible/blob/cb5d0cc79c1864b72d5ae40a926a1176b18f1b65/ftdetect/ansible.lua#L23) or [pearofducks/ansible-vim](https://github.com/pearofducks/ansible-vim/blob/master/ftdetect/ansible.vim#L36). They both `set ft=yaml.ansible`.

You *want* ansible scripts to have the yaml filetype because that's what they are; we want (neo)vim to keep recognizing them as yaml files. if we set `ft=ansible` only, we lose the syntax highlighting/`match_words`/etc defined for yaml files. 

But the plugin additionally [sets various useful options](https://github.com/mfussenegger/nvim-ansible/tree/main?tab=readme-ov-file#usage) like `keywordprg`/`path`/`iskeyword`/`isfname` for `ansible` file types. Hence the reason for multiple file types.

The problem is that right now, nvim-lspconfig diverges from standard vim behavior because filetype matching operates on the entire `vim.bo.filetype` as a whole, rather than splitting it into individual file types. And every server has to be configured for every possible combination of multiple filetypes. see:

https://github.com/neovim/nvim-lspconfig/blob/a3d9395455f2b2e3b50a0b0f37b8b4c23683f44a/lua/lspconfig/server_configurations/yamlls.lua#L4-L6

(notice that `yaml.ansible` is not included.

With this change, you can instead just specify `yaml`, and we'll match as long as *one* of your filetypes matches `yaml`

lua/lspconfig/util.lua
```diff
@@ -21,6 +21,21 @@ M.default_config = {
 -- global on_setup hook
 M.on_setup = nil
 
+---@param filetype string the filetype to check (can be a compound, dot-separated filetype; see |'filetype'|)
```
justinmk, line 24: this new function has no docstring explaining its purpose. and this PR has a wall of text discussing it.

how in the world is anyone supposed know when to use this

baod-rate: apologies, I didn't realize that it was unclear. I've added documentation. let me know if it's unclear

justinmk: doesn't look like there is a summary docstring. also a couple input => output examples would go a long way.

```
filetype "foo.bar" matches expected "..."
```

lua/lspconfig/server_configurations/elmls.lua
```diff
@@ -10,7 +10,7 @@ return {
     filetypes = { 'elm' },
     root_dir = function(fname)
       local filetype = api.nvim_buf_get_option(0, 'filetype')
-      if filetype == 'elm' or (filetype == 'json' and fname:match 'elm%.json$') then
+      if util.ft_matches(filetype, 'elm') or (util.ft_matches(filetype, 'json') and fname:match 'elm%.json$') then
```
justinmk, line 13: why can't we use vim.filetype.match with the `fname`, here? (note: the parens in the code below are important because match() has multi-value return)

    if 'elm' == (vim.filetype.match({filename=name}))

lua/lspconfig/util.lua
```diff
@@ -364,10 +377,8 @@ function M.get_other_matching_providers(filetype)
   for _, config in pairs(configs) do
     if not vim.tbl_contains(active_clients_list, config.name) then
       local filetypes = config.filetypes or {}
-      for _, ft in pairs(filetypes) do
-        if ft == filetype then
-          table.insert(other_matching_configs, config)
-        end
+      if M.ft_matches(filetype, filetypes) then
+        table.insert(other_matching_configs, config)
```
justinmk, line 381: these changes seem like they might be useful, but `ft_matches` could stay private, it doesn't need to be a new public function.

lua/lspconfig/server_configurations/elmls.lua
baodrate: Been a while since I looked at this code, but I think that should work. I just didn't want to change the semantics of the existing code. e.g. your suggested check wouldn't match files with a manually set `filetype`; although I'm not sure if that is a negligible difference or not. Would lspconfig even spawn a server for e.g. `foobar.txt` where the user calls `:set filetype=elm`?

lua/lspconfig/util.lua
baodrate: See: concerns raised in https://github.com/neovim/nvim-lspconfig/pull/3145#discussion_r1783179475

baodrate:
rebase is ready, pending answers to questions raised in https://github.com/neovim/nvim-lspconfig/pull/3145#discussion_r1783179475

justinmk:
> I think `vim.filetype.match` will also fail when `filetype` is set via autocmd rather than `filetype.lu`

`vim.filetype.get_option()` is designed to handle that situation. 

So, to be clear:

- it looks like your internal changes to util.lua in this PR could make sense
- but configs themselves should find another way to meet this use-case (and if you can find an satisfactory answer to that, we can document it in lspconfig's help/CONTRIBUTING.md/etc).
    - please take a close look at the functions on `vim.filetype.*`

lua/lspconfig/server_configurations/elmls.lua
justinmk: see https://github.com/neovim/nvim-lspconfig/pull/3145#issuecomment-2392302686

justinmk:
`lua/lspconfig/util.lua` is deprecated. This idea could make sense as an enhancement to Nvim core `vim.filetype` module, and/or `vim.lsp.config` or related features.

PR #3154, opened 2024-05-16
refactor: add compatible tbl_flatten and lsp_get_clients for new version 0.10

glepnir:
add compatible `util.tbl_flatten` and `lsp.get_active_clients` to avoid  deprecated error notify on 0.10 release.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -3,6 +3,7 @@ local validate = vim.validate
 local api = vim.api
 local lsp = vim.lsp
 local uv = vim.loop
+local nvim_nine = vim.fn.has 'nvim-0.9' == 1
 
 local is_windows = uv.os_uname().version:match 'Windows'
 
@@ -170,7 +171,7 @@ M.path = (function()
   end
 
   local function path_join(...)
-    return table.concat(vim.tbl_flatten { ... }, '/')
+    return table.concat(M.tbl_flatten { ... }, '/')
   end
 
   -- Traverse the path calling cb along the way.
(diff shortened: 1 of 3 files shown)


lua/lspconfig/util.lua
```diff
@@ -3,6 +3,7 @@ local validate = vim.validate
 local api = vim.api
 local lsp = vim.lsp
 local uv = vim.loop
+local nvim_nine_above = vim.verison().minor > 9
```
justinmk, line 6: lspconfig does not need to support 0.9.

also, use `has('nvim-0.9')`. checking only the minor version is fragile

glepnir: > lspconfig does not need to support 0.9.

take some days. Some package managers have not been updated yet

lua/lspconfig/util.lua
```diff
@@ -261,8 +262,12 @@ function M.search_ancestors(startpath, func)
   end
 end
 
+function M.tbl_flatten(t)
+  return nvim_nine and vim.tbl_flatten(t) or vim.iter(t):flatten():totable()
```
wookayin, line 266:
```suggestion
  return nvim_nine and vim.tbl_flatten(t) or vim.iter(t):flatten(math.huge):totable()
```

lua/lspconfig/util.lua
```diff
@@ -3,6 +3,7 @@ local validate = vim.validate
 local api = vim.api
 local lsp = vim.lsp
 local uv = vim.loop
+local nvim_nine = vim.fn.has 'nvim-0.9'
```
wookayin, line 6:
```suggestion
local nvim_nine = vim.fn.has('nvim-0.9') > 0
```
(a common error)

This evaluates to true in nvim-0.10 as well as in nvim-0.9.

Actually I would suggest: instead of using `has()`, we could do a nil check on `vim.iter`, see below comments.

lua/lspconfig/util.lua
```diff
@@ -261,8 +262,12 @@ function M.search_ancestors(startpath, func)
   end
 end
 
+function M.tbl_flatten(t)
```
wookayin, line 265: Consider:
```
M.tbl_flatten = vim.iter and (function()
   vim.iter(t):flatten(math.huge):totable()
end) or vim.tbl_flatten
```

wookayin (changes requested):

lua/lspconfig/util.lua
glepnir: just equal 1

wookayin:
As pointed in https://github.com/neovim/nvim-lspconfig/pull/3154#discussion_r1604304077, `nvim_nine` is true for all of nvim 0.9, 0.10, and 0.11, etc. So it ends up still using the deprecated `vim.tbl_flatten` in nvim-0.10, and `vim.iter()` not being called.

glepnir:
oops i didn't validate that. yup has seems like not correct

lorenzo-frittoli:
this is still broken

PR #3156, opened 2024-05-17
refactor: remove tbl_add_reverse_loopup

ofseed:
`vim.lsp.get_active_clients` will be deprecated in the future, and the deprecation warning will be shown in 0.11.


ofseed:
And the `vim.tbl_add_reverse_lookup` was also removed.

glepnir:
thanks but i already do it in #3154  could you just keep tbl_add_reverse_lookup

ofseed:
Of course

PR #3157, opened 2024-05-17
fix: use new version as check condition

glepnir:
`has('nvim-0.9')` also return 1 on 0.9 above ..

cc @wookayin


wookayin (approved):
Thanks for the quick fix! I can confirm it now works well.

One side note is the current behavior of `vim.iter:flatten()` is slightly different when user (mistakenly or intentionally) passes `nil` in the list: https://github.com/neovim/neovim/issues/28777, but hope this wouldn't be a big problem for most of the use cases.

PR #3163, opened 2024-05-19
docs: update ansiblels homepage url

kartikvashistha:
Updating the homepage for the `ansible-language-server` to the latest URL, as the currently stated URL points to an archived repository (whose repo description also now points to the latest URL as introduced by this change).

This can be further verified at the package's registry page here: https://www.npmjs.com/package/@ansible/ansible-language-server

### DIFF
--- a/doc/server_configurations.md
+++ b/doc/server_configurations.md
@@ -516,7 +516,7 @@ require'lspconfig'.angularls.setup{}
 
 ## ansiblels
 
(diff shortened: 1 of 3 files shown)


kartikvashistha:
Oops, I missed reading the above section in the docs 😅 . Should I revert the change under the `docs` folder, or let CI overwrite it?

PR #3169, opened 2024-05-22
fix(util) attempt to call `tbl_flatten` (a nil value)

Niclas3:
Problem: On nvim v10.0, function `tbl_flatten` doesn't return anything.
Reason:  In neovimv10.0 iter.lua:234, function `Iter:flatten` calls the 'error' and does not return anything value.


glepnir:
? ??  `tbl_flattern` is mark as depreacte in 0.10 we suggest use vim.iter

PR #3181, opened 2024-05-28
fix: avoid overriding global default capabilities

frangio:
A config's `default_config` acts as an override to the global `default_config` because of `'keep'` here:

https://github.com/neovim/nvim-lspconfig/blob/38de86f82efd9ba0881203767d6a8e1815abca28/lua/lspconfig/configs.lua#L60

Some configs were providing the entire capabilities table in their `default_config`. This is not necessary because the global `default_config` already includes them, and it was causing any of the user's changes to be overridden.

https://github.com/neovim/nvim-lspconfig/blob/38de86f82efd9ba0881203767d6a8e1815abca28/lua/lspconfig/util.lua#L19

The specific issue I was facing was that globally setting `snippetSupport = true` was ignored for rust_analyzer.

### DIFF
--- a/lua/lspconfig/server_configurations/elmls.lua
+++ b/lua/lspconfig/server_configurations/elmls.lua
@@ -1,9 +1,6 @@
 local util = require 'lspconfig.util'
-local lsp = vim.lsp
 local api = vim.api
 
-local default_capabilities = lsp.protocol.make_client_capabilities()
-default_capabilities.offsetEncoding = { 'utf-8', 'utf-16' }
 local elm_root_pattern = util.root_pattern 'elm.json'
 
 return {
@@ -22,6 +19,9 @@ return {
       skipInstallPackageConfirmation = false,
       disableElmLSDiagnostics = false,
       onlyUpdateDiagnosticsOnSave = false,
+    },
+    capabilities = {
+      offsetEncoding = { 'utf-8', 'utf-16' },
     },
   },
   docs = {
--- a/lua/lspconfig/server_configurations/arduino_language_server.lua
+++ b/lua/lspconfig/server_configurations/arduino_language_server.lua
@@ -1,9 +1,5 @@
 local util = require 'lspconfig.util'
 
-local default_capabilities = vim.lsp.protocol.make_client_capabilities()
-default_capabilities.textDocument.semanticTokens = vim.NIL
-default_capabilities.workspace.semanticTokens = vim.NIL
-
 return {
   default_config = {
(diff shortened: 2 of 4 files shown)


glepnir:
`snippetSupport ` is true by default

frangio:
That's not what I see. The return value of `vim.lsp.protocol.make_client_capabilities` says `snippetSupport = false` for me.

frangio:
It was false until https://github.com/neovim/neovim/pull/27339 which was merged to master only yesterday.

glepnir:
https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/protocol.lua#L741

I can't reproduce it show me true by default.

frangio:
https://github.com/neovim/neovim/blob/v0.10.0/runtime/lua/vim/lsp/protocol.lua#L744

Snippet support is not the point of this PR. I want to be able to globally set the default config and have that apply to all providers. That isn't working and this PR fixes that.

lua/lspconfig/server_configurations/elmls.lua
```diff
@@ -23,6 +24,7 @@ return {
       disableElmLSDiagnostics = false,
       onlyUpdateDiagnosticsOnSave = false,
     },
+    capabilities = default_capabilities,
```
glepnir, line 27: inline ` capabilities = { offsetEncoding =` and also other.

glepnir:
> Snippet support is not the point of this PR.

? you said you set this but it overide.

and this  not extend default capabilities  https://github.com/neovim/neovim/blob/7e44ab696a0488ad234b0915a8cf804fd6d79156/runtime/lua/vim/lsp/client.lua#L465

frangio:
The default capabilities are set in `util.lua`: https://github.com/neovim/nvim-lspconfig/blob/b124ef3bd4435a6db7ff03ea2f5a23e1e0487552/lua/lspconfig/util.lua#L19


And they get combined with each config here: https://github.com/neovim/nvim-lspconfig/blob/b124ef3bd4435a6db7ff03ea2f5a23e1e0487552/lua/lspconfig/configs.lua#L60

So there shouldn't be any missing capabilities. This is already how it's handled by most providers in lspconfig but the few included in this PR were not.

lua/lspconfig/server_configurations/elmls.lua
frangio: Done.

glepnir:
thanks

PR #3183, opened 2024-05-30
fix(eslint): useFlatConfig is not experimental anymore

saifulapm:
As [Support for the new ESLint flat config files has improved](https://github.com/microsoft/vscode-eslint?tab=readme-ov-file#version-305---pre-release)

```
eslint.experimental.useFlatConfig
```
should be 
```
eslint.useFlatConfig
```

I have tested with flat config. ```experimental``` not working anymore.

### DIFF
--- a/lua/lspconfig/server_configurations/eslint.lua
+++ b/lua/lspconfig/server_configurations/eslint.lua
@@ -71,9 +71,7 @@ return {
       validate = 'on',
       packageManager = nil,
       useESLintClass = false,
-      experimental = {
-        useFlatConfig = false,
-      },
+      useFlatConfig = false,
       codeActionOnSave = {
         enable = false,
         mode = 'all',
@@ -119,7 +117,7 @@ return {
         or vim.fn.filereadable(new_root_dir .. '/eslint.config.mts') == 1
         or vim.fn.filereadable(new_root_dir .. '/eslint.config.cts') == 1
       then
-        config.settings.experimental.useFlatConfig = true
+        config.settings.useFlatConfig = true
       end
 
       -- Support Yarn2 (PnP) projects


MrRoiz:
maybe related with this? 
image

This is all the config that I have for eslint 
image

d-mitrofanov-v:
It seems that the experimental setting shouldn't be deleted as it is [still used by old versions](https://github.com/microsoft/vscode-eslint?tab=readme-ov-file#version-305---pre-release) 
> The experimental settings eslint.experimental.useFlatConfig is deprecated and should only be used for ESLint versions >= 8.21 < 8.57.0.

d-mitrofanov-v:
@saifulapm @glepnir It seems that the experimental setting shouldn't be deleted as it is [still used by old versions](https://github.com/microsoft/vscode-eslint?tab=readme-ov-file#version-305---pre-release)

The experimental settings eslint.experimental.useFlatConfig is deprecated and should only be used for ESLint versions >= 8.21 < 8.57.0.

saifulapm:
Yes but if someone use old version, they can overwrite this in their own config. Default I think should need to be stick with latest version config.

MrRoiz:
> Yes but if someone use old version, they can overwrite this in their own config. Default I think should need to be stick with latest version config.

Is it not possible to support both and avoid that workaround in configs that needs to use those old versions?

saifulapm:
Yes.. we can keep both also.

saifulapm:
> maybe related with this? image
> 
> This is all the config that I have for eslint image

Can you share you full eslint lspconfig ?

MrRoiz:
@saifulapm well, that's all config that I sent to the eslint lsp actually, I'm using LazyVim, it just installs that lsp and calls the setup method with the data assigned to that key.

image

saifulapm:
@MrRoiz This issue happening because of [mason lspconfig](https://github.com/williamboman/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/eslint.lua)

I will submit a PR on mason repo. Until this is workaround

```lua
  eslint = {
    settings = {
      workingDirectories = { mode = "auto" },
      experimental = {
        useFlatConfig = false,
      },
    },
  },
```

MrRoiz:
@saifulapm not sure of how mason lspconfig can generate the issue, but anyways the workaround works. thanks! let me know if you figure out a solution to avoid that workaround.

RayGuo-ergou:
I think `eslint.experimental.useFlatConfig` should be added back, as they are deprecated not removed. 

see: https://github.com/microsoft/vscode-eslint/blob/main/%24shared/settings.ts#L161

thus the eslint-lsp server will still try to read it.

kesor:
This premature "fix" stole two days of my life. The new server is not yet released even, it is in pre-release, and the old server is broken because of this "fix". ROLL IT BACK.

kesor:
@glepnir please revert this.

glepnir:
it has been revert.

PR #3190, opened 2024-06-03
refactor(eslint): use vim.lsp.buf.code_action to implement :EslintFixAll

icholy:
The `lsp.util.buf_versions` table is deprecated and should not be used anymore. https://github.com/neovim/neovim/pull/28943. Simply switching to using `vim.b[bufnr].changedtick` is also problematic: https://github.com/neovim/neovim/issues/29163


icholy:
This doesn't have the same behaviour because `code_action` auto-fills the `context.diagnostics` value with the current line's diagnostics.

glepnir:
seems like those prs revert. so closed.

Issue #3191, opened 2024-06-04
sourcekit-lsp in monorepos


dirtyhabits97:
### Description

With the changes introduced in #3066 sourcekit-lsp fails in monorepos with multiple `Package.swift` files.

For example, I have the following structure:
```
coding_challenges/
  foo/Package.swift
  bar/Package.swift
  baz/Package.swift
```

When doing `cd coding_challenges/foo` and `nvim Package.swift` I'm seeing this in the `lsp.log` file:
```
[ERROR][2024-06-04 11:41:34] .../vim/lsp/rpc.lua:770	"rpc"	"/usr/bin/sourcekit-lsp"	"stderr"	"[2024-06-04 11:41:34.862] could not find manifest, or not a SwiftPM package: /Users/gonzalorehu/other_projects/coding_challenges\n"
[ERROR][2024-06-04 11:41:34] .../vim/lsp/rpc.lua:770	"rpc"	"/usr/bin/sourcekit-lsp"	"stderr"	"[2024-06-04 11:41:34.868] could not open compilation database for /Users/gonzalorehu/other_projects/coding_challenges/wc/Sources/WordCount.swift\n"
```

This is because the plugin now searches for `Package.swift` at the top of the git (mono)repo and there is none. It is not clear from #3066 why this order was modified, but reverting to the old order solves my issue:
```
root_dir = util.root_pattern('Package.swift', 'buildServer.json', 'compile_commands.json', '.git')
```

Moving `or util.find_git_ancestor(filename)` after `Package.swift` also seems to work:
```
    root_dir = function(filename, _)
      return util.root_pattern 'buildServer.json'(filename)
        or util.root_pattern('*.xcodeproj', '*.xcworkspace')(filename)
        -- better to keep it at the end, because some modularized apps contain multiple Package.swift files
        or util.root_pattern('compile_commands.json', 'Package.swift')(filename)
     -> or util.find_git_ancestor(filename)
    end,
```

In my opinion, we should maintain the original order, or at the very least push the `.git` search to the very end.
@wojciech-kulik thoughts? I saw you're running a complex setup [here](https://wojciechkulik.pl/ios/the-complete-guide-to-ios-macos-development-in-neovim), can you test if something breaks if we make this change?

wojciech-kulik:
Hi,

> It is not clear from https://github.com/neovim/nvim-lspconfig/pull/3066 why this order was modified

I left the comment here:
[...]

PR #3192, opened 2024-06-04
fix(sourcekit-lsp): use git as last resort for sourcekit-lsp.

dirtyhabits97:
Fixes a bug described in #3191

TL;DR: We should search for `Package.swift` before looking at the git root


wojciech-kulik (approved):

wojciech-kulik:
You should also update the order in docs, like here: https://github.com/neovim/nvim-lspconfig/pull/3066/files

wojciech-kulik (approved):

PR #3194, opened 2024-06-05
draft: fix(slint-lsp): set root dir to git ancestor

Tmpod:
Without this, the LSP doesn't resolve imports properly, at least on recent versions.


glepnir:
still have anything need change ?

Tmpod:
Actually no, I just forgot to unmark it as a draft :p

Issue #3087, opened 2024-03-22
Custom command name may not conform user command naming rules


blmarket:
### TLDR

I have a use case to register `commands` which is not actually user commands. In short, I wanna execute [lsp.start_client](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/manager.lua#L152) without calling [nvim_create_user_command](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/util.lua#L91).

### Minimal reproduction

note that I have my own LSP implementation, which is a part of the issue.

```lua
vim.cmd('packadd nvim-lspconfig')

local lspconfig = require('lspconfig')
local configs = require('lspconfig.configs')
local util = require("lspconfig.util")

-- My own lsp server for markdown
if not configs.blah then
  configs.blah = {
    default_config = {
      cmd = { 'blahblah' },
      filetypes = { 'markdown' },
      single_file_support = true,
      commands = {
        ['my_app/command1'] = { function(args, client_info)
        end },
        ['my_app/command2'] = { function(args, client_info)
        end },
      }
    },
  }
end

lspconfig.blah.setup({})

-- Codelens actions include `my_app/command1` which will be handled by custom commands
vim.api.nvim_create_autocmd('LspAttach', {
  callback = function(_)
    vim.api.nvim_create_autocmd({'TextChanged', 'InsertLeave'}, {
      callback = function(_)
        vim.lsp.codelens.refresh()
      end
    })
    vim.lsp.codelens.refresh()
  end
})
```

### The problem

With 0.9.5 it works fine, but with 0.10-dev it fails as it tries to register `my_app/command1` as user command and fail because of the command name. I don't know why 0.9.5 is working fine though...

### Context

I need a client-side integration to implement codelens action properly. (e.g. I'd like to use telescope to show the result etc...), and relying on [this neovim lsp client impl](https://github.com/neovim/neovim/blob/dc110cba3c0d48d7c9dbb91900f8be0cf6cf0c9b/runtime/lua/vim/lsp/client.lua#L864). So those commands are only to be run by codelens actions, not by user commands. I'm also fine if there's another way to add such custom command handlers without breaking neovim's user command name rules.

glepnir:
use vim.lsp.start instead

blmarket:
> use vim.lsp.start instead

Are there no LSP providing codelens with custom commands? Wondering this is something we can consider support in the future or never will be.

[...]

PR #3200, opened 2024-06-09
feat: do not process if commands[i] is func type

blmarket:
Context: https://github.com/neovim/nvim-lspconfig/issues/3087#issuecomment-2016041174

I want to set commands argument in [vim.lsp.ClientConfig](https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig) which has `table<string, fun>` type, which is not conforming to lspconfig's own commands type. lspconfig has to skip them for user command registrations.


glepnir:
lint fix.

blmarket:
Hello, can I get review again?

PR #3206, opened 2024-06-17
feat: add lsp-ai support

Robzz:
This PR adds support for the [lsp-ai](https://github.com/SilasMarvin/lsp-ai) language server. Since the server is language-agnostic, I've set the defaults to attach to all filetypes, and always run in single-file mode.

Do note that as is, this is not a full, functional configuration (it shouldn't error, just won't generate any completions), as there are parts for which I don't think there are good default values, so I just want to be sure you're good with this as long as I document it, since from a _quick_ look at the supported server list it looks like most (all ?) servers work out-of-the-box with the default config of an empty `setup{}` call. Then again this one is a bit unusual.  
In particular, a user would need to select an inference backend, which could either be some API requiring an API key or some other user-managed inference server (no sane defaults for us to provide here), or have `lsp-ai` do inference itself with `llama.cpp`, which requires a decent GPU as well as downloading a few GBs of model files from HuggingFace, which I don't think is a great default either.

### DIFF
--- a/lua/lspconfig/configs/lsp_ai.lua
+++ b/lua/lspconfig/configs/lsp_ai.lua
@@ -1,0 +1,27 @@
+return {
+  default_config = {
+    cmd = { 'lsp-ai' },
+    filetypes = {},
+    root_dir = nil,
+    single_file_support = true,
+    init_options = {
+      memory = {
+        file_store = vim.empty_dict(),
+      },
+      models = vim.empty_dict(),
+    },
+  },
+  docs = {
+    [[
+https://github.com/SilasMarvin/lsp-ai
+
+LSP-AI is an open source language server that serves as a backend for AI-powered functionality in your favorite code
+editors. It offers features like in-editor chatting with LLMs and code completions.
+
+
+You will need to provide configuration for the inference backends and models you want to use, as well as configure
+completion/code actions. See the [wiki docs](https://github.com/SilasMarvin/lsp-ai/wiki/Configuration) and
+[examples](https://github.com/SilasMarvin/lsp-ai/blob/main/examples/nvim) for more information.
+]],
+  },
+}


lua/lspconfig/server_configurations/lsp_ai.lua
```diff
@@ -0,0 +1,21 @@
+return {
+  default_config = {
+    cmd = { 'lsp-ai' },
+    filetypes = {},
+    root_dir = nil,
+    single_file_support = true,
+    init_options = {
+      memory = {
+        file_store = vim.fn.json_decode("{}"),
```
koozz, line 9: Fix for the lint issue:

```suggestion
        file_store = vim.fn.json_decode '{}',
```

Robzz: I did not notice that, that's fixed thanks.

justinmk:

Robzz:
Thanks for the heads up, will do and attempt to finish this in the next few days.

Robzz:
I finally got to play a bit with an up to date version of the language server to make sure everything works, seems OK. I still don't think there's much more that can be provided in terms of reasonable defaults, so this is the bare minimum required for the server to attach without complaints.

lua/lspconfig/configs/lsp_ai.lua
```diff
@@ -0,0 +1,27 @@
...
+  default_config = {
+    cmd = { 'lsp-ai' },
+    filetypes = {},
+    root_dir = nil,
+    single_file_support = true,
+    init_options = {
+      memory = {
+        file_store = vim.fn.json_decode '{}',
+      },
+      models = vim.fn.json_decode '{}',
```
justinmk, line 11:
```suggestion
      memory = {
        file_store = vim.empty_dict(),
      },
      models = vim.empty_dict(),
```

Robzz: Fixed and squashed everything together :+1:

lua/lspconfig/configs/lsp_ai.lua
```diff
@@ -0,0 +1,27 @@
...
+      },
+      models = vim.empty_dict(),
+    },
+  },
+  docs = {
+    [[
+https://github.com/SilasMarvin/lsp-ai
+
+LSP-AI is an open source language server that serves as a backend for AI-powered functionality in your favorite code
+editors. It offers features like in-editor chatting with LLMs and code completions.
```
justinmk, line 19: How does chat work? Is that a custom lsp request?

Robzz: There's one available (`document/textGeneration` IIRC) but it feels somewhat low-level, The more user friendly way is to define a prefix in config that enables the chat code action when present in the buffer.

lua/lspconfig/configs/lsp_ai.lua
```diff
@@ -0,0 +1,27 @@
...
+    root_dir = nil,
+    single_file_support = true,
+    init_options = {
+      memory = {
+        file_store = vim.empty_dict(),
+      },
+      models = vim.empty_dict(),
+    },
+  },
+  docs = {
```
khaneliman, line 14: Getting an issue downstream with this being a table instead of just a string, do we need to support the `desc` sometimes being a list of strings ?

justinmk: pr welcome

khaneliman: Was just double checking whether it was supposed to only be strings or you do expect tables for this, if it can be changed I can do a quick PR.

Robzz: Yeah that looks like a mistake on my part thanks for the PR :+1:

PR #3238, opened 2024-07-17
feat(leanls): add support for lakefile.toml-only packages

Julian:
This is partially a bugfix, as such packages would appear to function correctly because we fell back to `lean --server` -- but that is a lie, as it means `lake` doesn't control the LSP environment (and specifically it means that shared objects aren't correctly found for a package which prebuilds them with prebuildModules).

While we're here, remove said ancient Lean 4 logic for versions no one has used for years, simplifying the startup (and saving a subprocess). It'd have been better to fail loudly here rather than falling back to confusing errors.

### DIFF
--- a/lua/lspconfig/server_configurations/leanls.lua
+++ b/lua/lspconfig/server_configurations/leanls.lua
@@ -21,36 +21,11 @@ return {
         end
       end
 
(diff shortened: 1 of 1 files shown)


Julian:
(CC @hargoniX)

PR #3240, opened 2024-07-18
Added htmlangular filetype to emmet_language_server, emmet_ls, angularls and tailwindcss

Ask-786:
Added basic configurations to these servers for htmlangular filetype.


lua/lspconfig/server_configurations/tailwindcss.lua
```diff
@@ -25,6 +25,7 @@ return {
       'handlebars',
       'hbs',
       'html',
+      'htmlangular',
       -- 'HTML (Eex)',
```
glepnir, line 29: can you also remove these comments lines.

Ask-786: sure

PR #3241, opened 2024-07-18
feat!: remove `--preview` flag from Ruff server

dhruvmanila:
Now that the server is stabilized, the `--preview` flag isn't required to run it from [`v0.5.3`](https://github.com/astral-sh/ruff/releases/tag/0.5.3).

This PR removes the need of the `--preview` flag, update the documentation.

### Question

This is a breaking change because without the `--preview` flag, `ruff` will throw an error for previous versions. It's only from `0.5.3` that the flag isn't required.

Now, the reason the flag needs to be removed is because the flag itself would turn on preview features from both the linter and the formatter, as it should.

Can the maintainers guide me on how to go about making this kind of change? Is this ok to do so?

In our VS Code extension, we check the `ruff` version and only then use the `--preview` flag ([reference](https://github.com/astral-sh/ruff-vscode/blob/2dd2871fda974e4428bc34695dd711c3fb8a08cc/src/common/server.ts#L179-L185)).

### DIFF
--- a/lua/lspconfig/server_configurations/ruff.lua
+++ b/lua/lspconfig/server_configurations/ruff.lua
@@ -8,7 +8,7 @@ local root_files = {
 
 return {
   default_config = {
-    cmd = { 'ruff', 'server', '--preview' },
+    cmd = { 'ruff', 'server' },
     filetypes = { 'python' },
     root_dir = util.root_pattern(unpack(root_files)) or util.find_git_ancestor(),
     single_file_support = true,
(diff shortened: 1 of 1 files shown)


jdhao:
@dhruvmanila hi, I am following config from this change, but it does not work anymore.  With config in the doc, i see following error when I open python file:

>  LSP[ruff] Ruff received invalid client settings - falling back to default client settings.

My ruff version is 0.7.1, and instead the following config works 

```
require('lspconfig').ruff.setup({
  settings = {
    -- Ruff language server settings go here
  },
})

```

dhruvmanila:
@jdhao You only need to pass in the table if you want to add any client settings. For example,

```lua
require('lspconfig').ruff.setup({
  init_options = {
    settings = {
      lineLength = 100
    },
  }
})
```

Otherwise, you can just call setup without any arguments:

```lua
require('lspconfig').ruff.setup()
```

The example in the docs is mentioned to highlight where should the client settings go. Now, if it's empty, Neovim will serialize that into an empty _array_ (not an empty dictionary) and that's where the error is being raised. This is related to https://github.com/astral-sh/ruff/pull/11566 which solved it for an empty dictionary.

The config that you've provided works but if you include any client settings in it, they won't be considered because the server won't see it. Currently, the Ruff server supports reading settings only from `init_options.settings` field as the example states.

jdhao:
Okay, thanks for the explanation. But I think the general convention in neovim is that of empty Lua table is provided, the default setting will be used. So I am surprised that an empty table leads to errors. After providing some settings, I don't see the error anymore 👍

PR #3255, opened 2024-08-02
fix(texlab): remove deprecated texlab.auxDirectory

Thibaulltt:
I noticed this while debugging some texlab problems.

The configuration key 'texlab.auxDirextory' has been deprecated, in favor of 'texlab.build.auxDir', mirroring the latexmkrc's auxDir variable.

see: https://github.com/latex-lsp/texlab/wiki/Configuration#deprecated-texlabauxdirectory

### DIFF
--- a/lua/lspconfig/configs/texlab.lua
+++ b/lua/lspconfig/configs/texlab.lua
@@ -167,7 +167,6 @@ return {
           onSave = false,
           forwardSearchAfter = false,
         },
-        auxDirectory = '.',
         forwardSearch = {
           executable = nil,
           args = {},


justinmk:

Thibaulltt:
Hi, forgot to push my commit a couple weeks ago and had to rebase again. Should be good to go now :)

Thibaulltt:
Sorry for not mentionning it. It has a default value of `.` in texlab if not explicitly set, or if the user has no `latexmkrc` file. Since I use a `latexmkrc` file myself, this is auto-magically populated with the right value :)

Source: https://github.com/latex-lsp/texlab/wiki/Configuration#texlabbuildauxdirectory

Thibaulltt:
Should I maybe include it anyway and let the user override it in their own `setup({...})` call? Leaving it blank always ensures texlab is the one responsible for it, but I'd understand if you wanted a more complete config in this repo.

justinmk:
It's good to delegate to defaults where possible, we just need to mention that for future reference, when we change things. I will add your note to the commit message.

PR #3257, opened 2024-08-04
zk: implement ZkList command

WhyNotHugo:
Surfaces the `zk.list` LSP command in the same way as `zk.new` and `zk.index`.

### DIFF
--- a/lua/lspconfig/server_configurations/zk.lua
+++ b/lua/lspconfig/server_configurations/zk.lua
@@ -1,5 +1,13 @@
 local util = require 'lspconfig.util'
 
+local function find_zk_root(startpath)
+  for dir in vim.fs.parents(startpath) do
+    if vim.fn.isdirectory(vim.fs.joinpath(dir, '.zk')) == 1 then
+      return dir
+    end
+  end
+end
+
 return {
   default_config = {
     cmd = { 'zk', 'lsp' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/zk.lua
```diff
@@ -1,5 +1,13 @@
 local util = require 'lspconfig.util'
 
+local function find_zk_ancestor(startpath)
+  return util.search_ancestors(startpath, function(path)
```
glepnir, line 4: could you use vim.fs moudle replace util module here. It's helpful when we want remove some fs util functions in util.lua in future.

WhyNotHugo: Updated.

lua/lspconfig/server_configurations/zk.lua
```diff
@@ -1,5 +1,13 @@
 local util = require 'lspconfig.util'
 
+local function find_zk_root(startpath)
+  for dir in vim.fs.parents(startpath) do
+    if vim.fn.isdirectory(dir .. '/.zk') == 1 then
```
glepnir, line 5:
```suggestion
    if vim.fn.isdirectory(vim.fs.joinpath(dir, '.zk')) == 1 then
```

WhyNotHugo:
Thanks for the review

PR #3259, opened 2024-08-06
feat: add daedalus-language-server support

pm4rcin:
This PR adds support for https://github.com/kirides/DaedalusLanguageServer used by Gothic and Gothic II games. Please let me know if the `root_dir` was added correctly as I'm not sure if it's equivalent of steps described in README of mentioned repo.

### DIFF
--- a/lua/lspconfig/server_configurations/daedalus_ls.lua
+++ b/lua/lspconfig/server_configurations/daedalus_ls.lua
@@ -1,0 +1,33 @@
+local util = require 'lspconfig.util'
+local bin_name = 'DaedalusLanguageServer'
+
+if vim.fn.has 'win32' == 1 then
+  bin_name = bin_name + '.exe'
+end
+
+local root_files = {
+  'Gothic.src',
+  'Camera.src',
+  'Menu.src',
+  'Music.src',
+  'ParticleFX.src',
+  'SFX.src',
+  'VisualFX.src',
+}
+
+return {
+  default_config = {
+    cmd = { bin_name },
+    filetypes = { 'd' },
+    root_dir = util.root_pattern(unpack(root_files)),
+    settings = {
+      DaedalusLanguageServer = {
+        loglevel = 'debug',
+        inlayHints = { constants = true },
+        numParserThreads = 16,
+        fileEncoding = 'Windows-1252',
+        srcFileEncoding = 'Windows-1252',
+      },
+    },
+  },
+}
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/daedalus_ls.lua
```diff
@@ -0,0 +1,33 @@
+local util = require 'lspconfig.util'
+local bin_name = 'DaedalusLanguageServer'
+
+if vim.fn.has 'win32' == 1 then
+  bin_name = bin_name + '.exe'
+end
```
glepnir, line 6:
```suggestion

```

glepnir (commented):
you don't need do this now.

lua/lspconfig/server_configurations/daedalus_ls.lua
```diff
@@ -0,0 +1,33 @@
+local util = require 'lspconfig.util'
+local bin_name = 'DaedalusLanguageServer'
```
glepnir, line 2: inline

lua/lspconfig/server_configurations/daedalus_ls.lua
justinmk: also, if this line were kept, the operator is wrong (should be `..`)

pm4rcin: May I ask why should I remove it? I've checked some servers configs in repo and they had that. Or is it automatic now?

lua/lspconfig/server_configurations/daedalus_ls.lua
pm4rcin: I'm not sure what the request is but I guess I should add a newline between right? If yes then why you call it like that?

lua/lspconfig/server_configurations/daedalus_ls.lua
glepnir: which servers ?

glepnir: oh I see. they are also should be removed i will do it later. we use exepath to find server executable so no need it now

lua/lspconfig/server_configurations/daedalus_ls.lua
glepnir: I mean can just in cmd after win32 condition removed.

pm4rcin: Of course! I didn't see it since you highlighted those 2 lines together :D Fixed.

glepnir:
thanks

lua/lspconfig/server_configurations/daedalus_ls.lua
pm4rcin: Done.

PR #3261, opened 2024-08-07
feat: add tvm_ffi_navigator server

anirudhsundar:
Add the [TVM FFI navigator](https://github.com/tqchen/ffi-navigator) server configuration, which is used within the TVM project to jump to C++ definitions from python and vice versa

### DIFF
--- a/lua/lspconfig/server_configurations/tvm_ffi_navigator.lua
+++ b/lua/lspconfig/server_configurations/tvm_ffi_navigator.lua
@@ -1,0 +1,19 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'python', '-m', 'ffi_navigator.langserver' },
+    filetypes = { 'python', 'cpp' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/tvm_ffi_navigator.lua
```diff
@@ -0,0 +1,25 @@
...
+  default_config = {
+    cmd = { 'python', '-m', 'ffi_navigator.langserver' },
+    filetypes = { 'python', 'cpp' },
+    root_dir = function(fname)
+      local root_files = {
+        'pyproject.toml',
+        '.git',
+      }
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
+    end,
```
glepnir, line 13:
```suggestion
    root_dir = return util.root_pattern('pyproject.toml', '.git')
```

anirudhsundar: Yes, makes sense, thanks. I've made the changes

PR #3263, opened 2024-08-08
feat(texlab): give the find environments command a ui wrapper

DimitrisDimitropoulos:
This PR tries to improve #3225 by filtering the output and presenting it with `vim.lsp.util.open_floating_preview`. It is a pretty simple implementation that leverages builtin lua functions and basic API, therefore it should not be a burden to maintain. I have tested it locally without any major problems and i will keep testing, while making sure no issues pop up in this repo regarding this feature. I am really sure if the way the border style is setup is fitting. 
Here are some images:

I am totally open to any suggestions :)


DimitrisDimitropoulos:
#3274

PR #3265, opened 2024-08-09
fix(mesonlsp): update root directory pattern

on2e:
This PR adds a pattern for the meson.build files in the `root_pattern` functions of the `default_config` and `doc` tables for the `mesonlsp` and `swift_mesonls` servers.

meson.build files are the primary configuration files for the Meson build system and lspconfig should look for these files to determine the root of a project that uses Meson.

### DIFF
--- a/lua/lspconfig/server_configurations/mesonlsp.lua
+++ b/lua/lspconfig/server_configurations/mesonlsp.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'mesonlsp', '--lsp' },
     filetypes = { 'meson' },
-    root_dir = util.root_pattern('meson_options.txt', 'meson.options', '.git'),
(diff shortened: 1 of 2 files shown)


on2e:
Currently, lspconfig cannot find a root directory when opening a meson.build file with Neovim and failes to launch the `mesonlsp` server.

mochaaP:
This breaks subdirectory build files. (e.g. `nvim subdir/meson.build` will not detect the root directory correctly.)

justinmk:
Does re-ordering the `util.root_pattern` parameters fix it? Can you send a PR if so

mochaaP:
Tried it. Unfortunately, it does not.

A possible solution is to search using the previous pattern and fallback to find `meson.build` if all previous pattern failed to match.

PR #3266, opened 2024-08-12
feat: ruby_lsp works inside eruby files

adam12:
Since 0.17.5, Ruby LSP has supported running inside ERB files.

### DIFF
--- a/lua/lspconfig/server_configurations/ruby_lsp.lua
+++ b/lua/lspconfig/server_configurations/ruby_lsp.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'ruby-lsp' },
-    filetypes = { 'ruby' },
+    filetypes = { 'ruby', 'eruby' },
     root_dir = util.root_pattern('Gemfile', '.git'),
     init_options = {
       formatter = 'auto',


defsdoor:
What is your experience with ruby-lsp on .erb files ?  I'm running 0.17.12 and if I coerce nvim-lspconfig (via lazyvim) to use ruby-lsp for .erb (eruby) all it adds is errors on all <% - as if its not determining the file as Document::LanguageId::ERB

vinistock:
@defsdoor please check this issue https://github.com/Shopify/ruby-lsp/issues/2282. ERB support was added in v0.17.13, but if you have other gems constraining versions of `rbs`, `prism` or `sorbet-runtime` in your application, upgrading will fail and you will be unable to get the version of the language server that supports ERB.

adam12:
> What is your experience with ruby-lsp on .erb files ? I'm running 0.17.12 and if I coerce nvim-lspconfig (via lazyvim) to use ruby-lsp for .erb (eruby) all it adds is errors on all <% - as if its not determining the file as Document::LanguageId::ERB

I was able to use it without errors or warnings before submitting this PR, but we've noticed over in the ruby-lsp repo that there are similar errors which you are seeing.

I see those errors now locally, but I had updated a few things after patching lspconfig (as one does when they are procrastinating).

defsdoor:
Odd thing is that after poking things a bit it's now working fine for me on 0.17.12.....  And without this patch.

elken:
Following up from https://github.com/Shopify/ruby-lsp/pull/2432#issuecomment-2291318998 that it seems this PR should be reverted as it's proven quite a disruptive change that Neovim doesn't seem to be ready to handle yet.

PR #3271, opened 2024-08-20
feat: add tea-leaves support

svermeulen:
Added new server config for Teal language (named tea-leaves)


lua/lspconfig/server_configurations/tea_leaves.lua
```diff
@@ -0,0 +1,30 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = {
+      'tea-leaves',
+      -- Uncomment to enable logging in $HOME/.cache/tea-leaves
```
glepnir, line 7: It's better add these comments as description in docs.

PR #3272, opened 2024-08-20
test(ci): change to luajit-openresty

glepnir:
Problem: LuaJIT.org has stopped publishing release tarballs on their website.

Solution: use luajit-openresty instead of.


justinmk (approved):
No objection to this quickfix. Though we can in the future eliminate these steps (including luarocks) and migrate to Nvim's test suite, perhaps.

PR #3274, opened 2024-08-20
feat(clangd): add symbolInfo support

DimitrisDimitropoulos:
This PR adds support for the symbolInfo capability of clangd, for more documentation [see](https://clangd.llvm.org/extensions.html#symbol-info-request). It follows the same principle as #3263, although simpler. I think this is a great and simple addition, which will not add a ton of maintenance burden. Here are some images that showcase, what it does:
 

Thanks for your awesome work! :)

### DIFF
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -24,7 +24,7 @@ end
 local function symbol_info()
   local bufnr = vim.api.nvim_get_current_buf()
   local clangd_client = util.get_active_client_by_name(bufnr, 'clangd')
-  if not clangd_client then
+  if not clangd_client or not clangd_client.supports_method 'textDocument/symbolInfo' then
     return vim.notify('Clangd client not found', vim.log.levels.ERROR)
   end
   local params = vim.lsp.util.make_position_params()
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -27,11 +27,8 @@ local function symbol_info(bufnr)
   if not clangd_client then
     return vim.notify('Clangd client not found', vim.log.levels.ERROR)
   end
-  local pos = vim.api.nvim_win_get_cursor(0)
-  vim.lsp.buf_request(0, 'textDocument/symbolInfo', {
-    textDocument = { uri = vim.uri_from_bufnr(0) },
-    position = { line = pos[1] - 1, character = pos[2] },
-  }, function(err, res)
+  local params = vim.lsp.util.make_position_params()
+  vim.lsp.buf_request(0, 'textDocument/symbolInfo', params, function(err, res)
     if err or #res == 0 then
       -- Clangd always returns an error, there is not reason to parse it
       return
--- a/lua/lspconfig/server_configurations/clangd.lua
+++ b/lua/lspconfig/server_configurations/clangd.lua
@@ -83,7 +83,7 @@ return {
     },
     ClangdShowSymbolInfo = {
       function()
-        symbol_info(0)
+        symbol_info()
       end,
       description = 'Show symbol info',
     },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -21,6 +21,34 @@ local function switch_source_header(bufnr)
...
+  bufnr = util.validate_bufnr(bufnr)
+  local clangd_client = util.get_active_client_by_name(bufnr, 'clangd')
+  if not clangd_client then
+    return vim.notify('Clangd client not found', vim.log.levels.ERROR)
+  end
+  local pos = vim.api.nvim_win_get_cursor(0)
+  vim.lsp.buf_request(0, 'textDocument/symbolInfo', {
+    textDocument = { uri = vim.uri_from_bufnr(0) },
+    position = { line = pos[1] - 1, character = pos[2] },
+  }, function(err, res)
```
glepnir, line 34:
```suggestion
  local param = vim.lsp.util.make_position_params()
  vim.lsp.buf_request(0, 'textDocument/symbolInfo', param, function(err, res)
```

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -21,6 +21,34 @@ local function switch_source_header(bufnr)
   end
 end
 
+local function symbol_info(bufnr)
+  bufnr = util.validate_bufnr(bufnr)
+  local clangd_client = util.get_active_client_by_name(bufnr, 'clangd')
+  if not clangd_client then
```
glepnir, line 27:
```suggestion
  if not clangd_client or not client:support_method('textDocument/symbolInfo') then
```

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -21,6 +21,34 @@ local function switch_source_header(bufnr)
   end
 end
 
+local function symbol_info(bufnr)
+  bufnr = util.validate_bufnr(bufnr)
```
glepnir, line 25:
```suggestion
local function symbol_info(bufnr)
  local bufnr = vim.api.nvim_get_current_buf()
```
in usage you just pass 0 so no need receive a param

lua/lspconfig/server_configurations/clangd.lua
DimitrisDimitropoulos: I don't really understand what you suggest here, what is `client` and is there a method called `support_method`?

glepnir: sry a typo here `clang_client.supports_method`

lua/lspconfig/server_configurations/clangd.lua
```diff
@@ -21,6 +21,31 @@ local function switch_source_header(bufnr)
...
+    local name = string.format('name: %s', res[1].name) ---@type string
+    vim.lsp.util.open_floating_preview({ name, container }, '', {
+      height = 2,
+      width = math.max(string.len(name), string.len(container)),
+      focusable = false,
+      focus = false,
+      border = require('lspconfig.ui.windows').default_options.border or 'single',
+      title = 'Symbol Info',
+    })
+  end)
```
glepnir, line 46:
```suggestion
  clangd_client.request('textDocument/symbolInfo', params, function(err, res)
    if err or #res == 0 then
      -- Clangd always returns an error, there is not reason to parse it
      return
    end
    local container = string.format('container: %s', res[1].containerName) ---@type string
    local name = string.format('name: %s', res[1].name) ---@type string
    vim.lsp.util.open_floating_preview({ name, container }, '', {
      height = 2,
      width = math.max(string.len(name), string.len(container)),
      focusable = false,
      focus = false,
      border = require('lspconfig.ui.windows').default_options.border or 'single',
      title = 'Symbol Info',
    })
  end, bufnr)
```

DimitrisDimitropoulos:
Thanks for the review!

PR #3284, opened 2024-08-25
feat: improve default haxe_language_server init_options

tobil4sk:
Adds `.git` as a root directory pattern.

Previously, it would detect the root dir by matching with "*.hxml", however, it
would use "build.hxml" as the default `displayArguments` even though it may not
exist. This could cause the error:

```
haxe_language_server: -32603: Error: Could not process argument build.hxml (file not found)
Invalid character:
```

Now it will use the first ".hxml" file that is found in the project. It will
only do this if no `displayArguments` value has been set in the `setup()` call,
so it will still respect user set values.

If no hxml file is found, then it uses empty `displayArguments`, which is still
better than a broken configuration.

### DIFF
--- a/lua/lspconfig/server_configurations/haxe_language_server.lua
+++ b/lua/lspconfig/server_configurations/haxe_language_server.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'haxe-language-server' },
     filetypes = { 'haxe' },
-    root_dir = util.root_pattern '*.hxml',
+    root_dir = util.root_pattern('*.hxml', '.git'),
     settings = {
       haxe = {
         executable = 'haxe',
@@ -41,7 +41,7 @@ your project's root directory. If your file is named something different,
 specify it using the `init_options.displayArguments` setting.
 ]],
     default_config = {
-      root_dir = [[root_pattern("*.hxml")]],
+      root_dir = [[root_pattern("*.hxml", ".git")]],
     },
   },
 }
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/haxe_language_server.lua
```diff
@@ -1,18 +1,36 @@
 local util = require 'lspconfig.util'
 
+local function find_hxml(path)
+  return vim.fs.find(function(name)
```
glepnir, line 4: use vim.fs.root

lua/lspconfig/server_configurations/haxe_language_server.lua
```diff
@@ -1,18 +1,36 @@
...
         executable = 'haxe',
       },
     },
-    init_options = {
-      displayArguments = { 'build.hxml' },
-    },
+    init_options = {},
+    on_new_config = function(new_config, new_root_dir)
+      if new_config.init_options.displayArguments then
+        print 'displayArguments already set. Skipping'
```
glepnir, line 22: use vim.notify replace print

tobil4sk: This one I left in by mistake while debugging. I've replaced the other one with `vim.notify`.

lua/lspconfig/server_configurations/haxe_language_server.lua
tobil4sk: [`vim.fs.root`](https://neovim.io/doc/user/lua.html#vim.fs.root()) seems to return the parent directory, but I'd like to find a file. e.g. `/path/to/project/build.hxml`

PR #3288, opened 2024-08-31
Fixed typo in server_configurations.md

Ptitet:


glepnir:
modified tsserver.lua and then doc  will auto generate from ci

PR #3296, opened 2024-09-10
feat(ziggy): add ziggy and ziggy_schema support

rockorager:
Add support for the Ziggy data serialization format language server. The server supports both the ziggy file format and the schema format


glepnir:
thanks

PR #3299, opened 2024-09-12
feat: add c3-lsp support

xb-bx:
[c3-lsp](https://github.com/pherrymason/c3-lsp) support

### DIFF
--- a/lua/lspconfig/server_configurations/c3_lsp.lua
+++ b/lua/lspconfig/server_configurations/c3_lsp.lua
@@ -1,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/c3_lsp.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'c3lsp' },
+    root_dir = function(fname)
+      return util.root_pattern 'project.json'(fname) or util.root_pattern '.git'(fname)
```
glepnir, line 7: project.json and .git can combine into a table.

PR #3300, opened 2024-09-13
chore(nix): remove redundant flake-utils dependency; add nix-systems

NotAShelf:
Get rid of flake-utils, add nix-systems support & parameterize over system.

Tested devshell behaviour.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1694529238,
-        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
-        "owner": "numtide",
-        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


jacekpoz (approved):

Jappie3 (approved):

teto:
how does it differ from what we have now ? does it enable more platforms ? can't we do this with flake-utils ?

NotAShelf:
That's the point: we are trying to get rid of flake utils.

No point in having an entire library as an input when all we're using from its features is a single function that serves only ro abstract system, which we can already do ourselves.

diniamo (approved):
+1 for removing flake-utils. Even if the flake is more complex, flake-utils is bad. Flake-parts is the only abstraction I would agree to using. In this case though, the flake is simple enough that it could even be replaced by a shell.nix (which you might want to consider if you don't want to add more things to the flake).

NotAShelf:
That is a fair point, we could even get rid of systems input for a genAttrs with some default systems. I don't think the flake.nix is meant for for external party use in this case.

justinmk (approved):
waiting for @teto approval

teto:
I see this as harmless churn but it seems to make several persons happy so let's go for it.

Issue #2477, opened 2023-02-22
Conditionally disabling an LSP for certain buffers


LordMZTE:
### Requested feature

I'm looking to stop an LSP from attaching to a buffer (and being started if it's not already running) if a callback function I provide returns false. Specifically, I want to disable an LSP for a filename that matches a pattern.

justinmk:
one way would be to call stop() in your `on_attach` callback. See 

- https://github.com/neovim/nvim-lspconfig#suggested-configuration
- `:help vim.lsp.start_client()`

LordMZTE:
This isn't ideal, as the server is still started, but I guess it'll do.

glepnir:
set `auto_start` to false .create a Filetype event pattern is you filanme .then do config.luanch in callback

Issue #2437, opened 2023-02-06
Ability to disable lsp for some buffers


randomizedthinking:
### Requested feature

When working with repl, I have some utility buffers of the same type as the main files and the lsp server would generate many errors on them. I wish lsp can be turned off for such buffers.

The implementation could be simple. For instance, if a buffer is set with a local variable `b:lsp_disable`, then lsp would skip them. It would be nice if it is systematically implemented in nvim-lspconfig.

glepnir:
I think you can  simply use the vim diagnostic.disable with buffer number to disable all diagnostic.

respectTheCode:
The `vim.diagnostic.disable(0)` call is deprecated now and should be updated like this

```lua
local lsp_grp = vim.api.nvim_create_augroup("lsp_disable", { clear = true })
vim.api.nvim_create_autocmd({ "BufEnter", "BufNewFile" }, {
  group = lsp_grp,
  pattern = { ".env", ".env.*" },
  callback = function()
    vim.diagnostic.disable(vim.api.nvim_get_current_buf())
  end,
})
```

Issue #867, opened 2021-04-24
Is it possible to disable lsp in node_modules directory file?


ecosse3:
Hello,

I would like to disable LSP client or just efm langserver when I'm in the file which is inside node_modules directory.

Because of efm and prettier errors related to my config I can't read anything:
![image](https://user-images.githubusercontent.com/17691791/115966936-20121000-a530-11eb-8ba2-a6b480c08670.png)

Is it possible? I can provide my lsp/efm config if needed.
Maybe some particular vim BufEnter syntax to emit `:LspStop` inside node_modules dir file?

Thank you!

mjlbach:
Usage questions belong on our discourse, you can find the link in the readme.

________________________________
From: Łukasz Kurpiewski ***@***.***>
Sent: Saturday, April 24, 2021 10:08:10 AM
To: neovim/nvim-lspconfig ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [neovim/nvim-lspconfig] Is it possible to disable lsp in node_modules directory file? (#867)


Hello,

I would like to disable LSP client or just efm langserver when I'm in the file which is inside node_modules directory.

Because of efm and prettier errors related to my config I can't read anything:
[image]<https://user-images.githubusercontent.com/17691791/115966936-20121000-a530-11eb-8ba2-a6b480c08670.png>

Is it possible? I can provide my lsp/efm config if needed.
Maybe some particular vim BufEnter syntax to emit :LspStop inside node_modules dir file?

Thank you!

PR #3302, opened 2024-09-15
feat: allow conditional attachment of LSP servers

ronandalton:
This PR adds the ability for users to conditionally attach an LSP server to a buffer based on a user provided function that will be evaluated. This functionality is useful for disabling an LSP server for a particular project, or for files in a particular subdirectory. Currently there is no good way of doing this.

With this addition, disabling an LSP for a given project is easy:
```lua
server.should_attach = function()
  return not vim.g.disable_lsp
end
```

Then in the project's `.exrc` file the following can be added:
```vim
let g:disable_lsp = v:true
```

Related issues:
- https://github.com/neovim/nvim-lspconfig/issues/2477
- https://github.com/neovim/nvim-lspconfig/issues/2437
- https://github.com/neovim/nvim-lspconfig/issues/867

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -122,6 +122,14 @@ listed in `:help vim.lsp.start_client()` with the following unique entries:
     should likely be included in your new override. Some configurations
     use `on_new_config` to dynamically set or modify `cmd`.
 
+- {should_attach}
+
+    `function(filename, bufnr)` (default: function that returns true)
+
+    Called before the language server is attached to a buffer. If it evaluates
+    to false, the language server will not be attached. Useful for
+    conditionally enabling an LSP based on the current project.
+
 Note: all entries passed to `setup {}` override the entry in the default
 configuration. There is no composition.
 
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -14,6 +14,7 @@ local configs = {}
 --- @field autostart? boolean
 --- @field package _on_attach? fun(client: vim.lsp.Client, bufnr: integer)
 --- @field root_dir? string|fun(filename: string, bufnr: number)
+--- @field should_attach? fun(filename: string, bufnr: integer)
 
 --- @param cmd any
 local function sanitize_cmd(cmd)
--- a/lua/lspconfig/manager.lua
+++ b/lua/lspconfig/manager.lua
@@ -249,6 +249,10 @@ function M:try_add(bufnr, project_root)
   end
 
   local bufname = api.nvim_buf_get_name(bufnr)
+  if not self.config.should_attach(bufname, bufnr) then
+    return
+  end
+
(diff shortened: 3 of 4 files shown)


glepnir:
>  This functionality is useful for disabling an LSP server for a particular project, or for files in a particular subdirectory. Currently there is no good way of doing this.

emm i think there has a way to do .

```lua
vim.api.nvim_create_autocmd('LspAttach' ,
      callback = function(args)
              local bufnr = args.buf
              local bufname = vim.api.nvim_buf_get_name(bufnr)
              if should_not_attach then
                       vim.lsp.buf_detach_client(bufnr, args.data.client_id)
              end
      end

```

justinmk:
same answer as before https://github.com/neovim/nvim-lspconfig/issues/2477#issuecomment-1441603421

nvim-lspconfig is a bunch of configs. It's not a platform. It is not going to continue gaining random features.

Use `vim.lsp.start()` for full control.

ronandalton:
> emm i think there has a way to do .

Like I said in my PR, there isn't currently a good way of doing this. Starting the LSP and then stopping it again is not an ideal solution.

> nvim-lspconfig is a bunch of configs. It's not a platform. It is not going to continue gaining random features.
>Use vim.lsp.start() for full control.

lspconfig already has a bunch of code that handles launching of servers so the user doesn't need to worry about the technicalities of it. Using vim.lsp.start() directly would probably involve a bunch more custom code the user needs to write (and have the time and knowledge to know how to write). This PR only adds a few lines of code (really just 3) and solves this problem for these users. I don't really see how that can be a bad thing 🙂

glepnir:
this is not **stop** any client it just remove buffer of client on_attach

ronandalton:
Actually I tried that code but it doesn't seem to work. I just get an error (even after putting the callback in braces). `attempt to index local 'buf_state' (a nil value)`.

glepnir:
wrap in `vim.schedule(function() vim.lsp.buf_detach_client(bufnr, args.data.client_id) end)`

ronandalton:
Thanks, that worked. So just to clarify, this will stop the server from ever starting, correct? If this is the case then it seems to achieve the same goal as my code change. Given that this the solution you provided is robust, would it be a good idea to document it somewhere (perhaps in a FAQ if we have one)?

ronandalton:
The docs seem to say that LspAttach is triggered after an LSP client attaches to a buffer. That would make me think that the server was already started, unless the autocommands are run before doing that.

glepnir:
This will not prevent the server process from starting. It will just remove the buffer from the server's attatched_buffers. If you want to stop the client after remove buffer from server add check continue like 
`vim.tbl_isempty(client.attached_buffer) then vim.lsp.stop_client ` 

If you want more detailed control over the start and stop of the server, please read `:help vim.lsp.start` interface to customize , We focus on providing any useful capabilities.

PR #3303, opened 2024-09-16
Add C3 LSP

id3nom:
Language server for the C3 language.


glepnir:
duplicate there already have c3 lsp

PR #3307, opened 2024-09-17
Update server_configurations.md

tskj:
Note depreaction / rename in docs.

Should also arguably have a history of deprecations somewhere!


glepnir:
this is also not very useful . it's better show deprecate server in lspconfig  command this will very useful.

tskj:
I disagree. Yes, it's helpful to show the message where you put it and I don't suggest you remove it, but I strongly disagree that putting it in the documentation like I have suggested here isn't extremely useful.

It's very confusing for a user like me to come back to your documentation after many months and not find `tsserver` which I very vividly remember being here! And no explanation or note whatsoever about what happened to it. Remember, not every user knows everything you know about the project. Explicitness and clarity is useful and good.

glepnir:
I mean update the lspinfo and deprecate message . doc is genereate from bot . update there does not make sense wait for a while

tskj:
I am not entirely sure I understand what you mean by this, although I seem to gather that the file I changed is autogenerated? It would probably be best if you updated the correct place then, I'm not sure where that would be.

glepnir:
the document is generate from source lua file so edit it directly not a good idea now you can get the deprecate list in lspinfo instead search doc..

PR #3308, opened 2024-09-17
feat: show deprecate servers list in LspInfo

glepnir:
In order to smoothly transition to the vim.lsp.start interface in the future, adding checkhealth here is meaningless(because we already have checkhealth of lsp).  So shown deprecate servers list in LspInfo and update the server deprecate remove version to 0.2.1. 0.2.0 is tagged too quickly cause some servers not removed.

QQ_1726571465599

cc @tskj

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -34,18 +34,18 @@ local aliases = {
 }
 
 ---@return Alias
----@param name Get this alias, or nil to get all aliases that were used in the current session.
+---@param name string|nil get this alias, or nil to get all aliases that were used in the current session.
 M.server_aliases = function(name)
   if name then
     return aliases[name]
   end
   local used_aliases = {}
-  for name, alias in pairs(aliases) do
-    if deprecated.inconfig then
-      result[server_name] = deprecated
+  for sname, alias in pairs(aliases) do
+    if alias.inconfig then
+      used_aliases[sname] = alias
     end
   end
-  return result
+  return used_aliases
 end
 
 local mt = {}
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -39,8 +39,8 @@ M.server_aliases = function(name)
   if name then
     return aliases[name]
   end
-  local result = {}
-  for server_name, deprecated in pairs(aliases) do
+  local used_aliases = {}
+  for name, alias in pairs(aliases) do
     if deprecated.inconfig then
       result[server_name] = deprecated
     end
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -7,41 +7,51 @@ local M = {
 ---@class Alias
 ---@field to string The new name of the server
 ---@field version string The version that the alias will be removed in
----@param name string
----@return Alias
-local function server_alias(name)
-  local aliases = {
-    ['fennel-ls'] = {
-      to = 'fennel_ls',
-      version = '0.2.0',
-    },
-    ruby_ls = {
-      to = 'ruby_lsp',
-      version = '0.2.0',
-    },
-    ['starlark-rust'] = {
-      to = 'starlark_rust',
-      version = '0.2.0',
-    },
-    sumneko_lua = {
-      to = 'lua_ls',
-      version = '0.2.0',
-    },
-    tsserver = {
-      to = 'ts_ls',
-      version = '0.2.0',
-    },
-  }
+---@field inconfig? boolean need shown in lspinfo
+local aliases = {
+  ['fennel-ls'] = {
+    to = 'fennel_ls',
+    version = '0.2.1',
+  },
+  ruby_ls = {
+    to = 'ruby_lsp',
+    version = '0.2.1',
+  },
+  ['starlark-rust'] = {
+    to = 'starlark_rust',
+    version = '0.2.1',
+  },
+  sumneko_lua = {
+    to = 'lua_ls',
+    version = '0.2.1',
+  },
+  tsserver = {
+    to = 'ts_ls',
+    version = '0.2.1',
+  },
+}
 
-  return aliases[name]
+---@return Alias
+M.server_aliases = function(name)
+  if name then
+    return aliases[name]
+  end
+  local result = {}
+  for server_name, deprecated in pairs(aliases) do
+    if deprecated.inconfig then
+      result[server_name] = deprecated
+    end
+  end
+  return result
 end
 
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
-    local alias = server_alias(k)
+    local alias = M.server_aliases(k)
     if alias then
       vim.deprecate(k, alias.to, alias.version, 'lspconfig', false)
+      alias.inconfig = true
       k = alias.to
     end
 
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -32,6 +32,7 @@ local aliases = {
 }
 
 ---@return Alias
+---@param name Get this alias, or nil to get all aliases that were used in the current session.
 M.server_aliases = function(name)
   if name then
     return aliases[name]
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -4,6 +4,8 @@ local M = {
   util = require 'lspconfig.util',
 }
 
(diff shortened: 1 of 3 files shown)


lua/lspconfig.lua
```diff
@@ -7,41 +7,45 @@ local M = {
...
+    to = 'ts_ls',
+    version = '0.2.1',
+    inconfig = false,
+  },
+}
 
 local mt = {}
 function mt:__index(k)
+  if k == 'server_alias' then
+    return aliases
```
justinmk, line 42: this seems kind of strange. why not define `M.server_alias` ?

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -264,12 +264,22 @@ return function()
...
     '',
-    'Configured servers list: ' .. table.concat(util.available_servers(), ', '),
+    'Configured servers list: ' .. table.concat(available_servers, ', '),
   }
-
+  local deprecate_servers = {}
   vim.list_extend(buf_lines, matching_config_header)
+  local aliased = require('lspconfig').server_alias
+  for server_name, deprecate in pairs(aliased) do
+    if deprecate.inconfig then
```
justinmk, line 276: doesn't lspinfo.lua already show the "current connected" servers? why do we need to store state in `inconfig`, instead of just using `server_alias()` here just-in-time to see if a currently-connected server is deprecated?

glepnir: because available just show then server which list in configs. after deprecate there will show the new server name . like user configed `tsserver` but when they call LspInfo  available servers show `ts_ls` already.

lua/lspconfig.lua
```diff
@@ -7,41 +7,42 @@ local M = {
...
-    tsserver = {
-      to = 'ts_ls',
-      version = '0.2.0',
-    },
-  }
-
-  return aliases[name]
-end
+---@field inconfig boolean need shown in lspinfo
+M.aliases = {
```
justinmk, line 11: exposing this table globally means any code can modify it. When it was a function, it was read-only. Immutability is generally a good idea.

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -264,12 +264,21 @@ return function()
...
   end
 
+  local available_servers = util.available_servers()
   local matching_config_header = {
     '',
-    'Configured servers list: ' .. table.concat(util.available_servers(), ', '),
+    'Configured servers list: ' .. table.concat(available_servers, ', '),
   }
-
+  local deprecate_servers = {}
```
justinmk, line 272:
```suggestion
  local deprecated_servers = {}
```

lua/lspconfig.lua
```diff
@@ -7,41 +7,46 @@ local M = {
...
+}
 
-  return aliases[name]
-end
+M.server_aliases = setmetatable(aliases, {
+  __index = function(t, k)
+    return rawget(t, k) or nil
+  end,
+  __newindex = function()
+    vim.notify('Attempt to modify a frozen table aliases', vim.log.levels.ERROR)
```
justinmk, line 39: why are we doing this instead of just exposing it as a function?

glepnir: also thinking about this. one is closure one is freeze .  prefer use a closure ?

justinmk: freeze is uncommon, we should use plain old simple patterns except when we have a strong need to do something more clever.

lua/lspconfig.lua
```diff
@@ -7,41 +7,41 @@ local M = {
...
-  return aliases[name]
+M.server_aliases = function()
+  return aliases
 end
 
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
-    local alias = server_alias(k)
+    local alias = M.server_aliases()[k]
```
justinmk, line 41: if we return the entire list we didn't encapsulate anything, and the internal list can be modified.

```
---@return Alias
local function server_alias(name)
```

lua/lspconfig.lua
glepnir: Knowledge

lua/lspconfig/ui/lspinfo.lua
```diff
@@ -264,12 +264,21 @@ return function()
     end
   end
 
+  local available_servers = util.available_servers()
```
justinmk, line 267: what is the purpose of storing this in a new local?

glepnir: forgot to restore it. I wanted to do something here before.

lua/lspconfig.lua
glepnir: then in lspinfo there need a special param like `inconfig`  .. because there need a hole entry

glepnir: oh can be just nil

lua/lspconfig.lua
```diff
@@ -7,41 +7,51 @@ local M = {
...
+  },
+  tsserver = {
+    to = 'ts_ls',
+    version = '0.2.1',
+  },
+}
 
-  return aliases[name]
+---@return Alias
+M.server_aliases = function(name)
```
justinmk, line 35:
```suggestion
---@param name Get this alias, or nil to get all aliases that were used in the current session.
M.server_aliases = function(name)
```

lua/lspconfig.lua
```diff
@@ -7,41 +7,51 @@ local M = {
...
+  if name then
+    return aliases[name]
+  end
+  local result = {}
+  for server_name, deprecated in pairs(aliases) do
```
justinmk, line 40:
```suggestion
  local used_aliases = {}
  for old_name, alias in pairs(aliases) do
```

lua/lspconfig.lua
```diff
@@ -7,41 +7,51 @@ local M = {
...
+    if deprecated.inconfig then
+      result[server_name] = deprecated
+    end
+  end
+  return result
```
justinmk, line 45: didn't test, but something like this should work:

```suggestion
return vim.tbl_filter(function(v) return not not v.inconfig end, aliases)
```

glepnir: vim.iter is suggested ?

justinmk: I wasn't sure if lspconfig requires Nvim 0.10 now. But also, if there aren't multiple stages in the "pipeline", it's fine to use tbl_filter

glepnir: this is a map-like table not list-like table. tbl_filter will return a list-like table we need servername

lua/lspconfig.lua
```diff
@@ -4,44 +4,57 @@ local M = {
...
 
-  return aliases[name]
+---@return Alias
+---@param name Get this alias, or nil to get all aliases that were used in the current session.
+M.server_aliases = function(name)
+  if name then
+    return aliases[name]
+  end
+  local result = {}
+  for server_name, deprecated in pairs(aliases) do
```
justinmk, line 43:
```suggestion
  local used_aliases = {}
  for name, alias in pairs(aliases) do
```

lua/lspconfig.lua
glepnir:
```lua
function vim.tbl_filter(func, t)
  vim.validate({ func = { func, 'c' }, t = { t, 't' } })
  --- @cast t table<any,any>

  local rettab = {} --- @type table<any,any>
  for _, entry in pairs(t) do
    if func(entry) then
      rettab[#rettab + 1] = entry
    end
  end
  return rettab
end

```

vim.tbl_filter just return a list-like table with value in before `t` . then the server name as k will be ignored.

lua/lspconfig.lua
```diff
@@ -4,44 +4,57 @@ local M = {
...
-  return aliases[name]
+---@return Alias
+---@param name Get this alias, or nil to get all aliases that were used in the current session.
+M.server_aliases = function(name)
+  if name then
+    return aliases[name]
+  end
+  local used_aliases = {}
+  for name, alias in pairs(aliases) do
+    if deprecated.inconfig then
```
justinmk, line 44: code needs to be updated since the names were changed

justinmk (approved):

PR #3309, opened 2024-09-19
fix: add C3 library support

id3nom:
fix(c3_lsp): update root directory to be able to handle project in the following order - a C3 project 'project.json', a library '.c3l' dir, and ancestor 'git' repo.

### DIFF
--- a/lua/lspconfig/server_configurations/c3_lsp.lua
+++ b/lua/lspconfig/server_configurations/c3_lsp.lua
@@ -5,7 +5,7 @@ return {
     cmd = { 'c3lsp' },
     root_dir = function(fname)
       --look for a project directory
-      local path = util.root_pattern {"project.json"}(fname);
+      local path = util.root_pattern {'project.json', 'manifest.json'}(fname);
       if path ~= nil then
         return path;
       end
--- a/lua/lspconfig/server_configurations/c3_lsp.lua
+++ b/lua/lspconfig/server_configurations/c3_lsp.lua
@@ -4,7 +4,21 @@ return {
   default_config = {
     cmd = { 'c3lsp' },
     root_dir = function(fname)
-      return util.root_pattern { 'project.json', '.git' }(fname)
+      --look for a project directory
+      local path = util.root_pattern {"project.json"}(fname);
+      if path ~= nil then
+        return path;
+      end
+
+      --look for a library directory
+      path = util.root_pattern {"manifest.json"}(fname);
+      if path ~= nil then
+        return path;
+      end
+
+      --look for .git directory
+      return util.find_git_ancestors(fname);
+
     end,
     filetypes = { 'c3', 'c3i' },
   },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/c3_lsp.lua
```diff
@@ -4,7 +4,21 @@ return {
   default_config = {
     cmd = { 'c3lsp' },
     root_dir = function(fname)
-      return util.root_pattern { 'project.json', '.git' }(fname)
+      --look for a project directory
+      local path = util.root_pattern {"project.json"}(fname);
```
justinmk, line 8: why not do this? 
```suggestion
      local path = util.root_pattern {'project.json', 'manifest.json'}(fname);
```

lua/lspconfig/server_configurations/c3_lsp.lua
```diff
@@ -4,7 +4,14 @@ return {
...
     root_dir = function(fname)
-      return util.root_pattern { 'project.json', '.git' }(fname)
+      --look for a project directory
+      local path = util.root_pattern { 'project.json', 'manifest.json' }(fname)
+      if path ~= nil then
+        return path
+      end
+
+      --look for .git directory
+      return util.find_git_ancestors(fname)
```
glepnir, line 14:
```suggestion
      return util.root_patttern { 'project.json', 'manifest.json', '.git' }(fname)
```

id3nom: @glepnir will this also work if we are editing the branch inside the '.git' directory instead?

glepnir: this is an order will find every pattern until find root

lua/lspconfig/server_configurations/c3_lsp.lua
```diff
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'c3lsp' },
     root_dir = function(fname)
-      return util.root_pattern { 'project.json', '.git' }(fname)
+      return util.root_patttern { 'project.json', 'manifest.json', '.git' }(fname)
```
justinmk, line 7: this has a typo

glepnir: oops sry.

justinmk:
The [original form](https://github.com/neovim/nvim-lspconfig/pull/3309#discussion_r1766846520) of this PR is a hint that we may need to clean up more of the existing configs, since I assume that was copied from some other config.

glepnir:
This might take a while to clean up, as there are a lot of files. I'll come back here after I've prioritized C's Pr.

id3nom:
> The [original form](https://github.com/neovim/nvim-lspconfig/pull/3309#discussion_r1766846520) of this PR is a hint that we may need to clean up more of the existing configs, since I assume that was copied from some other config.

Yes indeed, it was based on existing configurations. tanks for your help @glepnir  and @justinmk

PR #3313, opened 2024-09-22
fix: root_dir support string value directly

glepnir:
root_dir in `vim.lsp.ClientConfig` can be a string type and in our annoation also mentioned `string|function` type. but actually root_dir does not support string type. add a type check of root_dir and then we can use `vim.fs.root` directly.


justinmk:
Why don't we fix the types instead of adding support for this? It's usually a mistake to support a string. Just require a list.

glepnir:
its `string|funtion` not list already add in annotation but string does not support . 

https://github.com/neovim/nvim-lspconfig/blob/f4fef355efa3c5d0813512480ee7b2c050b09fe4/lua/lspconfig/configs.lua#L8-L18

justinmk:
oh I see, `vim.lsp.ClientConfig` is part of Nvim lsp. so the function case is non-standard...

justinmk (approved):

PR #3315, opened 2024-09-22
Replace `cairo-language-server` with `scarb-cairo-language-server` and update installation docs

funkycadet:
This PR updates the Cairo Language Server (`cairo_ls`) configuration in `lspconfig` to use `scarb-cairo-language-server` instead of `cairo-language-server`. This change aligns with the new standard as Scarb is now the default package manager for Cairo, and the Scarb-distributed language server extension ensures compatibility with the version of the Cairo compiler used in Scarb.

* Replaced `cairo-language-server` with `scarb-cairo-language-server`, reflecting Scarb as the default package manager.
* Scarb's version of the language server is a launcher for the official Cairo Language Server, ensuring version compatibility with the compiler.
* Updated the documentation link for installing Cairo, following the latest official instructions.

This change simplifies user configuration and ensures they are working with the correct versions of the Cairo compiler and language server.

### DIFF
--- a/lua/lspconfig/server_configurations/cairo_ls.lua
+++ b/lua/lspconfig/server_configurations/cairo_ls.lua
@@ -1,6 +1,6 @@
 local util = require 'lspconfig.util'
 
-local bin_name = 'cairo-language-server'
+local bin_name = 'scarb-cairo-language-server'
 local cmd = { bin_name, '/C', '--node-ipc' }
 
 return {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/cairo_ls.lua
```diff
@@ -1,6 +1,6 @@
 local util = require 'lspconfig.util'
 
-local bin_name = 'cairo-language-server'
+local bin_name = 'scarb-cairo-language-server'
 local cmd = { bin_name, '/C', '--node-ipc' }
```
glepnir, line 4: these can be both removed just inline cmd into return cmd field.

funkycadet: Alright. Will do

glepnir:
thanks.

funkycadet:
its all good 🫡

PR #3316, opened 2024-09-23
fix(gopls): improve stability by fixing nil error when go env GOMODCACHE returns nothing

EJammy:
Currently, if `go env GOMODCACHE` prints nothing, the config will error. This PR will 1. retry with sync version 2. skip the special configuration case if mod_cache is nil. Even though @glepnir says fallback to sync will cause performance issue, which I agree, I think this is much better than just throwing an error. `go env GOMODCACHE` takes 0.001 ms on my machine.

Alternatively, we can remove 1. (the else case) and just keep mod_cache nil so that there are no performance issue, but we should definitely keep 2. (the nil check on mod_cache).

Related (partially fixes): #2733

### DIFF
--- a/lua/lspconfig/server_configurations/gopls.lua
+++ b/lua/lspconfig/server_configurations/gopls.lua
@@ -12,9 +12,11 @@ return {
         local result = async.run_command { 'go', 'env', 'GOMODCACHE' }
         if result and result[1] then
           mod_cache = vim.trim(result[1])
+        else
+          mod_cache = vim.fn.system 'go env GOMODCACHE'
         end
       end
-      if fname:sub(1, #mod_cache) == mod_cache then
+      if mod_cache and fname:sub(1, #mod_cache) == mod_cache then
         local clients = util.get_lsp_clients { name = 'gopls' }
         if #clients > 0 then
           return clients[#clients].config.root_dir


glepnir:
original issue seems like only on Windows ?

EJammy:
Seems like lots of people on Ubuntu encountered this issue. I think [this](https://github.com/neovim/nvim-lspconfig/issues/2733#issuecomment-2111245682) comment gives good insight. I bet this can be reproduced on any Linux system by getting go from snap and enabling AppArmor, but I got lazy lol... might try again later.

I just think it's pretty terrible from UX standpoint that it doesn't work out of the box and users have to look up the issue online to fix it. Also, the code just bothers me because `if result and result[1] then` check doesn't do anything, since if that condition is false mod_cache will be nil, and it will just error later.

Either way, I think it's better to consider the case where `go env GOMODCACHE` returns nothing.

glepnir:
I guess the possible reason is that the old GOPATH is not set, which causes GOMODCACHE to be empty?

EJammy:
I don't think that's the issue because running `go env GOMODCACHE` shows output. It's only when in async.run_command or, in the Ubuntu case, when redirecting output to a file in shell. I think the comment I link correctly identifies the problem on Ubuntu. Not sure why it is happening on Windows and Mac.

glepnir:
so problem in async.run_command ? i am consider replace it with vim.system but this need 0.10 basic version. but as a fallback i will merge this

PR #3317, opened 2024-09-23
fix(unocss): update filetypes

ryoppippi:
update filetypes of unocss lsp
Based on: 	
https://github.com/unocss/unocss/blob/35297359bf61917bda499db86e3728a7ebd05d6c/packages/vscode/src/autocomplete.ts#L12	
https://github.com/neovim/nvim-lspconfig/blob/dd329912c8d446240584a2dbcd3802af3a19105a/lua/lspconfig/server_configurations/tailwindcss.lua

### DIFF
--- a/lua/lspconfig/configs/unocss.lua
+++ b/lua/lspconfig/configs/unocss.lua
@@ -3,13 +3,31 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'unocss-language-server', '--stdio' },
+    -- copied from https://github.com/unocss/unocss/blob/35297359bf61917bda499db86e3728a7ebd05d6c/packages/vscode/src/autocomplete.ts#L12-L35
     filetypes = {
+      'erb',
+      'haml',
+      'hbs',
       'html',
+      'css',
+      'postcss',
+      'javascript',
       'javascriptreact',
-      'rescript',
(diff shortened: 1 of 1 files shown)


glepnir:
I am confused by similar behavior. Although this improves the out-of-the-box capabilities, it also bloats the table here. Many filetypes are not even in our built-in filetype support list. Maybe we should provide an append interface to achieve similar behavior instead of bloating it here.

ryoppippi:
@glepnir Thaks
So how can I do this.
Actually it is okay to configure by myself.
Or just add `astro` is fine for me.

glepnir:
Yes. Most of the time we recommend that users modify it locally. However, if the file type is supported by the built-in, it may be acceptable.

ryoppippi:
OK, so in this PR, I just add the ft which are listed in the official vscode extension

ryoppippi:
@glepnir Please check this out.

justinmk:
needs a rebase since https://github.com/neovim/nvim-lspconfig/pull/3330

ryoppippi:
Rebased done!

justinmk:
try moving the file to `lua/lspconfig/configs/unocss.lua`

ryoppippi:
@justinmk Hi! I just updated

PR #3318, opened 2024-09-23
feat(julials): add `JuliaActivateEnv` to activate a julia environment

mehalter:
This adds a command for the additional `julials` API for activating an environment. This is based on how extra functionality is added into the `texlab` configuration

### DIFF
--- a/lua/lspconfig/server_configurations/julials.lua
+++ b/lua/lspconfig/server_configurations/julials.lua
@@ -1,5 +1,58 @@
 local util = require 'lspconfig.util'
 
+local root_files = { 'Project.toml', 'JuliaProject.toml' }
+
+local function activate_env(path)
+  assert(vim.fn.has 'nvim-0.10' == 1, 'requires Nvim 0.10 or newer')
+  local bufnr = vim.api.nvim_get_current_buf()
+  local julials_clients = vim.lsp.get_clients { bufnr = bufnr, name = 'julials' }
+  assert(
+    #julials_clients == 0,
+    'method julia/activateenvironment is not supported by any servers active on the current buffer'
+  )
+  local function _activate_env(environment)
+    if environment then
+      for _, julials_client in ipairs(julials_clients) do
+        julials_client.notify('julia/activateenvironment', { envPath = environment })
+      end
+      vim.notify('Julia environment activated: \n`' .. environment .. '`', vim.log.levels.INFO)
+    end
+  end
+  if path then
+    path = vim.fs.normalize(vim.fn.fnamemodify(vim.fn.expand(path), ':p'))
+    local found_env = false
+    for _, project_file in ipairs(root_files) do
+      local file = vim.uv.fs_stat(vim.fs.joinpath(path, project_file))
+      if file and file.type then
+        found_env = true
+        break
+      end
+    end
+    if not found_env then
+      vim.notify('Path is not a julia environment: \n`' .. path .. '`', vim.log.levels.WARN)
+      return
+    end
+    _activate_env(path)
+  else
+    local depot_paths = vim.env.JULIA_DEPOT_PATH
+        and vim.split(vim.env.JULIA_DEPOT_PATH, vim.fn.has 'win32' == 1 and ';' or ':')
+      or { vim.fn.expand '~/.julia' }
+    local environments = {}
+    vim.list_extend(environments, vim.fs.find(root_files, { type = 'file', upward = true, limit = math.huge }))
+    for _, depot_path in ipairs(depot_paths) do
+      local depot_env = vim.fs.joinpath(vim.fs.normalize(depot_path), 'environments')
+      vim.list_extend(
+        environments,
+        vim.fs.find(function(name, env_path)
+          return vim.tbl_contains(root_files, name) and string.sub(env_path, #depot_env + 1):match '^/[^/]*$'
+        end, { path = depot_env, type = 'file', limit = math.huge })
+      )
+    end
+    environments = vim.tbl_map(vim.fs.dirname, environments)
+    vim.ui.select(environments, { prompt = 'Select a Julia environment' }, _activate_env)
+  end
+end
+
 local cmd = {
   'julia',
   '--startup-file=no',
@@ -45,9 +98,17 @@ return {
     cmd = cmd,
     filetypes = { 'julia' },
     root_dir = function(fname)
-      return util.root_pattern 'Project.toml'(fname) or util.find_git_ancestor(fname)
+      return util.root_pattern(unpack(root_files))(fname) or util.find_git_ancestor(fname)
     end,
     single_file_support = true,
+  },
+  commands = {
+    JuliaActivateEnv = {
+      activate_env,
+      description = 'Activate a Julia environment',
+      nargs = '?',
+      complete = 'file',
+    },
   },
   docs = {
     description = [[
@@ -70,6 +131,9 @@ Julia project, you must make sure that the project is instantiated:
 ```sh
 julia --project=/path/to/my/project -e 'using Pkg; Pkg.instantiate()'
 ```
+
+Note: The julia programming language searches for global environments within the `environments/`
+folder of `$JULIA_DEPOT_PATH` entries. By default this simply `~/.julia/environments`
     ]],
   },
 }


mehalter:
Looks like this fails the CI/CD check regarding the use of `dirname` and it potentially being used in the `root_dir`. I'm just confirming that the use of `dirname` does not go against the directive listed in the CI/CD check :)

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,34 @@
...
+        vim.notify('Julia environment activated: \n`' .. environment .. '`', vim.log.levels.INFO)
+      end
+    end)
+  else
+    vim.notify(
+      'method julia/activateenvironment is not supported by any servers active on the current buffer',
+      vim.log.levels.WARN
+    )
+  end
+end
```
glepnir, line 30: You can first check whether
```lua
if not julia_client then
      return vim.notify
end
```

mehalter: Revised the commit to follow this pattern. Thanks for the suggestion

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,34 @@
...
+  local julials_client = util.get_active_client_by_name(bufnr, 'julials')
+  if not julials_client then
+    vim.notify(
+      'method julia/activateenvironment is not supported by any servers active on the current buffer',
+      vim.log.levels.WARN
+    )
+    return
+  end
+  local julia_project_files = { 'Project.toml', 'JuliaProject.toml' }
+  local depot_envs = vim.fs.normalize(vim.env.JULIA_DEPOT_PATH or vim.fn.expand '~/.julia') .. '/environments'
```
justinmk, line 14: This kind of stuff (reading env var JULIA_DEPOT_PATH) is a bad sign. How are users going to know about that? That's a hidden assumption in the impl of this command. Should be documented at least, in the `docs` field below.

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,34 @@
 local util = require 'lspconfig.util'
 
+local function activate_env()
+  local bufnr = vim.api.nvim_get_current_buf()
+  local julials_client = util.get_active_client_by_name(bufnr, 'julials')
```
justinmk, line 5: use `vim.lsp.get_clients()`. This `util.get_active_client_by_name` function should be removed @glepnir

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,34 @@
...
+  if not julials_client then
+    vim.notify(
+      'method julia/activateenvironment is not supported by any servers active on the current buffer',
+      vim.log.levels.WARN
```
justinmk, line 9: Probably just assert()? The command shouldn't normally be available if there are no julia clients in the buffer, right?

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,34 @@
...
+    )
+    return
+  end
+  local julia_project_files = { 'Project.toml', 'JuliaProject.toml' }
```
justinmk, line 13: Here we are reinventing the root_dir on line 77--which also doesn't have JuliaProject.toml, which means this command has different expectations than the `root_dir` ...? Not good.

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -49,6 +78,14 @@ return {
     end,
     single_file_support = true,
   },
+  commands = {
+    JuliaActivateEnv = {
+      function()
+        activate_env()
+      end,
```
justinmk, line 85:
```suggestion
      activate_env
```

lua/lspconfig/server_configurations/julials.lua
mehalter: Ok, I can document it in the description. The environment variable `JULIA_DEPOT_PATH` is a special environment variable dictated by the programming language itself.

https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

lua/lspconfig/server_configurations/julials.lua
mehalter: `vim.notify` is a normal vim command that is available separately from the LSP client

lua/lspconfig/server_configurations/julials.lua
mehalter: Yeah, we can modify the `root_dir`. This is not necessarily the same as the root directory detection because it's not detecting a project root but an environment definition

lua/lspconfig/server_configurations/julials.lua
mehalter: Resolved in latest commit revision

lua/lspconfig/server_configurations/julials.lua
glepnir: In order to facilitate the transition of 0.9 users, which version will completely stop supporting 0.9x? 0.10.2 or 0.11.  then i can clean up utils including using vim.fs.

lua/lspconfig/server_configurations/julials.lua
justinmk: Yes, but when would `julials_client` ever be undefined? Commands defined in `commands` (I think?) are not defined unless their associated config is active for a buffer.

lua/lspconfig/server_configurations/julials.lua
mehalter: Updated in latest commit revision

justinmk: > which version will completely stop supporting 0.9x?

We can drop 0.9 support now, if we want. But that's separate from the first step, which is to remove all possible usages of vim.util *in the configs in this repo* :)

lua/lspconfig/server_configurations/julials.lua
mehalter: Ok, I added this documentation in the latest commit revision. Let me know if it sounds alright.

lua/lspconfig/server_configurations/julials.lua
mehalter: Oh I see, sounds good!

mehalter: Revised to being an `assert`

lua/lspconfig/server_configurations/julials.lua
justinmk: > It might be a good idea to keep the 0.9 support

Irrelevant for this PR. `vim.lsp.get_clients()` exists in 0.9

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,54 @@
 local util = require 'lspconfig.util'
 
+local function activate_env(path)
+  local bufnr = vim.api.nvim_get_current_buf()
+  local julials_clients = util.get_lsp_clients { bufnr = bufnr, name = 'julials' }
```
justinmk, line 5: `vim.lsp.get_clients()`

lua/lspconfig/server_configurations/julials.lua
glepnir: Nope. it only in 0.10.

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,54 @@
...
+        break
+      end
+    end
+    if not found_env then
+      vim.notify('Path is not a julia environment: \n`' .. path .. '`', vim.log.levels.WARN)
+      return
+    end
+    _activate_env(path)
+  else
+    local depot_paths = vim.env.JULIA_DEPOT_PATH and vim.split(vim.env.JULIA_DEPOT_PATH, ':')
```
justinmk, line 34: Is ":" the path separator that Julia uses on Windows?

This is why I'm so reluctant to merge these non-trivial command wrappers. They hide numerous subtle decisions that lead to bug reports later.

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,54 @@
...
+      or { vim.fn.expand '~/.julia' }
+    local environments = {}
+    vim.list_extend(environments, vim.fs.find(julia_project_files, { type = 'file', upward = true, limit = math.huge }))
+    for _, depot_path in ipairs(depot_paths) do
+      local depot_env = util.path.join(vim.fs.normalize(depot_path), 'environments')
```
justinmk, line 39: Use `vim.fs.joinpath`.

lua/lspconfig/server_configurations/julials.lua
mehalter: Ah, I misunderstood the previous discussion. I thought it should still use the `get_clients` utility since the rest of the codebase supports neovim 0.9 and then the 0.9 support would be dropped in a later PR

lua/lspconfig/server_configurations/julials.lua
mehalter: Yes that is the same path separator in windows as well. I apologize for the code doubling, it was me incorrectly implementing things the first time. But this is for Windows as well

lua/lspconfig/server_configurations/julials.lua
mehalter: resolved in latest commit

lua/lspconfig/server_configurations/julials.lua
mehalter: resolved in latest commit

lua/lspconfig/server_configurations/julials.lua
mehalter: > Irrelevant for this PR. `vim.lsp.get_clients()` exists in 0.9

I figured adding code that doesn't not match the current Neovim version support of the plugin would be irrelevant for this PR and a followup would make the blanket decision of which Neovim version to support. For now, I have removed the 0.9 support as requested

mehalter: If the idea is to remove all usages of `util` in the configuration. For now, should I include 0.9 support by checking for the old, deprecated function name?

lua/lspconfig/server_configurations/julials.lua
mehalter: I am going to go ahead and revert the use of `assert` because the original comment regarding the commands only being available when the language server is active is incorrect. I can revise it back to an `assert` if this is desired.

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,59 @@
...
+  if path then
+    path = vim.fs.normalize(path)
+    local found_env = false
+    for _, project_file in ipairs(julia_project_files) do
+      local file = (vim.uv or vim.loop).fs_stat(vim.fs.joinpath(path, project_file))
+      if file and file.type then
+        found_env = true
+        break
+      end
+    end
```
justinmk, line 31: can this loop be replaced by `vim.fs.find` ? Nvim versions before 0.10 are not supported (especially for commands)

lua/lspconfig/server_configurations/julials.lua
```diff
@@ -1,5 +1,59 @@
 local util = require 'lspconfig.util'
 
+local function activate_env(path)
```
justinmk, line 3:
```suggestion
local function activate_env(path)
  assert(vim.fn.has('nvim-0.10') == 1, 'requires Nvim 0.10 or newer')
```

then, remove the `vim.loop` usages below, `get_active_clients`, and any other outdated apis.

lua/lspconfig/server_configurations/julials.lua
justinmk: an assert() is sufficient.

lua/lspconfig/server_configurations/julials.lua
mehalter: I don't think so because `vim.fs.find` doesn't have a maximum depth. It has a `stop` parameter which stops when a given directory is reached but this really is only applicable doing an upwards search and there is no way to limit the depth it searches. Here we need to find project files just inside the directory that is passed in and not looking for them infinitely recursing down.

lua/lspconfig/server_configurations/julials.lua
mehalter: Removed support for <Neovim v0.10 in the command

PR #3322, opened 2024-09-28
fix(lua-language-server): update root directory pattern

### LINKED ISSUE #3165: possibly unintuitive lua_ls root_dir detection logic
### Description

I had some problems with the root dir detection of the `lua_ls` language server. I took a look at the [corresponding code](https://github.com/neovim/nvim-lspconfig/blob/9bda20fb967075355f253911bc066a8b5a03c77e/lua/lspconfig/server_configurations/lua_ls.lua#L17-L27) and was surprised that finding the git ancestor has lower priority than the `lua/` subdirectories (which, btw, is not mentioned in the [documentation](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#lua_ls)).

My specific problem with this: Consider the following directory structure:
```
HOME/
├─ workspaces/
│  ├─ lua/
│  ├─ python/
│  ├─ work/
│  │  ├─ SomeProjectWithLuaFiles/
│  │  │  ├─ .git/
```
The `lua/` and `python/` directories contain some miscellaneous language-specific projects I sometimes work on. The project directory under Work also contains some lua files. I expected the `.git` directory in the project to ensure that the project root dir is correctly detected to be `…/SomeProjectWithLuaFiles`. But since the [search for the `lua/` subdirectory](https://github.com/neovim/nvim-lspconfig/blob/9bda20fb967075355f253911bc066a8b5a03c77e/lua/lspconfig/server_configurations/lua_ls.lua#L22) is done first, my LSP detects `HOME/workspaces/` as the root directory.

I know I can easily change this, I was just wondering if this is intended? While I know that many lua projects have a `lua/` subdirectory, I feel that such language-specific workspace directories are not that uncommon, and that the search for the git ancestor should be done before looking for the `lua/` subdirectory.

Even if it is intended, I think this should be at least mentioned in the documentation for the default `lua_ls` configuration.

pitkling:
Fixes #3165 by returning the longer root path if both a `.git`- and a `lua`-ancestor are found (as [suggested by](https://github.com/neovim/nvim-lspconfig/issues/3165#issuecomment-2206937152) @powerman).

I also removed the documentation of the default `root_dir` pattern (assuming that now it will reference the source code, as in many other server configurations), since the root pattern calculation is more involved than the documentation indicated.

### DIFF
--- a/lua/lspconfig/configs/lua_ls.lua
+++ b/lua/lspconfig/configs/lua_ls.lua
@@ -19,13 +19,9 @@ return {
       if root and root ~= vim.env.HOME then
         return root
       end
-      local root_lua = util.root_pattern 'lua/'(fname)
-      local root_git = util.find_git_ancestor(fname)
-      if string.len(root_lua or '') >= string.len(root_git or '') then
-        return root_lua
-      else
-        return root_git
-      end
+      local root_lua = util.root_pattern 'lua/'(fname) or ''
+      local root_git = util.find_git_ancestor(fname) or ''
+      return #root_lua >= #root_git and root_lua or root_git
     end,
     single_file_support = true,
     log_level = vim.lsp.protocol.MessageType.Warning,
--- a/lua/lspconfig/configs/lua_ls.lua
+++ b/lua/lspconfig/configs/lua_ls.lua
@@ -19,11 +19,13 @@ return {
       if root and root ~= vim.env.HOME then
         return root
       end
-      root = util.root_pattern 'lua/'(fname)
-      if root then
-        return root
+      local root_lua = util.root_pattern 'lua/'(fname)
+      local root_git = util.find_git_ancestor(fname)
+      if string.len(root_lua or '') >= string.len(root_git or '') then
+        return root_lua
+      else
+        return root_git
       end
-      return util.find_git_ancestor(fname)
     end,
     single_file_support = true,
     log_level = vim.lsp.protocol.MessageType.Warning,
@@ -83,8 +85,5 @@ See `lua-language-server`'s [documentation](https://luals.github.io/wiki/setting
 * [Lua.workspace.library](https://luals.github.io/wiki/settings/#workspacelibrary)
 
 ]],
-    default_config = {
-      root_dir = [[root_pattern(".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml", "stylua.toml", "selene.toml", "selene.yml", ".git")]],
-    },
   },
 }


lua/lspconfig/server_configurations/lua_ls.lua
```diff
@@ -19,11 +19,13 @@ return {
...
-      root = util.root_pattern 'lua/'(fname)
-      if root then
-        return root
+      local root_lua = util.root_pattern 'lua/'(fname)
+      local root_git = util.find_git_ancestor(fname)
+      if string.len(root_lua or '') >= string.len(root_git or '') then
+        return root_lua
+      else
+        return root_git
       end
```
glepnir, line 28:
```suggestion
      local root_lua = util.root_pattern 'lua/'(fname) or ''
      local root_git = util.find_git_ancestor(fname) or ''
      return #root_lua >= #root_git and root_lua or root_git
```

pitkling: P.S.: The failing check seems to be because of the commit message format. I'll squash the second commit when there's nothing else I should change.

glepnir: Is there any use for the length check here? This should be a fallback priority issue. Shouldn't root_git continue to search only when root_lua fails to find it?

pitkling: To be honest not sure. That's why I opened #3165 a while back to ask this. If root_git is only used if root_lua fails (which was the case before), the directory structure I outlined in #3165 detects the (in my case at least) wrong root directory. As I mentioned there, the directory layout I use and that causes the problem seems not that uncommon. But then I'm only developing in Lua occasionally, so maybe that is okay and intended.

Anyway, in case this change is ok, one question concerning your suggested code change: This would now return the empty string instead of `nil` if neither a lua nor git ancestor is detected. Is that consistent with what `root_dir` is supposed to return?

pitkling:
> needs a rebase since #3330

@justinmk: Done, thanks for the heads up!

justinmk:
> I also removed the documentation of the default `root_dir` pattern

That's fine for now. We should really just autogenerate the root_dir docs by copying the literaly lua code that assigns `root_dir`, directly into the docs. Tracked in https://github.com/neovim/nvim-lspconfig/issues/2011 if anyone wants to help with that :)

folke:
This change is not incorrect. For luals to work properly with `lua/*` its root directory should be the parent of the `lua` directory.

folke:
Nevermind, this change does make sense, but seems to break for some people.
See https://github.com/folke/lazydev.nvim/issues/70

folke:
Fixed in #3391

PR #3325, opened 2024-09-29
fix(fennel-ls): use closest cfg as root directory and fallback to git repo root

justiceenunciate:
## Changes
* `fennel-ls` root directory  looks for closest `flsproject.fnl` first, and if it cannot find one, the root of the nearest git repository is used.
* The `fennel-ls` description now references this file and links to the configuration.

Special thanks to @XeroOl for helping me get started here. 🙂

### DIFF
--- a/lua/lspconfig/configs/fennel_ls.lua
+++ b/lua/lspconfig/configs/fennel_ls.lua
@@ -5,7 +5,10 @@ return {
     cmd = { 'fennel-ls' },
     filetypes = { 'fennel' },
     root_dir = function(dir)
-      return util.find_git_ancestor(dir)
+      local has_fls_project_cfg = function(path)
+        return util.path.is_file(vim.fs.joinpath(path, 'flsproject.fnl'))
+      end
+      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git')
     end,
     settings = {},
     capabilities = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/server_configurations/fennel_ls.lua
```diff
@@ -5,7 +5,10 @@ return {
     cmd = { 'fennel-ls' },
     filetypes = { 'fennel' },
     root_dir = function(dir)
-      return util.find_git_ancestor(dir)
+      local has_fls_project_cfg = function(x)
+        return util.path.is_file(util.path.join(x, "flsproject.fnl"))
```
glepnir, line 9:
```suggestion
        return util.path.is_file(vim.fs.joinpath(x, "flsproject.fnl"))
```

lua/lspconfig/server_configurations/fennel_ls.lua
```diff
@@ -5,7 +5,10 @@ return {
...
+      end
+      return util.search_ancestors(dir, has_fls_project_cfg) or util.find_git_ancestor(dir)
```
glepnir, line 11:
```suggestion
      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git')
```

glepnir (commented):
vim.fs have more higher priority since we plan to use vim.fs module replace most of functions in util.lua.  `is_file` is wrapper of `uv.fs_stat` so keep .

PR #3335, opened 2024-10-02
feat: expose config definition as `config_def`

justinmk:
## Problem:
Users/plugins may want to use the config definition without actually activating the config via setup(). https://github.com/neovim/nvim-lspconfig/pull/3330#issuecomment-2387285872

## Solution:
- Expose `config_def` field and document it.
- Also undeprecate some stuff that doesn't yet have a documented alternative.

---

TODO: configs.lua sets `M.filetypes = config.filetypes` and other fields in ad-hoc fashion "for :LspInfo" but it's not clear when those fields are actually populated, and they don't source from `config_def`...

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -26,6 +26,9 @@ local function sanitize_cmd(cmd)
   end
 end
 
+---@param t table
+---@param config_name string
+---@param config_def table Config definition read from `lspconfig.configs.<name>`.
 function configs.__newindex(t, config_name, config_def)
   validate {
     name = { config_name, 's' },
@@ -178,7 +181,7 @@ function configs.__newindex(t, config_name, config_def)
       end)
     end
 
-    -- Used by :LspInfo
+    -- Used by :LspInfo (evil, mutable aliases?)
     M.get_root_dir = get_root_dir
     M.filetypes = config.filetypes
     M.handlers = config.handlers
@@ -290,7 +293,9 @@ function configs.__newindex(t, config_name, config_def)
 
   M.commands = config_def.commands
   M.name = config_name
-  M.document_config = config_def
+  -- Expose the (original?) values of a config (non-active, or before `setup()`).
+  M.config_def = config_def
+  M.document_config = config_def -- For back-compat.
 
   rawset(t, config_name, M)
 end
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,7 @@
 root = true
 
 [*]
+max_line_length = 100
 indent_style = space
 indent_size = 2
 tab_width = 8
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -28,20 +28,26 @@ QUICKSTART                                                *lspconfig-quickstart*
 5. Run `:checkhealth lsp` to see the status or to troubleshoot.
 
 ==============================================================================
-THE SETUP METHOD                                               *lspconfig-setup*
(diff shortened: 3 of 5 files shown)


lua/lspconfig/configs.lua
```diff
@@ -178,7 +181,7 @@ function configs.__newindex(t, config_name, config_def)
       end)
     end
 
-    -- Used by :LspInfo
+    -- Used by :LspInfo (evil, mutable aliases?)
     M.get_root_dir = get_root_dir
     M.filetypes = config.filetypes
     M.handlers = config.handlers
```
justinmk, line 187: It's not clear why these fields don't merge the info from `config_def`. @glepnir do you know?

glepnir: should be merged with user config.  looks like a mistake here.  but the handlers are no longer managed by lspconfig. So it is useless.

justinmk: > but the handlers are no longer managed by lspconfig. So it is useless.

Interesting, can some code be deleted?

glepnir: see https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/denols.lua

we suggest use vim.lsp.handlers[method] = xxx, but lspinfo can not capture this override.

glepnir: If we want to show the override handler, I think we can locate the modified function file through debug.info in checkhealth? then i think we can remove handlers manage here.

justinmk: I'm trying to understand:

> handlers are no longer managed by lspconfig

in the denols.lua config you linked, I still see `config.handlers` being assigned.

glepnir: More like a bridge. It passes to vim.lsp.handlers via config. lspconfig should not do this. It is better to use vim.lsp.handlers directly in config.

there only `jdtls eslint denols relay_lsp koka grammarly` use config.handlers.

justinmk: could you create an issue (or PR to lspconfig.txt) that describes how users / config authors can do that to migrate their `lspconfig.handlers` ?

glepnir:
```diff
diff --git a/runtime/lua/vim/lsp/health.lua b/runtime/lua/vim/lsp/health.lua
index 18066a84d..faf43ddcd 100644
--- a/runtime/lua/vim/lsp/health.lua
+++ b/runtime/lua/vim/lsp/health.lua
@@ -159,6 +159,16 @@ local function check_position_encodings()
   end
 end

+function M.check_handlers_override()
+  local handlers = vim.lsp.handlers
+  for name, handler in ipairs(handlers) do
+    local source = debug.getinfo(handler, 'S').source
+    if not source:find('vim/lsp/handlers.lua') then
+      report_info(('Handler of method %s override in: %s '):format(name, source))
+    end
+  end
+end
+
 --- Performs a healthcheck for LSP
 function M.check()
   check_log()

```

an example which improve our checkhealth of vim.lsp to show the override handlers location. then we don't need to show it in lspinfo .this way is more like we want.

glepnir: `just `vim.lsp.handler[method]` replace `config_def.handlers[method]`  yes ..for more clear we need an issue or Pr for track..

justinmk: > for more clear we need an issue or Pr for track

yes,don't want to forget this thread.

PR #3338, opened 2024-10-02
feat: assert minimum required Nvim version

dundargoc:


dundargoc:
I anticipate the most annoying thing about this will be to remember to update this value. Maybe we could add a CI check for it.

lua/lspconfig.lua
```diff
@@ -76,4 +76,14 @@ function mt:__index(k)
   return configs[k]
 end
 
+local minimum_neovim_version = '0.10'
+if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
+  local msg = string.format(
+    'Neovim version is too low. Current version: %s. Required version: %s',
```
justinmk, line 82: need to mention lspconfig, I don't think notify() does that automatically?
```suggestion
    'nvim-lspconfig requires Nvim version %s, but you are running: %s',
```

lua/lspconfig.lua
```diff
@@ -76,4 +76,14 @@ function mt:__index(k)
   return configs[k]
 end
 
+local minimum_neovim_version = '0.10'
+if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
+  local msg = string.format(
+    'nvim-lspconfig requires Nvim version %s, but you are running: %s',
+    vim.version(),
+    vim.version.parse(minimum_neovim_version)
```
justinmk, line 84:
```suggestion
    vim.version.parse(minimum_neovim_version),
    vim.version()
```

lua/lspconfig.lua
```diff
@@ -76,4 +76,14 @@ function mt:__index(k)
...
 end
 
+local minimum_neovim_version = '0.10'
+if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
+  local msg = string.format(
+    'nvim-lspconfig requires Nvim version %s, but you are running: %s',
+    vim.version.parse(minimum_neovim_version),
+    vim.version()
+  )
+  vim.notify(msg, vim.log.levels.WARN)
```
justinmk, line 86: since we are using vim.uv, should it just be an assert?

dundargoc: I'm confused. What does this have to do with `vim.uv` (other than that the PR for it happened to inspire this one)? I imagine this to be a general-purpose warning that we can continually use whenever we need to bump the minimum version, rather than a one-time thing.

justinmk: since https://github.com/neovim/nvim-lspconfig/commit/c25b9dd6bddb125b538f3f01cb47b4a6e606f6c5 , there will be an error for any Nvim version that doesn't have vim.uv

so showing a warning here means that code will still run and produce an error

dundargoc: Oh derp, you're right. I understand what you mean. Yes, aborting/throwing an error sounds like the safest bet in that case.

dundargoc: Still not entirely sure if `vim.uv` is relevant, it just happens to be the last thing that broke (but in the future there might be some other function), so don't really see the value in hardcoding it to the existence of `vim.uv`. Let's throw an error instead of warning? I'll update the code and we can take a look.

dundargoc: What the hell, the output from `vim.notify` is awful when used with `vim.log.levels.ERROR`. We can't use this lol.

justinmk: > so don't really see the value in hardcoding it to the existence of `vim.uv`

it's just an example. don't need to hardcode anything ,just change vim.notify() to assert()

dundargoc: Sure, done

mehalter:
Just a heads up, the error message parsing doesn't work in older versions of Neovim. In Neovim <0.10 parsing `vim.version` as a string gives `table: 0x<hex>`

mehalter:
We could backport the new `__tostring()` method into the `lspconfig` utils so that it can be used the same until 0.10 is no longer supported at least

```lua
function version_string()
  local ret = table.concat({ self.major, self.minor, self.patch }, '.')
  if self.prerelease then
    ret = ret .. '-' .. self.prerelease
  end
  if self.build and self.build ~= vim.NIL then
    ret = ret .. '+' .. self.build
  end
  return ret
end
```

justinmk:
Thanks, yeah that would be helpful.

mehalter:
I can open a PR with this in like an hour if no one gets to it by then

mehalter:
Actually, I got a moment so I opened a quick one: https://github.com/neovim/nvim-lspconfig/pull/3345

PR #3339, opened 2024-10-02
feat(lspinfo): replace :LspInfo with :checkhealth

justinmk:
Problem:
:LspInfo has its own "inner platlform" of highlights, mappings etc. And it doesn't integrate with :checkhealth.

Solution:
- Move the lspinfo code to a healthcheck.
- LspInfo features such as "q" to quit, highlights, etc., should be added to :checkhealth in Nvim core.


glepnir:
It is definitely the direction to move forward. But the timing is not quite right. Because core checkhealth needs some improve.  avoid some noisy issues, it's better after core checkhealth improved.

justinmk:
> Because core checkhealth needs some improve. avoid some noisy issues, it's better after core checkhealth improved.

which issues exactly should block this?

glepnir:
the configured servers list and depreceted servers list is missing in checkhealth. may need  an integration.

justinmk:
> the configured servers list and depreceted servers list is missing in checkhealth. may need an integration

? The code in this PR is the same as the old LspInfo implementation. So configured & deprecated should display, or there's a bug.

lua/lspconfig/health.lua
```diff
@@ -277,29 +254,7 @@ return function()
     vim.list_extend(buf_lines, { '', 'Deprecated servers: ' .. table.concat(deprecated_servers, ' ') })
```
justinmk, line 254: deprecated servers

glepnir: Oh yes. i didn't look at code carefully.

glepnir (approved):

.editorconfig
```diff
@@ -1,7 +1,7 @@
 root = true
 
 [*]
-max_line_length = 100
+max_line_length = 120
```
dundargoc, line 4: Super nitpicky, but any real reason why line length is 120 for this repo but 100 in neovim? Not really biased towards any value, just that the discrepancy raised eyebrows.

justinmk: No reason except that's what this repo defined for .stylua so I just copied it here. Changing it might involve some churn

lua/lspconfig/health.lua
```diff
@@ -1,12 +1,14 @@
...
 local util = require 'lspconfig.util'
 
 local error_messages = {
-  cmd_not_found = 'Unable to find executable. Please check your path and ensure the server is installed',
-  no_filetype_defined = 'No filetypes defined, Please define filetypes in setup()',
+  cmd_not_found = 'Unable to find executable. Check your $PATH and ensure the server is installed.',
+  no_filetype_defined = 'No filetypes defined. Define filetypes in setup().',
   root_dir_not_found = 'Not found.',
-  async_root_dir_function = 'Asynchronous root_dir functions are not supported in :LspInfo',
+  async_root_dir_function = 'Asynchronous root_dir functions are not supported by `:checkhealth lspconfig`',
```
justinmk, line 11: I need help with confirming that these messages still work correctly and are displayed nicely. Probably they should be converted to `health.error()` calls.

daviareias:
any tips to restore the highlights? It's really hard to read the buffer returned by `LspInfo` now

daviareias:
> what highlights specifically

Sorry I didn't explain my situation correctly.

before when I use to type LspInfo, it would show a nice floating window, now it shows a buffer like this:
Screenshot 2024-10-08 at 4 49 53 PM

I find it a little hard to read, and couldn't find how to customize it by putting in a floating window like this:
Screenshot 2024-10-08 at 4 51 18 PM

glepnir:
This is how it works now. CheckHealth is displayed in the normal window.

justinmk:
You still didn't mention which highlights.

The door is open for adding features to `:checkhealth`, that needs to be done in Nvim core.

nvim-enjoyer:
removing the floating window option and moving to a checkhealth looks way worse. 

i get the idea of moving to `:che` but could that not have been done AFTER floating window support in nvim core???

glepnir:
Please be patient. I will try to add a floating window for checkhealth in core and hope it will be in time for 0.11.

909oce:
> Please be patient. I will try to add a floating window for checkhealth in core and hope it will be in time for 0.11.
> 
> Update: early draft it works.
> 
> QQ_1730802329055

is this available in the -git version? if so, is there any documentation on how to enable it please?

GrzegorzKozub:
I was using `require('lspconfig.ui.windows')` to set the border of the LSP information float that is displayed when I hit the `K` button like so:

```lua
require('lspconfig.ui.windows').default_options.border = 'rounded'
```

 I prefer using the `NormalFloat` highlight the same as my `Normal` (editor background) highlight and just have a rounded border.

Anyone knows if there's another way to change this border? The default seems `none`.

justinmk:
> I also noticed that some configs in this repo still use this API, for example:

Thanks for mentioning that. Fixed in https://github.com/neovim/nvim-lspconfig/pull/3454

> I was using `require('lspconfig.ui.windows')` to set the border of the LSP information float that is displayed when I hit the `K` button like so:

K is unrelated to this repo. This repo is just a collection of configs. K is a default mapping provided by Nvim core, which calls `vim.lsp.buf.hover()`. See also https://github.com/neovim/neovim/pull/31074

PR #3340, opened 2024-10-02
refactor!: replace util.path.join with vim.fs.joinpath

dundargoc:
We should rely on builtin functionality if possible to minimize code duplication.

### DIFF
--- a/lua/lspconfig/configs/apex_ls.lua
+++ b/lua/lspconfig/configs/apex_ls.lua
@@ -7,7 +7,7 @@ return {
     on_new_config = function(config)
       if not config.cmd and config.apex_jar_path then
         config.cmd = {
-          vim.env.JAVA_HOME and util.path.join(vim.env.JAVA_HOME, 'bin', 'java') or 'java',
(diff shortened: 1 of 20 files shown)


dundargoc:
Unsure what general strategy is for removing lspconfig functions. This seems used fairly frequently: https://github.com/search?q=util.path.join+language%3ALua&type=code. Worth deprecating, or do we just go for it?

justinmk:
> what general strategy is for removing lspconfig functions

we should keep the util stuff around to avoid breaking downstream. meanwhile, in this repo:

- CI should fail if any config uses a deprecated `util` function
- we should remove all docs that suggest a deprecated `util` function

justinmk:
To be clear, my above comment is in favor of this PR. When you asked about "removing lspconfig functions" I interpreted that to mean removing the `util.xx` entirely

justinmk:
OTOH https://github.com/neovim/nvim-lspconfig/issues/3342 is making me think that 0.9 support is still worthwhile, if it doesn't add too much burden.

PR #3345, opened 2024-10-03
fix: support printing version tables in neovim <0.10

mehalter:
Follow up to #3338

### DIFF
--- a/lua/lspconfig.lua
+++ b/lua/lspconfig.lua
@@ -48,6 +48,20 @@ M.server_aliases = function(name)
   return used_aliases
 end
 
+-- Temporary port of Nvim 0.10 vim.version:tostring.
+---@param version vim.Version
+local function version_string(version)
+  assert(version.major and version.minor and version.patch, 'invalid vim.Version table')
+  local ret = table.concat({ version.major, version.minor, version.patch }, '.')
+  if version.prerelease then
+    ret = ret .. '-' .. version.prerelease
+  end
+  if version.build and version.build ~= vim.NIL then
+    ret = ret .. '+' .. version.build
+  end
+  return ret
+end
+
 local mt = {}
 function mt:__index(k)
   if configs[k] == nil then
@@ -80,8 +94,8 @@ local minimum_neovim_version = '0.9'
 if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
   local msg = string.format(
     'nvim-lspconfig requires Nvim version %s, but you are running: %s',
-    vim.version.parse(minimum_neovim_version),
-    vim.version()
+    minimum_neovim_version,
+    vim.version and version_string(vim.version()) or 'older than v0.5.0'
   )
   error(msg)
 end


lua/lspconfig/util.lua
```diff
@@ -428,4 +428,18 @@ function M.strip_archive_subpath(path)
   return path
 end
 
+-- Replace with built in vim.version tostring that was added in neovim v0.10
+-- Used mainly for printing out of date version, so after v0.10 is no longer
+-- supported at least
+function M.version_string(version)
```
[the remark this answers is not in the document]
mehalter: added the valid luadoc as well as the assertion for validating the table

lua/lspconfig/util.lua
```diff
@@ -428,4 +428,20 @@ function M.strip_archive_subpath(path)
   return path
 end
 
+-- Replace with built in vim.version tostring that was added in neovim v0.10
+-- Used mainly for printing out of date version, so after v0.10 is no longer
+-- supported at least
+---@param version vim.Version
+function M.version_string(version)
```
justinmk, line 435: don't want a new "public" API, please make this a `local function` in lspconfig.lua

lua/lspconfig/util.lua
```diff
@@ -428,4 +428,20 @@ function M.strip_archive_subpath(path)
...
+  assert(version.major and version.minor and version.patch, 'version_string called with an invalid vim.Version table')
```
justinmk, line 436:
```suggestion
  assert(version.major and version.minor and version.patch, 'invalid vim.Version table')
```

lua/lspconfig/util.lua
```diff
@@ -428,4 +428,20 @@ function M.strip_archive_subpath(path)
   return path
 end
 
+-- Replace with built in vim.version tostring that was added in neovim v0.10
+-- Used mainly for printing out of date version, so after v0.10 is no longer
+-- supported at least
```
justinmk, line 433:
```suggestion
-- Temporary port of Nvim 0.10 vim.version:tostring.
```

mehalter: Revised commit with recommended rewording

lua/lspconfig/util.lua
mehalter: resolved in latest commit

lua/lspconfig/util.lua
mehalter: This is true, I moved it to being a local function. I also added checking to make sure the minimum version is able to be parsed as well

lua/lspconfig.lua
```diff
@@ -78,10 +92,12 @@ end
 
 local minimum_neovim_version = '0.9'
 if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
+  local minimum_vim_version = vim.version.parse(minimum_neovim_version)
```
justinmk, line 95: `vim.version.parse` was not available in older versions, that's why it was intentionally not used.

mehalter: oh interesting, `vim.version.parse` was used. I was just moving it. How would you like to resolve it?

mehalter: Ok, I moved it to just use the string rather than using `vim.version.parse` like the code was using originally

lua/lspconfig.lua
```diff
@@ -80,8 +94,8 @@ local minimum_neovim_version = '0.9'
 if vim.fn.has('nvim-' .. minimum_neovim_version) == 0 then
   local msg = string.format(
     'nvim-lspconfig requires Nvim version %s, but you are running: %s',
-    vim.version.parse(minimum_neovim_version),
-    vim.version()
+    minimum_neovim_version,
+    version_string(vim.version())
```
justinmk, line 98: `vim.version()` has been around [longer](https://github.com/neovim/neovim/commit/61437c20b5936de1f833d155b7287d9151e763d8) than `.parse()`, so that's why it's ok to use it. But to be extra safe, we could do:

```suggestion
    version_string(vim.version and vim.version() or 'older than v0.5.0')
```

mehalter: I like it, I'll add this with some fixes

PR #3347, opened 2024-10-04
Revert "fix: autostart=false: attach when editing new, nonexistent file"

glepnir:
Reverts neovim/nvim-lspconfig#2712


glepnir:
this pr will cause invalid buffer id

PR #3352, opened 2024-10-04
fixup: quick update, squash later

dundargoc:


lua/lspconfig/configs.lua
```diff
@@ -1,6 +1,6 @@
 local util = require 'lspconfig.util'
 local async = require 'lspconfig.async'
-local api, validate, lsp, uv, fn = vim.api, vim.validate, vim.lsp, (vim.uv or vim.loop), vim.fn
+local api, validate, lsp, uv, fn = vim.api, vim.validate, vim.lsp, vim.uv, vim.fn
```
justinmk, line 3: does 0.9 have vim.uv ?

dundargoc: No. I was just testing if the CI catches this fact. Should've done this in private but was too lazy. Will close PR after confirmation.

dundargoc:
Yup, the 0.9.5 works correctly. We could limit the `v0.9.5` test to a single OS but it's quick so probably not necessary.

PR #3354, opened 2024-10-05
feat(regal): add regal/startDebugging handler

rinx:
Regal has codelens feature to start a DAP debugging session.
https://github.com/StyraInc/regal/blob/f70b8928568a4286bef0bd531b8f06810d1e38f7/internal/lsp/server.go#L755

To handle this request, this pull request adds `regal/startDebugging` handler to default config to regal.lua.

### DIFF
--- a/lua/lspconfig/configs/regal.lua
+++ b/lua/lspconfig/configs/regal.lua
@@ -8,6 +8,26 @@ return {
       return util.root_pattern '*.rego'(fname) or util.find_git_ancestor(fname)
     end,
     single_file_support = true,
+    handlers = {
+      ['regal/startDebugging'] = function(_, result)
+        if not result then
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/regal.lua
```diff
@@ -8,6 +8,28 @@ return {
       return util.root_pattern '*.rego'(fname) or util.find_git_ancestor(fname)
     end,
     single_file_support = true,
+    handlers = {
+      ['regal/startDebugging'] = function(_, result)
```
glepnir, line 12: just `vim.lsp.handler[method] = function(_, result)` since we plan to remove config.handlers in future.

rinx: I didn't know about that. Thanks.

glepnir:
Isn't this possible natively or in the doc. It's bundled with a third party plugin.

rinx:
It makes sense. 👍 
I'll do it in the other plugin and maybe write some docs about it.

PR #3356, opened 2024-10-06
docs: hide modeline

janvhs:
This commit puts the modeline setting into a comment, so it appears hidden in the GitHub Makrdown preview

Super simple and straight forward, but I just noticed when copying the settings for ZLS

### DIFF
--- a/doc/configs.md
+++ b/doc/configs.md
@@ -14278,4 +14278,6 @@ require'lspconfig'.zls.setup{}
 
 
 
-vim:ft=markdown
+<!--
+  vim:ft=markdown
+-->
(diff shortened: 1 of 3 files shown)


justinmk:
the template is here: https://github.com/neovim/nvim-lspconfig/blob/master/scripts/README_template.md , try editing that?

doc/configs.md
```diff
@@ -14278,4 +14278,6 @@ require'lspconfig'.zls.setup{}
 
 
 
-vim:ft=markdown
+<!--
+  vim:ft=markdown
```
justinmk, line 14282: should not have leading whitespace

PR #3358, opened 2024-10-06
fix(harper_ls): filetype "csharp" -> "cs"

sharpchen:


justinmk:
need sanity check from @elijah-potter

elijah-potter:
Looks good to me. Thanks for catching that.

PR #3361, opened 2024-10-08
feat: add basics_ls.lua

antonk52:
Basics language server provides completion for buffer words, paths, and snippets. The goal is to allow current neovim users to test the built-in lsp completion without 3rd party plugins until neovim gains support for adding custom completion sources.


glepnir:
looks interesting. i may need give it a try thanks

PR #3363, opened 2024-10-09
fix(typo): correct typo in npm installation example

vimalsonara:
Corrected the typo in the npm installation example

- Changed the incorrect `npm install [-g]` command to `npm install -g` for the packages:

1. @biomejs/biome
2. @ast-grep/cli
3. rome


justinmk:
The brackets indicate that -g is "optional". This is a common pseudo-syntax, but maybe it's confusing to use it here.

PR #3366, opened 2024-10-12
feat: poryscript-pls

lhearachel:
Adds minimal `poryscript-pls` configuration to the registry.

Function handlers expected by the LSP implementation are not included to keep this config minimal.

### DIFF
--- a/lua/lspconfig/configs/poryscript_pls.lua
+++ b/lua/lspconfig/configs/poryscript_pls.lua
@@ -1,0 +1,20 @@
+local util = require('lspconfig.util')
+
+return {
+  default_config = {
+    cmd = { 'poryscript-pls' },
+    filetypes = { 'pory' },
+    root_dir = util.find_git_ancestor,
+    single_file_support = true,
+  },
+  docs = {
+    description = [[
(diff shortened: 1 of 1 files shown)


justinmk:
Approximately how many users does https://github.com/huderlem/poryscript-pls have? It has zero stars. We need to have some standard of minimum popularity for nvim-lspconfig configs.

lhearachel:
> Approximately how many users does https://github.com/huderlem/poryscript-pls have? It has zero stars. We need to have some standard of minimum popularity for nvim-lspconfig configs.

I would totally understand wanting to revert this merge due to lack of overall usage; most of the users of the LSP, as I understand, are VSCode users (and thus use an extension in the Marketplace) rather than nvim users.

justinmk:
Sure, that's fine, it sounds like there are actual users.  We just need to avoid having 100s of random configs that nobody actually uses.

lhearachel:
Retroactive reference point: [here](https://marketplace.visualstudio.com/items?itemName=karathan.poryscript) is a link to the VSC Marketplace extension.

PR #3367, opened 2024-10-12
fix: command LspStop can receive server name

### LINKED ISSUE #3362: `:LspStop config-name` (as opposed to `:LspStop client-id`)
### Description

Apologies if there has already been an issue for this - I couldn't find anything out there that seemed similar. 

There are some occasions where I would like to be able to stop a specific LSP (like `eslint`), but leave all other language servers running on a buffer (`volar`, `tailwindcss`, etc.)

Currently, the only way that I know of to disable LSPs on a buffer is with `:LspStop`, but this stops all language servers.

I think it would be a nice feature to modify this to accept LSP server names like `:LspStop eslint` so we can stop specific LSPs that are attached to buffer and leave all the others attached.

glepnir:
Receive server name as command argument then close this server if find it.

Fix #3362

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -110,22 +110,30 @@ end, {
 
 api.nvim_create_user_command('LspStop', function(info)
   local current_buf = vim.api.nvim_get_current_buf()
-  local server_id, force
+  local server_id, force, server_name
   local arguments = vim.split(info.args, '%s')
   for _, v in pairs(arguments) do
     if v == '++force' then
       force = true
     elseif v:find '^[0-9]+$' then
       server_id = v
+    else
+      server_name = v
     end
   end
 
   if not server_id then
     local servers_on_buffer = require('lspconfig.util').get_lsp_clients { bufnr = current_buf }
+    local found = false
     for _, client in ipairs(servers_on_buffer) do
-      if client.attached_buffers[current_buf] then
+      if client.attached_buffers[current_buf] and (server_name and (server_name == client.config.name) or true) then
         client.stop(force)
+        found = true
       end
+    end
+
+    if server_name and not found then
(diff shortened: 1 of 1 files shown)


plugin/lspconfig.lua
```diff
@@ -110,20 +110,22 @@ end, {
...
     end
   end
 
   if not server_id then
     local servers_on_buffer = require('lspconfig.util').get_lsp_clients { bufnr = current_buf }
     for _, client in ipairs(servers_on_buffer) do
-      if client.attached_buffers[current_buf] then
+      if client.attached_buffers[current_buf] and (server_name and (server_name == client.config.name) or true) then
         client.stop(force)
       end
```
justinmk, line 130: if server_name is provided but not found, we should show a `config "foo" not found` message to indicate that. silent failure is usually a bad UX.

glepnir: make sense

plugin/lspconfig.lua
```diff
@@ -110,23 +110,31 @@ end, {
...
     for _, client in ipairs(servers_on_buffer) do
-      if client.attached_buffers[current_buf] then
+      if client.attached_buffers[current_buf] and (server_name and (server_name == client.config.name) or true) then
         client.stop(force)
+        found = true
       end
     end
+
+    if server_name and not found then
+      vim.notify(('LSP server "%s" not found'):format(server_name), vim.log.levels.WARN)
```
justinmk, line 136: Is it really the server or just the config name? "LSP server" seems misleading, it sounds like the server program name could not be found?
```suggestion
      vim.notify(('nvim-lspconfig: config "%s" not found'):format(server_name), vim.log.levels.WARN)
```

justinmk (approved):
Nice!

justinmk:
Oh, we forgot to update the docs :)

PR #3372, opened 2024-10-14
fix(biome): remove 'css' duplicate

leakedmemory:
Problem:
There was 2 instances of the css string inside filetypes.

Solution:
Remove the duplicate and rearrange the filetypes in alphabetical order.


justinmk:
We could add a lint rule that ensures these kinds of items are alphabetized...

PR #3377, opened 2024-10-16
Make Uiua LSP work on ".ua" files instead of ".uiua" files

JoaoFelipe3:
`.ua` is the file type this is supposed to work on.

### DIFF
--- a/lua/lspconfig/configs/uiua.lua
+++ b/lua/lspconfig/configs/uiua.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'uiua', 'lsp' },
-    filetypes = { 'uiua' },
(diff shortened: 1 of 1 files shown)


justinmk:
can you provide a reference and perhaps insight as to why @ckafi may have chosen the other way?

JoaoFelipe3:
[Here](https://github.com/uiua-lang/uiua-modules) is a list of various Uiua modules. Note that these all use ".ua".

@ckafi probably didn't know that the most-used file extension was ".ua".

ckafi:
@JoaoFelipe3 you're mixing concepts here. `filetype` is not the same as the file ending. If you open a `main.rs` for example, the filetype will be `rust` not `rs`.

JoaoFelipe3:
Oh my god

PR #3389, opened 2024-10-19
fix(fennel-ls): use cwd when git repo parent is not found

justiceenunciate:
## Context
For the `fennel-ls` lspconfig, I recently changed how the  `fennel-ls` configuration file is found, i.e. nearest config -> git root. 

However, when there is no `fennel-ls` configuration available, and all parent directories of the current buffer do contain git repositories, then the LSP won't be loaded at all. This PR fixes that.

## Changes
* Use `cwd` for `fennel-ls`  when git repository parent is unavailable

## Questions
1. Can I write a test so this doesn't fail again? If so, where should that live?
2. If `root_dir` is `nil`, should the LSP fail to load? I'm not sure if there are legitimate cases to ever send `nil` to an LSP but I thought I'd ask if a wider, structural solution is appropriate.

Thanks for your help!

### DIFF
--- a/lua/lspconfig/configs/fennel_ls.lua
+++ b/lua/lspconfig/configs/fennel_ls.lua
@@ -8,7 +8,7 @@ return {
       local has_fls_project_cfg = function(path)
         return util.path.is_file(vim.fs.joinpath(path, 'flsproject.fnl'))
       end
-      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git')
+      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git') or vim.fn.getcwd()
     end,
     settings = {},
     capabilities = {


lua/lspconfig/configs/fennel_ls.lua
```diff
@@ -8,7 +8,7 @@ return {
       local has_fls_project_cfg = function(path)
         return util.path.is_file(vim.fs.joinpath(path, 'flsproject.fnl'))
       end
-      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git')
+      return util.search_ancestors(dir, has_fls_project_cfg) or vim.fs.root(0, '.git') or vim.fn.getcwd()
```
glepnir, line 11: use vim.uv.cwd() would be better.

justinmk:
Isn't this something that could be done by default? But IIRC that was discussed and there were reasons against it. I can't find the discussion though. Why is this reasonable for e.g. fennel and pyright, but not others?

mfussenegger:
> Isn't this something that could be done by default?

At least from core perspective `nil` is legitimate for servers who support single-file's without project setup.
Not every server does - I think that's why lspconfig introduced a dedicated flag to indicate if they do.

And afaik a `cwd` as fallback was mostly removed because it led to users accidentally having their $HOME indexed by a language server.

justinmk:
Thanks for those reminders!

> `cwd` as fallback was mostly removed because it led to users accidentally having their $HOME indexed by a language server.

Right, and then this PR reintroduces that issue...

justinmk:
> At least from core perspective `nil` is legitimate for servers who support single-file's without project setup.
> Not every server does - I think that's why lspconfig introduced a dedicated flag to indicate if they do.

So that is a potential way forward here. 

Previous "single file support" PRs, for reference: https://github.com/neovim/nvim-lspconfig/pulls?q=is%3Apr+single+file+is%3Aclosed

PR #3391, opened 2024-10-22
fix(lua_ls): when neither lua or git root found, return nil. See #3322

folke:
Due to #3322, when neither a `/lua` directory or `.git` root dir exists, the empty (and invalid) `''` root dir was returned.

Instead it should return `nil` so that single-file mode can trigger instead.


pitkling (approved):
Looks good to me. Returning the empty string in my original PR #3322 only happened because of a code simplification (see also my question concerning that simplification at the end of [this comment](https://github.com/neovim/nvim-lspconfig/pull/3322#discussion_r1779451552), where I basically asked about whether there is a difference in returning `nil` or the empty string.).

PR #3394, opened 2024-10-23
feat: ada_ls

glacambre:
The configuration for the Ada Language Server was first added in #171 and removed in #3310. The removal happened due to misunderstandings, it was thought at the time that the default language server configuration could not work on its own (#1683), it turns out that this was actually caused by a bug in the ALS that was fixed a long time ago. This means the default ALS configuration can be re-introduced.

However, in the meantime, a new neovim plugin for Ada was created by and the name "als" was borrowed, thus we have to use a new name, ada_ls, in order to avoid breaking this plugin.

This reverts commit 7b8b0b3ddd0ed6eddc93982753acaddc578defac.

### DIFF
--- a/lua/lspconfig/configs/ada_ls.lua
+++ b/lua/lspconfig/configs/ada_ls.lua
@@ -1,0 +1,32 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'ada_language_server' },
+    filetypes = { 'ada' },
+    root_dir = util.root_pattern('Makefile', '.git', '*.gpr', '*.adc'),
+  },
+  docs = {
+    description = [[
+https://github.com/AdaCore/ada_language_server
+
+Installation instructions can be found [here](https://github.com/AdaCore/ada_language_server#Install).
+
+Can be configured by passing a "settings" object to `ada_ls.setup{}`:
+
+```lua
+require('lspconfig').ada_ls.setup{
+    settings = {
+      ada = {
+        projectFile = "project.gpr";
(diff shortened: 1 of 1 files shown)


justinmk:
> the name "als" was borrowed, thus we have to use a new name, ada_language_server

nit: `ada_ls` ? that seems to be the typical convention here

lua/lspconfig/configs/ada_ls.lua
```diff
@@ -0,0 +1,37 @@
+local util = require 'lspconfig.util'
+local bin_name = 'ada_language_server'
+
+if vim.fn.has 'win32' == 1 then
+  bin_name = 'ada_language_server.exe'
+end
```
justinmk, line 6: ternary saves some code and is clearer. but i thought that this was handled automatically? @glepnir

glacambre: This might be a remnant from when I first added the config in 2020, perhaps it's not necessary anymore. I'll try to double check.

glacambre: Confirmed that this works fine by default, PR updated. Thanks for catching this.

glepnir: it is no longer needed. we use exepath with server.cmd now see #2595

glacambre:
Thanks for the merge! I'll try to prototype the changes I suggested in https://github.com/neovim/nvim-lspconfig/pull/3310 so that we can decide whether they make sense/are worth having.

PR #3411, opened 2024-10-31
fix: correct type annotation on lspconfig.Config::on_new_config

tristan957:
It takes a new_root_dir parameter as documented in the help docs.


glepnir:
thx

PR #3414, opened 2024-11-01
feat(buf): new lsp for protobuf

gogamid:
As may know, bufbuild team is working on the new lsp for protobufs which is included in the cli itself. Compared to buf-language-server, this lsp supports more features. Take a look at this [PR](https://github.com/bufbuild/buf/pull/3316) for more info. This config sets up the new buf lsp.


lua/lspconfig/configs/buflsp.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+
+    cmd = { 'buf', 'beta', 'lsp' },
+    filetypes = { 'proto' },
+    root_dir = function(fname)
+      return require('lspconfig.util').root_pattern('buf.work.yaml', '.git')(fname)
+    end,
```
sprsquish, line 10: Looks like the lua linter failed with an unused var. You can also simplify root_dir since root_pattern returns a function.
```suggestion
return {
  default_config = {

    cmd = { 'buf', 'beta', 'lsp' },
    filetypes = { 'proto' },
    root_dir = require('lspconfig.util').root_pattern('buf.work.yaml', '.git'),
```

lua/lspconfig/configs/buflsp.lua
```diff
@@ -0,0 +1,21 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+
+    cmd = { 'buf', 'beta', 'lsp' },
+    filetypes = { 'proto' },
+    root_dir = function(fname)
+      return require('lspconfig.util').root_pattern('buf.work.yaml', '.git')(fname)
```
lithammer, line 9: `buf.work.yaml` seems to be old and the new preferred file is `buf.yaml`.

https://buf.build/docs/configuration/v2/buf-yaml/

gogamid: thanks for the suggestion, replaced :)

gogamid:
finally fixed the linter problems :D

sprsquish (commented):
Thanks for posting this!

qq: are you having problems with the LSP timing out and exiting? When I was playing with this I had to set `--timeout=0` or the LSP would exit. 
I'd also suggest adding `--log-format=text`. The default is color and it makes the logs hard to read

gogamid:
@sprsquish yeah, i got occasional crash, i thought because the lsp is still in beta phase. But good suggestions!

lua/lspconfig/configs/buflsp.lua
```diff
@@ -0,0 +1,17 @@
+return {
+  default_config = {
+
```
glepnir, line 3:
```suggestion
  default_config = {
```

lua/lspconfig/configs/buflsp.lua
```diff
@@ -0,0 +1,17 @@
...
+    cmd = { 'buf', 'beta', 'lsp', '--timeout=0', '--log-format=text' },
```
glepnir, line 4:
```suggestion
  default_config = {
    cmd = { 'buf', 'beta', 'lsp', '--timeout=0', '--log-format=text' },
```

glepnir:
the old bufls can mark as deprecated ? and  current server file name is a bit confusing. I think it is better to use xx_ls.lua as the file name if there is no special server command name.

justinmk (approved):

sprsquish (approved):

sprsquish (approved):

glepnir:
https://github.com/bufbuild/buf-language-server this repo is archived so could you also add a deprecate of bufls in lua/lspconfig.lua aliases table ?

gogamid:
@glepnir done

PR #3417, opened 2024-11-03
fix(fennel-ls): enable single_file_support

justiceenunciate:
## Context
`fennel-ls` will not load when a configuration file is not present, i.e. `flsproject.fnl`. This fixes that by adding `single_file_support`.

This PR addresses comments on from my [first attempted PR](https://github.com/neovim/nvim-lspconfig/pull/3389).

## Changes
* Add `single_file_support` to `fennel-ls`.


doc/configs.md
```diff
@@ -3493,6 +3493,10 @@ Default config:
   ```lua
   {}
   ```
+  - `single_file_support` : 
+  ```lua
+  true
+  ```
```
glepnir, line 3499: you don't need add this in doc

PR #3434, opened 2024-11-12
refactor(util): replace some path util by using vim.fs

glepnir:


glepnir:
closed since @dundargoc ❤️ working on

PR #3437, opened 2024-11-17
bug with opening projects embedded in other projects

kkrime:
In the following senario;

```
outter project/
├─ .git
├─ README.me
├─ src/
├─ inner project/
│  ├─ .git
│  ├─ README.me
│  ├─ src/
```

Assuming both the `outter project` and `inner project` are of the same langauge;
if you were to open the `outter project` first, then open the `inner project`, both projects would have the same `root directory`

This is causing issues with other plugins that rely on `root directory` to be accurate; 
https://github.com/ahmedkhalf/project.nvim/blob/8c6bad7d22eef1b71144b401c9f74ed01526a4fb/lua/project_nvim/project.lua#L25

This commit fixes the above described issue

### DIFF
--- a/lua/lspconfig/manager.lua
+++ b/lua/lspconfig/manager.lua
@@ -5,23 +5,6 @@ local uv = vim.uv or vim.loop
 local async = require 'lspconfig.async'
 local util = require 'lspconfig.util'
 
----@param client vim.lsp.Client
----@param root_dir string
----@return boolean
-local function check_in_workspace(client, root_dir)
-  if not client.workspace_folders then
-    return false
-  end
-
-  for _, dir in ipairs(client.workspace_folders) do
-    if (root_dir .. '/'):sub(1, #dir.name + 1) == dir.name .. '/' then
-      return true
-    end
-  end
-
-  return false
-end
-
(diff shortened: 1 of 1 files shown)


glepnir:
plugin should check server support workspace folder first..

kkrime:
@glepnir can you elaborate on what you mean, I'm new to this code base?

glepnir:
nothing should be changed in there. see spec of workpace folder report to plugin maybe https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders

kkrime:
@glepnir thank you for clarifying. I have a question it's on the same lines as this issue;

When there are 2 projects (seperate not one being a sub project of another) and `:LspInfo` is run, the `root directory` is different for each project, however the `root directory` is the same when calling `vim.lsp.get_clients()`.

My question is, should the output from `vim.lsp.get_clients()` be the same as that of `:LspInfo` including the `root directory`?

kkrime:
Ignore the above question @glepnir I looked at the code and I can see that `:LspInfo` and `vim.lsp.get_clients()` go through seperate code paths

PR #3438, opened 2024-11-17
feat: allow list of client ids or config names in LspRestart and LspStop

guilhas07:
## Problems:
- `:LspStop` doesn't support list of client ids or config names.
- `:LspRestart` doesn't support config names although supports list of client ids
- Command completion uses ids and names, not allowing the user to hit enter immediately after
finding the desired server

## Solution:
- Unify `LspStop` and `LspRestart` implementation supporting lists of client ids and
config names
- Command completion only returns config names
- Modify docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,9 +122,9 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (deprecated alias to `:che lspconfig`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config. Pass `autostart = false` to your `.setup{}` call for a language server if you would like to launch clients solely with this command. Defaults to all servers matching current buffer filetype.
-* `:LspStop {<client_id> or <config_name>} ...` Stops the servers with the given client-ids or config names. Defaults to
+* `:LspStop [<client_id_or_name> ...]` Stops the given server(s). Defaults to
 stopping all servers active on the current buffer. To force stop add `++force`
-* `:LspRestart {<client_id> or <config_name>} ...` Restarts the clients with the given client-ids or config names, and attempts to reattach to all
+* `:LspRestart [<client_id_or_name> ...]` Restarts the given client(s), and attempts to reattach to all
 previously attached buffers. 
 
 ## Contributions
--- a/README.md
+++ b/README.md
@@ -122,8 +122,10 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (deprecated alias to `:che lspconfig`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config. Pass `autostart = false` to your `.setup{}` call for a language server if you would like to launch clients solely with this command. Defaults to all servers matching current buffer filetype.
-* `:LspStop <client_id>` Defaults to stopping all buffer clients.
-* `:LspRestart <client_id>` Defaults to restarting all buffer clients.
+* `:LspStop {<client_id> or <config_name>} ...` Stops the servers with the given client-ids or config names. Defaults to
+stopping all servers active on the current buffer. To force stop add `++force`
+* `:LspRestart {<client_id> or <config_name>} ...` Restarts the clients with the given client-ids or config names, and attempts to reattach to all
+previously attached buffers. 
 
 ## Contributions
 
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -382,13 +382,13 @@ Launches the requested (configured) client, but only if it successfully
 resolves a root directory. Note: Defaults to all configured servers matching
 the current buffer filetype.
 
-:LspStop {[client_id] or [config_name]} ...                        *:LspStop*
+:LspStop [client_id] or [config_name] ...                        *:LspStop*
 Stops the servers with the given client-ids or config names. Defaults to
 stopping all servers active on the current buffer. To force stop language
 servers: >
     :LspStop ++force
 
-:LspRestart {[client_id] or [config_name]} ...                     *:LspRestart*
+:LspRestart [client_id] or [config_name] ...                     *:LspRestart*
 Restarts the clients with the given client-ids or config names, and attempts
 to reattach to all previously attached buffers.
 
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -382,14 +382,15 @@ Launches the requested (configured) client, but only if it successfully
 resolves a root directory. Note: Defaults to all configured servers matching
 the current buffer filetype.
 
-:LspStop [client_id] or [config_name]                              *:LspStop*
-Stops the server with the given client-id or config name. Defaults to stopping
-all servers active on the current buffer. To force stop a language server: >
-    :LspStop <client_id> ++force
+:LspStop {[client_id] or [config_name]} ...                        *:LspStop*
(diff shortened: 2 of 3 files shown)


plugin/lspconfig.lua
```diff
@@ -28,23 +28,48 @@ local lsp_complete_configured_servers = function(arg)
   end, util.available_servers()))
 end
 
-local lsp_get_active_client_ids = function(arg)
+local lsp_get_active_clients = function(arg)
   local clients = vim.tbl_map(function(client)
-    return ('%d:%s'):format(client.id, client.name)
+    return ('%s'):format(client.name)
```
glepnir, line 33: why remove id ? if there are multiple servers which does not support workspace folders, it will be difficult to distinguish which one want to stop

guilhas07: Are saying that its possible to receive multiple entries like:
- 1 lua_ls
- 2 lua_ls 

If that is the case than I will revert this.

Is there a way to show the completion with the id and name but on complete only pass the id so the user
only needs to hit enter?

plugin/lspconfig.lua
```diff
@@ -28,23 +28,48 @@ local lsp_complete_configured_servers = function(arg)
...
 local get_clients_from_cmd_args = function(arg)
   local result = {}
+  local managed_clients = util.get_managed_clients()
+  local clients = {}
+  for _, client in pairs(managed_clients) do
+    clients[client.name] = client
+  end
+
+  local err_msg = ''
+  arg = arg:gsub('[%a-_]+', function(name)
```
guilhas07, line 51: Is this enough to match all valid config names ?

guilhas07 (commented):
When testing I found that lua_ls didn't close on first `LspStop` or `LspRestart` . Should `LspRestart` use force=true ? https://github.com/neovim/nvim-lspconfig/blob/b55037d7660c6463042f7992159034070091d14f/plugin/lspconfig.lua#L104

plugin/lspconfig.lua
guilhas07: Are saying that its possible to receive multiple entries like:
- 1 lua_ls
- 2 lua_ls 

If that is the case than I will revert this.

Is there a way to show the completion with the id and name but on complete only pass the id so the user
only needs to hit enter?

glepnir: revert .  if you want complete with input id need imporve  current custom complete  and split cmdline content to get id then complete with server name

plugin/lspconfig.lua
glepnir: nope.. some server file name without underline..

justinmk:
>  Command completion uses ids and names, not allowing the user to hit enter immediately after
>   finding the desired server

Fixing that is welcome, and should be the focus of this PR.

Supporting lists is out of scope. These commands are meant for interactive use, and the use-case for passing multiple args simulatenously is uncommon, thus not worth the code complexity (nor the extra documentation). For programmatic use the api should be used.

plugin/lspconfig.lua
guilhas07: But this matches server names without underscore too so that is not a problem. I meant if there are characters that I'm currently not matching that should be included.

guilhas07:
> Supporting lists is out of scope. These commands are meant for interactive use, and the use-case for passing multiple args simulatenously is uncommon, thus not worth the code complexity (nor the extra documentation). For programmatic use the api should be used.

With these changes both LspRestart and LspStop use the same parsing function, simplifying the LspStop code (went from 47 to 25 loc), and adding the possibility to use config name for LspRestart and error logging. Overall the diff is around the same loc.
```
 plugin/lspconfig.lua | 93 ++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------
 1 file changed, 48 insertions(+), 45 deletions(-)
```
In terms of added complexity in the parsing function, only error logging and substitution from `config_name` to `client id` was added, which in my opinion its easy to follow (although I can add more comments if its necessary). 
> Fixing that is welcome, and should be the focus of this PR.

This was what motivate me to also add the list, because if it works correctly it would be very fast to autocomplete two Lsp
servers. Also, the LspRestart command already supported it (though not documented).

Do you have any pointers how to fix it ? I presume that I can insert some logic when accepting a completion so it uses the id.

README.md
```diff
@@ -122,8 +122,10 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (deprecated alias to `:che lspconfig`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config. Pass `autostart = false` to your `.setup{}` call for a language server if you would like to launch clients solely with this command. Defaults to all servers matching current buffer filetype.
-* `:LspStop <client_id>` Defaults to stopping all buffer clients.
-* `:LspRestart <client_id>` Defaults to restarting all buffer clients.
+* `:LspStop {<client_id> or <config_name>} ...` Stops the servers with the given client-ids or config names. Defaults to
+stopping all servers active on the current buffer. To force stop add `++force`
+* `:LspRestart {<client_id> or <config_name>} ...` Restarts the clients with the given client-ids or config names, and attempts to reattach to all
```
justinmk, line 127:
```suggestion
* `:LspStop <client_id_or_name> [<client_id_or_name> ...]` Stops the given server(s). Defaults to
stopping all servers active on the current buffer. To force stop add `++force`
* `:LspRestart <client_id_or_name> [<client_id_or_name> ...]` Restarts the given client(s), and attempts to reattach to all
```

doc/lspconfig.txt
```diff
@@ -382,14 +382,15 @@ Launches the requested (configured) client, but only if it successfully
...
 
-:LspStop [client_id] or [config_name]                              *:LspStop*
-Stops the server with the given client-id or config name. Defaults to stopping
-all servers active on the current buffer. To force stop a language server: >
-    :LspStop <client_id> ++force
-
-:LspRestart [client_id]                                          *:LspRestart*
-Restarts the client with the given client id, and attempts to reattach to all
-previously attached buffers.
+:LspStop {[client_id] or [config_name]} ...                        *:LspStop*
```
justinmk, line 385: lua syntax is not used for vim commands documentation. `{..}` means something different in vim docs.

```suggestion
:LspStop [client_id_or_config_name] ...                        *:LspStop*
```

doc/lspconfig.txt
```diff
@@ -382,14 +382,15 @@ Launches the requested (configured) client, but only if it successfully
...
+Stops the servers with the given client-ids or config names. Defaults to
+stopping all servers active on the current buffer. To force stop language
+servers: >
+    :LspStop ++force
+
+:LspRestart {[client_id] or [config_name]} ...                     *:LspRestart*
```
justinmk, line 391:
```suggestion
:LspRestart [client_id] or [config_name] ...                     *:LspRestart*
```

justinmk (approved):
LGTM, let's try it. Hoping @guilhas07 will followup on any bugs :)

guilhas07:
> LGTM, let's try it. Hoping @guilhas07 will followup on any bugs :)

:saluting_face:

plugin/lspconfig.lua
glepnir: there still not revert.

guilhas07: I believe it's best this way. This enhances the normal case. If a user needs to close a specific id he needs to know it using LspInfo anyway.

glepnir: nope

PR #3439, opened 2024-11-17
feat(vale_ls): add reStructuredText to filetypes

antznin:
Vale supports parsing reStructuredText files, as documented here: https://vale.sh/docs/topics/scoping/#restructuredtext

It uses rst2html in the background.

### DIFF
--- a/lua/lspconfig/configs/vale_ls.lua
+++ b/lua/lspconfig/configs/vale_ls.lua
@@ -3,7 +3,7 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'vale-ls' },
-    filetypes = { 'markdown', 'text', 'tex' },
+    filetypes = { 'markdown', 'text', 'tex', 'rst' },
     root_dir = util.root_pattern '.vale.ini',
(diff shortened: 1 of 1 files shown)


glepnir:
seems like rst does not include in our builtin filetype detect lists ? if it need user regsiter this filetype by themself i think this also can be done in local

antznin:
Not sure I understand what you mean: this change is what I had to add to `vale_ls.lua` for `rst` for `vale_ls` to parse reStructuredText files. Running `:set filetype?` on rst files returns `filetype=rst` (neovim v0.10.1).

antznin:
Same result when I open an rst file with `nvim --noplugin`, so this shouldn't be specific to my config I think?

glepnir:
aha good to know i just don't find it in detect.lua if `set ft?` can return correct filetype let's go :)

antznin:
Thanks for the tip, yes it does return `filetype=rst` with `nvim --clean -u NONE`. I forgot to update `server_configurations.md`, I can do that too.

glepnir:
np thanks.

Issue #3433, opened 2024-11-12
Pylyzer requires Erg as dependency, suggest adding this info


fishBone000:
### Description

Pylyzer needs Erg to function correctly, whose path to its binary and libs should be stored in env var $ERG_PATH:
> [ERROR][2024-11-11 12:08:17] .../vim/lsp/rpc.lua:770	"rpc"	"/home/fishbone/.local/share/nvim/mason/bin/pylyzer"	"stderr"	"\27[91m[ERR] ERG_PATH not found\27[m\n"

I can resolve this by extracting tar from [Erg release](https://github.com/erg-lang/erg/releases/tag/v0.6.47) to my `~/.erg`, then export env var by `vim.fn.setenv("ERG_PATH", vim.fn.getenv("HOME") .. "/.erg")`

However no such info regarding Erg is provided in `lspconfig-all`. I suggest adding this, or even better, adding `erg_path` option to let lspconfig handle exporting the env var automatically.

fishBone000:
File structure of Erg tarball:
```plain
tar -tf erg-x86_64-unknown-linux-gnu.tar.gz                                                                                                                                     exit 1  47s
erg
lib/
lib/std/
lib/std/exception.er
...and more lib/ files
```

justinmk:
send a pr

PR #3441, opened 2024-11-18
fix: add pylyzer dependenncy info (erg) (#3433)

fishBone000:
Problem: Pylyzer needs Erg as dependency, such info is missing in configs.md
Solution: Add such info in doc, and add `ergPath` setup option

### DIFF
--- a/lua/lspconfig/configs/pylyzer.lua
+++ b/lua/lspconfig/configs/pylyzer.lua
@@ -23,11 +23,8 @@ return {
         checkOnType = false,
       },
     },
-    ergPath = vim.fn.getenv('HOME') .. '/.erg',
     on_new_config = function(new_config, _)
-      if new_config.ergPath then
-        new_config.cmd_env.ERG_PATH = new_config.ergPath
-      end
+      new_config.cmd_env.ERG_PATH = vim.fn.getenv('HOME') .. '/.erg'
     end,
   },
   docs = {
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/pylyzer.lua
```diff
@@ -23,12 +23,25 @@ return {
         checkOnType = false,
       },
     },
+    ergPath = vim.fn.getenv('HOME') .. '/.erg',
+    on_new_config = function(new_config, _)
+      if new_config.ergPath then
+        new_config.cmd_env.ERG_PATH = new_config.ergPath
+      end
+    end,
```
justinmk, line 31: This saves the user basically nothing, while also adding a special-case interface to this config.

It would be more useful to just document what you're doing here, instead of adding a special `ergPath` field.

fishBone000: I thought the `cmd_env` is gonna work? I found it in [bqnlsp's default config](https://github.com/neovim/nvim-lspconfig/blob/f012c1b176f0e3c71f40eb309bdec0316689462e/lua/lspconfig/configs/bqnlsp.lua#L24).

I think it would still be nice if I set the `ERG_PATH` in default config, so users who aren't familiar with lsp config can use it out of box:
```lua
      vim.fn.setenv("ERG_PATH", vim.fn.getenv("HOME") .. "/.erg")
    end
```
Is this OK?

justinmk: `cmd_env` is a standard field, no objection to using that. My objection is about the custom `ergPath` thing.



> still be nice if I set the `ERG_PATH` in default config, so users who aren't familiar with lsp config can use it out of box

Sure, yes. And some sort of brief hint in the docs.

fishBone000:
Will mark it ready for review after I test if it's working.

lua/lspconfig/configs/pylyzer.lua
```diff
@@ -23,12 +23,22 @@ return {
         checkOnType = false,
       },
     },
+    on_new_config = function(new_config, _)
+      new_config.cmd_env.ERG_PATH = vim.fn.getenv('HOME') .. '/.erg'
```
justinmk, line 27:
```suggestion
      new_config.cmd_env.ERG_PATH = vim.fs.joinpath(vim.uv.os_homedir(), '/.erg')
```

justinmk:
Feel free to reopen targeting `lsp/*`. Not accepting changes to `lua/lspconfig/configs/*`

PR #3443, opened 2024-11-20
fix: follow upstream deprecate non-method clients functions

glepnir:
see  https://github.com/neovim/neovim/commit/454ae672aad172a299dcff7c33c5e61a3b631c90

### DIFF
--- a/lua/lspconfig/configs/denols.lua
+++ b/lua/lspconfig/configs/denols.lua
@@ -1,4 +1,4 @@
-local util = require 'lspconfig.util'
+local util = require('lspconfig.util')
 local lsp = vim.lsp
 
 local function buf_cache(bufnr, client)
@@ -6,7 +6,7 @@ local function buf_cache(bufnr, client)
     command = 'deno.cache',
     arguments = { {}, vim.uri_from_bufnr(bufnr) },
   }
-  client.request('workspace/executeCommand', params, function(err, _result, ctx)
+  util.client_proxy(client).request('workspace/executeCommand', params, function(err, _, ctx)
     if err then
       local uri = ctx.params.arguments[2]
       vim.api.nvim_err_writeln('cache command failed for ' .. vim.uri_to_fname(uri))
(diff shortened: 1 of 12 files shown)


lua/lspconfig/configs/denols.lua
```diff
@@ -96,7 +96,7 @@ return {
   commands = {
     DenolsCache = {
       function()
-        local clients = util.get_lsp_clients { bufnr = 0, name = 'denols' }
+        local clients = util.get_lsp_clients({ bufnr = 0, name = 'denols' })
```
justinmk, line 99: is stylua doing this? not required by our style guide.

lua/lspconfig/configs/julials.lua
```diff
@@ -1,19 +1,19 @@
...
+  local julials_clients = vim.lsp.get_clients({ bufnr = bufnr, name = 'julials' })
   assert(
     #julials_clients > 0,
     'method julia/activateenvironment is not supported by any servers active on the current buffer'
   )
   local function _activate_env(environment)
     if environment then
       for _, julials_client in ipairs(julials_clients) do
-        julials_client.notify('julia/activateenvironment', { envPath = environment })
+        util.client_proxy(julials_client).notify('julia/activateenvironment', { envPath = environment })
```
justinmk, line 16: Definitely not in favor of this PR if it requires `client_proxy` in the configs themselves. That just introduces even more migration later.

If there's a way to make it transparent to configs, that's fine. Otherwise we should just wait until we can bump the minimum required Nvim to 0.11

lua/lspconfig/configs/denols.lua
glepnir: seems like a bug of stylua . just run `stylua ./` in shell

lua/lspconfig/configs/julials.lua
glepnir: Yes. The current command design is too fragile. The client is necessary here. So it is better to provide a default wrapper function with the `client` and `bufnr` as the parameter so that we can easily handle any upstream changes on the client object internally.  

```lua
      switch_source_header(client, bufnr),   
````

PR #3450, opened 2024-11-24
refactor: use vim.lsp.start instead of vim.lsp.start_client

MariaSolOs:
This PR replaces the usage of `vim.lsp.start_client` by `vim.lsp.start`, delegating the logic of client reuse to the editor.

Also adding support for a `silent` option to prevent error notifications on init.

### DIFF
--- a/lua/lspconfig/manager.lua
+++ b/lua/lspconfig/manager.lua
@@ -85,7 +85,8 @@ end
 --- @param new_config lspconfig.Config
 --- @param root_dir string
 --- @param single_file boolean
-function M:_start_client(bufnr, new_config, root_dir, single_file)
+--- @param silent boolean
+function M:_start_client(bufnr, new_config, root_dir, single_file, silent)
   -- do nothing if the client is not enabled
   if new_config.enabled == false then
     return
@@ -129,6 +130,7 @@ function M:_start_client(bufnr, new_config, root_dir, single_file)
 
   local client_id = lsp.start(new_config, {
     bufnr = bufnr,
+    silent = silent,
     reuse_client = function(existing_client)
       if (self._clients[root_dir] or {})[existing_client.name] then
         self:_notify_workspace_folder_added(root_dir, existing_client)
@@ -153,9 +155,11 @@ end
 ---@param root_dir string
 ---@param single_file boolean
 ---@param bufnr integer
+---@param silent boolean
+function M:add(root_dir, single_file, bufnr, silent)
   root_dir = util.path.sanitize(root_dir)
   local new_config = self.make_config(root_dir)
-  self:_start_client(bufnr, new_config, root_dir, single_file)
+  self:_start_client(bufnr, new_config, root_dir, single_file, silent)
 end
 
 --- @return vim.lsp.Client[]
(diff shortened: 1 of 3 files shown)


glepnir:
yes. i remember wrong version just tried on 0.9

MariaSolOs:
> yes. i remember wrong version just tried on 0.9

Thanks for checking!

lua/lspconfig/manager.lua
```diff
@@ -150,88 +128,46 @@ function M:_start_new_client(bufnr, new_config, root_dir, single_file)
...
-  self:_start_new_client(bufnr, new_config, root_dir, single_file)
-end
-
---- @private
---- @param bufnr integer
---- @param new_config lspconfig.Config
---- @param root_dir string
---- @param client vim.lsp.Client
---- @param single_file boolean
-function M:_attach_after_client_initialized(bufnr, new_config, root_dir, client, single_file)
```
glepnir, line 188: When neovim reads the session it may contain multiple projects. This needs to be checked after the first server start.

MariaSolOs: The logic of `_attach_or_spawn` is:
1. Call `lsp.buf_attach_client` (which neovim does inside `vim.lsp.start` so no need to do it here)
2. Add the client to the `_clients` cache
3. Send a `workspace/didChangeWorkspaceFolders` notification if the server supports it. 

The implementation in this PR keeps 2 and 3, event after the server is first initialized. Or are you referring to something else?

glepnir: vsplit has two buffers in two projects. mksession starts neovim with -S session.vim. The first buffer triggers FileType to start a server. Note that this is an asynchronous server that needs initialization time. At this point you are not sure if the server supports workspacefolder. But the second buffer has triggered FileType again. The embed process is processed continuously. So the second buffer should wait for the server started by the first buffer to complete initialization and check whether to connect to it or start a new one.

MariaSolOs: I see. I've replaced the code using the timer by adding [this hook](https://github.com/neovim/nvim-lspconfig/blob/2eccb418f2f972ad1167f8491a8180acdeb02384/lua/lspconfig/manager.lua#L105) to `on_init`.

glepnir: if I remember correctly, this is a hook for modifying the parameters of the server itself. It runs after the root judgment. The problem is that if the server has not initialized the second buffer, the same event loop will run the luanch function. Can we know that the first server started supports workspacefolder? Shouldn't there be a synchronous waiting problem here? I have forgotten some of the code here. I simply tested this Pr code and it works as I expected. But I still don't understand the logic here.

glepnir (approved):

justinmk (approved):
Thank you, each incremental code reduction like this gets us closer to upstreaming to core. Nice naming improvements too.

MariaSolOs:
> Thank you, each incremental code reduction like this gets us closer to upstreaming to core. Nice naming improvements too.

That’s the goal indeed! And the name changes were mostly because I had to stare at the code for a while to understand what it was doing 😅

PR #3451, opened 2024-11-24
fix(lua_ls): don't set user config dir as root

RayJameson:
root_files may be in user's config dir which causes lua_ls to recursievly scan all config dirs, if you edit lua file like this: ~/.config/awesome/rc.lua


justinmk:
pretty sure this has been discussed 1 million times. search closed issues/prs, and provide references that explain why I'm wrong

PR #3452, opened 2024-11-25
fix: check existing_client support workspaceFolder

glepnir:
if existing_client does not support workpaceFolder should spawn a new server instance

### DIFF
--- a/lua/lspconfig/manager.lua
+++ b/lua/lspconfig/manager.lua
@@ -58,12 +58,10 @@ end
 --- @param root_dir string
 --- @param client vim.lsp.Client
 function M:_notify_workspace_folder_added(root_dir, client)
-  if is_dir_in_workspace_folders(client, root_dir) then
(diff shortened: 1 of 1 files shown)


lithammer:
This looks like something that should be upstreamed to core?

glepnir:
`reuse_client` just a condition check  in the upstream. Maybe other possibilities are considered as public API. Not sure if coding check workpaceFolder in core is appropriate. But we need it here because it is compatible with release version

```lua
function lsp.start(config, opts)
  opts = opts or {}
  local reuse_client = opts.reuse_client or reuse_client_default
  local bufnr = resolve_bufnr(opts.bufnr)

  for _, client in pairs(all_clients) do
    if reuse_client(client, config) then
      if lsp.buf_attach_client(bufnr, client.id) then
```

justinmk (approved):

PR #3459, opened 2024-11-26
Revert "fix: check existing_client support workspaceFolder"

glepnir:
Reverts neovim/nvim-lspconfig#3452


Marat-Gumerov:
My neovim was spawning new omnisharp instances for every single .cs file I open in my project.
It looks like this commit fixes the issue. Thank you :)

Issue #2079, opened 2022-08-23
Replace functions from `utils` with upstream functions


ranjithshegde:
### Language server

N/A

### Requested feature

1. `path`
2. `search_ancestors`
3. `root_pattern`
4. `find_git_ancestors`
5. `find_mercurial_ancestors`
6. `find_node_modules_ancestors`
7. `find_package_json_ancestors`

All of the above can be replaced with relevant functions from `vim.fs` module.

## Proposal
1. Soft deprecate it after the release of `nvim 0.8` + stabilization period. 
2. Internally redirect the calls to the relevant calls from `vim.fs` module
3. replaces the functions inside `server_configurations/[server].lua` wherever relevant
4. Provide examples and comprehensive documentation

### Other clients which have this feature

N/A

justinmk:
@ranjithshegde this would still be a welcome change, if you are keen

glepnir:
there are some external compatibility issues. those functions not only used  in our repo.. so  I am a little worried about the issues that may arise from the changes.

justinmk:
> those functions not only used in our repo.

Where are they used? lspconfig was never intended to be a platform. It's intended to be data-only. And we need to move towards that. This is the way we will be able to improve quality and improve maintainability of these configs.

glepnir:
I knwo and agree your thought..but it used in some other plugins or user config .just a quick search you will see 
https://github.com/search?q=lspconfig.util&type=code&p=3

justinmk:
We don't necessarily need to remove `lspconfig.util.root_pattern`, we can start by updating their implementations to use vim.fs and minimizing code where possible. And where possible, show a deprecation message that explains how to use `vim.fs` directly instead of e.g. `lspconfig.util.root_pattern`. And update all of the examples in lspconfig docs to use `vim.fs` directly, instead of `lspconfig.util`.

justinmk:
For example, `vim.fs.root()` exists now.

glepnir:
yes.. just worried that some util functions  like util.pathjoin etc  abused and cause incompatibility issues. So I have not refactored here.

dundargoc:
Since we don't actually test/run any of the lua configs in CI to my knowledge, the best we can hope for is to make a string search of changed lines on PRs and complain if the deprecated function is used.

justinmk:
The "Proposal" in the main description of this issue is updated and correct. To be clear:

[...]

PR #3460, opened 2024-11-26
refactor: remove internal implementation of util.path.dirname

dundargoc:
Instead, just return the result of vim.fs.dirname.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -155,20 +155,7 @@ M.path = (function()
   --- @param path T
   --- @return T
   local function dirname(path)
-    local strip_dir_pat = '/([^/]+)$'
-    local strip_sep_pat = '/$'
-    if not path or #path == 0 then
-      return path
-    end
-    local result = path:gsub(strip_sep_pat, ''):gsub(strip_dir_pat, '')
-    if #result == 0 then
-      if iswin then
-        return path:sub(1, 2):upper()
-      else
-        return '/'
-      end
-    end
-    return result
+    return vim.fs.dirname(path)
   end
 
   local function path_join(...)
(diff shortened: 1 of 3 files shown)


glepnir:
LGTM.. Actually it can be removed. see https://github.com/neovim/nvim-lspconfig/pull/3434
Although the test passed. But the test here is a bit weak. I am afraid that hasty merge will break some uses. . So it has been in draft state.

glepnir (approved):

dundargoc:
> Actually it can be removed.

@justinmk pulled out a deprecation strategy in https://github.com/neovim/nvim-lspconfig/issues/2079#issuecomment-2393805975, I interpret his comment to mean we shouldn't remove any functions from util yet.

glepnir:
Yes. A safe approach :D. But I want to make lspconfig become a pure configuration collection. It would be helpful to exploit some of the functions of lspconfig step by step? If possible, is it a good idea to use deprecate here? Point to vim.fs.dirname so that any other plugins or users who use this function will remove the dependency on util.path.dirname.

justinmk:
This is a good example of why removing `util.dirname` gains almost nothing. Keeping the skeleton around is low cost. Because we'll just freeze it once everything is upstreamed. And meanwhile, shrinking the actual implementation gives us a clear picture of which logic still doesn't have an upstream equivalent. The mere presence of `util.dirname` doesn't harm anything, though, yes, we will need to update something like `.github/ci/run_sanitizer.sh` to prevent future uses of `util.dirname` in configs.

As we hollow-out the guts of `util`, it will become almost empty. We can later decide whether we want to delete it entirely, and that can be done as one big "breaking" step instead of many incremental "breaking" steps.

So this is a good change, and we should continue to make this kind of change. The possible future breaking change can wait until later. There is no need to do it right now. Let's just focus on shrinking the implementation, first.

lua/lspconfig/util.lua
```diff
@@ -155,20 +155,7 @@ M.path = (function()
   --- @param path T
   --- @return T
   local function dirname(path)
```
justinmk, line 157: a `@deprecated` token would help too

PR #3470, opened 2024-11-27
refactor: group deprecated functions at the end

dundargoc:
This makes it easier to assess how much functionality still needs to be deprecated.


lua/lspconfig/util.lua
```diff
@@ -407,4 +395,10 @@ function M.strip_archive_subpath(path)
   return path
 end
 
+---@deprecated use `vim.fs.dirname` instead
+M.dirname = vim.fs.dirname
```
justinmk, line 399: shouldn't this be `M.path.dirname` ?

PR #3475, opened 2024-11-28
refactor: deprecate util.path.is_dir

dundargoc:


dundargoc:
Dependent on https://github.com/neovim/nvim-lspconfig/pull/3474

PR #3476, opened 2024-11-28
refactor: deprecate util.find_mercurial_ancestor

dundargoc:

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -247,15 +247,6 @@ function M.find_git_ancestor(startpath)
   end)
 end
 
-function M.find_mercurial_ancestor(startpath)
-  return M.search_ancestors(startpath, function(path)
-    -- Support Mercurial directories
-    if vim.fn.isdirectory(M.path.join(path, '.hg')) == 1 then
-      return path
-    end
-  end)
-end
-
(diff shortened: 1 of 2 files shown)


dundargoc:
Dependent on https://github.com/neovim/nvim-lspconfig/pull/3475

lua/lspconfig/util.lua
```diff
@@ -405,4 +396,9 @@ function M.path.exists(filename)
   return stat and stat.type or false
 end
 
+--- @deprecated use `vim.fs.find('.hg', { path = startpath, upward = true })` instead
+function M.find_mercurial_ancestor(startpath)
+  return vim.fs.find('.hg', { path = startpath, upward = true })
```
glepnir, line 401: A limit should be set here. And it is possible that multiple matching directories will be returned. The first one found may not be the correct root directory

dundargoc: Wait what? How does one identify a "real" root directory? I assumed it meant the first parent directory containing ".hg"?

glepnir: Oh.. I misread this. I was confused and thought it was node search. Sorry I haven't slept for 35 hours so this is a bit confusing. nvm

glepnir (approved):

PR #3477, opened 2024-11-28
refactor: deprecate util.find_node_modules_ancestor

dundargoc:

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -297,9 +297,14 @@ below returns a function that takes as its argument the current buffer path.
 - `util.find_git_ancestor`: a function that locates the first parent directory
   containing a `.git` directory. >
    root_dir = util.find_git_ancestor
-- `util.find_node_modules_ancestor`: a function that locates the first parent
-  directory containing a `node_modules` directory. >
-   root_dir = util.find_node_modules_ancestor
+- Locate the first parent dir containing a "node_modules" dir: >lua
+  vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
+<
+  If you have Nvim 0.10 or newer then >lua
+  vim.fs.root(root_dir, "node_modules")
+<
+  can be used instead.
+  - Note: The old `util.find_node_modules_ancestor` API is deprecated and will be removed.
 - `util.find_package_json_ancestor`: a function that locates the first parent
   directory containing a `package.json`. >
    root_dir = util.find_package_json_ancestor
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -247,14 +247,6 @@ function M.find_git_ancestor(startpath)
   end)
 end
 
-function M.find_node_modules_ancestor(startpath)
-  return M.search_ancestors(startpath, function(path)
-    if vim.fn.isdirectory(M.path.join(path, 'node_modules')) == 1 then
-      return path
-    end
-  end)
-end
-
 function M.find_package_json_ancestor(startpath)
   return M.search_ancestors(startpath, function(path)
     local jsonpath = M.path.join(path, 'package.json')
@@ -399,6 +391,11 @@ end
 --- @deprecated use `vim.fs.find('.hg', { path = startpath, upward = true })[1]` instead
 function M.find_mercurial_ancestor(startpath)
   return vim.fs.find('.hg', { path = startpath, upward = true })[1]
+end
+
+--- @deprecated use `vim.fs.find('node_modules', { path = startpath, upward = true })[1]` instead
+function M.find_node_modules_ancestor(startpath)
+  return vim.fs.find('node_modules', { path = startpath, upward = true })[1]
 end
 
 return M
--- a/.github/ci/run_sanitizer.sh
+++ b/.github/ci/run_sanitizer.sh
@@ -15,7 +15,7 @@ if git diff --pickaxe-all -U0 -G "${SEARCH_PATTERN}" "${REF_BRANCH}" "${PR_BRANC
   exit 1
 fi
 
-SEARCH_PATTERN='(util\.path\.dirname|util\.path\.sanitize|util\.path\.exists|util\.path\.is_file|util\.path\.is_dir|util\.find_mercurial_ancestor)'
+SEARCH_PATTERN='(util\.path\.dirname|util\.path\.sanitize|util\.path\.exists|util\.path\.is_file|util\.path\.is_dir|util\.find_mercurial_ancestor|util\.find_node_modules_ancestor)'
 
 if git diff --pickaxe-all -U0 -G "${SEARCH_PATTERN}" "${REF_BRANCH}" "${PR_BRANCH}" -- '*.lua' | grep -Ev '\.lua$' | grep -E "^\+.*${SEARCH_PATTERN}" ; then
   echo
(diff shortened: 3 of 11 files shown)


doc/lspconfig.txt
```diff
@@ -297,9 +297,6 @@ below returns a function that takes as its argument the current buffer path.
 - `util.find_git_ancestor`: a function that locates the first parent directory
   containing a `.git` directory. >
    root_dir = util.find_git_ancestor
-- `util.find_node_modules_ancestor`: a function that locates the first parent
```
justinmk, line 300: Let's leave some breadcrumbs, for now. We can decide later how to present/word it.

```
- Locate the first parent dir containing a "node_modules" dir: >
  vim.fs.find('node_modules', { path = root_dir, upward = true })
>
  - Replaces the old `util.find_node_modules_ancestor` API which is deprecated and will be removed.
```

dundargoc: Tbh we already write that `find_...` is deprecated in favor of `vim.fs.find` so I guess I can just revert it and leave it be for now?

justinmk: The code snippet in the above example was helpful for me. It really helps not to have to track this stuff down later, why don't we just leave a breadcrumb? It is zero-cost.

dundargoc: I think you misunderstood, I was agreeing with you about leaving a breadcrumb. Your suggestion sounds fine, let's go with that 👍

lua/lspconfig/util.lua
```diff
@@ -405,4 +388,14 @@ function M.path.exists(filename)
...
 end
 
+--- @deprecated use `vim.fs.find('.hg', { path = startpath, upward = true })` instead
+function M.find_mercurial_ancestor(startpath)
+  return vim.fs.find('.hg', { path = startpath, upward = true })
+end
+
+--- @deprecated use `vim.fs.find('node_modules', { path = startpath, upward = true })` instead
+function M.find_node_modules_ancestor(startpath)
+  return vim.fs.find('node_modules', { path = startpath, upward = true })
```
glepnir, line 398: one thing to note. I remember there was an issue about util.find_node_modules_ancestor

```
mkdir -p /home/mike/test/a/node_modules
mkdir /home/mike/test/node_modules
```

run `:lua print( require'lspconfig'.util.find_node_modules_ancestor('/home/mike/test/a'))` the result shouldbe test not a

so does there  need `limit =math.huge` with a stop and return the last item in the result list ?  🤔

dundargoc: Oh wow, so the first "root marker" we find might not be the actual root marker? That's fucked up lol. OK, will test it out, thanks for the heads up.

glepnir: >  I assume this is not correct behavior for ".git" though?

yes something like git submodule

glepnir: it should return `test` not `test/a` ...i mean parent is actually root in this situation. maybe something like

```lua
  local resulst = vim.fs.find('node_modules', { path = startpath, upward = true, limit =math.huge, stop = vim.env.HOME })
 if #result > 2 then
      -- if find more and last is $HOME return #result - 1 should be root
    return { result[#result - 1] }

```

dundargoc: What do you mean by "should" though? Wouldn't we have gotten 5000 complaints already if any web related lsp wasn't working correctly?

dundargoc: That sounds like a recipe for disaster to fix something that will affect so many at the same time. I think we should just focus on pure refactor/rewrite for this PR and tackle the fix in a future PR.

glepnir: Related https://github.com/neovim/nvim-lspconfig/issues/1908. Took me some time to find it. I closed the issue because the original author did not respond. The plan is to fix it during the refactoring. So when I saw this PR, I thought of this issue. Of course, it can also be fixed in another PR.  not a big problem because there are only a few similar issues but  solved at the workspaceFolder level. But it is a bit strange that some of them do not support worksapceFolder .  it's fine.

glepnir: yes np and thanks. just it appear in my mind

lua/lspconfig/util.lua
```diff
@@ -401,4 +393,9 @@ function M.find_mercurial_ancestor(startpath)
   return vim.fs.find('.hg', { path = startpath, upward = true })
 end
 
+--- @deprecated use `vim.fs.find('node_modules', { path = startpath, upward = true })` instead
+function M.find_node_modules_ancestor(startpath)
+  return vim.fs.find('node_modules', { path = startpath, upward = true })
```
glepnir, line 398: vim.fs.find return a table before we return string.

lua/lspconfig/util.lua
dundargoc: I think that issue seems more about the documentation of `find_node_modules_ancestor` rather than any problems it has as a helper function related to `lspconfig`. I think the `find_node_modules_ancestor` behaves "correctly" in that it seems to do what most people/servers expect. I could be wrong, but my interpretation is that we don't need to "fix" anything in this case.

lua/lspconfig/util.lua
dundargoc: Oh shit, you're right. Thanks.

lua/lspconfig/util.lua
glepnir: > I think the find_node_modules_ancestor behaves

emm I think it should return test instead of test/a , servers that support workspaceFolder, if you open test first and then open the file in a, the server will be reused. There is no problem here. Most people do this. But if you open a first and then open the file in test, it will be messed up. However, this is an isolated case, but it is still a problem.

lua/lspconfig/util.lua
dundargoc: I guess I need to wrap all `vim.fs.find` with `next()`?

lua/lspconfig/util.lua
glepnir: But as you said, this has nothing to do with PR. We shouldn't focus on this. Just keep doing it. No one will do it anyway. 😮‍💨

dundargoc:
Should we wait for the minimum supported version to be 0.10 before we do this because of `vim.fs.root`? Right now, the replacement is `next(vim.fs.root(root_dir, 'node_module') or {})` which isn't super elegant.

glepnir:
next will return key and value .. `vim.F.if_nil(unpack(...` or  currentl behavior is always first item so just next(result) ~= nil and result[1] or nil

dundargoc:
OK, I think this should work in all instances:

```
vim.fs.find('node_modules', {path='/Users/dundargoc/test/a', upwards=true})[1]
```

It's not pretty.

glepnir:
yeah .but it work 😊 can improve on next fix `test` not `test/a`

justinmk:
> hould we wait for the minimum supported version to be 0.10 before we do this because of `vim.fs.root`? Right now, the replacement we can use is `vim.fs.find('node_modules', {path='/Users/dundargoc/test/a', upwards=true})[1]`

No, let's move forward with this. But let's also mention `vim.fs.root` in the "breadcrumbs" note, for users of Nvim 0.10.

dundargoc:
Done

dundargoc:
@justinmk done

justinmk (approved):

PR #3483, opened 2024-12-01
refactor: replace all instances of vim.uv with vim.loop

dundargoc:
We still support neovim 0.9 currently, so we can't use vim.uv. Also add
a check so we don't accidentally reintroduce it.

### DIFF
--- a/lua/lspconfig/manager.lua
+++ b/lua/lspconfig/manager.lua
@@ -1,6 +1,5 @@
 local api = vim.api
 local lsp = vim.lsp
-local uv = vim.uv or vim.loop
 
 local async = require 'lspconfig.async'
 local util = require 'lspconfig.util'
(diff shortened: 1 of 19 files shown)


lua/lspconfig/manager.lua
```diff
@@ -1,6 +1,5 @@
 local api = vim.api
 local lsp = vim.lsp
-local uv = vim.uv or vim.loop
```
justinmk, line 3: why revert these?

justinmk (approved):

lua/lspconfig/manager.lua
dundargoc: Eh, why not? I thought it'd be simpler to just use vim.loop until we don't need it anymore, this compat hack isn't really needed. Can unrevert if you prefer

glepnir (approved):

lua/lspconfig/manager.lua
dundargoc: Also I get nervous when I see "uv" because I know we can't use "vim.uv", but then I see that "uv" isn't "vim.uv", we annoyed me.

PR #3484, opened 2024-12-01
refactor: remove outdated code

dundargoc:
Lspconfig requires neovim version 0.9 at the time of writing this, so
discard any version checks and code for older versions.

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -6,17 +6,6 @@ vim.g.lspconfig = 1
 local api, lsp = vim.api, vim.lsp
 local util = require('lspconfig.util')
 
-if vim.fn.has 'nvim-0.8' ~= 1 then
-  local version_info = vim.version()
-  local warning_str = string.format(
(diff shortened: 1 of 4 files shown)


plugin/lspconfig.lua
```diff
@@ -6,17 +6,6 @@ vim.g.lspconfig = 1
 local api, lsp = vim.api, vim.lsp
 local util = require('lspconfig.util')
 
-if vim.fn.has 'nvim-0.8' ~= 1 then
-  local version_info = vim.version()
-  local warning_str = string.format(
-    '[lspconfig] requires neovim 0.8 or later. Detected neovim version: 0.%s.%s',
```
justinmk, line 12: this should be bumped to 0.9, not removed entirely

we need an "active" check like this somewhere, to avoid confusion and bug reports.

dundargoc: Does the version check in lsoconfig.lua not suffice? I assumed we only needed one check.

justinmk (approved):

PR #3486, opened 2024-12-01
feat(bqls): Add bqls support

joshbode:
Adds support for BigQuery language server.

thanks @kitagry

### DIFF
--- a/lua/lspconfig/configs/bqls.lua
+++ b/lua/lspconfig/configs/bqls.lua
@@ -1,0 +1,22 @@
+return {
+  default_config = {
+    cmd = { 'bqls' },
+    filetypes = { 'sql' },
+    root_dir = function(fname)
+      return vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
+    end,
+    single_file_support = true,
+    settings = {},
+  },
+  docs = {
+    description = [[
+https://github.com/kitagry/bqls
+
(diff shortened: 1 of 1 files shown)


glepnir:
Thanks. But we have some rules for new servers. Please see https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md#adding-a-server-to-lspconfig

could you provide some more information about this server how popular it is

joshbode:
I would say not very popular (yet!) but maybe @kitagry has some more context (there is a [plugin](https://github.com/kitagry/bqls.nvim) that adds the config dynamically).

I was hoping that adding support would increase the exposure of this LS, which appears to provide better support for BigQuery than the existing [`sqlls`](https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#sqlls) server.

glepnir:
It's a great server. But it's only 7 stars at the moment. And it doesn't have the same download volume as other editors like vscode. So I think we can wait a little longer to avoid our repo becoming bloated with some small and unpopular server configuration.

el-iot:
I'm also using this language server in neovim and would love to see this merged.

justinmk:
please rebase. also try to avoid use of `util.lua`, see other configs for reference

joshbode:
thank you, @justinmk :)

PR #3491, opened 2024-12-03
fix(svelte): prefer git over package.json

igorlfs:
When using the Svelte LSP on a monorepo, the preference for a `package.json` may set an incorrect root dir.

To address that, prefer finding the git "root" instead.

### DIFF
--- a/lua/lspconfig/configs/svelte.lua
+++ b/lua/lspconfig/configs/svelte.lua
@@ -4,7 +4,7 @@ return {
   default_config = {
     cmd = { 'svelteserver', '--stdio' },
     filetypes = { 'svelte' },
-    root_dir = util.root_pattern('package.json', '.git'),
+    root_dir = util.root_pattern('.git', 'package.json'),
   },
   docs = {
     description = [[


glepnir:
package.json is a common file for the frontend. The priority here should be fine for most people. I think you can adjust it locally

igorlfs:
> package.json is a common file for the frontend.

Indeed, it is, that's why it's kept as a fallback. This change doesn't affect those who rely on a package.json, but fixes the issue with monorepos

Besides, there's precedent for preferring `.git` over `package.json`:

https://github.com/neovim/nvim-lspconfig/blob/2c007c79eedfca0ec652c50d571cb0c52d1577ba/lua/lspconfig/configs/solidity.lua#L7

https://github.com/neovim/nvim-lspconfig/blob/2c007c79eedfca0ec652c50d571cb0c52d1577ba/lua/lspconfig/configs/somesass_ls.lua#L8

https://github.com/neovim/nvim-lspconfig/blob/2c007c79eedfca0ec652c50d571cb0c52d1577ba/lua/lspconfig/configs/prismals.lua#L12

Issue #3499, opened 2024-12-10
Fail to start kotlin-language-server due to suffix extension


KevinNitroG:
### Description

In the kotlin-language-server README, they say the executable file can be suffixed with sh, cmd, or bat (https://github.com/fwcd/kotlin-language-server#figuring-out-the-dependencies). However, the current nvim-lspconfig for kotlin LSP on Windows will be suffixed with .bat, which failed to start on my machine.

![Image](https://github.com/user-attachments/assets/47d26e5e-f37d-4d8a-871e-9491be14b9a9)

Current nvim-lspconfig: https://github.com/neovim/nvim-lspconfig/blob/5a812abc65d529ea7673059a348814c21d7f87ff/lua/lspconfig/configs/kotlin_language_server.lua#L4C1-L7C1

I tried to remove the suffix extension (`cmd = { 'kotlin-language-server' }`), and it worked fine. Also `kotlin-language-server.cmd` exist on my PATH (installed via mason), not `kotlin-language-server.bat`

justinmk:
Looks like a bug, that part of the config is 3 years old. Send a PR to fix it?

KevinNitroG:
Alright I'm gonna create a PR for this

PR #3500, opened 2024-12-11
fix(kotlin_language_server): fix wrong cmd

KevinNitroG:
Change .bat extension to .cmd because .bat is not available.

Ref: #3499

### DIFF
--- a/lua/lspconfig/configs/kotlin_language_server.lua
+++ b/lua/lspconfig/configs/kotlin_language_server.lua
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 local bin_name = 'kotlin-language-server'
 if vim.fn.has 'win32' == 1 then
-  bin_name = bin_name .. '.bat'
+  bin_name = bin_name .. '.cmd'
 end
 
 --- The presence of one of these files indicates a project root directory


lua/lspconfig/configs/kotlin_language_server.lua
```diff
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 local bin_name = 'kotlin-language-server'
 if vim.fn.has 'win32' == 1 then
-  bin_name = bin_name .. '.bat'
+  bin_name = bin_name .. '.cmd'
 end
```
justinmk, line 6: can't we just delete this entire block? cc @glepnir

KevinNitroG: Yes I think so. Also other configs. Because on my machine it still works when not declaring extension.

image

KevinNitroG: I just changed .bat to .cmd because I see other config check for win32 to add extension

justinmk: > I see other config check for win32 to add extension

those are probably just old

glepnir: cmd already process by exepath. no need if block

KevinNitroG: > cmd already process by exepath. no need if block

So I will remove the if block then?

Do you want me to create another PR for removing if block also?

justinmk: > So I will remove the if block then?

yes



> Do you want me to create another PR for removing other configs' if block also?

sure, unless @glepnir knows of a reason we didn't do that already

PR #3501, opened 2024-12-12
fix(tinymist): fallback to single file in some case

Yakkhini:
When the `root_dir` is null, tinymist should still runs to give a baseline usage.

Ref: #3094 in this PR the modify of `root_dir` is rejected, but set single file support field to true looks just fine.

### DIFF
--- a/lua/lspconfig/configs/tinymist.lua
+++ b/lua/lspconfig/configs/tinymist.lua
@@ -5,7 +5,7 @@ return {
     cmd = { 'tinymist' },
     filetypes = { 'typst' },
     root_dir = util.find_git_ancestor,
-    single_file_support = false,
+    single_file_support = true,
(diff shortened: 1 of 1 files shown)


Yakkhini:
@Myriad-Dreamin

glepnir:
Does tinymist support null root dir ? could you also provide a link about this.

Yakkhini:
> Does tinymist support null root dir ? could you also provide a link about this.

Just like @Myriad-Dreamin says above. Actually before this PR the tinymist would not start in most times since that lots people not put their typst codes into a git repo. ;) (https://github.com/Myriad-Dreamin/tinymist/issues/528) I tried to override the default `single_file_support` to true and it works.

glepnir:
okay thx

PR #3502, opened 2024-12-13
refactor: deprecate util.find_package_json_ancestor

dundargoc:

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -246,18 +246,9 @@ function M.find_git_ancestor(startpath)
   end)
 end
 
-function M.find_package_json_ancestor(startpath)
-  return M.search_ancestors(startpath, function(path)
-    local jsonpath = M.path.join(path, 'package.json')
-    if (vim.loop.fs_stat(jsonpath) or {}).type == 'file' then
-      return path
-    end
-  end)
-end
-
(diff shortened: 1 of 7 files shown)


justinmk (approved):

lua/lspconfig/util.lua
```diff
@@ -397,4 +388,9 @@ function M.find_node_modules_ancestor(startpath)
   return vim.fs.find('node_modules', { path = startpath, upward = true })[1]
 end
 
+--- @deprecated use `vim.fs.find('package.json', { path = startpath, upward = true })[1]` instead
+function M.find_package_json_ancestor(startpath)
+  return vim.fs.find('package.json', { path = startpath, upward = true })[1]
```
glepnir, line 393: do we also need `type = file` ? before use fs_stat to validate it's a file

dundargoc: I know, but I don't like that because it's longer/less copypastable and also because `vim.fs.root` (which will replace this eventually) don't actually check if root marker is a file or not. So I think if we can get away with just checking for existence then that will make things simpler in the future.

glepnir: yes not a big deal. go.

glepnir (approved):

PR #3505, opened 2024-12-13
Fix for vim.fs.find regression not returning parent directory

jacobrreed:
fixed #3503

PR #3506, opened 2024-12-13
docs: update following #3505

guilhas07:
Problem: 
Annotation and docs were not update accordingly in #3505 

Solution:
Update them


justinmk:
Thanks!

PR #3507, opened 2024-12-13
refactor: deprecate util.find_git_ancestor

dundargoc:

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -236,16 +236,6 @@ function M.root_pattern(...)
   end
 end
 
-function M.find_git_ancestor(startpath)
-  return M.search_ancestors(startpath, function(path)
-    -- Support git directories and git files (worktrees)
-    local gitpath = M.path.join(path, '.git')
-    if vim.fn.isdirectory(gitpath) == 1 or (vim.loop.fs_stat(gitpath) or {}).type == 'file' then
-      return path
-    end
-  end)
-end
-
 function M.insert_package_json(config_files, field, fname)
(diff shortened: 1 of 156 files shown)


lua/lspconfig/util.lua
```diff
@@ -393,4 +383,9 @@ function M.find_package_json_ancestor(startpath)
   return vim.fs.dirname(vim.fs.find('package.json', { path = startpath, upward = true })[1])
 end
 
+--- @deprecated use `vim.fs.find('.git', { path = startpath, upward = true })[1]` instead
+function M.find_git_ancestor(startpath)
+  return vim.fs.find('.git', { path = startpath, upward = true })[1]
```
glepnir, line 388: use `vim.fs.dirname` to get parent dir of result  `{ "/Users/mw/workspace/neovim/.git" }`

justinmk: https://github.com/neovim/nvim-lspconfig/pull/3505

dundargoc:
pinging @jacobrreed @folke to take a look in case I accidentally destroy the internet again

lua/lspconfig/util.lua
glepnir: Considering this is a very common situation, maybe it is better to add parent = true to the filter table

justinmk (approved):

lua/lspconfig/util.lua
dundargoc: You mean to vim.fs.find? But we already have vim.fs.root that does this, I don't think we need to add anything to vim.fs.find.

PR #3510, opened 2024-12-14
refactor!: remove util.path.escape_wildcards

dundargoc:

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -97,10 +97,6 @@ end
 
 -- Some path utilities
 M.path = (function()
-  local function escape_wildcards(path)
-    return path:gsub('([%[%]%?%*])', '\\%1')
-  end
-
   --- @param path string
   --- @return boolean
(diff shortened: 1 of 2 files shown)


lua/lspconfig/util.lua
```diff
@@ -378,6 +377,11 @@ function M.path.exists(filename)
   return stat and stat.type or false
 end
 
+--- @deprecated
+function M.path.escape_wildcards(path)
+  return path:gsub('([%[%]%?%*])', '\\%1')
```
justinmk, line 382: should this use the local one from line 214?

justinmk (approved):

lua/lspconfig/util.lua
dundargoc: > we could try removing this , I doubt anyone else is using it

Yeah, this is barely used when I did code search.

dundargoc: OK, removing it instead.

PR #3511, opened 2024-12-14
refactor!: remove util.path.is_absolute

dundargoc:

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -107,16 +107,6 @@ M.path = (function()
     end
   end
 
-  --- @param filename string
-  --- @return boolean
-  local function is_absolute(filename)
-    if iswin then
-      return filename:match '^%a:' or filename:match '^\\\\'
-    else
-      return filename:match '^/'
-    end
-  end
-
   local function path_join(...)
     return table.concat(M.tbl_flatten { ... }, '/')
   end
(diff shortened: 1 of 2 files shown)


lua/lspconfig/util.lua
```diff
@@ -370,6 +359,13 @@ M.path.dirname = vim.fs.dirname
 --- @deprecated use `vim.fs.normalize` instead
 M.path.sanitize = vim.fs.normalize
 
+--- @deprecated use `vim.fn.isabsolutepath(filename) == 1` instead
+--- @param filename string
+--- @return boolean
+function M.path.is_absolute(filename)
+  return vim.fn.isabsolutepath(filename) == 1
```
justinmk, line 366: this was added in Nvim 0.11 https://github.com/neovim/neovim/commit/37952bf7b442cac794c4663f2e0123e7d72bc443 , so doesn't help us here.

but we could try removing this, it's possible that no one is using it.

justinmk (approved):

lua/lspconfig/util.lua
dundargoc: > this was added in Nvim 0.11 [neovim/neovim@37952bf](https://github.com/neovim/neovim/commit/37952bf7b442cac794c4663f2e0123e7d72bc443) , so doesn't help us here.

Doh, I should've checked. Thanks

> but we could try removing `path.is_absolute`, it's possible that no one is using it.

ThePrimagen uses it in his config, so it's worth to remove it just to annoy him lol

justinmk (approved):

PR #3513, opened 2024-12-14
fix ruff and ruff_lsp returning a function from root pattern

### LINKED ISSUE #3512: Error: [lspconfig] unhandled error: vim/fs.lua:0: path: expected string, got function
### Description

Getting this warning with basedpyright:

```
[lspconfig] unhandled error: vim/fs.lua:0: path: expected string, got function
```

I'm using LazyVim with this config:

https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/python.lua

apostolos-geyer:
fixes #3512 

the issue was that `util.root_pattern` returns a function, which i believe is intended to be called with fname as its argument.

this causes root_dir to return a funcref when called which causes the attached issue as well as causing an error on LspStart ruff.

### DIFF
--- a/lua/lspconfig/configs/crystalline.lua
+++ b/lua/lspconfig/configs/crystalline.lua
@@ -5,7 +5,8 @@ return {
     cmd = { 'crystalline' },
     filetypes = { 'crystal' },
     root_dir = function(fname)
-      return util.root_pattern 'shard.yml' or vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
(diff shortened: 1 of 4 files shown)


apostolos-geyer:
sorry if therse any guidelines i missed on this i forgot to read it first i've never contributed to any foss before this was just broken for me and saw someone had the same issue so thought i should push my fix

justinmk:
Thank you! Found 2 other cases:

```
crystalline.lua
janet_lsp.lua

```

PR #2110, opened 2022-09-05
fix(lua-language-server): check if root dir is home directory

uga-rosa:
There have been cases reported where the absence of this check causes the home directory to be set as the workspace, resulting in an error message. 
```
LSP[sumneko_lua] Your workspace is set to '/home/uga'. Lua language server refused to load this directory. Please check your configuration.[learn more here](https://github.com/sumneko/lua-language-server/wiki/Why-scanning-home-folder)
```

PR #3515, opened 2024-12-15
fix(lua_ls): return longest root path

### LINKED ISSUE #3508: bug(lua_ls): incorrect root_dir due to early return when root_files is found which ignore longer root_git (and root_lua)
### Description

The issue is similar to https://github.com/neovim/nvim-lspconfig/issues/3165 but now it seems it's caused by `root_files`. The issue occurs with the following directory structure:

```sh
HOME/
├─ workspaces/ # `root`, actual `root_dir`
│  ├─ stylua.toml # one of `root_files`
│  ├─ lua/
│  ├─ python/
│  ├─ work/
│  │  ├─ SomeProjectWithLuaFiles/ # `root_git`, expected `root_dir`
│  │  │  ├─ .git/
```

Whenever I open a file under `../SomeProjectWithLuaFiles/`, I expect `root_dir` is set to `../SomeProjectWithLuaFiles/`. However somehow `root_dir` is set to `HOME/workspaces/`. After quick investigation, it seems `stylua.toml` makes the LSP set `root_dir` to `HOME/workspaces/` because after I removed `stylua.toml` the `root_dir` is now `../SomeProjectWithLuaFiles/`. Looking at `lua_ls` config, it seems this is caused by early return from the following lines:

https://github.com/neovim/nvim-lspconfig/blob/3cb6c05779459eb5fe9cddbb47675593594f0415/lua/lspconfig/configs/lua_ls.lua#L19-L21

https://github.com/neovim/nvim-lspconfig/pull/3322 suggest setting `root_dir` to the longer root path if both a `.git`- and a `lua`-ancestor are found, which is not satisfied in this case.

Please let me know if this is confirmed to be an issue. I am happy to help with a PR.

mangkoran:
This commit doesn't address the case when root path == vim.env.HOME as the case lack return when it's true (#2110)

Fix #3508

### DIFF
--- a/lua/lspconfig/configs/lua_ls.lua
+++ b/lua/lspconfig/configs/lua_ls.lua
@@ -15,16 +15,15 @@ return {
     cmd = { 'lua-language-server' },
     filetypes = { 'lua' },
     root_dir = function(fname)
-      local root = util.root_pattern(unpack(root_files))(fname)
-      if root and root ~= vim.env.HOME then
-        return root
-      end
+      local root_file = util.root_pattern(unpack(root_files))(fname) or ''
       local root_lua = util.root_pattern 'lua/'(fname) or ''
       local root_git = vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1]) or ''
-      if root_lua == '' and root_git == '' then
+      if root_file == '' and root_lua == '' and root_git == '' then
         return
       end
-      return #root_lua >= #root_git and root_lua or root_git
+      local root = #root_file >= #root_lua and root_file or root_lua
+      root = #root >= #root_git and root or root_git
+      return root
     end,
     single_file_support = true,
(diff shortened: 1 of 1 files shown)


glepnir (approved):

pitkling (approved):

mangkoran:
Hi @justinmk is there other thing that need to be addressed for https://github.com/neovim/nvim-lspconfig/issues/3508?

mangkoran:
Hi @justinmk is there any other thing that blocking this PR for being merged?

mangkoran:
Hi @dundargoc @justinmk could we have this PR merged to fix https://github.com/neovim/nvim-lspconfig/issues/3508 ?

I read the linked discussion regarding util.root_pattern deprecation, however it seems it's [postponed](https://github.com/neovim/nvim-lspconfig/pull/3569#issuecomment-2605053357) for now.

PR #3520, opened 2024-12-21
refactor: deprecate util.path.join

dundargoc:

### DIFF
--- a/lua/lspconfig/configs/jdtls.lua
+++ b/lua/lspconfig/configs/jdtls.lua
@@ -8,19 +8,19 @@ local env = {
 }
 
 local function get_cache_dir()
-  return env.XDG_CACHE_HOME and env.XDG_CACHE_HOME or util.path.join(env.HOME, '.cache')
+  return env.XDG_CACHE_HOME and env.XDG_CACHE_HOME or env.HOME .. '/.cache'
 end
 
 local function get_jdtls_cache_dir()
-  return util.path.join(get_cache_dir(), 'jdtls')
+  return get_cache_dir() .. '/jdtls'
 end
 
 local function get_jdtls_config_dir()
-  return util.path.join(get_jdtls_cache_dir(), 'config')
+  return get_jdtls_cache_dir() .. '/config'
 end
 
 local function get_jdtls_workspace_dir()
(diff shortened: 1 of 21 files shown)


dundargoc:
@justinmk I'd like for you to check this is ok, don't want to put in too much time on something that might get yeeted.

lua/lspconfig/util.lua
```diff
@@ -107,10 +107,6 @@ M.path = (function()
     end
   end
 
-  local function path_join(...)
-    return table.concat(M.tbl_flatten { ... }, '/')
```
justinmk, line 111: Didn't even support `\` slashes so this was just cosmetic...

lua/lspconfig/configs/jdtls.lua
```diff
@@ -8,19 +8,19 @@ local env = {
...
 end
 
 local function get_jdtls_config_dir()
-  return util.path.join(get_jdtls_cache_dir(), 'config')
+  return get_jdtls_cache_dir() .. '/config'
 end
 
 local function get_jdtls_workspace_dir()
-  return util.path.join(get_jdtls_cache_dir(), 'workspace')
+  return get_jdtls_cache_dir() .. '/workspace'
```
justinmk, line 23: I could imagine java or java tooling using a .bat/.cmd wrapper on Windows that possibly uses a windows shell command that doesn't support `/` slashes. But the lspconfig join() didn't support `\` slashes, so we aren't regressing anything here.

justinmk (approved):

lua/lspconfig/util.lua
dundargoc: Tbf even neovim's `vim.fs.joinpath` doesn't use `\` or provide the option to do that. I guess the usecase is being able to concatenate paths without wrapping a table around it? Idk.

justinmk: yeah and avoiding double slashes "//" (technically valid, but maybe some plugins/scripts break on it). And communicates "intention".

For Nvim core, avoiding `\` slashes is probably good, because it should be discouraged. Most of the time, `\` is cargo-culting.

justinmk (approved):
I tagged a new release. Krampus will get anyone who is not using the tagged release

PR #3524, opened 2024-12-22
refactor: deprecate util.path.is_descendant

dundargoc:

### DIFF
--- a/lua/lspconfig/configs/intelephense.lua
+++ b/lua/lspconfig/configs/intelephense.lua
@@ -1,5 +1,9 @@
 local util = require 'lspconfig.util'
 
+local function is_descendant(root, path)
+  return vim.startswith(vim.fs.normalize(path), vim.fs.normalize(root))
+end
+
 return {
   default_config = {
     cmd = { 'intelephense', '--stdio' },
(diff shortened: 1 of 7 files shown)


lua/lspconfig/configs/intelephense.lua
```diff
@@ -1,5 +1,9 @@
 local util = require 'lspconfig.util'
 
+local function is_descendant(root, path)
+  return vim.startswith(vim.fs.normalize(path), vim.fs.normalize(root))
```
justinmk, line 4: If we don't have an answer in `vim.fs` then we probably should wait until we do. Maybe it could be named `vim.fs.has()`

dundargoc: IMO something like `vim.fs.relpath` that return path relative to another (similar to `:~` in fnamemodify but generalized) would probably be more useful. Haven't fully thought about it, but I imagine it'd return `nil` in the case it isn't a descendant.

Actually, most uses of is_descendant in our configs could already be replaced by fnamemodify, it is only rust_analyzer that requires a generalized `is_descendant`. We could move this function to the rust_analyzer file and just use fnamemodify for the other?

justinmk: We could but it would be better to get stuff into `vim.fs` so that we can use it in the configs, instead of having to redo the configs again ...

dundargoc:
Closing as some functionality is needed in core for this to be smooth.

PR #3525, opened 2024-12-22
refactor!: remove public interface of util.path.traverse_parents

dundargoc:


dundargoc:
Removed public function it since none of the configs actually use this and I don't think it's used much in the wild anyway, so figured this should be fine.

glepnir:
If this function is not used internally, it can be safely removed. I think we never recommend any configuration to use the function encapsulated here. vim.fs is always our direction

glepnir (approved):

justinmk (approved):

Issue #3496, opened 2024-12-09
Duplicate `node_modules` appears in path in astro and volar lsp


Parsifa1:
### Description

```
Error executing vim.schedule lua callback: ...rapped-nightly/share/nvim/runtime/lua/vim/lsp/client.lua:551: RPC[Error] code_name = InternalError, message = "Request initialize failed with message: Can't find typescript.js or tsserverlibrary.js in \"/Users/parsifa1/Project/blog/node_modules/node_modules/typescript/lib\""
```

![Image](https://github.com/user-attachments/assets/5d466fba-6048-4bf9-a61b-208f55e06c47)

after logging:

![Image](https://github.com/user-attachments/assets/eee5dfcc-bcbf-43f1-82b2-624764e8e823)

the `project_root` seems to have different return

using neovim nightly

PR #3527, opened 2024-12-24
fix(mdx-analyzer): Duplicate node_modules folder causing errors, and filetypes option.

MuntasirSZN2:
In #3496, astro ls and volar lsp was fixed for having duplicate node_modules folder, but mdx-analyzer doesn't. This pr fixes that. Not only that, the `filetypes` option had `markdown.mdx`, but now vim.bo.filetype returns `mdx`, thats fixed also.

### DIFF
--- a/lua/lspconfig/configs/mdx_analyzer.lua
+++ b/lua/lspconfig/configs/mdx_analyzer.lua
@@ -2,7 +2,7 @@ local util = require 'lspconfig.util'
 
 local function get_typescript_server_path(root_dir)
   local project_root = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/mdx_analyzer.lua
```diff
@@ -2,13 +2,13 @@ local util = require 'lspconfig.util'
 
 local function get_typescript_server_path(root_dir)
   local project_root = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
```
glepnir, line 4:
```suggestion
  local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
```

PR #3535, opened 2025-01-01
feat(ccls): add CclsSwitchSourceHeader

MaskRay:
ccls 0.20241108 has added support for the extension from clangd.

### DIFF
--- a/lua/lspconfig/configs/ccls.lua
+++ b/lua/lspconfig/configs/ccls.lua
@@ -3,22 +3,22 @@ local util = require 'lspconfig.util'
 local function switch_source_header(bufnr)
   bufnr = util.validate_bufnr(bufnr)
   local client = util.get_active_client_by_name(bufnr, 'ccls')
-  local params = { uri = vim.uri_from_bufnr(bufnr) }
-  if client then
-    client.request('textDocument/switchSourceHeader', params, function(err, result)
-      if err then
-        error(tostring(err))
-      end
-      if not result then
-        print 'Corresponding file cannot be determined'
-        return
-      end
-      vim.api.nvim_command('edit ' .. vim.uri_to_fname(result))
-    end, bufnr)
-  else
-    print 'method textDocument/switchSourceHeader is not supported by any servers active on the current buffer'
+  if not client then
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/ccls.lua
```diff
@@ -1,5 +1,25 @@
...
+      if not result then
+        print 'Corresponding file cannot be determined'
+        return
+      end
+      vim.api.nvim_command('edit ' .. vim.uri_to_fname(result))
+    end, bufnr)
+  else
+    print 'method textDocument/switchSourceHeader is not supported by any servers active on the current buffer'
+  end
+end
```
glepnir, line 21:
```suggestion
local function switch_source_header(bufnr)
  bufnr = util.validate_bufnr(bufnr)
  local client = util.get_active_client_by_name(bufnr, 'ccls')
  if not client then
    vim.notify('method textdocument/switchsourceheader is not supported by any servers active on the current buffer')
  end
  local params = vim.lsp.util.make_text_document_params(bufnr)
  client.request('textdocument/switchsourceheader', params, function(err, result)
    if err then
      error(tostring(err))
    end
    if not result then
      vim.notify('corresponding file cannot be determined')
      return
    end
    vim.cmd.edit(vim.uri_to_fname(result))
  end, bufnr)
end

```

lua/lspconfig/configs/ccls.lua
```diff
@@ -1,5 +1,25 @@
...
+      error(tostring(err))
+    end
+    if not result then
+      vim.notify('corresponding file cannot be determined')
+      return
+    end
+    vim.cmd.edit(vim.uri_to_fname(result))
+  end, bufnr)
+end
+
```
glepnir, line 21:
```suggestion
```

MaskRay:
@glepnir The fixup! commit introduced an issue. The method name was changed from `textDocument/switchSourceHeader` to `textdocument/switchsourceheader`. The latter form is invalid and not accepted by ccls...

glepnir:
oh a typo. in nightly we have constant method variables..but they are not available in 0.10x ..fix in https://github.com/neovim/nvim-lspconfig/pull/3536

PR #3536, opened 2025-01-02
fix(ccls): typo on request method

glepnir:
Problem: method prefix is textDocument

Solution: make d to upper

### DIFF
--- a/lua/lspconfig/configs/ccls.lua
+++ b/lua/lspconfig/configs/ccls.lua
@@ -7,7 +7,7 @@ local function switch_source_header(bufnr)
     vim.notify('method textdocument/switchsourceheader is not supported by any servers active on the current buffer')
   end
   local params = vim.lsp.util.make_text_document_params(bufnr)
-  client.request('textdocument/switchsourceheader', params, function(err, result)
+  client.request('textDocument/switchsourceheader', params, function(err, result)
     if err then
       error(tostring(err))
     end


lua/lspconfig/configs/ccls.lua
```diff
@@ -7,7 +7,7 @@ local function switch_source_header(bufnr)
     vim.notify('method textdocument/switchsourceheader is not supported by any servers active on the current buffer')
   end
   local params = vim.lsp.util.make_text_document_params(bufnr)
-  client.request('textdocument/switchsourceheader', params, function(err, result)
+  client.request('textDocument/switchsourceheader', params, function(err, result)
```
MaskRay, line 10: The correct spelling is `textDocument/switchSourceHeader` and the message above needs to be fixed as well.

lua/lspconfig/configs/ccls.lua
```diff
@@ -7,7 +7,7 @@ local function switch_source_header(bufnr)
     vim.notify('method textdocument/switchsourceheader is not supported by any servers active on the current buffer')
```
MaskRay, line 7: here

lua/lspconfig/configs/ccls.lua
glepnir: will fix on refactor PR there lots of duplicate codes on clangd.

glepnir: `switchsourceheader` is correct? in clangd is `switchSourceHeader`

MaskRay: `textDocument/switchSourceHeader`, the same as clangd :)

I had used the function copied from clangd for a while https://github.com/MaskRay/Config/blob/1f431c999079e900bbe7958a00578cba4c4b3b67/home/.config/nvim/lua/my/util.lua#L41  until I attempted to upstream it...

PR #3542, opened 2025-01-05
fix(msbuild): more root patterns

sharpchen:


glepnir:
does there any information about these filetypes are supported by this server?

justinmk (approved):
shouldn't `*.sln` be prioritized (i.e. make it first in the list)?

sharpchen:
> does there any information about these filetypes are supported by this server?

https://github.com/tintoy/msbuild-project-tools-vscode?tab=readme-ov-file#file-extensions

And I might misunderstand the `root_parttern`, does this implies the server only auto start when the current working directory matches any of the patterns, if not then `sln` and `slnx` might be redundant.

> shouldn't *.sln be prioritized (i.e. make it first in the list)?

right, updated.

justinmk (approved):

PR #3544, opened 2025-01-05
fix(slangd): use .slang instead of .shaderslang for slang

mTvare6:
The extension used by slang is .slang.
[Examples] and [docs] for reference.

[Examples]: https://github.com/shader-slang/slang/tree/master/examples
[docs]: https://shader-slang.com/slang/user-guide/get-started.html

### DIFF
--- a/lua/lspconfig/configs/slangd.lua
+++ b/lua/lspconfig/configs/slangd.lua
@@ -7,7 +7,7 @@ end
 return {
   default_config = {
     cmd = { bin_name },
-    filetypes = { 'hlsl', 'shaderslang' },
+    filetypes = { 'hlsl', 'slang' },
     root_dir = function(fname)
       return vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
     end,


lithammer:
This isn't a file extension though. It's a `filetype`[^1].

[^1]: https://vimhelp.org/filetype.txt.html

mTvare6:
My bad, but neovim itself doesn't have a shaderslang filetype yet either way.

lithammer:
Seems like the filetype was added to [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter). So you would need that and enable "slang" for his to make sense.

> The filetype `shaderslang` is a invention of myself, but I plan to also use it for a tree-sitter parser [nvim-treesitter/nvim-treesitter#5659](https://github.com/nvim-treesitter/nvim-treesitter/pull/5659) and a TBD PR to vim for filetype detection. `shaderslang` is used because Vim/Neovim already has support for a filetype called `slang` which I assume is for https://en.wikipedia.org/wiki/S-Lang with extension `*.sl` while `shaderslang` uses `*.slang`.


- https://github.com/neovim/nvim-lspconfig/pull/2907
- https://github.com/nvim-treesitter/nvim-treesitter/pull/5659

mTvare6:
Vim now supports [shaderslang](https://github.com/vim/vim/pull/16387)

PR #3546, opened 2025-01-06
fix(clojure_lsp): adds current folder as fallback for root_dir

guilhermefelipecgs:
This is necessary to allow navigation within the dependencies.

### DIFF
--- a/lua/lspconfig/configs/clojure_lsp.lua
+++ b/lua/lspconfig/configs/clojure_lsp.lua
@@ -4,7 +4,13 @@ return {
   default_config = {
     cmd = { 'clojure-lsp' },
     filetypes = { 'clojure', 'edn' },
-    root_dir = util.root_pattern('project.clj', 'deps.edn', 'build.boot', 'shadow-cljs.edn', '.git', 'bb.edn'),
+    root_dir = function(pattern)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/clojure_lsp.lua
```diff
@@ -4,7 +4,13 @@ return {
   default_config = {
     cmd = { 'clojure-lsp' },
     filetypes = { 'clojure', 'edn' },
-    root_dir = util.root_pattern('project.clj', 'deps.edn', 'build.boot', 'shadow-cljs.edn', '.git', 'bb.edn'),
+    root_dir = function(pattern)
+      local fallback = vim.loop.cwd()
```
glepnir, line 8: does single_file_support not work with this server?

justinmk:
This has been discussed many times. Search existing PRs/issues.

https://github.com/neovim/nvim-lspconfig/blob/8121483b8132b7053120fafd83728178fb3febf6/CONTRIBUTING.md?plain=1#L46

lua/lspconfig/configs/clojure_lsp.lua
guilhermefelipecgs: It actually works, I'll switch to this option.

PR #3551, opened 2025-01-09
feat: gh-actions-language-server

### LINKED ISSUE #2737: Github Actions Language Server
### Language server

[Github Actions Language Server](https://github.com/actions/languageservices/tree/main/languageserver)

### Requested feature

There seems to be an [language server for Github Actions](https://github.com/actions/languageservices/tree/main/languageserver). I haven't tried, and was wondering if someone has successfully used it.
Would be awesome to include it as a config here if possible

_No response_

disrupted:
---
name: Pull Request
about: Submit a pull request
title: 'add GitHub Actions language server'
---

Closes #2737

### DIFF
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -6,6 +6,13 @@ return {
     filetypes = { 'yaml.github' },
     root_dir = util.root_pattern('.github'),
     single_file_support = true,
+    capabilities = {
+      workspace = {
+        didChangeWorkspaceFolders = {
+          dynamicRegistration = true,
+        },
+      },
+    },
   },
   docs = {
     description = [[
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -20,6 +20,16 @@ https://github.com/lttb/gh-actions-language-server
 
 Language server for GitHub Actions.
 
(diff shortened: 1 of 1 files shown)


comatory:
@disrupted Like I wrote [here](https://github.com/lttb/gh-actions-language-server/issues/1#issue-2777049498), for me it did not work without `bun` runtime so mayb e we could add it to the instructions as well.

comatory:
Okay—so the server NPM package removed the requirement on `bun` (it was a mistake). Please ignore my previous comment.

chrisgrieser:
I just gave this one a try, and noticed that the config does not work initially, due to the usage of `yaml.github` as a filetype. While I get the idea to not start this language server on every yaml file, I do not think that the filetype `yaml.github` is automatically in nvim by default? I have to manually set them in my github action files, at least.

disrupted:
> I just gave this one a try, and noticed that the config does not work initially, due to the usage of `yaml.github` as a filetype. While I get the idea to not start this language server on every yaml file, I do not think that the filetype `yaml.github` is automatically in nvim by default? I have to manually set them in my github action files, at least.

that's true. I recommend defining this pattern for GitHub workflows:

```lua
vim.filetype.add({
  pattern = {
    ['.*/%.github[%w/]+workflows[%w/]+.*%.ya?ml'] = 'yaml.github',
  },
})
```

We would either have to upstream this into Neovim core ftdetect or add a section to the docs

justinmk:
> We would either have to upstream this into Neovim core ftdetect or add a section to the docs

The `docs` item in these configs is intended for that, it's definitely worth mentioning there.

PR #3553, opened 2025-01-11
Feat: add lua 3rd party language servers

antonk52:
Per title. These servers make it possible to not need plugins like null-ls or none-ls if those are only used for lua tools.

### DIFF
--- a/lua/lspconfig/configs/stylua3p_ls.lua
+++ b/lua/lspconfig/configs/stylua3p_ls.lua
@@ -1,0 +1,16 @@
+local util = require('lspconfig.util')
+
+return {
+  default_config = {
+    cmd = { 'stylua-3p-language-server' },
+    filetypes = { 'lua' },
+    root_dir = util.root_pattern('.stylua.toml', 'stylua.toml'),
+  },
+  docs = {
+    description = [[
+https://github.com/antonk52/lua-3p-language-servers
+
+3rd party Language Server for Stylua lua formatter
+]],
+  },
+}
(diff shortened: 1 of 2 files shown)


lua/lspconfig/configs/stylua3p_ls.lua
```diff
@@ -0,0 +1,16 @@
+local util = require('lspconfig.util')
+
+return {
+  default_config = {
+    cmd = { 'stylua-3p-language-server' },
+    filetypes = { 'lua' },
+    root_dir = util.root_pattern('.stylua.toml'),
```
chrisgrieser, line 7: can also be a `stylua.toml`: file https://github.com/JohnnyMorganz/StyLua#finding-the-configuration

lua/lspconfig/configs/stylua3p_ls.lua
```diff
@@ -0,0 +1,16 @@
+local util = require('lspconfig.util')
+
+return {
+  default_config = {
+    cmd = { 'stylua-3p-language-server' },
+    filetypes = { 'lua' },
```
chrisgrieser, line 6: stylua also supports `luau`: https://luau.org/

antonk52: the LS does not currently support it

lua/lspconfig/configs/stylua3p_ls.lua
antonk52: Thanks, fixed

glepnir (approved):

chrisgrieser (approved):

PR #3554, opened 2025-01-13
fix(pyright): use activated virtual environment if possible

laddertosky:
Problem:
Pyright will use system python by default, even if we already have a virtual environment activated.

Solution:
Check current if `VIRTUAL_ENV` exists in environment variable, if so, set the default pythonPath with the binary inside.

### DIFF
--- a/lua/lspconfig/configs/pyright.lua
+++ b/lua/lspconfig/configs/pyright.lua
@@ -40,6 +40,21 @@ local function set_python_path(path)
   end
 end
 
+local python_bin = function(venv)
+  if vim.fn.has('win32') == 1 then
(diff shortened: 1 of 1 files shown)


laddertosky:
update the pr title to match commit lint

lithammer:
You know you can reword a commit, right? You don't have to close the pull request and open a new one for every lint failure.

laddertosky:
Does this still work for remote commit? I thought git commit --amend only works for local commit.

justinmk:
After --amend you need to force-push to your PR branch. Ask chatgpt or similar for more info.

PR #3557, opened 2025-01-13
fix(pyright): use activated virtual environment if possible

laddertosky:
Problem:
Pyright will use system python by default, even if we already have a virtual environment activated.

Solution:
Check current if VIRTUAL_ENV exists in environment variable, if so, set the default pythonPath with the binary inside.

### DIFF
--- a/lua/lspconfig/configs/pyright.lua
+++ b/lua/lspconfig/configs/pyright.lua
@@ -40,6 +40,21 @@ local function set_python_path(path)
   end
 end
 
+local python_bin = function(venv)
+  if vim.fn.has('win32') == 1 then
+    return venv .. '\\Scripts\\python.exe'
+  end
+  return venv .. '/bin/python'
+end
+
+local try_local_python_path = function()
+  local venv_path = os.getenv('VIRTUAL_ENV')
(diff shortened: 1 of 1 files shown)


lithammer:
I can't reproduce this. Works fine for me? What version of pyright?

lua/lspconfig/configs/pyright.lua
```diff
@@ -40,6 +40,21 @@ local function set_python_path(path)
   end
 end
 
+local python_bin = function(venv)
+  if vim.fn.has('win32') == 1 then
+    return venv .. '\\Scripts\\python.exe'
```
justinmk, line 45:
```suggestion
    return venv .. '/Scripts/python.exe'
```

laddertosky:
I figure out my mistake, sorry for taking your time to check. The problem is I have two different virtual environments in two projects and I forgot to switch when I navigate to the second one.

justinmk:
> I can't reproduce this. Works fine for me? What version of pyright?

Yeah, if the venv is "activated" then $PATH would have been modified, so searching for `python` should already work.

PR #3558, opened 2025-01-13
feat(gh_actions_ls): use `root_dir` to decide whether to attach

chrisgrieser:
cc @disrupted 

This saves the user the extra setup work of copypasting the `vim.filetype.add` snippet, making this language server work out of the box.

Follow-up to #3551, on this suggestion: https://github.com/neovim/nvim-lspconfig/pull/3551/files#r1912833011

### DIFF
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -5,10 +5,7 @@ return {
     cmd = { 'gh-actions-language-server', '--stdio' },
     filetypes = { 'yaml' },
     root_dir = function(filename)
-      if not filename:find('/%.github/workflows/.+%.ya?ml') then
-        return
-      end
-      return util.root_pattern('.github')
+      return filename:find('/%.github/workflows/.+%.ya?ml') and util.root_pattern('.github') or nil
     end,
     single_file_support = false,
     capabilities = {
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -3,9 +3,14 @@ local util = require 'lspconfig.util'
 return {
   default_config = {
     cmd = { 'gh-actions-language-server', '--stdio' },
-    filetypes = { 'yaml.github' },
-    root_dir = util.root_pattern('.github'),
-    single_file_support = true,
+    filetypes = { 'yaml' },
+    root_dir = function(filename)
+      if not filename:find('/%.github/workflows/.+%.ya?ml') then
+        return
+      end
+      return util.root_pattern('.github')
+    end,
+    single_file_support = false,
     capabilities = {
       workspace = {
         didChangeWorkspaceFolders = {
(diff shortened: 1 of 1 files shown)


disrupted:
I agree that this is probably a sensible default and less manual maintenance for users. Although personally, I am not a fan of the runtime check and will likely continue using the special filetype as it keeps things tidier.
We should also decide on a common strategy. [gitlab_ci_ls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#gitlab_ci_ls) is a very similar candidate

lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -3,9 +3,14 @@ local util = require 'lspconfig.util'
...
     cmd = { 'gh-actions-language-server', '--stdio' },
-    filetypes = { 'yaml.github' },
-    root_dir = util.root_pattern('.github'),
-    single_file_support = true,
+    filetypes = { 'yaml' },
+    root_dir = function(filename)
+      if not filename:find('/%.github/workflows/.+%.ya?ml') then
+        return
+      end
+      return util.root_pattern('.github')
```
glepnir, line 11:
```suggestion
    return filename:find('/%.github/workflows/.+%.ya?ml') and util.root_pattern('.github') or nil
```

justinmk:
> We should also decide on a common strategy. [gitlab_ci_ls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#gitlab_ci_ls) is a very similar candidate

Agreed. I have never seen a `.gitlab` file on any project I'm familiar with, so the runtime check is probably the right approach.

justinmk:
> That was meant for the `.github` folder that contains workflows etc. I believe the root pattern function works on folders as well as files, does it not?

Oh sorry I saw `yaml.github` in https://github.com/neovim/nvim-lspconfig/pull/3551#issuecomment-2585464445 and didn't look closely enough. I see now that it's only defining a filetype with that name.

In any case, avoiding the need for users to manually call `vim.filetype.add()` in their configs, is worthwhile. However it may require the `single_file_support=false` trick as mentioned in this PR.

lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -3,9 +3,11 @@ local util = require 'lspconfig.util'
...
   default_config = {
     cmd = { 'gh-actions-language-server', '--stdio' },
-    filetypes = { 'yaml.github' },
-    root_dir = util.root_pattern('.github'),
-    single_file_support = true,
+    filetypes = { 'yaml' },
+    root_dir = function(filename)
+      return filename:find('/%.github/workflows/.+%.ya?ml') and util.root_pattern('.github') or nil
+    end,
+    single_file_support = false,
```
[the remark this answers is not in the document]
chrisgrieser: sure

lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -3,9 +3,16 @@ local util = require 'lspconfig.util'
...
-    single_file_support = true,
+    filetypes = { 'yaml' },
+
+    -- Only attach to yaml files that are GitHub workflows instead of all yaml
+    -- files. (A nil root_dir and no single_file_support results in the LSP not
+    -- attaching.) For details, see #3558
+    root_dir = function(filename)
+      return filename:find('/%.github/workflows/.+%.ya?ml') and util.root_pattern('.github') or nil
+    end,
+    single_file_support = false,
```
justinmk, line 14:
```suggestion
    -- Disabling "single file support" is a hack to avoid enabling this LS for
    -- every random yaml file. The `root_dir` definition specifies the kind of 
    -- yaml file that is required for this LS.
    single_file_support = false,
```

PR #3565, opened 2025-01-16
feat(asm_lsp): add NASM config support and single_file support

iguanacucumber:
Hi, This simple PR add supports NASM config files, it also enable single_file support

### DIFF
--- a/lua/lspconfig/configs/asm_lsp.lua
+++ b/lua/lspconfig/configs/asm_lsp.lua
@@ -1,13 +1,10 @@
-local util = require 'lspconfig.util'
-
 return {
   default_config = {
     cmd = { 'asm-lsp' },
     filetypes = { 'asm', 'vmasm' },
     single_file_support = true,
     root_dir = function(fname)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/asm_lsp.lua
```diff
@@ -1,16 +1,20 @@
...
+
 return {
   default_config = {
     cmd = { 'asm-lsp' },
     filetypes = { 'asm', 'vmasm' },
+    single_file_support = true,
     root_dir = function(fname)
-      return vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
+      return util.root_pattern('.asm-lsp.toml')(fname)
+        or vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
```
glepnir, line 10:
```suggestion
      return vim.fs.dirname(vim.fs.find({ '.asm-lsp.toml', '.git' }, { path = fname, upward = true })[1])
```

PR #3566, opened 2025-01-17
refactor(gh_actions_ls): allow .forgejo and .gitea as root_dir

Crashdummyy:
The projects [forgejo](https://forgejo.org/) and [gitea](https://about.gitea.com/) share the same syntax even across the composite actions.
There are only negligible differences IMHO which besides the actions author should be aware of.
Besides its only a matter of time until those become implemented as well.

It'd be a nice addition to let it attach for those systems as well


justinmk:
seems fine but should be called out in the `docs`

justinmk (approved):

PR #3567, opened 2025-01-18
refactor: comment util.path.is_descendant to prepare deperecation in future

dundargoc:
We don't deprecate it currently as the suggested replacement
(vim.fs.relpath) isn't available on the minimum supported neovim
version.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -5,7 +5,7 @@ local nvim_eleven = vim.fn.has 'nvim-0.11' == 1
 
 local iswin = vim.loop.os_uname().version:match 'Windows'
 
-local M = {}
+local M = { path = {} }
 
 M.default_config = {
   log_level = lsp.protocol.MessageType.Warning,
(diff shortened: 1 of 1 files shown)


lua/lspconfig/util.lua
```diff
@@ -294,7 +243,52 @@ function M.strip_archive_subpath(path)
...
 
---- Functions that can be removed once minimum required neovim version is high enough
+--- Public functions that can be deprecated once minimum required neovim version is high enough
+
+local function is_fs_root(path)
+  if iswin then
+    return path:match '^%a:$'
+  else
+    return path == '/'
+  end
```
glepnir, line 253:
```suggestion
  return iswin and path:match '^%a:$' or path == '/'
```

lua/lspconfig/util.lua
```diff
@@ -294,7 +243,52 @@ function M.strip_archive_subpath(path)
...
+end
+
+-- Traverse the path calling cb along the way.
+local function traverse_parents(path, cb)
```
glepnir, line 257: how about this 🤔

```lua
local function traverse_parents(path, callback)
    local real_path = vim.loop.fs_realpath(path)
    if not real_path then
        return nil, nil
    end

    if type(callback) ~= "function" then
        error("callback must be a function")
    end

    local dir = real_path

    while dir do
        local parent = vim.fs.dirname(dir)
        if not parent or parent == dir then
            break
        end

        local should_stop = callback(parent, real_path)
        if should_stop then
            return parent, real_path
        end

        if is_fs_root(parent) then
            break
        end

        dir = parent
    end

    return nil, nil
end
```

dundargoc: Do we really need to refactor this function? We'll remove it once `is_descendant` is removed anyway, so think we can treat this function as frozen.

glepnir: No need as far as I remember this function is not used in lspconfig 😄

lua/lspconfig/util.lua
dundargoc: I think this is dangerous, if `iswin` is true and `path:match` returns `nil` then it will incorrectly set `path` to `/`. Example:

```lua
print(true and nil or 'oof')
```

prints `oof` even though the intention is to return `nil`

glepnir: Oh i forget there nil trap yes

glepnir (approved):

justinmk (approved):
Thanks @dundargoc for driving this. Now we have a very clear migration path for nvim-lspconfig.

PR #3569, opened 2025-01-19
refactor!: deprecate util.root_pattern

dundargoc:
The primary (and potentially breaking) change is that
`util.root_pattern` traverses once per provided file pattern, while
`vim.fs.find` only traverses once. This means that the search will no
longer be prioritized by file order, which may break user configurations
who unkowingly rely on this behavior.

This will also be breaking for language servers that rely on code
dependencies that is archived but still needs to be read. More context:
https://github.com/neovim/nvim-lspconfig/pull/1687.

### DIFF
--- a/doc/lspconfig.txt
+++ b/doc/lspconfig.txt
@@ -245,7 +245,7 @@ ADDING NEW SERVERS                                    *lspconfig-new*
 
 The steps for adding and enabling a new server configuration are:
 
-1. Define the configuration (see also |vim.fs.root()|): >lua
+1. Define the configuration: >lua
    local lspconfig = require 'lspconfig'
    local configs = require 'lspconfig.configs'
 
@@ -256,7 +256,7 @@ The steps for adding and enabling a new server configuration are:
          cmd = {'/home/neovim/lua-language-server/run.sh'},
          filetypes = {'lua'},
          root_dir = function(fname)
-           return lspconfig.util.find_git_ancestor(fname)
+           return vim.fs.root(fname, '.git')
          end,
          settings = {},
        },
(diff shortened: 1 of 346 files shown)


dundargoc:
I couldn't come up with a good way to replicate the functionality provided by `strip_archive_subpath` so think we might just want to cut our losses here. @mfussenegger had a big brain suggestion about a related lsp feature that did something similar (by being able to read code from any URI or something?), so might want to use that in the future in core instead.

doc/lspconfig.txt
```diff
@@ -354,8 +358,8 @@ for some project structures. Example (for Kotlin): >lua
     'build.gradle.kts', -- Gradle
   }
   root_dir = function(fname)
-    local primary = util.root_pattern(unpack(root_files))(fname)
-    local fallback = util.root_pattern(unpack(fallback_root_files))(fname)
+    local primary = vim.fs.dirname(vim.fs.find(root_files, { path = fname, upward = true })[1])
+    local fallback = vim.fs.dirname(vim.fs.find(fallback_root_files, { path = fname, upward = true })[1])
```
justinmk, line 362: is `vim.fs.root()` not relevant here? I know the configs are not using it for back-compat, but our docs could recommend stuff available in Nvim 0.10/0.11

justinmk (approved):
Before merging this, let's wait a ~day to see if master is stable enough to tag a release.

Also: hoping that once we bump the minimum Nvim to 0.10, we can migrate these cases to `vim.fs.root`. If `vim.fs.root` doesn't work then we need to find out what it's missing ASAP :)

dundargoc:
Looking more closely: I think `vim.fs.find` and `util.root_pattern` behaves differently in that `find` traverses once and `root_pattern` traverses once per provided filepattern, meaning that the ordering of the files used to matter. This would be a breaking change for users/configs unknowingly relying on this behavior, but it's impossible to say beforehand which servers it'd be without user input. I suggest we merge this anyway so we can at least evaluate the actual effect out there and react accordingly.

I'll add this info to the commit message as well.

justinmk:
> meaning that the ordering of the files used to matter.

Er, that's very much intentional: https://github.com/neovim/nvim-lspconfig/pull/2885

Changing it to "random" would be even worse than reverting than PR.

Related?:

- https://github.com/neovim/nvim-lspconfig/pull/3321
- https://github.com/neovim/nvim-lspconfig/pull/3515

dundargoc:
What about the following suggestion?

1. Add option in `vim.fs.find` for this functionality. As an example, python's `os.walk` has a `topdown` that adjust whether the traversal is breadth first or depth first. A similar option for `vim.fs.find` would suit it well I think.
2. Make it depth first by default for `vim.fs.root`.

This approach seems like the simplest way to move forward, but I understand if you're hesitant to add more options to vim.fs.find.

justinmk:
> What about the following?

Those SGTM. 

> Edit: in the meantime, this problem can be worked around by doing multiple calls to `vim.fs.find`. Aka `find(priority1) or find(priority2) or ...`. It's clunky but not super awful.

Maybe we should keep `root_pattern` around for now. There's no real reason to avoid it until we have an answer in core.

dundargoc:
Oof. I understand.

Closing for now until there's a solution in core as described in https://github.com/neovim/nvim-lspconfig/pull/3569#issuecomment-2605033349.

PR #3570, opened 2025-01-20
Adding structurizr LS

Tacsiazuma:


justinmk:
No comments from others indicating desire for this, and the LSP repo has no stars, so this needs to mature a bit before we include it.

PR #3573, opened 2025-01-21
fix(gh_actions_ls): Fix incorrect rootdir resolution

Crashdummyy:
Regression of #3566

### DIFF
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -7,18 +7,17 @@ return {
     -- files. (A nil root_dir and no single_file_support results in the LSP not
     -- attaching.) For details, see #3558
     root_dir = function(filename)
-      local patterns = {
-        { pattern = '/%.github/workflows/.+%.ya?ml', root = '.github' },
-        { pattern = '/%.forgejo/workflows/.+%.ya?ml', root = '.forgejo' },
-        { pattern = '/%.gitea/workflows/.+%.ya?ml', root = '.gitea' },
+      local dirs_to_check = {
+        '.github/workflows',
+        '.forgejo/workflows',
+        '.gitea/workflows',
       }
 
-      for _, entry in ipairs(patterns) do
-        if filename:find(entry.pattern) then
-          local match = vim.fs.find(entry.root, { path = filename, upward = true })[1]
-          if match then
-            return match
-          end
+      local dir = vim.fs.dirname(filename)
+      for _, subdir in ipairs(dirs_to_check) do
+        local match = vim.fs.find(subdir, { path = dir, upward = true })[1]
+        if match and vim.fn.isdirectory(match) == 1 and vim.fs.dirname(filename) == match then
+          return match
         end
       end
 
--- a/lua/lspconfig/configs/gh_actions_ls.lua
+++ b/lua/lspconfig/configs/gh_actions_ls.lua
@@ -9,9 +9,19 @@ return {
     -- files. (A nil root_dir and no single_file_support results in the LSP not
     -- attaching.) For details, see #3558
     root_dir = function(filename)
-      return filename:find('/%.(github|forgejo|gitea)/workflows/.+%.ya?ml')
-          and util.root_pattern('.github', '.forgejo', '.gitea')(filename)
-        or nil
+      local patterns = {
+        { pattern = '/%.github/workflows/.+%.ya?ml', root = '.github' },
+        { pattern = '/%.forgejo/workflows/.+%.ya?ml', root = '.forgejo' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -9,9 +9,19 @@ return {
...
-        or nil
+      local patterns = {
+        { pattern = '/%.github/workflows/.+%.ya?ml', root = '.github' },
+        { pattern = '/%.forgejo/workflows/.+%.ya?ml', root = '.forgejo' },
+        { pattern = '/%.gitea/workflows/.+%.ya?ml', root = '.gitea' },
+      }
+
+      for _, entry in ipairs(patterns) do
+        if filename:find(entry.pattern) then
+          return util.root_pattern(entry.root)(filename)
```
glepnir, line 20: using vim.fs.find on there please see `:help vim.fs.find`

dundargoc: This is how I changed in my PR: https://github.com/dundargoc/nvim-lspconfig/blob/refactor/root_pattern/lua/lspconfig/configs/gh_actions_ls.lua#L1-L13. Just copy that and adjust as necessary.

dundargoc: Wait, what is actually happening here though? What's with all the patterns and stuff? Does the following not solve the problem?

```lua
root_dir = function(fname)
  return vim.fs.dirname(vim.fs.find({ '.github', '.forgejo', '.gitea' }, { path = fname, upward = true })[1])
end,
```

dundargoc: By the way, there were complications with the root_pattern PR so you can keep using it for now if you still want to. I'll take a look at codespell warning.

dundargoc: Will merge when green: https://github.com/neovim/nvim-lspconfig/pull/3577

dundargoc: > What do you mean by "complications" ?

Unrelated to your PR: https://github.com/neovim/nvim-lspconfig/pull/3569#issuecomment-2605033349

> Do I need to change anything ?

No, you don't need to change anything. This looks fine.

lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -9,9 +7,22 @@ return {
...
-        or nil
+      local patterns = {
+        { pattern = '/%.github/workflows/.+%.ya?ml', root = '.github' },
+        { pattern = '/%.forgejo/workflows/.+%.ya?ml', root = '.forgejo' },
+        { pattern = '/%.gitea/workflows/.+%.ya?ml', root = '.gitea' },
+      }
+
+      for _, entry in ipairs(patterns) do
+        if filename:find(entry.pattern) then
+          local match = vim.fs.find(entry.root, { path = filename, upward = true })[1]
```
glepnir, line 18: use `vim.fs.dirname`

dundargoc:
What exactly is the regression? When I test out gh_actions on master it seems to work for me.

dundargoc:
What repo/file are you testing? I'd like to test it.

dundargoc:
Does the following work? I tried it and autocompletion worked for me

```lua
 root_dir = require('lspconfig.util').root_pattern('.github/workflows', '.forgejo/workflows', '.gitea/workflows'),
```

dundargoc:
> First of all I thought I should not use root_pattern anymore.

Yes, but then I said in https://github.com/neovim/nvim-lspconfig/pull/3573#discussion_r1924349023 that because of reasons we can still use root_pattern. The current approach of vim.fs.dirname(... is also fine.

dundargoc (approved):
OK, this works for me. There might be an easier way to achieve this but fuck it, let's go with it for now.

lua/lspconfig/configs/gh_actions_ls.lua
```diff
@@ -9,9 +7,21 @@ return {
...
+        '.github/workflows',
+        '.forgejo/workflows',
+        '.gitea/workflows',
+      }
+
+      local dir = vim.fs.dirname(filename)
+      for _, subdir in ipairs(dirs_to_check) do
+        local match = vim.fs.find(subdir, { path = dir, upward = true })[1]
+        if match and vim.fn.isdirectory(match) == 1 and vim.fs.dirname(filename) == match then
+          return match
```
justinmk, line 20: What is the intention of this logic? Is it something that needs to be in a `util` function, and potentially find its way into `vim.fs` ?

The commit messages don't give any insight.

justinmk: `vim.fs.relpath()` in Nvim 0.11 may address that. Maybe a code comment here would be helpful

PR #3581, opened 2025-01-22
fix: fix incorrect root finding

dundargoc:


justinmk (approved):

justinmk:
After this we should tag a release

lua/lspconfig/configs/volar.lua
```diff
@@ -1,7 +1,7 @@
 local util = require 'lspconfig.util'
 
 local function get_typescript_server_path(root_dir)
-  local project_root = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
+  local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
   return project_root and (project_root .. '/typescript/lib') or ''
```
lolzballs, line 5: I think this should be:

```lua
return project_root and (project_root .. '/node_modules/typescript/lib') or ''
```

This applies to the change in `astro.lua` as well.

rafaelpirolla:
It doesn't seem to find the typescript libs in the mason dir anymore after this.

Issue #3583, opened 2025-01-23
vim.validate is deprecated. Feature will be removed in Nvim 1.0


Mte90:
### Description

```
- use vim.validate(name, value, validator, optional_or_msg) instead.
- stack traceback:
        /home/mte90/.vim/plugged/nvim-lspconfig/lua/lspconfig/configs.lua:37
        /home/mte90/.vim/plugged/nvim-lspconfig/lua/lspconfig.lua:89
        /home/mte90/.config/nvim/lua/plugin/lsp.lua:57
        [C]:-1
        /home/mte90/.config/nvim/lua/plugins.lua:136
        [C]:-1
        lua:4
    - stack traceback:
        /home/mte90/.vim/plugged/nvim-lspconfig/lua/lspconfig/configs.lua:77
        /home/mte90/.config/nvim/lua/plugin/lsp.lua:57
        [C]:-1
        /home/mte90/.config/nvim/lua/plugins.lua:136
        [C]:-1
        lua:4
    - stack traceback:
        /home/mte90/.vim/plugged/nvim-lspconfig/lua/lspconfig/configs.lua:37
        /home/mte90/.vim/plugged/nvim-lspconfig/lua/lspconfig.lua:89
        /home/mte90/.config/nvim/lua/plugin/lsp.lua:80
        [C]:-1
        /home/mte90/.config/nvim/lua/plugins.lua:136
        [C]:-1
        lua:4
```

dundargoc:
Right, this is the output from `:checkhealth vim.deprecated`. We decided to not remove the deprecated `vim.validate` calls in https://github.com/neovim/nvim-lspconfig/pull/3517#discussion_r1890341110 since that file was going to be removed in the future anyway, but we hadn't taken the vim.deprecated output into consideration. As I see it our options here are

1. Tell users to ignore `vim.deprecated` output. This would mark this as wontfix
2. Do a version check and call correct version of `vim.validate`
3. Just remove the validation

I think we should go with option 3 and just remove the validation from `configs.lua`. That file is effectively frozen anyway IIUC, so I don't think we need to worry too much about incorrect input.

Mte90:
I just spotted because the majority of my checkhealth is generated by this deprecation as I have various LSP plugins.

PR #3584, opened 2025-01-23
fix: remove validation from configs.lua

dundargoc:
This is because using the old syntax for vim.validate causes
`:checkhealth vim.deprecated` to be flooded with deprecated messages.

It would also be possible to do a version check and use the newer syntax
for vim.validate, but since configs.lua will be replaced by
vim.lsp.config in the future there is little need to future-proof it.

Closes https://github.com/neovim/nvim-lspconfig/issues/3583.

### DIFF
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -5,7 +5,7 @@
 
 local util = require 'lspconfig.util'
 local async = require 'lspconfig.async'
-local api, validate, lsp, fn = vim.api, vim.validate, vim.lsp, vim.fn
+local api, lsp, fn = vim.api, vim.lsp, vim.fn
 local tbl_deep_extend = vim.tbl_deep_extend
 
 local configs = {}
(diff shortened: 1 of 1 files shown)


dundargoc:
I suppose we could keep the old syntax for the validation and do a reverse version check, aka only validate if neovim is less than v0.11 but don't know how much we'd gain from that.

lua/lspconfig/configs.lua
```diff
@@ -57,14 +49,7 @@ function configs.__newindex(t, config_name, config_def)
...
-    for k, v in pairs(config_def.commands) do
-      validate {
-        ['command.name'] = { k, 's' },
-        ['command.fn'] = { v[1], 'f' },
-      }
-    end
-  else
+  if not config_def.commands then
     config_def.commands = {}
   end
```
glepnir, line 54:
```suggestion
  config_def.commands = config_def.commands or {}
```

glepnir (approved):

Issue #3588, opened 2025-01-26
Removal of `util.available_servers` from public API in #3571 - recommended alternative?


spanishpear:
### Description

👋 Heya!

It seems that #3571 made `util.available_servers` a private function (as part of #2079), which is commonly used in user configurations to add capabilities to all servers.

For example - my configuration contains:
```lua
local language_servers = require("lspconfig").util.available_servers()
for _, ls in ipairs(language_servers) do
    require('lspconfig')[ls].setup({
        capabilities = capabilities, -- custom capabilities, excluded from snippet
        on_attach = on_attach, -- custom on_attach, excluded from snippet
        settings = servers[ls] -- map of settings per server , excluded from snippet
    })
end
```

Additionally - it's used/recommended in common community plugins - such as:
 - nvim-ufo minimal configuration [here](https://github.com/kevinhwang91/nvim-ufo?tab=readme-ov-file#minimal-configuration).
 - other examples from github search [here](https://github.com/search?q=%22util.available_servers%22&type=code)

For now, I've migrated my personal config to simply use the private method, but is there a preferred alternative to achieve the same goal?
I may simply be missing an obvious API that allows me to do this - - so apologies if so :) 

I've raised this as an issue, as the question template points to stack-overflow, which feels incorrect for a question for maintainers. Apologies if this is not the correct forum, I could not see a discussions tab.

Thank you!

dundargoc:
Hmm, if this is useful enough it might warrant a core functionality. Or perhaps this is already possible to do with in core. @justinmk @gpanders any input?

justinmk:
Thanks for raising this, it helps to know what is actually being used.

(Side-note: `available_servers` is poorly named, it actually returns *configs*, which configure aspects of the client *and* the server.)

Starting with Nvim 0.11, the way to configure "all configs" is to use the special `"*"` name:

```lua
vim.lsp.config('*', {
  capabilities = { … },
  settings = { … },
  on_attach = { … },
})
```

# Proposal

- We can restore the old name of `util.available_servers()` avoid disturbing people. Although by now most people may have already migrated, so renaming it again might be worse...
- We should avoid breaking changes to `lspconfig.util`; instead, we want to ensure that it is 100% NOT needed for Nvim 0.11.
    - We are in the process of soft-deprecating the entire `lspconfig.util` module.
[...]

PR #3589, opened 2025-01-27
revert: "refactor!: make available_servers function private"

dundargoc:
This reverts commit e118ce58dab72c17216292eef7df4cee3cf60885.

It turns out `util.available_servers` is used more than anticipated, so
we revert the privatization for the time being.

Closes https://github.com/neovim/nvim-lspconfig/issues/3588.

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -223,7 +223,8 @@ function M.get_managed_clients()
   return clients
 end
 
-function M._available_servers()
+--- @deprecated use `vim.lsp.config` in Nvim 0.11+ instead.
+function M.available_servers()
   local servers = {}
   local configs = require 'lspconfig.configs'
   for server, config in pairs(configs) do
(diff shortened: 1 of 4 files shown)


lua/lspconfig/util.lua
```diff
@@ -223,7 +223,7 @@ function M.get_managed_clients()
   return clients
 end
 
-function M._available_servers()
+function M.available_servers()
```
justinmk, line 226:
```suggestion
--- @deprecated use `vim.lsp.confg` in Nvim 0.11+ instead.
function M.available_servers()
```

future: we should probably add `use vim.lsp.confg in Nvim 0.11+` to all util.lua deprecation notices

dundargoc: Wait, people using nvim 0.10 won't have a way to remove this warning. I guess it's an acceptable tradeoff?

justinmk: Our nvim-lspconfig deprecation plan didn't really consider concerns about non-breaking soft *warnings* for Nvim 0.10, AFAIR. It doesn't strike me as too important, as long as their code doesn't break.

justinmk (approved):

PR #3592, opened 2025-01-28
fix(angularls): set Angular core version for backwards compatibility

### LINKED ISSUE #3590: Angular standalone components backwards compatibility
## Current behavior

Angular language server complains about standalone components in a project that does not use it.

`Component AppComponent is standalone, and cannot be declared in an NgModule. Did you mean to import it instead? [-996008]`

## How to reproduce

Install node/npm

```bash
npx @angular/cli@15 new angular-app
cd angular-app
nvim --clean -u minimal.lua angular-app/src/app/app.module.ts
```

## Minimal init.lua

```lua
-- Run this file as `nvim --clean -u minimal.lua`

for name, url in pairs {
  'https://github.com/neovim/nvim-lspconfig',
} do
  local install_path = vim.fn.fnamemodify('nvim_issue/' .. name, ':p')
    vim.fn.system { 'git', 'clone', '--depth=1', url, install_path }
  end
  vim.opt.runtimepath:append(install_path)
end

-- ADD INIT.LUA SETTINGS _NECESSARY_ FOR REPRODUCING THE ISSUE  
require'lspconfig'.angularls.setup{}

```

## Proposed fix

Pass `--angularCoreVersion` to the default cmd:

```lua
local function get_probe_dir(root_dir)
  local project_root = require("lspconfig.util").find_node_modules_ancestor(root_dir)

  return project_root and (project_root .. "/node_modules") or ""
end

local function get_angular_core_version(root_dir)
  local project_root = require("lspconfig.util").find_node_modules_ancestor(root_dir)

  local package_json = project_root .. "/package.json"

  if not vim.loop.fs_stat(package_json) then
    return nil
  end

  local cmd = "jq -r '.dependencies[\"@angular/core\"]' " .. package_json
  local handle = io.popen(cmd)
  local angular_core_version = handle:read("*a")
  handle:close()

  return angular_core_version
end

return {
    local probe_dir = get_probe_dir(new_root_dir)
    local angular_core_version = get_angular_core_version(new_root_dir)

    new_config.cmd = {
      "ngserver", "--stdio", "--tsProbeLocations", probe_dir, "--ngProbeLocations", probe_dir, "--angularCoreVersion",
      angular_core_version,
    }
  end,
}
```

## Notes

Initializing a new Angular project with cli version 19 seems to work fine, but more tests would be appreciated.

mactep:
Closes #3590.

@justinmk what would be the preferred way to read the json? I followed some examples on the codebase.

### DIFF
--- a/lua/lspconfig/configs/angularls.lua
+++ b/lua/lspconfig/configs/angularls.lua
@@ -9,7 +9,27 @@ local function get_probe_dir(root_dir)
   return project_root and (project_root .. '/node_modules') or ''
 end
 
+local function get_angular_core_version(root_dir)
+  local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
+
+  local package_json = project_root .. '/package.json'
+  if not vim.loop.fs_stat(package_json) then
+    return ''
+  end
+
+  local contents = io.open(package_json):read '*a'
+  local json = vim.json.decode(contents)
+  if not json.dependencies then
+    return ''
+  end
+
+  local angular_core_version = json.dependencies['@angular/core']
+
+  return angular_core_version
+end
+
 local default_probe_dir = get_probe_dir(vim.fn.getcwd())
+local default_angular_core_version = get_angular_core_version(vim.fn.getcwd())
 
 return {
   default_config = {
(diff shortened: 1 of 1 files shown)


justinmk:
Strange that the lsp server doesn't do this kind of thing on its own, based on the root dir or workspace folder.

PR #3594, opened 2025-01-29
fix(angularls): handle nil while checking angular core version

Ask-786:
Problem:
Causes warning message while opening a directory without node modules


mactep:
Nice catch. Thanks for the quick fix.

@justinmk can you check this one as well?

PR #3595, opened 2025-01-29
feat: dts-lsp

golovinengineer:
The dts-lsp server does not meet the requirement of having 100 stars on its GitHub repository or VSCode extension. Additionally, I have no affiliation with it. I originally created this configuration for personal use and am sharing it in case others find it useful.

### DIFF
--- a/lua/lspconfig/configs/dts_lsp.lua
+++ b/lua/lspconfig/configs/dts_lsp.lua
@@ -10,13 +10,18 @@ return {
   },
   docs = {
     description = [[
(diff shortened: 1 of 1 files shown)


justinmk:
The 100 stars requirement is a rough guideline to avoid spam. 

I had to click through 4 different links to find out what devicetree is, maybe the docs here can link to the zephyer project or something like that?

golovinengineer:
Yes, sure. I should have thought about it.

PR #3603, opened 2025-02-06
don't skip VIMRUNTIME in lua lsp if ~/.config/nvim/.luarc.json exists

jyn514:
.luarc is created by lua-lsp itself (e.g. if you hit "disable diagnostics" in a code action), at which point the lsp will start showing errors that the `vim` global is undefined after the next restart.

### DIFF
--- a/lua/lspconfig/configs/lua_ls.lua
+++ b/lua/lspconfig/configs/lua_ls.lua
@@ -48,7 +48,7 @@ require'lspconfig'.lua_ls.setup {
   on_init = function(client)
     if client.workspace_folders then
       local path = client.workspace_folders[1].name
-      if vim.loop.fs_stat(path..'/.luarc.json') or vim.loop.fs_stat(path..'/.luarc.jsonc') then
+      if path ~= vim.fn.stdpath('config') and (vim.loop.fs_stat(path..'/.luarc.json') or vim.loop.fs_stat(path..'/.luarc.jsonc')) then
         return
       end
     end


lua/lspconfig/configs/lua_ls.lua
```diff
@@ -48,7 +48,7 @@ require'lspconfig'.lua_ls.setup {
   on_init = function(client)
     if client.workspace_folders then
       local path = client.workspace_folders[1].name
-      if vim.loop.fs_stat(path..'/.luarc.json') or vim.loop.fs_stat(path..'/.luarc.jsonc') then
+      if path ~= vim.fn.stdpath('config') and (vim.loop.fs_stat(path..'/.luarc.json') or vim.loop.fs_stat(path..'/.luarc.jsonc')) then
```
justinmk, line 51: the PR title says `if ~/.config/nvim/.luarc.json exists` but here you are comparing `path ~= vim.fn.stdpath('config')` ("NOT equal to ~/.config/nvim") ... ?

jyn514: the PR says "DON'T" skip the runtime if this directory exists. if this condition is taken, then the code does an early return, and the runtime is never loaded. so we want the condition to be false when the directory exists, hence ~= instead of ==.

gwuen:
Going through my LSP config, I just stumbled upon this change in the documentation, and I'm wondering: Doesn't the same argument hold true for any other directory holding Neovim configuration? Including subdirectories like `lua/my_config/` or really any plugin, which might very well include its own `.luarc.json`.

To also include subdirectories, one may check for `not path:find(vim.fn.stdpath("config"), 1, true)`. But frankly, I think there are too many scenarios to cover them all with one snippet. Previously, the logic was straightforward: If the user changed the workspace-specific server config, they were expected to include all the non-standard libraries themselves. The new snippet might lead to confusion about why this is different for one specific directory.

For a more predictable behavior, I think it would be better to either revert the change or come up with a conditional that covers not just one of many cases. I'd like to hear your thoughts, maybe there's also a good reason for keeping it the way it is that I just overlooked.

PR #3586, opened 2025-01-25
feat: oxc-language-server

marekvospel:
This pull request adds the minimal working config for [`oxlint`](https://oxc.rs) [github](https://github.com/oxc-project/oxc).
The configuration is highly inspired by the `biome` config. The `oxc_language_server` binary is part of the `oxlint` package (It has 2 binaries, `oxlint` and `oxlint_language_server`).

Sadly I was unable to generate the config.md and config.txt docs files. After executing the command in CONTRIBUTING.md, I got no error and the file was unchanged...

Related to #3062

PR #3604, opened 2025-02-06
feat(oxc_language_server): add OxcFixAll command

marekvospel:
This pull request adds the `OxcFixAll` command for basic `--fix` functionality. (as mentioned in https://github.com/neovim/nvim-lspconfig/pull/3586#issuecomment-2613930725)

Note this functionality was added to the language server very recently (https://github.com/oxc-project/oxc/pull/8858) and only works with `oxc_language_server` >=`0.15.10`. (Still not in mason-registry, see https://github.com/mason-org/mason-registry/pull/8717)

### DIFF
--- a/lua/lspconfig/configs/oxlint.lua
+++ b/lua/lspconfig/configs/oxlint.lua
@@ -19,7 +19,7 @@ return {
     commands = {
       OxcFixAll = {
         function()
-          local client = util.get_active_client_by_name(0, 'oxlint')
+          local client = vim.lsp.get_clients({ bufnr = 0, name = 'oxlint' })[0]
           if client == nil then
             return
           end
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/oxlint.lua
```diff
@@ -15,6 +15,27 @@ return {
     },
     root_dir = util.root_pattern('.oxlintrc.json'),
     single_file_support = false,
+
+    commands = {
+      OxcFixAll = {
+        function()
+          local client = util.get_active_client_by_name(0, 'oxlint')
```
justinmk, line 22: use `vim.lsp.get_client({name=...})`, there is no need for `get_active_client_by_name`

@dundargoc this would be an easy one to deprecate

marekvospel: I couldn't find documentation of `vim.lsp.get_client` perhaps you meant `vim.lsp.get_clients`?

PR #3606, opened 2025-02-06
docs(atlas): add treesitter instructions

net:


lua/lspconfig/configs/atlas.lua
```diff
@@ -49,6 +49,20 @@ vim.filetype.add({
 })
 ```
 
+Optionally, tell treesitter to treat Atlas filetypes as HCL for better syntax highlighting:
+
+```lua
+vim.treesitter.language.register('hcl', 'atlas-config')
```
justinmk, line 55: This is too much. What does this have to do with a LSP config? There is already a `vim.filetype.add` note above, which should already deal with this. And that note itself is arguably off topic.

PR #3610, opened 2025-02-09
fix(angularls): trim leading characters from version

### LINKED ISSUE #3609: Angularls regression from recent changes
### Description
I recently updated nvim-lspconfig and noticed issues with my `angularls` lsp on new projects. There are errors being reported that shouldnt be with new versions of angular. It looks as though some backwards compatibility options recently added may be effecting newer projects.

## Bug
Angular 19+ components are default standalone, but with the new nvim-lspconfig updates these error show for new projects when they shouldnt.

Image

## Reproduce
1. Install latest `nvim-lspconfig`
2. Install `angularls`(i use mason-lspconfig)
3. Start new angular 19+ project, `ng new`
4. Install `npm install --save-dev @angular/language-service` (needed for lsp to work in project)
5. See errors like above screen shot in components with imports that are defaulte standalone

The incorrect errors from the lsp stop if I pin `nvim-lspconfig` to commit ` commit = "1f941b3"` in my lazy config. (This is the commit before the angular prs).

Im on linux/popos

## Related PRs/commits
https://github.com/neovim/nvim-lspconfig/pull/3592
https://github.com/neovim/nvim-lspconfig/commit/9962e25a23f0e992116b99a9a6400947d5a7177b <-- This seems to be the culprit as pinning any commits after result in the errors described

mactep:
Closes #3609

### DIFF
--- a/lua/lspconfig/configs/angularls.lua
+++ b/lua/lspconfig/configs/angularls.lua
@@ -29,6 +29,8 @@ local function get_angular_core_version(root_dir)
 
   local angular_core_version = json.dependencies['@angular/core']
 
+  angular_core_version = angular_core_version and angular_core_version:match('%d+%.%d+%.%d+')
+
   return angular_core_version
 end
 


whoop-t (approved):
lgtm

PR #3615, opened 2025-02-13
fix(eslint): update openDoc os call for windows

Kushal-Chandar:
Problem:
os.execute misinterprets quoted url as window title

Solution:
add an empty title argument

### DIFF
--- a/lua/lspconfig/configs/eslint.lua
+++ b/lua/lspconfig/configs/eslint.lua
@@ -131,16 +131,8 @@ return {
     end,
     handlers = {
       ['eslint/openDoc'] = function(_, result)
-        if not result then
-          return
-        end
-        local sysname = vim.loop.os_uname().sysname
-        if sysname:match 'Windows' then
-          os.execute(string.format('start %q', result.url))
-        elseif sysname:match 'Linux' then
-          os.execute(string.format('xdg-open %q', result.url))
-        else
-          os.execute(string.format('open %q', result.url))
+        if result then
+          vim.ui.open(result.url)
         end
         return {}
       end,


Kushal-Chandar:
opens_cmd
opens_browser
difference

Kushal-Chandar:
@pitkling can you take a look

lua/lspconfig/configs/eslint.lua
```diff
@@ -136,7 +136,7 @@ return {
         end
         local sysname = vim.loop.os_uname().sysname
         if sysname:match 'Windows' then
```
justinmk, line 138: all of this should be replaced with a single call to `vim.ui.open()`

Kushal-Chandar: I think that was added in 10.0. I can make the change. Can i make the change? The README says 9.0+.

justinmk: We'll update that soon. But anyway, if people want to use this command, they need Nvim 0.10

Kushal-Chandar: NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1713484068

  fall-back for $VIM: "C:/Program Files (x86)/nvim/share/nvim"


vim.ui.open when run from vim environment doesn't do anything.

and in my lsp, the second argument is simply not present.
image


i don't have lua 5.1 installed on my system

Kushal-Chandar: +add it works for files doesn't work for urls

justinmk: `vim.ui.open()` absolutely works for URLs. Try this:

    :lua vim.ui.open('https://google.com')

justinmk: why are you passing `cmd=osurl` ? don't do that.

Kushal-Chandar: Hi, turns out it was an issue with qutebrowser. works with edge on windows

PR #3620, opened 2025-02-18
Add config for turbo-language-server

trekdemo:
The language server supports writing [Turbo](https://turbo.hotwired.dev). 

There is a corresponding [VS Code plugin][1] with 700+ downloads.

[1]: https://marketplace.visualstudio.com/items?itemName=marcoroth.turbo-lsp&ssr=false#overview

### DIFF
--- a/lua/lspconfig/configs/turbo_ls.lua
+++ b/lua/lspconfig/configs/turbo_ls.lua
@@ -1,0 +1,24 @@
+return {
+  default_config = {
+    cmd = { 'turbo-language-server', '--stdio' },
+    filetypes = { 'html', 'ruby', 'eruby', 'blade', 'php' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/turbo_ls.lua
```diff
@@ -0,0 +1,26 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'turbo-language-server', '--stdio' },
+    filetypes = { 'html', 'ruby', 'eruby', 'blade', 'php' },
+    root_dir = util.root_pattern('Gemfile', '.git'),
```
glepnir, line 7: you can use `vim.fs.root` here

trekdemo: Thanks, @glepnir; I did not know about `vim.fs.root()`.

PR #3621, opened 2025-02-19
improve `util.root_pattern()`

tmillr:
- Accept root patterns matching broken symlinks. If the dirent exists, that should be good enough.

- perf: don't check/resolve/follow symlinks when matching root patterns

- Remove pointless conditionals such as `if match ~= nil then return match`

- Remove pointless check for file existence (`vim.fn.glob()` already guarantees this)

- perf: avoid unnecessary table creations (e.g. used only to concatenate strings; `string.format()` or `..` should be preferred for simple/static string concatenations)

- perf: avoid synchronous `vim.loop.fs*()` calls where possible (they tend to be slower than their builtin counterparts)

- Account for the fact that functions like `string.match()` and `string.gsub()` return multiple values (e.g. use `return (s:gsub(...))` to return the modified string)

- Add types to some of the utils

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -20,17 +20,19 @@ M.default_config = {
 -- global on_setup hook
 M.on_setup = nil
 
+--- @param bufname string
+--- @return boolean
 function M.bufname_valid(bufname)
-  if bufname:match '^/' or bufname:match '^[a-zA-Z]:' or bufname:match '^zipfile://' or bufname:match '^tarfile:' then
-    return true
-  end
-  return false
+  return (bufname:find '^/' or bufname:find '^%a:' or bufname:find '^zipfile://' or bufname:find '^tarfile:') ~= nil
 end
 
+--- @param bufnr integer
+--- @return integer
 function M.validate_bufnr(bufnr)
   if nvim_eleven then
     validate('bufnr', bufnr, 'number')
   end
+
   return bufnr == 0 and api.nvim_get_current_buf() or bufnr
 end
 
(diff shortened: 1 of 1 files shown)


justinmk (approved):
Most (or all) of util.lua is on the path to deprecation. However, improvements to it, especially simplifications, are still helpful because it makes it easier to understand what is missing from "upstream" Nvim stdlib.

tmillr:
It's basically ready now, but going through the file one more time, I observed and decided to fix the bug in `root_pattern()`, only to go down the issue-ref rabbit-hole and realize that [said bug is actually feature now](#2885).

The way `root_pattern()` was modified to prefer pattern order over directory depth feels very wrong to me, and I think the issue that caused it to get changed should have instead been:

1. **Fixed locally** (e.g. by removing certain patterns, or calling `root_pattern()` multiple times separately). There was never anything keeping people from calling it multiple times with certain (higher-precedence) patterns at a time if that was absolutely needed, and the algorithm would have been the same (as it is today).
2. Fixed by introducing another, separate, root pattern util which works differently (i.e. as it does today)

And it seems to have led to some [ugly/hacky code](1110787f1b464888c59a044c48c5119d14078044) in order to workaround the new (and IMO unintuitive) behavior, and it has caused at least a couple bugs as well (#3508).

But that change was over a year ago, and perhaps things were different then. The way I understand it now is that the `root_dir` callback drives/defines a workspace and its members/roots? Or in other words, it is the only way to achieve a multi-root workspace (maybe some servers determine this internally on their own, but I know at least `lua_ls` doesn't). This is another reason why I think `root_pattern()` should've stayed the way it was. A `.luarc.json` in a nested subdir should reuse the server but also create a new/separate workspace member/context, and it should probably even do the same upon finding a nested `.git`. But there may be some nuance here that I am unaware of.

But anyway, even though it should definitely be changed back to how it was, I understand (now) that that could potentially cause some servers' configuration to break as they may have been changed in the meantime in order to adapt. So just let me know whether to change it, or leave it (hopefully to be addressed later).

justinmk:
> and I think [the issue that caused it to get changed](https://github.com/neovim/nvim-lspconfig/pull/2885) should have instead been:
> 
> 1. **Fixed locally** (e.g. by removing certain patterns, or calling `root_pattern()` multiple times separately). There was never anything keeping people from calling it multiple times with certain (higher-precedence) patterns at a time if that was absolutely needed, and the algorithm would have been the same (as it is today).
> 2. Fixed by introducing another, separate, root pattern util which works differently (i.e. as it does today)

Thanks for those notes. That will help to articulate in documentation.

Meanwhile, please confirm that https://github.com/neovim/neovim/pull/33485 matches the behavior you expect.

Thanks!

PR #3626, opened 2025-02-26
fix(sourcekit): use textDocument/diagnostic over publishDiagnostics

MainKt:
Problem:
nvim doesn't recognize that sourcekit supports `textDocument/diagnostic` requests, causing it to rely on `textDocument/publishDiagnostics`, which results in delayed diagnostics.

Solution:
Manually configure `client.server_capabilities.diagnosticProvider` in `on_init` to enable `textDocument/diagnostic` support. Override the `textDocument/publishDiagnostics` handler to prevent duplicate diagnostics. 

fix neovim/neovim#31873

~~I'm not sure if there's a way to configure `server_capabilities` without `on_init` :sweat_smile:~~

### DIFF
--- a/lua/lspconfig/configs/sourcekit.lua
+++ b/lua/lspconfig/configs/sourcekit.lua
@@ -15,6 +15,14 @@ return {
       local t = { objc = 'objective-c', objcpp = 'objective-cpp' }
       return t[ftype] or ftype
     end,
+    capabilities = {
+      textDocument = {
+        diagnostic = {
+          dynamicRegistration = true,
+          relatedDocumentSupport = true,
+        },
+      },
+    },
   },
   docs = {
     description = [[


lua/lspconfig/configs/sourcekit.lua
```diff
@@ -15,6 +15,15 @@ return {
       local t = { objc = 'objective-c', objcpp = 'objective-cpp' }
       return t[ftype] or ftype
     end,
+    on_init = function(client)
+      client.server_capabilities.diagnosticProvider = {
```
justinmk, line 19: existing configs show examples of how to declare capabilities:

https://github.com/neovim/nvim-lspconfig/blob/6b63bdf2399b9bedf93297d98419550523a9ad68/lua/lspconfig/configs/clangd.lua#L65

MainKt: Thanks, I updated the config to dynamically register `textDocument/diagnostics`

PR #3628, opened 2025-02-26
fix(turbo_ls): Fix `root_dir`

okuramasafumi:
Currently, `root_dir` in `configs.md` is

- `root_dir` : ```lua "/home/runner/work/nvim-lspconfig/nvim-lspconfig" ```

This is wrong, so this commit fixes this.

### DIFF
--- a/lua/lspconfig/configs/turbo_ls.lua
+++ b/lua/lspconfig/configs/turbo_ls.lua
@@ -1,8 +1,10 @@
+local util = require 'lspconfig.util'
+
 return {
   default_config = {
     cmd = { 'turbo-language-server', '--stdio' },
     filetypes = { 'html', 'ruby', 'eruby', 'blade', 'php' },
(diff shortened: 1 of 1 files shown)


glepnir:
vim.fs.root is we suggest now.

okuramasafumi:
@glepnir OK, then how can we fix the wrong `root_dir` with `vim.fs.root`?

justinmk:
This looks like a docs issue. The docs for this field are usually generated here: https://github.com/neovim/nvim-lspconfig/blob/6b63bdf2399b9bedf93297d98419550523a9ad68/scripts/docgen.lua#L165

but it's returning early here: https://github.com/neovim/nvim-lspconfig/blob/6b63bdf2399b9bedf93297d98419550523a9ad68/scripts/docgen.lua#L148-L149

Need to fix docgen.lua.

PR #3639, opened 2025-03-07
feat(tinymist): implement simple workspace commands

DimitrisDimitropoulos:
This PR implements most of the simple workspace commands for tinymist in a "factory" way. It mainly implements the commands that are responsible for exporting to different formats and those that return info regarding the server and the document. Hope this is helpful.


lua/lspconfig/configs/tinymist.lua
```diff
@@ -1,3 +1,73 @@
...
+  local cmd_display = export_type or info_type
+  ---Execute the Tinymist command, supporting both 0.10 and 0.11 exec methods
+  ---@return nil
+  local function run_tinymist_command()
+    local bufnr = vim.api.nvim_get_current_buf()
+    local client = require 'lspconfig.util'.get_active_client_by_name(bufnr, 'tinymist')
+    if not client then
+      return vim.notify('No Tinymist client attached to the current buffer', vim.log.levels.ERROR)
+    end
+    local arguments = { vim.uri_to_fname(vim.uri_from_bufnr(bufnr)) }
```
glepnir, line 18: why not `api.nvim_buf_get_name`

lua/lspconfig/configs/tinymist.lua
```diff
@@ -1,3 +1,73 @@
...
+    end
+    if vim.fn.has 'nvim-0.11' == 1 then
+      -- For Neovim 0.11+
+      return client:exec_cmd({
+        title = title_str,
+        command = command_name,
+        arguments = arguments,
+      }, { bufnr = bufnr }, handler)
+    else
+      -- For Neovim < 0.11
```
justinmk, line 36: No, we aren't going to add new features for Nvim 0.10 or older.

lua/lspconfig/configs/tinymist.lua
```diff
@@ -7,6 +77,7 @@ return {
     end,
     single_file_support = true,
   },
+  commands = tinymist_commands,
```
justinmk, line 80: No. Just use the normal nvim_create_user_command api in a on_attach. This old 'commands' thing is deprecated.

lua/lspconfig/configs/tinymist.lua
```diff
@@ -7,6 +77,7 @@ return {
...
   docs = {
```
justinmk, line 81: docs should mention the commands?

DimitrisDimitropoulos:
I think all the issues have been addressed

Myriad-Dreamin:
I see a minor issue but this might be caused that I'm not oftenly playing with neovim. If I understand it correctly, if a too old binary is used, the related command to call will fail or miss some option hardly. They sometimes have to do that to work on some typst doucments that only get compiled in the old compiler.

Therefore, you should provide the commands available respecting dynamic information of tinymist. From my knowledge about lsp standard, this can be achieved correctly by checking the provided commands in the server capability.

DimitrisDimitropoulos:
If understand you point correctly, this shouldn't really be an issue since `exec_cmd` checks if the workspace command is supported, [see](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/client.lua?plain=1#L953-L966).

lua/lspconfig/configs/tinymist.lua
```diff
@@ -6,11 +50,33 @@ return {
...
+      end
+    end,
   },
   docs = {
     description = [[
 https://github.com/Myriad-Dreamin/tinymist
 An integrated language service for Typst [taɪpst]. You can also call it "微霭" [wēi ǎi] in Chinese.
+
+Currently some of Tinymist's workspace commands are supported, namely: exportSvg, exportPng, exportPdf, exportMarkdown,
+exportText, exportQuery, exportAnsiHighlight, getServerInfo, getDocumentTrace, getWorkspaceLabels and getDocumentMetrics
```
justinmk, line 79: ... what are the command names that the user can use?

DimitrisDimitropoulos: Done.

PR #3646, opened 2025-03-15
feat: muon

annacrombie:
Adds muon as a language server for meson files

### DIFF
--- a/lua/lspconfig/configs/muon.lua
+++ b/lua/lspconfig/configs/muon.lua
@@ -1,0 +1,20 @@
+local async = require('lspconfig.async')
+
+return {
+  default_config = {
+    cmd = { 'muon', 'analyze', 'lsp' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/muon.lua
```diff
@@ -0,0 +1,20 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'muon', 'analyze', 'lsp' },
+    filetypes = { 'meson' },
+    root_dir = function(fname)
+      local obj = vim.system({ 'muon', 'analyze', 'root-for', fname }):wait()
```
glepnir, line 8: there has async.run_command for async parse

glepnir: nope here  use coroutine control. When return resume the subsequent execution.

Issue #3645, opened 2025-03-14
There are two QML configs for the same language-server


lgrosz:
### Description

There are two configs for the same lsp.

- [qmlls](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/qmlls.lua)
- [qml_lsp](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/qml_lsp.lua)

[Here](https://doc.qt.io/qt-6/qtqml-tooling-qmlls.html) is the official Qt page for the language server. The preferred name appears to be "qmlls"

> Note: qmlls is currently in development, see [Known Limitations](https://doc.qt.io/qt-6/qtqml-tooling-qmlls.html#known-limitations) for more details.

The sources appears to live in [QtDeclarative](https://code.qt.io/cgit/qt/qtdeclarative.git/tree/).

I suppose I expect to see

- one config, qmlls.lua
- links to at least the Qt page describing the language server, https://doc.qt.io/qt-6/qtqml-tooling-qmlls.html
- if the source is linked, it should probably point to the upstream code.qt.io url, https://code.qt.io/cgit/qt/qtdeclarative.git/tree/

justinmk:
Thanks for noting that. Please send a PR that deletes `qml_lsp.lua` and improves the other one as you suggested.

We don't maintain these configs (there are 1000s of them), we need actual users to do that.

lgrosz:
Will do. I'll look into the contribution guidelines and submit a PR.

lgrosz:
Upon further investigation, I was just wrong. These are different projects, though it appears `qml_lsp` is archived. `qmlls` links to the GitHub mirror, which is fine, it's probably never going anywhere. I'm just going to close this issue.

justinmk:
> it appears `qml_lsp` is archived. `qmlls` links to the GitHub mirror, which is fine, it's probably never going anywhere

Cleaning up old things is useful, it avoids users having to discover the best path on their own. Thus deleting `qml_lsp` is worthwhile.

And if there's a better link for `qmlls`, updating it is helpful.

lgrosz:
Sounds good, I will make the changes.

PR #3647, opened 2025-03-16
Chore and doc tweaks to QML-related LSPs

lgrosz:
```git
* ... docs(qmlls): point to official sources
* ... chore(qml_lsp)!: remove config
```


lgrosz:
Closes #3645

PR #3649, opened 2025-03-19
fix(tblgen): find tablegen_compile_commands.yml

anirudhsundar:
The `tblgen_lsp_server` doesn't seem to load any specific compilation database server and hence we need to explicitly specify the command-line argument to pick the correct database path.

### DIFF
--- a/lua/lspconfig/configs/tblgen_lsp_server.lua
+++ b/lua/lspconfig/configs/tblgen_lsp_server.lua
@@ -1,8 +1,19 @@
 local util = require 'lspconfig.util'
 
+local function get_command()
+  local cmd = { 'tblgen-lsp-server' }
+  local files = vim.fs.find('tablegen_compile_commands.yml', { path = vim.fn.expand('%:p:h'), upward = true })
+  if #files > 0 then
+    local file = files[1]
+    table.insert(cmd, '--tablegen-compilation-database=' .. vim.fs.dirname(file) .. '/tablegen_compile_commands.yml')
+  end
+
+  return cmd
+end
+
 return {
   default_config = {
-    cmd = { 'tblgen-lsp-server' },
+    cmd = get_command(),
     filetypes = { 'tablegen' },
     root_dir = function(fname)
       return util.root_pattern 'tablegen_compile_commands.yml'(fname)


lua/lspconfig/configs/tblgen_lsp_server.lua
```diff
@@ -1,8 +1,14 @@
 local util = require 'lspconfig.util'
 
+local get_compile_commands_path = function()
+  return '--tablegen-compilation-database='
+    .. vim.fs.dirname(vim.fs.find('.git', { path = vim.fn.expand('%:p:h'), upward = true })[1])
```
glepnir, line 5: if there is not .git folder what happen here

lua/lspconfig/configs/tblgen_lsp_server.lua
```diff
@@ -1,8 +1,14 @@
 local util = require 'lspconfig.util'
 
+local get_compile_commands_path = function()
```
justinmk, line 3:
```suggestion
local function get_compile_commands_path()
```

lua/lspconfig/configs/tblgen_lsp_server.lua
anirudhsundar: Thanks for the catch, I've modified the code to return an empty string if the `.git` directory is not found so that it falls back to LSP invocation without any arguments.

I'm only checking for the existence of `.git` directory and assuming that `tablegen_compile_commands.yml` would be present in that directory. If that is missing, the LSP server would handle it gracefully.

anirudhsundar:
Friendly ping for review

lua/lspconfig/configs/tblgen_lsp_server.lua
```diff
@@ -1,8 +1,18 @@
 local util = require 'lspconfig.util'
 
+local function get_compile_commands_path()
+  local files = vim.fs.find('tablegen_compile_commands.yml', { path = vim.fn.expand('%:p:h'), upward = true })
+  if #files > 0 then
+    local file = files[1]
+    return '--tablegen-compilation-database=' .. vim.fs.dirname(file) .. '/tablegen_compile_commands.yml'
+  end
+
+  return ''
```
justinmk, line 10: This will send a literal empty string arg to `tblgen-lsp-server`, are you sure it handles that?

anirudhsundar: Ah okay, I did not know that (I'm learning both lua and nvim-lsp as I'm doing this). 

`tblgen-lsp-server ""` doesn't work on the shell, but weirdly enough, I did not get any errors when I tested the LSP on a directory that does not have any `tablgen_compile_commands.yml` file. Maybe the error was printed in LSP log and I missed it.

Anyway, thanks for the catch. I modified the function to build the whole command instead of just the argument. This new version should definitely work in all cases as far as I can tell.

Let me know if this makes sense, thanks.

anirudhsundar:
@justinmk @glepnir Is the current version good to be merged

PR #3650, opened 2025-03-19
fix(biome): improved biomejs detection

maximtrp:
Biome configuration files (`biome.json`, `biome.jsonc`) are [not required](https://biomejs.dev/guides/configure-biome/#configuration-file-resolution) for Biome to function within a project. If no configuration file is present, Biome defaults to its built-in settings. Therefore, I suggest to check for the Biome package in the `package.json` file in addition to looking for configuration files.

### DIFF
--- a/lua/lspconfig/configs/biome.lua
+++ b/lua/lspconfig/configs/biome.lua
@@ -17,7 +17,11 @@ return {
       'typescriptreact',
       'vue',
     },
-    root_dir = util.root_pattern('biome.json', 'biome.jsonc'),
+    root_dir = function(fname)
+      local root_files = { 'biome.json', 'biome.jsonc' }
+      root_files = util.insert_package_json(root_files, 'biome', fname)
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/biome.lua
```diff
@@ -17,7 +17,11 @@ return {
       'typescriptreact',
       'vue',
     },
-    root_dir = util.root_pattern('biome.json', 'biome.jsonc'),
+    root_dir = function(fname)
+      local root_file = { 'biome.json', 'biome.jsonc' }
+      root_file = util.insert_package_json(root_file, 'biome', fname)
+      return util.root_pattern(unpack(root_file))(fname)
```
glepnir, line 23: could you please use `vim.fs.find` and vim.fs.root instead

glepnir: root_pattern

PR #3657, opened 2025-03-21
feat: postgrestools

psteinroe:
adds support for the postgres language server (https://github.com/supabase-community/postgres_lsp)


lua/lspconfig/configs/postgrestools.lua
```diff
@@ -0,0 +1,19 @@
+local util = require 'lspconfig.util'
+
+return {
+  default_config = {
+    cmd = { 'postgrestools', 'lsp-proxy' },
+    filetypes = {
+      'sql',
+    },
+    root_dir = util.root_pattern('postgrestools.jsonc'),
```
glepnir, line 9: use vim.fs.root instead of root_pattern

psteinroe: thanks for the quick review! fixed it

sharpchen:
I remember postgre_lsp was in beta, so since it renamed its executable, maybe just update on the original config?

justinmk:
yes, just update `postgres_lsp` instead of adding a new one.

psteinroe:
thanks for the feedback! I updated postgres_lsp instead.

PR #3661, opened 2025-03-26
feat: upgrade for golangci-lint v2

Mi7teR:
`--out-format=json` was replaced by `--out.json.path=stdout`
https://github.com/golangci/golangci-lint/blob/86cc7c6bfbe4d6ca515f9c45b58cfa03cfd0c3da/docs/src/docs/product/migration-guide.mdx#command-line-flags

PR #3662, opened 2025-03-26
fix(golangci-lint): fix typo in commang argument

Mi7teR:
fix for https://github.com/neovim/nvim-lspconfig/pull/3661
thanks to [@ravi-dd](https://github.com/neovim/nvim-lspconfig/pull/3661#discussion_r2014241733)


sherif-fanous:
I think you also need to pass `,"--show-stats=false" ` otherwise you end up with an error message similar to the following one

`invalid character '6' after top-level value`

I've overridden my command to `command = { "golangci-lint", "run", "--output.json.path=stdout", "--show-stats=false" },` and now it runs with no errors

PR #3666, opened 2025-03-27
feat(async): add `run_job`, a coroutine-less version of `run_command`

TheRealLorenz:
Needed for neovim 0.11 port

3 configurations use `async.run_command`, but this function cannot be used outside of a coroutine context. This is a replacement that's made exactly for that

Needed for #3659

### DIFF
--- a/lua/lspconfig/async.lua
+++ b/lua/lspconfig/async.lua
@@ -57,6 +57,47 @@ function M.run_command(cmd)
   return stdout and stdout or nil
 end
 
+---@param cmd string[]
+---@param on_done function(string[]?)
+---@param on_error? function(integer, string[]?)
+function M.run_job(cmd, on_done, on_error)
+  local stdout = {}
+  local stderr = {}
+
+  local jobid = vim.fn.jobstart(cmd, {
+    on_stdout = function(_, data, _)
+      data = table.concat(data, '\n')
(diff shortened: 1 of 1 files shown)


lua/lspconfig/async.lua
```diff
@@ -57,6 +57,47 @@ function M.run_command(cmd)
...
 end
 
+---@param cmd string[]
+---@param on_done function(string[]?)
+---@param on_error? function(integer, string[]?)
+function M.run_job(cmd, on_done, on_error)
+  local stdout = {}
+  local stderr = {}
+
+  local jobid = vim.fn.jobstart(cmd, {
```
justinmk, line 67: Can `vim.system` be used instead?

justinmk:
> 3 configurations use `async.run_command`

Why ?

TheRealLorenz:
> > 3 configurations use `async.run_command`
> 
> Why ?

Well I don't know Go that much, but its configuration runs `go env GOMODCACHE` to get the cache folder location. There's also `rust_analyzer` and `muon`.

lua/lspconfig/async.lua
TheRealLorenz: I think `vim.system` might actually be better, the syntax is easier. Maybe I could inline the calls and close this PR.

PR #3674, opened 2025-03-29
fix: format doc/configs.txt as proper :help-compatible help file

### LINKED ISSUE #3600: :help pages filetype anchors
### Description

The helppage for `:h lspconfig-all` is written with `markdown` syntax rather than `help` syntax and is aptly marked as such with a vim comment at the bottom.

However helpages has `iskeyword` set to `iskeyword=!-~,^*,^|,^",192-255` but usual default for other files (like markdown) is `iskeyword=@,48-57,_,192-255`.

This is problematic because it makes it difficult to follow the links at the the top of the helppage which look like this:

```markdown
- [agda_ls](#agda_ls)
- [aiken](#aiken)
```

1. `K` does not work in plain neovim as it doesn't support markdown anchors
2. `*` does not work because it selects across the word boundary characters like `]` `(`

Possible fix: also set `iskeyword` in the footer comment.

o6ivp:
Fixes #3600

This PR reformats `doc/configs.txt` to follow Neovim `:help` conventions.

### What it does

- Adds proper `*tag*` definitions for each server (`*pyright*`, `*gopls*`, etc.)
- Ensures `:help lspconfig-all`, `:help ada_ls` etc. work
- Fixes broken `K` key and helptags indexing
- Removes duplicate tags, fixes formatting, adds section headers

### Why this is needed

Currently, `configs.txt` is not valid as a `:help` file and cannot be indexed properly (see #3600).  
This PR does not change content — only formatting.

Note: This file is normally generated, but this PR intentionally updates it to fix structure and tagging.


o6ivp:
Thanks for the review. I have corrected the issue in #3675.
I would be happy to have your feedback again.

PR #3684, opened 2025-04-02
feat: add PL/I support

VitGottwald:


justinmk:
Not accepting new configs in `lua/lspconfig/configs/`, but can merge this if you move it to `lsp/*` and convert it to the new format (see other configs for example)

VitGottwald:
Converted it to the new format as requested and tested locally.

PR #3686, opened 2025-04-03
fix(eslint): Find closest ESLint directory to avoid version mismatch

### LINKED ISSUE #3685: ESLint: Possible issue with root directory when working with different ESLint versions
### Description

Hi there!  

First off, thanks for the awesome project—it makes configuring LSPs in Neovim so much easier!  

### Context  
I ran into an issue where VSCode correctly picked up the closest ESLint version in my project, but Neovim’s ESLint LSP always used the root directory’s version instead.  

After troubleshooting, I found that my inner directory had ESLint `^9.20.1`, while the project root had `7.32.0`. Neovim defaulted to the root’s version instead of the closest one. I resolved this by tweaking my config:  

```lua
local function get_eslint_closest_dir()
  local cwd = vim.fn.getcwd()
  local eslint_node_modules = vim.fn.finddir('node_modules/eslint', cwd .. ';')

  if eslint_node_modules == '' then
    return nil
  end

  return eslint_node_modules == 'node_modules/eslint' and cwd
    or eslint_node_modules:match '(.*)/node_modules/eslint'
end

return {
  single_file_support = true,
  on_new_config = function(config, new_root_dir)
    new_root_dir = get_eslint_closest_dir() or new_root_dir
    -- (continuing from on_new_config in nvim-lspconfig)
}
```

It seems like `new_root_dir` is always set to the project’s root when opening a file.  

So I was wondering—does this behavior come from the `on_new_config` callback, the ESLint LSP itself, or is it something `nvim-lspconfig` should handle? 

Thanks in advance! 😊

estacioneto:
Fixes #3685, where projects with different ESLint versions incorrectly used the farthest version instead of the closest.

### DIFF
--- a/lua/lspconfig/configs/eslint.lua
+++ b/lua/lspconfig/configs/eslint.lua
@@ -47,6 +47,28 @@ local root_file = {
   'eslint.config.cts',
 }
 
+--- Find the closest directory containing the ESLint library.
+--- Using the repository root as the ESLint library location is a common practice
+--- but handling different versions in monorepos can cause the LSP not to point to
+--- the correct instance.
+---
+--- @return string|nil (path) The path to the closest directory containing the ESLint library, or nil if not found.
+local function get_eslint_closest_dir()
+  local cwd = vim.fn.getcwd()
+  local eslint_node_modules = vim.fn.finddir('node_modules/eslint', cwd .. ';')
+
+  if eslint_node_modules == '' then
+    return nil
+  end
+
+  if eslint_node_modules == 'node_modules/eslint' then
+    return cwd
+  end
+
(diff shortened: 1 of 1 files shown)


MrRoiz:
Not sure if this is the issue but after updating `nvim-lspconfig` my eslint lsp does not shows any diagnosis 🤔

justinmk:
reverted : https://github.com/neovim/nvim-lspconfig/pull/3689

estacioneto:
Interesting @MrRoiz... Can you share the `:LspLog` so I can figure out what made it not show the diagnosis? I'm curious on how it broke your setup

PR #3688, opened 2025-04-04
feat(fountain): added fountain-lsp-server

nooneknowspeter:
adds support for [fountain-lsp-server](https://github.com/oparaskos/fountain-lsp-server)
it depends on the package from the mason registry(sent a pull request) or can be built from source

### DIFF
--- a/doc/configs.md
+++ b/doc/configs.md
@@ -108,6 +108,7 @@ Nvim by running `:help lspconfig-all`.
 - [flux_lsp](#flux_lsp)
 - [foam_ls](#foam_ls)
 - [fortls](#fortls)
+- [fountain-lsp-server](#fountain-lsp-server)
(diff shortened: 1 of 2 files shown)


justinmk:
> it depends on the package from the mason registry

huh?

https://github.com/oparaskos/fountain-lsp-server has no activity (not even discussions) for 1+ months, and no commits for 2 years.

If users actually request this we can revisit it.

nooneknowspeter:
> > it depends on the package from the mason registry
> 
> huh? what is the URL for this?
> 
> https://github.com/oparaskos/fountain-lsp-server has no activity (not even discussions) for 1+ months, and no commits for 2 years.
> 
> If users actually request this we can revisit it.

This is the pr i mentioned -> https://github.com/mason-org/mason-registry/pull/9584

I understand

PR #3690, opened 2025-04-05
fix: improve path escaping in wildcard function

ryoppippi:
The escape_wildcards function was enhanced to handle a broader range of special characters that could cause issues when used in pattern matching. This includes additional wildcards, pattern matching characters, shell special characters, the escape character itself, and whitespace.

I faced this issue when I work on React Router and Tanstack 	Router.
They adapt file-based routing, and they use special characters like `$` as filename.
https://reactrouter.com/how-to/file-route-conventions#optional-segments

Before this fix, it takes like 15 seconds to open files like `foo.$.ts`, but this fixes!

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -124,7 +124,7 @@ local function escape_wildcards(path)
   -- - Shell special characters: ` ' " ; | & < >
   -- - Whitespace: %s (spaces, tabs, etc.)
   -- The gsub function replaces any matches with the same character preceded by a backslash
-  return path:gsub('([%[%]%?%*%$%(%)%^%%%.%+%-%~\\`\'";|&<>%s])', '\\%1')
+  return path:gsub('([' .. vim.pesc('[]?*$()^%.+-~') .. '\\`\'";|&<>%s])', '\\%1')
 end
 
 function M.root_pattern(...)
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -117,7 +117,14 @@ function M.search_ancestors(startpath, func)
 end
 
 local function escape_wildcards(path)
-  return path:gsub('([%[%]%?%*])', '\\%1')
+  -- Escape all potentially problematic special characters
+  -- Covers:
+  -- - Wildcards and pattern matching: [ ] ? * $ ( ) ^ % . + - ~
+  -- - Escape character: \
+  -- - Shell special characters: ` ' " ; | & < >
+  -- - Whitespace: %s (spaces, tabs, etc.)
+  -- The gsub function replaces any matches with the same character preceded by a backslash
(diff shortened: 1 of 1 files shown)


justinmk:
I see that this is used in `root_pattern`.

Note that all of util.lua will be removed: https://github.com/neovim/nvim-lspconfig/issues/2079

But in order to fully remove it, we need to figure out what's still missing in Nvim core. 

Nvim/Vim has `shellescape()`, but that seems not relevant because here the glob is passed to `vim.fn.glob()`. Seems like `globescape()` is missing.

lua/lspconfig/util.lua
```diff
@@ -117,7 +117,14 @@ function M.search_ancestors(startpath, func)
...
 local function escape_wildcards(path)
-  return path:gsub('([%[%]%?%*])', '\\%1')
+  -- Escape all potentially problematic special characters
+  -- Covers:
+  -- - Wildcards and pattern matching: [ ] ? * $ ( ) ^ % . + - ~
+  -- - Escape character: \
+  -- - Shell special characters: ` ' " ; | & < >
+  -- - Whitespace: %s (spaces, tabs, etc.)
+  -- The gsub function replaces any matches with the same character preceded by a backslash
+  return path:gsub('([%[%]%?%*%$%(%)%^%%%.%+%-%~\\`\'";|&<>%s])', '\\%1')
```
justinmk, line 127: can this be simplified using `vim.pesc()` ? 
```suggestion
  return path:gsub('([' .. vim.pesc('[]?*$()^%.+-~') .. '\\`\'";|&<>%s])', '\\%1')
```

ryoppippi: nice I didn't know that!

ryoppippi:
> Note that all of util.lua will be removed: https://github.com/neovim/nvim-lspconfig/issues/2079

Okay! I remember. But this issue is critical for me right now. I hope this PR fixes my loading issue!

justinmk:
I didn't test my code suggestion. Can you confirm that you tested the latest version of this PR?

justinmk:
still waiting on an answer to my last comment.

also, please rebase

ryoppippi:
Oh sorry. did I need to write a test code? it works fine on my local!

slavek-kucera:
It looks like this change has completely broken the functionality.
`root_pattern` no longer works if any of the directories in the path have e.g. `-` in its name.

justinmk:
reverted https://github.com/neovim/nvim-lspconfig/pull/3729

@ryoppippi if you want to try a more targeted change, please feel free

ryoppippi:
I stop using `root_pattern` and migrate my dotfiles to use `lua/lsp/foo.lua` style lsp config. but I still face an issue when I open files like `foo/bar.$.tsx` like that. Do you have any ideas?

ryoppippi:
Okay, I found neoconf using this root_pattern or vim.fn.glob function internally

PR #3692, opened 2025-04-05
feat: require Nvim 0.10+

justinmk:
Pinned announcement: https://github.com/neovim/nvim-lspconfig/issues/3693

cc @dundargoc


dundargoc (approved):
Looks good. There's a bunch of code and docs we can remove now, but we can deal with that later if you don't wanna handle it now.

PR #3711, opened 2025-04-13
fix(configs): remove utils in config with globs as markers

TheRealLorenz:
Based on #2079.

The majority of configs which rely on `lspconfig.util` use `util.root_pattern` to match against wildcards. I've opened an issue here https://github.com/neovim/neovim/issues/33444 to discuss about this.

In the meantime I'll remove `lspconfig.util` where it's trivial to do so.

### DIFF
--- a/lsp/ada_ls.lua
+++ b/lsp/ada_ls.lua
@@ -18,13 +18,18 @@
 --- })
 --- ```
 
-local util = require 'lspconfig.util'
-
 return {
   cmd = { 'ada_language_server' },
   filetypes = { 'ada' },
   root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern('Makefile', '.git', 'alire.toml', '*.gpr', '*.adc')(fname))
+    on_dir(vim.fs.root(bufnr, function(name, _)
+      local patterns = { 'Makefile', '.git', 'alire.toml', '*.gpr', '*.adc' }
+      for _, pattern in ipairs(patterns) do
+        if vim.glob.to_lpeg(pattern):match(name) ~= nil then
+          return true
+        end
+      end
+      return false
+    end))
   end,
 }
--- a/lsp/autotools_ls.lua
+++ b/lsp/autotools_ls.lua
@@ -9,15 +9,19 @@
 ---
 --- Language server for autoconf, automake and make using tree sitter in python.
 
-local util = require 'lspconfig.util'
-
 local root_files = { 'configure.ac', 'Makefile', 'Makefile.am', '*.mk' }
 
 return {
   cmd = { 'autotools-language-server' },
   filetypes = { 'config', 'automake', 'make' },
   root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern(unpack(root_files))(fname))
+    on_dir(vim.fs.root(bufnr, function(name, _)
+      for _, pattern in ipairs(root_files) do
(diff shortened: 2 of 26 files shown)


justinmk (approved):

TheRealLorenz:
I'm not entirely sure if this migration is ok. It seems like `util.root_pattern` and `vim.fs.root` behave differently, the first looks for the first pattern in the entire path before checking the others, while `vim.fs.root` I think searches every pattern in every folder before going up.

TheRealLorenz:
I don't know if these is relevant for *most* of the configs, but if that is the case we would probably need a flag to change the behaviour of `vim.fs.root()` to specify the precedence of operations.

One example is `sourcekit.lua` 

https://github.com/TheRealLorenz/nvim-lspconfig/blob/deacc03f6480b69c718f70a787e7ccb928c8a64c/lsp/sourcekit.lua#L11-L21

Here it was explicitly said that the precedence was important, so I called `vim.fs.root()` multiple times.

lsp/ada_ls.lua
```diff
@@ -22,8 +20,13 @@ local util = require 'lspconfig.util'
 return {
   cmd = { 'ada_language_server' },
   filetypes = { 'ada' },
-  root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern('Makefile', '.git', 'alire.toml', '*.gpr', '*.adc')(fname))
+  root_markers = function(name, _)
```
justinmk, line 23: I don't think `root_markers` accepts a function, based on `:help vim.lsp.Config`.

lsp/autotools_ls.lua
```diff
@@ -15,8 +13,12 @@ local root_files = { 'configure.ac', 'Makefile', 'Makefile.am', '*.mk' }
 return {
   cmd = { 'autotools-language-server' },
   filetypes = { 'config', 'automake', 'make' },
-  root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern(unpack(root_files))(fname))
+  root_markers = function(name, _)
+    for _, pattern in ipairs(root_files) do
+      if vim.glob.to_lpeg(pattern):match(name) ~= nil then
```
justinmk, line 18: If this is the pattern that's actually unavoidable in every config, it's a strong hint that we need to resolve https://github.com/neovim/neovim/issues/33444 (by enhancing `vim.fs.find()`, and/or accepting a function for `root_markers`, and/or supporting wildcards directly in `root_dir` / `root_markers`).

justinmk:
> I'm not entirely sure if this migration is ok. It seems like `util.root_pattern` and `vim.fs.root` behave differently

See also https://github.com/neovim/nvim-lspconfig/pull/3651 . Deciding on well-defined behavior will require some thought.

justinmk:
needs a rebase after https://github.com/neovim/nvim-lspconfig/pull/3767 and https://github.com/neovim/nvim-lspconfig/pull/3766

justinmk:
https://github.com/neovim/nvim-lspconfig/pull/3820 removes `root_pattern` for non-glob cases.

TheRealLorenz:
With https://github.com/neovim/neovim/pull/33771 we should be close to getting this PR ready. There's still an issue with `util.insert_package_json()`, as in:

https://github.com/neovim/nvim-lspconfig/blob/bedca8b426b2fee0ccac596d167d71bbe971253f/lsp/stylelint_lsp.lua#L56-L59

How would this be tackled?

justinmk:
Not sure about `util.insert_package_json`, but can worry about that later :)

My first thought would be to propose a `vim.fs` enhancement that generalizes it.

PR #3712, opened 2025-04-13
docs: change contributing.md to new style lsp config

yochem:
Problem:
Contributing guide describes old style of lsp configuration.

Solution:
Change to explain new style.


justinmk:
Thanks! I have an incoming update to lspconfig.txt also.

Issue #3705, opened 2025-04-11
Just to have a reminder, these are the LSPs which rely on `on_new_config`:


justinmk:
Just to have a reminder, these are the LSPs which rely on `on_new_config`:

- `angularls`
- `apex_ls`
- `astro`
- `bqnlsp`
- `cadence`
- `codeqlls`
- `drools_lsp`
- `eslint`
- `glint`
- `haxe_language_server`
- `jsonnet_ls`
- `leanls`
- `mdx_analyzer`
- `omnisharp`
- `openedge_ls`
- `powershell_es`
- `relay_lsp`
- `sourcery`
- `tailwindcss`
- `vdmj`
- `volar`
- `zls`

These are the ones which have `single_file_support = false`:

- `biome`
- `bitbake_ls`
- `ccls`
- `delphi_ls`
- `efm`
- `gh_actions_ls`
- `matlab_ls`
- `oxlint`
- `vectorcode_server`

_Originally posted by @TheRealLorenz in https://github.com/neovim/nvim-lspconfig/issues/3659#issuecomment-2776613000_

justinmk:
- https://github.com/neovim/nvim-lspconfig/pull/3666 not needed because `vim.system()` can be used instead

ybc37:
`efm` doesn't has `single_file_support = false`. Its explicitly set to `true`: <https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/efm.lua#L7>.

Is it in this list by mistake and could be migrated to `vim.lsp.config`?

TheRealLorenz:
> `efm` doesn't has `single_file_support = false`. Its explicitly set to `true`: <https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/efm.lua#L7>.
> 
> Is it in this list by mistake and could be migrated to `vim.lsp.config`?

I don't use `efm` myself, so I thought to wait until neovim 0.11.1 in order to also port the documentation which said that older version of `efm ` didn't have single file support.

rockyzhang24:
@ybc37 You could use `workspace_required` to deal with `single_file_support`, ref: https://github.com/neovim/neovim/pull/31824

TheRealLorenz:
> @ybc37 You could use `workspace_required` to deal with `single_file_support`, ref: https://github.com/neovim/neovim/pull/31824

It's not available in the current neovim release.

justinmk:
[...]

PR #3713, opened 2025-04-13
feat(gh_actions_ls): add `vim.lsp.config` support

chrisgrieser:
`gh_actions_ls` is missing in the `lsp` folder, likely due to it using a workaround with single file support to prevent it from attaching to every yaml file, even though it should only attach to certain yaml files, without defining an extra `yaml.github` filetype (see #3558).

Using `.github/workflows` and co as `root_markers` however prevents this issue, allowing us to use the LSP like a regular LSP without it attaching to unrelated yaml files. Thus, there is no anymore problem using `git_actions_ls` with `vim.lsp.config`.

cc @disrupted

### DIFF
--- a/lsp/gh_actions_ls.lua
+++ b/lsp/gh_actions_ls.lua
@@ -1,0 +1,31 @@
+---@brief
+-- https://github.com/lttb/gh-actions-language-server
+--
+-- Language server for GitHub Actions.
+--
+-- The projects [forgejo](https://forgejo.org/) and [gitea](https://about.gitea.com/)
+-- design their actions to be as compatible to github as possible
+-- with only [a few differences](https://docs.gitea.com/usage/actions/comparison#unsupported-workflows-syntax) between the systems.
+-- The `gh_actions_ls` is therefore enabled for those `yaml` files as well.
+--
+-- The `gh-actions-language-server` can be installed via `npm`:
+--
+-- ```sh
+-- npm install -g gh-actions-language-server
+-- ```
+return {
+  cmd = { 'gh-actions-language-server', '--stdio' },
+  filetypes = { 'yaml' }, -- the `root_markers` prevent attaching to every yaml file
+  root_markers = {
+    '.github/workflows',
+    '.forgejo/workflows',
+    '.gitea/workflows',
+  },
+  capabilities = {
+    workspace = {
+      didChangeWorkspaceFolders = {
+        dynamicRegistration = true,
+      },
+    },
+  },
+}


justinmk:
Nice, thanks! So we can cross this from https://github.com/neovim/nvim-lspconfig/issues/3705

gstokkink:
Just a headsup, this LSP config did not work for me until I also added the following to the config:

```
          init_options = {
            sessionToken = <whatever>
          },
```

Maybe that needs to be documented?

Also, diagnostic error messages don't show up, but maybe there's something broken in my setup, though this only occurs for the particular LSP as far as I can see.

chrisgrieser:
@gstokkink That's weird, when I tested it earlier, it seemed to work. But I can confirm I also get the error, and diagnostics not showing up anymore. I am not sure what causes this behavior though, `vim.lsp.config` and `lspconfig[server].setup` should be *mostly* equivalent?

justinmk:
> `vim.lsp.config` and `lspconfig[server].setup` should be _mostly_ equivalent?

yes. and I don't see how `init_options` could be related to their differences anyway.

does the old config still work? maybe something changes in the LS.

gstokkink:
Weird thing is that sometimes the diagnostics _do_ show up, but I haven't yet figured out what causes this erratic behaviour. Maybe some caching issue?

justinmk:
Reminds me of https://github.com/neovim/neovim/issues/32247

Maybe https://github.com/neovim/neovim/issues/33391#issuecomment-2799960824 provides a hint that could be used here?

gstokkink:
Yeah it definitely seems to have to do with previous edits. For one particular workflow now the diagnostic symbols show up in the gutter but nothing else is shown. For other workflows it works fine.

gstokkink:
Alright, fixed it. I had to set 'event' to `VeryLazy` instead of `LspAttach` for the diagnostic plugin I use: https://github.com/rachartier/tiny-inline-diagnostic.nvim. Point about the `sessionToken` still stands, however.

justinmk:
A small update to the docstring would be welcome

chrisgrieser:
Aside from the session-token issue, the LS still attaches to some yaml files it's not supposed to attach. I am very sorry, I assume I messed up my testing of the change by somehow still running `lspconfig[server].setup` instead of `vim.lsp.config`.

gstokkink:
Yeah, I noticed it attaching to other YAML files as well. Not sure how this can happen, are the root markers not being evaluated correctly?

justinmk:
Look at the old config, and figure out how to make it work with the new one.

chrisgrieser:
> Yeah, I noticed it attaching to other YAML files as well. Not sure how this can happen, are the root markers not being evaluated correctly?

Yeah that one is due to the `vim.lsp.config` not supporting "disabling" single file mode. nvim 0.11.1 (or nightly riht now) will support `workspace_required = true`, which I just added via PR.

gstokkink:
@chrisgrieser thanks! I've resorted to using the `yaml.github` workaround in the meantime 🙂

jhoogstraat:
The issue is not so much the `sessionToken`, but `init_options` not being present:
https://github.com/actions/languageservices/blob/e2ec264801c7166e011b35424de4dc0a2aa33811/languageserver/src/connection.ts#L54

chrisgrieser:
@gstokkink @jhoogstraat 
I made a PR which should fix both issues, if you want to have a look: #3857

PR #3716, opened 2025-04-13
feat(biome): add `vim.lsp.config` support

chrisgrieser:
Another one to check off the list at #3705.

`single_file_support = false` was introduced in PR #2984. 

As far as I can tell, it is not relevant anymore: when using `vim.lsp.config`'s `root_markers`, the LSP is simply not attached when the root marker is not found, so the problem that #2984 was solving does not occur.

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -1,0 +1,27 @@
+---@brief
+-- https://biomejs.dev
+--
+-- Toolchain of the web. [Successor of Rome](https://biomejs.dev/blog/annoucing-biome).
+--
+-- ```sh
+-- npm install [-g] @biomejs/biome
+-- ```
+
+return {
+  cmd = { 'biome', 'lsp-proxy' },
+  filetypes = {
+    'astro',
+    'css',
+    'graphql',
+    'javascript',
+    'javascriptreact',
+    'json',
+    'jsonc',
+    'svelte',
+    'typescript',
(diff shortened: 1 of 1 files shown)


igorlfs:
Hello!

> As far as I can tell, it is not relevant anymore: when using vim.lsp.config's root_markers, the LSP is simply not attached when the root marker is not found, so the problem that https://github.com/neovim/nvim-lspconfig/pull/2984 was solving does not occur.

Unfortunately, that does not seem to be the case. The LSP still gets attached, although with a nil root directory.

```
vim.lsp: Active Clients ~
- biome (id: 1)
  - Version: 0.0.0
  - Root directory: nil
  - Command: { "biome", "lsp-proxy" }
  - Settings: {}
  - Attached buffers: 1
```

justinmk:
does it improve anything if you set `workspace_required=true` in the config?

```lua'
vim.lsp.config('biome', {
  workspace_required = true,
})
```

igorlfs:
> does it improve anything if you set `workspace_required=true` in the config?
> 
> ```
> vim.lsp.config('biome', {
>   workspace_required = true,
> })
> ```

Yep, that fixes the issue

chrisgrieser:
My bad, I misunderstood how `root_markers` was supposed to work. 

Should this PR be reverted until `workspace_required` is released on stable?

PR #3718, opened 2025-04-13
Revert "feat(biome): add `vim.lsp.config` support"

chrisgrieser:
Reverts neovim/nvim-lspconfig#3716, due to the issues reported there. 

I am sorry, I think I messed up something when testing the change on my end.


justinmk:
er, why not add `workspace_required` to the config? https://github.com/neovim/nvim-lspconfig/pull/3716#issuecomment-2800006283

chrisgrieser:
> why not add `workspace_required` to the config?

Isn't `workspace_required` only on nightly? I am on 0.11 and cannot find it on `vim.lsp.ClientConfig` / `vim.lsp.Config`. I assumed lsp-config targets stable rather than nightly?

justinmk:
it will be in 0.11.1

nvim-lspconfig targets stable, but I think passing unknown fields is allowed, so adding it is harmless until 0.11.1 is released (soon).

you can try the nightly: https://github.com/neovim/neovim/releases/tag/nightly

chrisgrieser:
> nvim-lspconfig targets stable, but I think passing unknown fields is allowed, so adding it is harmless until 0.11.1 is released

Well, in that case, should I re-add the `lsp/` file for `biome`, but with `workspace_required = true`? (Sorry for all the back-and-forth on reverts here 🙈)

maximtrp:
@chrisgrieser  Why not use `root_dir` here as in my pull request (#3650)? Biome does not require a config to run.

chrisgrieser:
@maximtrp I wasn't aware of the reasoning for that. Anyway, my PR is now reverted anyway, so in a PR adding `lsp/biome.lua` for 0.11.1, that can be rectified.

justinmk:
SGTM

PR #3719, opened 2025-04-13
Revert "feat(gh_actions_ls): add `vim.lsp.config` support"

chrisgrieser:
Reverts neovim/nvim-lspconfig#3713, due to the various issues reported there making the LSP not properly usable with `vim.lsp.config` yet.


justinmk:
Thanks for the followup. 

If the config doesn't work , people don't have to use it. Or they can fix it. No need to revert it.

PR #3725, opened 2025-04-14
Use lspconfig to demonstrate configuring pylsp

idbrii:
Instead of using vim.lsp, show how to configure pylsp with lspconfig.

Test
Run the example code from `:lua`. Also try with E402. See warnings appear and disappear as I reconfigure.


antonk52:
Can you please expand on why this change is needed? lspconfig is in process of moving away from the `lspconfig[name].setup` interface in favour of builtin `vim.lsp.config`. This server can already be configured using the builtin `config` function.

PR #3726, opened 2025-04-15
feat(atlas): add support for `atlas-rule` filetype

luantranminh:
Following #3539, this PR adds a new `atlas-rule` file type.


luantranminh:
@justinmk Thank you for reminding me. I updated the `lsp/*` instead.

justinmk:
This currently only changes the docs. Is that intended?

PR #3730, opened 2025-04-15
remove .git from root_markers

cbackas:
https://github.com/neovim/nvim-lspconfig/issues/3728

So far I've opted not to touch any LSPs with `root_dir` functions that may or may not include checks for `.git`.  I split the commits up (mostly correctly) into LSPs that had more than 1 root_marker and LSPs that had `.git` as their only root marker, I wasn't sure if we'd end up wanting to keep the `.git` marker on there if its all there is? Also the `nixd` LSP has a root_marker of `git` instead of `.git`, I don't know anything about nixd so didn't touch it, but maybe that was a typo?


justinmk:
Thanks for looking at this! I'm expecting a lot of "my config doesn't work" after this change, so need to prepare for that and also confirm that @neovim/core are generally aligned with this change (see https://github.com/neovim/nvim-lspconfig/issues/3728 for motivation).

Will make a decision on this in another week or 2.

cbackas:
> Thanks for looking at this! I'm expecting a lot of "my config doesn't work" after this change, so need to prepare for that and also confirm that @neovim/core are generally aligned with this change (see #3728 for motivation).
> 
> Will make a decision on this in another week or 2.

Sounds good, hope you don't think I'm trying to nag you with the merges from master, I'm just dogfooding my fork right now and want the new LSPs that are being merged :)

justinmk:
> I'm just dogfooding my fork right now and want the new LSPs that are being merged :)

Perfect, that's the right way to do it :)

cbackas:
@justinmk seems like option 2 was at least the winner to make it into nvim main, should I redo this PR and instead replace all the `.git`s with `{ '.git` }` or would it be premature since root_marker priority wont be actually released for while?

TheRealLorenz:
> @justinmk seems like option 2 was at least the winner to make it into nvim main, should I redo this PR and instead replace all the `.git`s with `{ '.git' }` or would it be premature since root_marker priority wont be actually released for while?
> 
> 
> 
> @TheRealLorenz if you're already planning to blow through all the configs and add priorities maybe I should close this? but I can help if you'd like

This shouldn't be needed as the default behaviour is directory-up first for each pattern. If `.git` is the last item in `root_markers` it will be evaluated only if other markers aren't found in every directory going up to root.

cbackas:
> This shouldn't be needed as the default behaviour is directory-up first for each pattern. If `.git` is the last item in `root_markers` it will be evaluated only if other markers aren't found in every directory going up to root.

Awesome I'll rebuild nvim and see if things are workin how I need them with main branch of lspconfig and then close this

cbackas:
Ok yeah then the priority list didn't change anything for this problem, they both still attach...

> but after some more thought, this is more like a problem with those insane configs. if they don't want to be enabled under specific conditions, those configs should figure out when they should / should not be enabled.
>
> after all, `root_markers` is simply a way to _find the top-level directory_. whether a config enables itself is decided by `filetypes`, and/or `root()` in complex cases. 

My understanding was that `workspace_requires` would enable these "insane configs" without having extra complexity other than the git root marker needing to not exist. I guess for now if we can't/won't axe git root markers here, in my config i'll just have to implement my own little `root_dir` function `vim.lsp.config()` for these 2 LSPs, not the end of the world. 

Although yeah I'd love to see a world where the config merging actually merges lists as well as tables haha but Lewis's callback pitch seems... flexible.... lol I'll watch for how that shakes out

justinmk:
> . I guess for now if we can't/won't axe git root markers here, in my config i'll just have to implement my own little `root_dir` function in the `vim.lsp.config()` calls for these 2 LSPs, not the end of the world.

we could axe it for those specific configs, or define `root_dir()` or something for those specific configs. there are 3-4 of those typescript-related configs that I haven't mentally digested, need someone to look into it.

PR #3731, opened 2025-04-15
feat(eslint): add vim.lsp.config support

crnvl96:
This PR crosses `eslint` from #3705

Compared locally against the current implementation from `lua/lspconfig/configs/eslint.lua` ans seems ok, but I'm open for more specific checks or any kind of improvement you may have.

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -1,85 +1,35 @@
+--- @brief
+---
+--- https://github.com/hrsh7th/vscode-langservers-extracted
+---
+--- `vscode-eslint-language-server` is a linting engine for JavaScript / Typescript.
+--- It can be installed via `npm`:
+---
+--- ```sh
+--- npm i -g vscode-langservers-extracted
+--- ```
+---
+--- `vscode-eslint-language-server` provides an `EslintFixAll` command that can be used to format a document on save:
+--- ```lua
+--- vim.lsp.config('eslint', {
+---   --- ...
+---   on_attach = function(client, bufnr)
+---     vim.api.nvim_create_autocmd("BufWritePre", {
+---       buffer = bufnr,
+---       command = "EslintFixAll",
+---     })
+---   end,
+--- })
+--- ```
+---
+--- See [vscode-eslint](https://github.com/microsoft/vscode-eslint/blob/55871979d7af184bf09af491b6ea35ebd56822cf/server/src/eslintServer.ts#L216-L229) for configuration options.
+---
+--- Messages handled in lspconfig: `eslint/openDoc`, `eslint/confirmESLintExecution`, `eslint/probeFailed`, `eslint/noLibrary`
+---
+--- Additional messages you can handle: `eslint/noConfig`
+
 local util = require 'lspconfig.util'
 local lsp = vim.lsp
-
--- @brief
---
--- https://github.com/hrsh7th/vscode-langservers-extracted
---
--- `vscode-eslint-language-server` is a linting engine for JavaScript / Typescript.
--- It can be installed via `npm`:
---
--- ```sh
--- npm i -g vscode-langservers-extracted
--- ```
---
--- `vscode-eslint-language-server` provides an `EslintFixAll` command that can be used to format a document on save:
--- ```lua
--- vim.lsp.config('eslint', {
---   --- ...
---   on_attach = function(client, bufnr)
---     vim.api.nvim_create_autocmd("BufWritePre", {
---       buffer = bufnr,
---       command = "EslintFixAll",
---     })
---   end,
--- })
--- ```
---
--- See [vscode-eslint](https://github.com/microsoft/vscode-eslint/blob/55871979d7af184bf09af491b6ea35ebd56822cf/server/src/eslintServer.ts#L216-L229) for configuration options.
---
--- Messages handled in lspconfig: `eslint/openDoc`, `eslint/confirmESLintExecution`, `eslint/probeFailed`, `eslint/noLibrary`
---
--- Additional messages you can handle: `eslint/noConfig`
-
-local function fix_all(opts)
-  opts = opts or {}
-
-  local bufnr = opts.bufnr or vim.api.nvim_get_current_buf()
-  vim.validate('bufnr', bufnr, 'number')
-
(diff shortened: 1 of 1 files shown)


lsp/eslint.lua
```diff
@@ -0,0 +1,202 @@
+-- https://github.com/hrsh7th/vscode-langservers-extracted
```
justinmk, line 1: missing `@brief`. look at other configs

lsp/eslint.lua
```diff
@@ -0,0 +1,202 @@
...
+    'svelte',
+    'astro',
+  },
+  on_init = function(client)
+    vim.api.nvim_create_user_command('EslintFixAll', function()
+      fix_all({ client = client, sync = true })
+    end, {})
+  end,
+  -- https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats
+  root_dir = function(bufnr, on_dir)
```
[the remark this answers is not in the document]
crnvl96: My only concern with that change is that, although deprecated, it's possible to define an eslint config [inside a package.json file](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated#configuration-file-formats). 

This is handled dynamically by lspconfig, which checks for [specific patterns](https://github.com/crnvl96/nvim-lspconfig/blob/17cd6119f78ffaa2324b767a6f5740af22161e07/lua/lspconfig/util.lua#L126-L141) in the user's package.json file. Since the property `root_markers` only accepts a [list of strings](https://neovim.io/doc/user/lsp.html#vim.lsp.Config), we would lose the ability to support package.json-based ESLint configs if migrating from `root_dir`.

The current `root_dir` implementation allows for this more complex pattern matching, while `root_markers` would be limited to exact file names.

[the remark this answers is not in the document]
dorage: I didn't think of that. Thank you.

marks0mmers:
Thanks for the work on this, I was just about to begin my attempt at this change, and then saw yours!

lsp/eslint.lua
```diff
@@ -0,0 +1,204 @@
...
+-- See [vscode-eslint](https://github.com/microsoft/vscode-eslint/blob/55871979d7af184bf09af491b6ea35ebd56822cf/server/src/eslintServer.ts#L216-L229) for configuration options.
+--
+-- Messages handled in lspconfig: `eslint/openDoc`, `eslint/confirmESLintExecution`, `eslint/probeFailed`, `eslint/noLibrary`
+--
+-- Additional messages you can handle: `eslint/noConfig`
+
+local function fix_all(opts)
+  opts = opts or {}
+
+  local bufnr = util.validate_bufnr(0)
```
seblyng, line 37:
```suggestion
  validate('bufnr', bufnr, 'number')
```

crnvl96: The current code uses `util.validate_bufnr` intentionally. This function acts as a compatibility layer, ensuring buffer number validation works correctly even on Neovim v0.10, which `nvim-lspconfig` still supports.

The proposed `vim.validate` function was introduced later and is not available or might behave differently in v0.10. Therefore, sticking with `util.validate_bufnr` is necessary for maintaining backward compatibility.

seblyng: vim.lsp.config was introduced in 0.11, meaning that this code should only run on 0.11, so it's a perfectly fine change

crnvl96: Sometimes, the obvious is so clear that it becomes hard to see :sweat_smile:

My bad, and thank you

seblyng (approved):

lsp/eslint.lua
```diff
@@ -0,0 +1,206 @@
...
+    request = function(buf, method, params)
+      client:request_sync(method, params, nil, buf)
+    end
+  else
+    request = function(buf, method, params)
+      client:request(method, params, nil, buf)
+    end
+  end
+
+  request(bufnr, 'workspace/executeCommand', {
```
justinmk, line 58: use client:exec_cmd ?

lsp/eslint.lua
```diff
@@ -0,0 +1,206 @@
...
+
+  local client = opts.client or vim.lsp.get_clients({ bufnr = bufnr, name = 'eslint' })[1]
+
+  if not client then
+    return
+  end
+
+  local request
+
+  if opts.sync then
```
justinmk, line 48: what is the point of this option? you only call it one way, below. and client:exec_cmd should be used anyway

lsp/eslint.lua
```diff
@@ -0,0 +1,206 @@
...
+  opts = opts or {}
+
+  local bufnr = opts.bufnr or vim.api.nvim_get_current_buf()
+  vim.validate('bufnr', bufnr, 'number')
+
+  local client = opts.client or vim.lsp.get_clients({ bufnr = bufnr, name = 'eslint' })[1]
+
+  if not client then
+    return
+  end
```
justinmk, line 44: if the client doesn't exist why would the command be called? the command is defined on attach. and it's already passed by your call below, so get_clients isn't needed.

```suggestion
  local client = assert(opts.client)
```

seblyng: I think the intention was to just copy it without changing too much from the old config, but I agree.

This entire method could essentially just be inlined with 1-2 lines or so where the command is created

crnvl96:
Totally agree with your suggestions, they should be fixed now!

Actually, my goal was mainly to port the old config with as few changes as possible.

seblyng (approved):

lsp/eslint.lua
```diff
@@ -0,0 +1,182 @@
+local util = require 'lspconfig.util'
+local lsp = vim.lsp
+
+-- @brief
```
justinmk, line 4: move this to the top and use `---` (triple hyphen) for the entire comment

lsp/eslint.lua
```diff
@@ -0,0 +1,182 @@
...
+    'javascript.jsx',
+    'typescript',
+    'typescriptreact',
+    'typescript.tsx',
+    'vue',
+    'svelte',
+    'astro',
+  },
+  on_init = function(client)
+    vim.api.nvim_create_user_command('EslintFixAll', function()
```
justinmk, line 48: commands must be prefixed with "Lsp"
```suggestion
    vim.api.nvim_create_user_command('LspEslintFixAll', function()
```

PR #3734, opened 2025-04-16
Feature: migrate builtin commands

### LINKED ISSUE #2216: LspStop should remove related autocmds

### LINKED ISSUE #3714: migrate :Lsp commands to vim.lsp.config
## Problem

The `:Lsp*` commands aren't compatible with vim.lsp.config, e.g. `:LspStart`:

https://github.com/neovim/nvim-lspconfig/blob/9e6f0534c49e9d67982d115591977f6e8fcce1e7/plugin/lspconfig.lua#L70-L73

cc @TheRealLorenz 

## Proposal

- on Nvim 0.11, commands like `:LspStart` should call vim.lsp.config, never the legacy lspconfig api.
- on Nvim 0.11, the legacy lspconfig api should always throw an `error('nvim-lspconfig 2.x only supports vim.lsp.config. If you dont want to migrate, pin to nvim-lspconfig 1.x')`

TheRealLorenz:
fix #3714
fix #2216

Things to note:

- `LspStart` starts the selected LSP and attaches to every buffer which has the right filetype. Should I only attach it to the current buffer?

- `LspStop` stops the selected LSP.

- `LspRestart` restart the selected LSPs, but I don't get how and why it used a timer.

- Should we add `LspEnable`? Maybe in a future PR.

EDIT:

Should I add checks for neovim 0.11?

Something like:
```lua
  if not has_eleven then
    error('nvim-lspconfig 2.x only supports vim.lsp.config. If you dont want to migrate, pin to nvim-lspconfig 1.x')
  end
```

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -67,6 +67,45 @@ end
 -- script in scriptnames to be executed is lspconfig.
 api.nvim_create_user_command('LspInfo', ':checkhealth vim.lsp', { desc = 'Alias to `:checkhealth vim.lsp`' })
 
+if vim.version.ge(vim.version(), { 0, 11, 2 }) then
+  local complete_client = function(arg)
+    return vim
+      .iter(vim.lsp.get_clients())
+      :map(function(client)
+        return client.name
+      end)
+      :filter(function(name)
+        return name:sub(1, #arg) == arg
+      end)
+      :totable()
+  end
+
+  api.nvim_create_user_command('LspRestart', function(info)
+    for _, name in ipairs(info.fargs) do
+      if vim.lsp.config[name] == nil then
+        vim.notify(("Invalid server name '%s'"):format(info.args))
+      else
+        vim.lsp.enable(name, false)
+      end
+    end
+
+    local timer = assert(vim.uv.new_timer())
+    timer:start(500, 0, function()
+      for _, name in ipairs(info.fargs) do
+        vim.schedule_wrap(function(x)
+          vim.lsp.enable(x)
+        end)(name)
+      end
+    end)
+  end, {
+    desc = 'Restart the given client(s)',
+    nargs = '+',
+    complete = complete_client,
+  })
+
+  return
+end
+
 api.nvim_create_user_command('LspStart', function(info)
   local server_name = string.len(info.args) > 0 and info.args or nil
   if server_name then
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -80,6 +80,29 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
       :totable()
   end
 
+  local complete_config = function(arg)
+    return vim
+      .iter(vim.api.nvim_get_runtime_file(('lsp/%s*.lua'):format(arg), true))
+      :map(function(path)
+        local file_name = path:match('[^/]*.lua$')
+        return file_name:sub(0, #file_name - 4)
+      end)
+      :totable()
+  end
+
+  api.nvim_create_user_command('LspStart', function(info)
+    if vim.lsp.config[info.args] == nil then
+      vim.notify(("Invalid server name '%s'"):format(info.args))
+      return
+    end
+
+    vim.lsp.enable(info.args)
+  end, {
+    desc = 'Enable and launch a language server',
+    nargs = '?',
+    complete = complete_config,
+  })
+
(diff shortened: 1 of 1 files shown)


TheRealLorenz:
> I don't think `LspStart` and `LspRestart` will work when `root_dir` is a function with this. I think you must do something like this:
> 
> ```lua
> if type(config.root_dir) == 'function' then
>   ---@param root_dir string
>   config.root_dir(bufnr, function(root_dir)
>     config.root_dir = root_dir
>     vim.schedule(function()
>       vim.lsp.start(config, {
>         bufnr = bufnr,
>         reuse_client = config.reuse_client,
>         _root_markers = config.root_markers,
>       })
>     end)
>   end)
> else
>   vim.lsp.start(config, {
>     bufnr = bufnr,
>     reuse_client = config.reuse_client,
>     _root_markers = config.root_markers,
>   })
> end
> ```
> 
> Inspiration taken from neovim source code, in `lsp_enable_callback` function in `vim/lsp.lua`

Thank you, I totally missed that. The `vim.lsp` API is a little but cumbersome as for now.

TheRealLorenz:
Opened a PR at https://github.com/neovim/neovim/pull/33578 to make `:LspStart` less ugly.

plugin/lspconfig.lua
```diff
@@ -3,129 +3,129 @@ if vim.g.lspconfig ~= nil then
...
 end, {
   desc = 'Manually restart the given language client(s)',
-  nargs = '?',
-  complete = lsp_get_active_clients,
+  nargs = '*',
+  complete = complete_client,
 })
 
-api.nvim_create_user_command('LspStop', function(info)
+vim.api.nvim_create_user_command('LspStop', function(info)
```
justinmk, line 128: Nvim 0.11+, `LspStop` can just do this?:

    vim.lsp.enable(..., false)

TheRealLorenz: Yes, I'll add it. Should I re-enable the config in `LspStart`?

TheRealLorenz:
@justinmk On further thinking, probably these commands do not suit `vim.lsp` quite well (at least semantically).

We could use `LspEnable` (with optional ++now), `LspDisable`, `LspStart` and `LspStop` "systemctl-like"

TheRealLorenz:
What's the current situation of this after https://github.com/neovim/neovim/pull/33702? @justinmk 

It should simplify logic as:

`:LspStart xxx` as `vim.lsp.enable(xxx)`
`:LspStop xxx` as `vim.lsp.enable(xxx, false)`
`:LspRestart xxx` as `vim.lsp.enable(xxx, false); vim.lsp.enable(xxx)`

Probably it would be useful to rename the commands when (and if) they're upstreamed to neovim core.

justinmk:
> It should simplify logic as:

yes, exactly 👍 



> rename the commands when (and if) they're upstreamed to neovim core.

yeah, most likely the upstreamed form will be `:Lsp <subcommand>`

justinmk:
It probably makes sense to add a `if vim.fn.has('nvim-0.11.2') == 1 then` section at the top of this file, which defines the commands for Nvim 0.11.2+, and then `return` early. Then the old code can just stay as it is.

TheRealLorenz:
Is there a way to provide completion for `:LspStart` ? I didn't find one as the configs are lazily resolved on request if I'm correct.

justinmk (approved):
looks like a good starting point, we can improve it if needed

justinmk:
> provide completion for `:LspStart` ? I didn't find one as the configs are lazily resolved on request if I'm correct.

oh, we could copy the logic from here: https://github.com/neovim/neovim/blob/97a6259442526e33a73201557b1cb74ccdb64eef/runtime/lua/vim/lsp.lua#L443-L454

TheRealLorenz:
> > provide completion for `:LspStart` ? I didn't find one as the configs are lazily resolved on request if I'm correct.
> 
> oh, we could copy the logic from here: https://github.com/neovim/neovim/blob/97a6259442526e33a73201557b1cb74ccdb64eef/runtime/lua/vim/lsp.lua#L443-L454

That would be fine for getting configs under `lsp/` but wouldn't cover configs defined by the user in his dotfiles as `vim.lsp.config('foo', {})`, am I right?

justinmk:
> wouldn't cover configs defined by the user in his dotfiles as `vim.lsp.config('foo', {})`, am I right?

right, but we can think about that later.

PR #3737, opened 2025-04-16
feat(volar): add vim.lsp.config support

seblyng:
This PR crosses volar from https://github.com/neovim/nvim-lspconfig/issues/3705

Seems to work for me on one of my projects at least


lsp/volar.lua
```diff
@@ -0,0 +1,97 @@
+local function get_typescript_server_path(root_dir)
+  local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
+  return project_root and (project_root .. '/node_modules/typescript/lib') or ''
```
glepnir, line 3: please use vim.fs.joninpath

seblyng: Woops didn't notice. Fixing

seblyng: Fixed

PR #3738, opened 2025-04-16
feat(astroJS): add vim.lsp.config support

Grahf0085:
This PR crosses astroJS from https://github.com/neovim/nvim-lspconfig/issues/3705

Someone else testing on their machine would be appreciated :))


cbackas:
works for me

PR #3741, opened 2025-04-17
fix(astro): using vim.fs.jointhpath concat path

glepnir:
Problem: hardcord path symbol in get_typescript_server_path

Solution: using vim.fs.joinpath to concat path

### DIFF
--- a/lsp/astro.lua
+++ b/lsp/astro.lua
@@ -1,6 +1,6 @@
 local function get_typescript_server_path(root_dir)
   local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
-  return project_root and (project_root .. '/node_modules/typescript/lib') or ''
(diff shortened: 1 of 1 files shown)


lsp/astro.lua
```diff
@@ -1,6 +1,6 @@
 local function get_typescript_server_path(root_dir)
   local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
-  return project_root and (project_root .. '/node_modules/typescript/lib') or ''
+  return project_root and vim.fs.joinpath(project_root, 'node_modules', 'typescript', 'lib') or ''
```
justinmk, line 3: string literals don't need to be joined. just use "/".

     vim.fs.joinpath(project_root, 'node_modules/typescript/lib')

glepnir: right it gsub path separator according to the system.

justinmk: "/" works everywhere. platform-specific slash isn't needed.

PR #3746, opened 2025-04-18
feat: angularls

BBboy01:
This PR crosses angularls from #3705

mainly same with `lua/lspconfig/configs/angularls.lua`

- removed`on_new_config` as I test locally it doesn't matter
- changed default probe paths like how angular vscode extension is (https://github.com/angular/vscode-ng-language-service/blob/main/client/src/client.ts#L304)


lsp/angularls.lua
```diff
@@ -0,0 +1,87 @@
...
+-- angularls will get module by `require.resolve(PROBE_PATH, MODULE_NAME)` of nodejs
+local ts_probe_dirs = vim.iter({ extension_path, default_probe_dir }):join(',')
+local ng_probe_dirs = vim
+  .iter({ extension_path, default_probe_dir })
+  :map(function(p)
+    return vim.fs.joinpath(p, '/@angular/language-server/node_modules')
+  end)
+  :join(',')
+
+---@brief
```
justinmk, line 55: should be at top of config

lsp/angularls.lua
```diff
@@ -0,0 +1,87 @@
...
+
+  local angular_core_version = json.dependencies['@angular/core']
+
+  angular_core_version = angular_core_version and angular_core_version:match('%d+%.%d+%.%d+')
+
+  return angular_core_version
+end
+
+local default_probe_dir = get_probe_dir()
+local default_angular_core_version = get_angular_core_version()
```
justinmk, line 34: evaluating this here (and getcwd() above) means that if the user changes the CWD later, that won't be used by the config.

It seems like `cwd` should be defined as a function for this config, instead of a list. That makes it more "dynamic".

The old config https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/angularls.lua had this issue too, though, so I guess this can be revisited later...

justinmk:
This config has problems which broke the doc generator. I mitigated the issue in https://github.com/neovim/nvim-lspconfig/pull/3747/commits/e8963b5b0b48324b2da4e5cc0db8447dca3bd659 , but the logic needs attention...

PR #3648, opened 2025-03-18
fix(tailwindcss): improved tailwindcss detection

maximtrp:
Currently, the Tailwind CSS language server is enabled for any workspace containing a `.git` folder, a `node_modules` folder, or a `package.json` file, which is not ideal. This PR introduces a specific check for the `tailwindcss` package within `package.json` file using `insert_package_json` method from `util` (this approach is used in `eslint` config). Since the latest version of Tailwind CSS (v4) no longer requires `postcss.config.*` and `tailwind.config.*` files, these can no longer be relied upon as sole criteria for enabling Tailwind CSS LSP.

PR #3751, opened 2025-04-19
feat(tailwind): add `vim.lsp.config` support

r4zendev:
Should remove `tailwindcss` from the list in #3705

Tested locally, modifies server settings the same way as it did in pre-0.11 setup, LSP works as expected.

### DIFF
--- a/lsp/tailwindcss.lua
+++ b/lsp/tailwindcss.lua
@@ -1,0 +1,123 @@
+---@brief
+--- https://github.com/tailwindlabs/tailwindcss-intellisense
+---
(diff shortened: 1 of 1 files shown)


maximtrp:
The latest version of Tailwind does not require `tailwind.config.*` or `postcss.config.*` files. So these markers are not reliable (see my PR #3648). I have tested this config on one of my recent projects based on Tailwind v4 and it does not work. We'd better check `package.json` for `tailwindcss` package.

justinmk:
SGTM, pr welcome

PR #3753, opened 2025-04-21
feat(roslyn_ls): add initial config as `vim.lsp.config`

616b2f:


lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+        ["workspace/projectInitializationComplete"] = function(_, _, ctx)
+            vim.notify("Roslyn project initialization complete", vim.log.levels.INFO, { title = "roslyn.nvim" })
+
+            local buffers = vim.lsp.get_buffers_by_client_id(ctx.client_id)
+            for _, buf in ipairs(buffers) do
+                vim.lsp.util._refresh("textDocument/diagnostic", { bufnr = buf })
+            end
+
+            ---NOTE: This is used by rzls.nvim for init
+            vim.api.nvim_exec_autocmds("User", { pattern = "RoslynInitialized", modeline = false })
```
seblyng, line 45: While it is probably nice for `rzls.nvim`, I don't feel like this should exist in a minimal config for `nvim-lspconfig`

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+        end,
+    }
+end
+
+---@type vim.lsp.ClientConfig
+return {
+  name = "roslyn_ls",
+  offset_encoding = 'utf-8',
+  ---@param dispatchers vim.lsp.rpc.Dispatchers
+  cmd = function (dispatchers)
```
seblyng, line 86: IMO, `cmd` should not be set by default, and should instead be set explicitly by the user. I think an example should be using `--stdio` flag for starting up the server

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+    -- enable inlay hints if LSP server supports it
+    if client.supports_method("textDocument/inlayHint") then
+      vim.lsp.inlay_hint.enable(true, { bufnr=bufnr })
+    end
+
+    -- enable semantic tokens highligting hints
+    if client.supports_method("textDocument/semanticTokens") then
+      client.server_capabilities.semanticTokensProvider = true
+    end
+
```
seblyng, line 113: `nvim-lspconfig` should not set inlay hint or semantic tokens to true/false IMO, so this can just be removed

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+  capabilities = {
+      -- HACK: Enable filewatching to later just not watch any files
+      -- This is to not make the server watch files and make everything super slow in certain situations
+      workspace = {
+          didChangeWatchedFiles = {
+              dynamicRegistration = true,
+              -- enable file watcher capabilities for lsp clients
+              relativePatternSupport = true,
+          },
+      },
```
seblyng, line 155: This should also just be removed IMO. I have some config for turning it off/on in the plugin because filewatching can be very slow. However, I don't think `nvim-lspconfig` should have any of this

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+      -- HACK: Doesn't show any diagnostics if we do not set this to true
+      textDocument = {
+          diagnostic = {
+              dynamicRegistration = true,
+          },
+      },
```
seblyng, line 161: This is a hack, since neovim has this turned off (probably because there is something that is not supported), but this is required for diagnostics to appear. I am not sure if there are any downsides to "hack" this, but I haven't noticed any problems with this

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+
+local function roslyn_handlers()
+  return {
+        ["workspace/projectInitializationComplete"] = function(_, _, ctx)
+            vim.notify("Roslyn project initialization complete", vim.log.levels.INFO, { title = "roslyn.nvim" })
+
+            local buffers = vim.lsp.get_buffers_by_client_id(ctx.client_id)
+            for _, buf in ipairs(buffers) do
+                vim.lsp.util._refresh("textDocument/diagnostic", { bufnr = buf })
+            end
```
seblyng, line 42: This is strictly not required and also sort of a hack, but diagnostics will not appear immediately if we are not refreshing here. I have not investigated why though. I personally don't have anything against this being here

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+            })
+            return vim.NIL
+        end,
+        ["workspace/_roslyn_projectNeedsRestore"] = function(_, result, ctx)
+            local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
+
+            ---@diagnostic disable-next-line: param-type-mismatch
+            client:request("workspace/_roslyn_restore", result, function(err, response)
+                if err then
+                    vim.notify(err.message, vim.log.levels.ERROR, { title = "roslyn.nvim" })
```
seblyng, line 59: The title for `vim.notify` calls should at least be the same for all calls if we should have it

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,195 @@
...
+            local csproj_file = fs.joinpath(path, name)
+            on_init_project(client, {csproj_file})
+          end
+          return match
+        end)
+      end
+
+      assert(root_dir, "no solution and no csproj found")
+
+      client.root_dir = root_dir
```
seblyng, line 143: I am not sure about changing root_dir in `on_attach`. I don't know if it will have any effect

seblyng (commented):
When I forked the `roslyn.nvim` plugin, there was no easy way to start the server reliably because it didn't support --stdio or named pipes from the client. However, now it supports both.

While I will probably still use `roslyn.nvim` because it does a bit more clever root_dir detection (IMO), I think a minimal config could probably be good enough for a lot of people

lsp/roslyn_ls.lua
justinmk: Sounds related to discussion in https://github.com/neovim/neovim/issues/33391

This is fine for now.

lsp/roslyn_ls.lua
justinmk: Should go in `before_init`

seblyng: I think instead we can just use `on_init`, and just search in the current `root_dir` for a sln or a csproj file, because the root_markers should already have found a root_dir. `before_init` is also too early to send out the notify about opening a solution or project that is required. So something like this should work:
```lua
    on_init = {
        function(client)
            local root_dir = client.config.root_dir

            for entry, type in vim.fs.dir(root_dir) do
                if type == "file" and vim.endswith(entry, ".sln") then
                    on_init_sln(client, entry)
                end
            end

            for entry, type in vim.fs.dir(root_dir) do
                if type == "file" and vim.endswith(entry, ".csproj") then
                    on_init_project(client, { entry })
                end
            end
        end
    }
```

Looping twice to prefer a solution if we found both

lsp/roslyn_ls.lua
616b2f: agree, will remove it

lsp/roslyn_ls.lua
616b2f: I do think it would make sense to have a default command and using `--stdio` here make sense, as it's simpler to handle here.

Said that, it's tricky here to define a proper default `cmd`, because there is no common executable name that can be used here, I need to have another look on how the server is distributed, as there are platform specific nuget packages and some which are platform independent (which contain only `.dll` files that have to be called with `dotnet` command).

seblyng: > Said that, it's tricky here to define a proper default cmd, because there is no common executable name that can be used here, I need to have another look on how the server is distributed, as there are platform specific nuget packages and some which are platform independent (which contain only .dll files that have to be called with dotnet command).

Exactly, that is why I don't think it makes sense to have a default cmd value, and instead have docs on an example

lsp/roslyn_ls.lua
616b2f: I have to test this, as the current `on_attach` worked and I had an issue with jump to dependent nuget package sources where I added a workaround, not sure if this would still work with `on_init`, also the referenced issue could be a good addition, it may make sense to wait till it merged.

lsp/roslyn_ls.lua
616b2f: I personally had issues to setup `omnisharp` back then because there was no default `cmd` and having something that works for some people and can be used as an example to write your own `cmd` is IMO better then having only something documented which should only theoretically work, but this is only my very personal opinion.

I just checked and in the platform specific nuget packages there are `Microsoft.CodeAnalysis.LanguageServer.exe` and `Microsoft.CodeAnalysis.LanguageServer` executables in the `/content/LanguageServer` dir. So this could be used instead, so if the user put it in the path it can be picked up properly.

Try to test this also and adjust here after that.

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,161 @@
+--- borrowed some source from: https://github.com/seblj/roslyn.nvim
```
seblyng, line 1: This can be removed in favor of some proper documentation with `---@brief`

lsp/roslyn_ls.lua
seblyng: Yeah I sort of agree with it being cumbersome to setup without a default cmd, but when there is bad support in package managers for this server (I haven't seen a single one have it), it is also annoying having to add things to path

I don't have any big objections to this if others think it should be there. There at least needs to be documentation on how to install it and how to make the default cmd work

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,161 @@
...
+
+local uv = vim.uv
+local fs = vim.fs
+
+---@param client vim.lsp.Client
+---@param target string
+local function on_init_sln(client, target)
+  vim.notify('Initializing: ' .. target, vim.log.levels.INFO, { title = 'roslyn_ls' })
+  ---@diagnostic disable-next-line: param-type-mismatch
+  client:request('solution/open', {
```
seblyng, line 13: You can use `client:notify` and then not pass an empty function as last argument here and for the `project/open`

```lua
client:notify("solution/open", {
    solution = vim.uri_from_fname(solution),
})
```

justinmk:
try rebasing. the lint CI is failing because this pr branch is out of date

lsp/roslyn_ls.lua
616b2f: This is where I hope we can merge https://github.com/williamboman/mason.nvim/pull/1781 in future, I use this currently in combination with this package definition https://github.com/616b2f/mason-registry/blob/main/packages/roslyn/package.yaml.

Other alternative is to compile from source, or download and extract nuget files manually, none of which is convenient IMO.

I will try to write something up in the documentation part

lsp/roslyn_ls.lua
justinmk: If you are trying to conditionally enable the LS, `root_dir` function is the right approach: https://github.com/neovim/neovim/issues/32037#issuecomment-2825599872

Simply don't call the callback if the LS should not be enabled for a buffer.

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,186 @@
+---@brief
+---
+-- https://github.com/dotnet/roslyn
```
justinmk, line 3:
```suggestion
--- https://github.com/dotnet/roslyn
```

for all lines in this comment

justinmk (approved):

lsp/roslyn_ls.lua
```diff
@@ -0,0 +1,186 @@
...
+--
+-- To install the server, compile from source or download as nuget package.
+-- Go to `https://dev.azure.com/azure-public/vside/_artifacts/feed/vs-impl/NuGet/Microsoft.CodeAnalysis.LanguageServer.<platform>/overview`
+-- replace `<platform>` with one of the following `linux-x64`, `osx-x64`, `win-x64`, `neutral` (for more info on the download location see https://github.com/dotnet/roslyn/issues/71474#issuecomment-2177303207).
+-- Download and extract it (nuget's are zip files).
+-- - if you shose `neutral` nuget version, then you have to change the `cmd` like so:
```
seblyng, line 9:
```suggestion
-- - if you chose `neutral` nuget version, then you have to change the `cmd` like so:
```

seblyng (approved):
There are of course things to make solution resolution better, but this looks good to me as a minimal config that fits this repo

616b2f:
> There are of course things to make solution resolution better, but this looks good to me as a minimal config that fits this repo

Yes sure, we can also improve in future when we encounter actual issues or have some meaningful improvements regarding speed or reliability.

PR #3754, opened 2025-04-21
ci(lint): enforce "Lsp" command name prefix

### LINKED ISSUE #3706: CI: lint enforces "Lsp" prefix on commands
### Description

All config-defined commands should be prefixed with `Lsp`.

justinmk:
- enforce "Lsp" command name prefix
- enforce buffer-local command definition

@crnvl96 please double-check that I didn't break the config :)

fix #3706

### DIFF
--- a/.github/ci/run_sanitizer.sh
+++ b/.github/ci/run_sanitizer.sh
@@ -1,24 +1,53 @@
 #!/usr/bin/env bash
+
+# USAGE: To run locally:
+#   bash .github/ci/run_sanitizer.sh origin/master HEAD
+
 set -e
 
 REF_BRANCH="$1"
 PR_BRANCH="$2"
 
-# checks for added lines that contain search pattern and prints them
-SEARCH_PATTERN='(path\.dirname|fn\.cwd)'
+# Enforce buffer-local commands.
+_check_cmd_buflocal() {
+  if git grep -P 'nvim_create_user_command' -- 'lsp/*.lua' ; then
+    echo
+    echo 'Define commands with nvim_buf_create_user_command (buffer-local), not nvim_create_user_command'
+    exit 1
+  fi
+}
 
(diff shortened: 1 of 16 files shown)


crnvl96:
- Should we use the `bufnr` buitin param that `on_attach` receives by default?

```lua
    vim.api.nvim_buf_create_user_command(bufnr, 'LspEslintFixAll', function()
      client:exec_cmd({
```

- Since the function properties (such as `on_init` and `on_attach`) are overwritten when the user provides a custom configuration, a small section on the help docs explaining how to deal with such situation would be useful.

As a quick illustration of what I mean:

*default config of eslint*:

```lua
 on_attach = function(client)
    vim.api.nvim_buf_create_user_command(0, 'LspEslintFixAll', function()
      local bufnr = vim.api.nvim_get_current_buf()

      client:exec_cmd({
        title = 'Fix all Eslint errors for current buffer',
        command = 'eslint.applyAllFixes',
        arguments = {
          {
            uri = vim.uri_from_bufnr(bufnr),
            version = lsp.util.buf_versions[bufnr],
          },
        },
      }, { bufnr = bufnr })
    end, {})
  end,
```

If the user want to setup some custom configuration, like this:

```lua
on_attach = function(client)
        buffer = bufnr,
        command = 'LspEslintFixAll',
    })
 end,
```

The execution of `LspEslintFixAll` would fail, because the original command itself would be overwritten.

The solution I've came up withis the following:

```lua
local eslint_on_attach = vim.lsp.config.eslint.on_attach

vim.lsp.config('eslint', {
    cmd = { 'vscode-eslint-language-server', '--stdio' },
        eslint_on_attach(client, bufnr)

            buffer = bufnr,
            command = 'LspEslintFixAll',
        })
    end,
})
```

And my idea would be adding an example just like this on the docs.

What do you think, @justinmk? If this is reasonable, I can work on a PR soon

PR #3757, opened 2025-04-22
feat(mdx_analyzer): add vim.lsp.config support

MuntasirSZN:
Should remove from #3705

Tested with latest stable.


lsp/mdx_analyzer.lua
```diff
@@ -0,0 +1,25 @@
...
+---@brief
+---https://github.com/mdx-js/mdx-analyzer
+---
+---`mdx-analyzer`, a language server for MDX
+
+return {
+  cmd = { 'mdx-language-server', '--stdio' },
+  filetypes = { 'mdx' },
+  root_markers = { 'package.json' },
+  single_file_support = true,
```
justinmk, line 15: this is not a valid field in vim.lsp.config. should be `workspace_required = false` (requires Nvim 0.11.1+)

lsp/mdx_analyzer.lua
```diff
@@ -0,0 +1,25 @@
+local function get_typescript_server_path(root_dir)
+  local project_root = vim.fs.dirname(vim.fs.find('node_modules', { path = root_dir, upward = true })[1])
+  return project_root and vim.fs.joinpath(project_root, 'node_modules', 'typescript', 'lib') or ''
+end
+
+---@brief
```
justinmk, line 6: should be a top of file

PR #3770, opened 2025-04-24
fix: add pylyzer dependency info (erg) (#3433)

fishBone000:
Pylyzer requires Erg as dependency. This info is missing in default config and doc.

### DIFF
--- a/lsp/pylyzer.lua
+++ b/lsp/pylyzer.lua
@@ -5,8 +5,7 @@
 --- `pylyzer`, a fast static code analyzer & language server for Python.
 ---
 --- `pylyzer` requires Erg as dependency, and finds it via `ERG_PATH` environment variable.
---- In the default config `ERG_PATH` is set to `~/.erg`, as shown in the `on_new_config` function.
---- Add a `on_new_config` on your own if you want to change it.
+--- This config sets `ERG_PATH="~/.erg"`. Set `cmd_env` if you want to change it.
 --- To install Erg, simply extract tarball/zip from [Erg releases](https://github.com/erg-lang/erg/releases/latest)
 --- to the the path where you want to install it, e.g. `~/.erg`.
 return {
--- a/lsp/pylyzer.lua
+++ b/lsp/pylyzer.lua
@@ -3,6 +3,12 @@
 --- https://github.com/mtshiba/pylyzer
 ---
 --- `pylyzer`, a fast static code analyzer & language server for Python.
+---
+--- `pylyzer` requires Erg as dependency, and finds it via `ERG_PATH` environment variable.
+--- In the default config `ERG_PATH` is set to `~/.erg`, as shown in the `on_new_config` function.
+--- Add a `on_new_config` on your own if you want to change it.
+--- To install Erg, simply extract tarball/zip from [Erg releases](https://github.com/erg-lang/erg/releases/latest)
+--- to the the path where you want to install it, e.g. `~/.erg`.
 return {
   cmd = { 'pylyzer', '--server' },
   filetypes = { 'python' },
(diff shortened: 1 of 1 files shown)


fishBone000:
Opened this PR as a fixer to #3441 (update the file in `/lsp/` instead of `/lua/lspconfig/...`)

lsp/pylyzer.lua
```diff
@@ -22,4 +28,7 @@ return {
       checkOnType = false,
     },
   },
+  on_new_config = function(new_config, _)
```
justinmk, line 31: no such thing in vim.lsp.config. try `before_init` ? and please test that it actually works.

lsp/pylyzer.lua
```diff
@@ -22,4 +28,7 @@ return {
...
+    new_config.cmd_env.ERG_PATH = vim.fs.joinpath(vim.uv.os_homedir(), '/.erg')
```
justinmk, line 32: don't need the slash if you are using joinpath()
```suggestion
    new_config.cmd_env.ERG_PATH = vim.fs.joinpath(vim.uv.os_homedir(), '.erg')
```

lsp/pylyzer.lua
justinmk: Sure, though that means the CWD "sticks" and won't dynamically update. I don't use this config so I have no strong opinion :)

lithammer: I mean it wasn't really doing anything dynamic anyway. It was essentially just setting `ERG_PATH=~/.erg`. Unless I'm missing something...?

And `before_init` happens after starting the binary. So setting `cmd_env` wouldn't have any effect?

justinmk: > `before_init` happens after starting the binary

oh. Then `cmd` should be defined as a function (again, assuming this functionality is important for users of this config)

lithammer: > > `before_init` happens after starting the binary
> 
> 
> 
> oh. Then `cmd` should be defined as a function (again, assuming this functionality is important for users of this config)

I'm not sure about this though. Just going by the docs:

> Callback invoked before the LSP "initialize" phase, where params contains the parameters being sent to the server and config is the config that **was** passed to vim.lsp.start(). You can use this to modify parameters before they are sent.

Seems to indicate that `vim.lsp.start()` has already been called.

lithammer: With that said. I suppose `cmd` being a function could be useful when you want to start a server relative to the resolved root.

For example you might want to call `./node_modules/.bin/biome` relative to that root.

Though I suppose that could also be solved by having relative paths be relative to workspace root instead of cwd (though relative paths aren't supported at all at the moment I think?). 

Or do the VS Code way (or CoC, don't remember which). By using some magic marker like `<root_dir>/node_modules/.bin/biome`.

fishBone000:
Fixed, tested and it works.

Not sure what's about the `CWD` you were talking about though? I kind of get the idea but can't see how it's related to this `pylyzer` config?

lsp/pylyzer.lua
```diff
@@ -3,6 +3,12 @@
 --- https://github.com/mtshiba/pylyzer
 ---
 --- `pylyzer`, a fast static code analyzer & language server for Python.
+---
+--- `pylyzer` requires Erg as dependency, and finds it via `ERG_PATH` environment variable.
+--- In the default config `ERG_PATH` is set to `~/.erg`, as shown in the `on_new_config` function.
+--- Add a `on_new_config` on your own if you want to change it.
```
justinmk, line 9:
```suggestion
--- This config sets `ERG_PATH="~/.erg"`. Set `cmd_env` if you want to change it.
```

lithammer:
> Not sure what's about the `CWD` you were talking about though? I kind of get the idea but can't see how it's related to this `pylyzer` config?

Yeah sorry. I digressed.

PR #3773, opened 2025-04-24
feat: Add gnls config

nickdiego:
Add support for Microsoft's GN language server [1].

[1] https://github.com/microsoft/gnls

### DIFF
--- a/lsp/gnls.lua
+++ b/lsp/gnls.lua
@@ -1,0 +1,19 @@
+---@brief
+---
+--- https://github.com/microsoft/gnls
+---
+--- Microsoft's language server for GN build files.
+---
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/gnls.lua
```diff
@@ -0,0 +1,25 @@
+local util = require 'lspconfig.util'
```
justinmk, line 1: not accepting new configs in lspconfig/configs/.

lsp/gnls.lua
```diff
@@ -0,0 +1,5 @@
+return {
```
justinmk, line 1: missing doc. look at other configs.

nickdiego: I'm still missing something. Let me have a look. sorry for the noise.

PR #3775, opened 2025-04-25
fix(mesonlsp): improved root detection

SpikedPaladin:
This commit enhances root directory detection of mesonlsp language server.
Current behavior:
`root` is set on first meson file user is opened
New behavior:
`root` is set to parent of `meson.build` with `project` function call

### DIFF
--- a/lsp/mesonlsp.lua
+++ b/lsp/mesonlsp.lua
@@ -3,8 +3,35 @@
 --- https://github.com/JCWasmx86/mesonlsp
 ---
 --- An unofficial, unendorsed language server for meson written in C++
+
+---Checks if a given path contains a valid Meson project root file
+---Looks for 'meson.build' file which contains 'project()' function
+local meson_matcher = function(_, path)
+  local pattern = 'meson.build'
+  local f = vim.fn.glob(table.concat({ path, pattern }, '/'))
+  if f == '' then
+    return false
+  end
+  for line in io.lines(f) do
+    -- skip meson comments
+    if not line:match '^%s*#.*' then
+      local str = line:gsub('%s+', '')
+      if str ~= '' then
+        if str:match '^project%(' then
+          return true
+        else
+          break
+        end
+      end
+    end
+  end
+  return false
+end
+
 return {
   cmd = { 'mesonlsp', '--lsp' },
   filetypes = { 'meson' },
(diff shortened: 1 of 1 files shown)


lua/lspconfig/configs/mesonlsp.lua
```diff
@@ -1,10 +1,32 @@
-local util = require 'lspconfig.util'
+local meson_matcher = function(path)
```
justinmk, line 1: not accepting changes to old configs

lsp/mesonlsp.lua
```diff
@@ -3,8 +3,33 @@
 --- https://github.com/JCWasmx86/mesonlsp
 ---
 --- An unofficial, unendorsed language server for meson written in C++
+
+local meson_matcher = function(_, path)
```
justinmk, line 7: docstring should give an example of the line this is trying to find. shouldn't have to study the code to figure that out.

SpikedPaladin: I don't understand what you mean, can you explain in more detail please?

justinmk:
```suggestion
--- Decides if a file is a meson file by looking for lines like:
--- "foo.bar" or "baz"
local meson_matcher = function(_, path)
```

replace "foo.bar" with an actual example

justinmk: what is an example of text content that this looks for ?

SpikedPaladin: Done

lsp/mesonlsp.lua
```diff
@@ -3,8 +3,35 @@
...
+  end
+  for line in io.lines(f) do
+    -- skip meson comments
+    if not line:match '^%s*#.*' then
+      local str = line:gsub('%s+', '')
+      if str ~= '' then
+        if str:match '^project%(' then
+          return true
+        else
+          break
```
justinmk, line 23: `vim.filetype.match()` exists for this purpose. However,

    vim.filetype.match({ filename = f, contents = 'project' })

seems to return "meson" even if the file doesn't contain "project", which isn't what you want here.

SpikedPaladin: For mesonlsp root_dir is required to be at folder with 'project' file. If root will be in different place all file will be underlined with red. I just take code from vala_ls which finds project root perfectly.

PR #3780, opened 2025-04-26
feat(ts_ls): provide command to run source actions

thatsmydoing:
Source actions are code actions that apply to the whole file. They are not exposed via `vim.lsp.buf.code_action()` and must be requested explicitly.

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -29,6 +29,8 @@
 --- }
 --- ```
 ---
+--- Source actions such as organizing imports and removing unused code are available via `:LspTypescriptSourceAction`.
+---
 --- ### Vue support
 ---
 --- As of 2.0.0, Volar no longer supports TypeScript itself. Instead, a plugin
@@ -93,4 +95,19 @@ return {
       return vim.NIL
     end,
   },
+  on_attach = function(client)
+    -- ts_ls provides code actions that have a prefix `source.` that apply to the whole file. These will only appear in
+    -- `vim.lsp.buf.code_action()` if specified in `context.only`.
+    vim.api.nvim_buf_create_user_command(0, 'LspTypescriptSourceAction', function()
+      local source_actions = vim.tbl_filter(function(action)
+        return vim.startswith(action, 'source.')
+      end, client.server_capabilities.codeActionProvider.codeActionKinds)
+
+      vim.lsp.buf.code_action({
+        context = {
+          only = source_actions,
+        },
+      })
+    end, {})
+  end,
 }


justinmk:
> They are not exposed via `vim.lsp.buf.code_action()`

Thanks for confirming that :)

lsp/ts_ls.lua
```diff
@@ -78,4 +81,41 @@ return {
     'typescript.tsx',
   },
   root_markers = { 'tsconfig.json', 'jsconfig.json', 'package.json', '.git' },
+  on_attach = function(client)
+    -- ts_ls provides code actions that have a prefix `source.` that apply to the whole file. These are not discoverable
+    -- via the normal `vim.lsp.buf.code_action()` so add a command that allows access to them.
+    local function get_source_actions()
+      local codeActionKinds = client.server_capabilities.codeActionProvider.codeActionKinds
```
justinmk, line 88: Why aren't these discoverable via `code_action()` ? Can we enhance `code_action()` ?

thatsmydoing: The typescript language server only returns these code actions if they're specified in `context.only`.

So there's nothing to fix on the nvim side. The command implementation actually does use `code_action()`.

lsp/ts_ls.lua
```diff
@@ -78,4 +81,41 @@ return {
...
+      local params
+      if #opts.fargs == 0 then
+        -- ask the server what available actions there are
+        params = {
+          context = {
+            only = get_source_actions(),
+          },
+        }
+      else
+        -- with a specific action, apply it immediately
```
justinmk, line 110: This is an extra interface. Configs generally shouldn't invent new interfaces. The `code_actions()` menu is already sufficient. If someone wants to map something, they can just call `code_action()` with the right parameters.

lsp/ts_ls.lua
justinmk: > The typescript language server only returns these code actions if they're specified in `context.only`.

But why? Is it because `code_action()` is sending position info, or diagnostics info? Do we need a way to tell `code_action()` not to send that info?

thatsmydoing: That is really just how they've implemented it. You can see the code checks it [here](https://github.com/typescript-language-server/typescript-language-server/blob/0a034dae6c777733f4c2f6d5b151c5ef781fb662/src/lsp-server.ts#L779) where `kinds` is populated from [context.only](https://github.com/typescript-language-server/typescript-language-server/blob/0a034dae6c777733f4c2f6d5b151c5ef781fb662/src/lsp-server.ts#L769). The [refactor and quickfix actions](https://github.com/typescript-language-server/typescript-language-server/blob/0a034dae6c777733f4c2f6d5b151c5ef781fb662/src/lsp-server.ts#L770-L775) on the other hand have an extra `!kinds` check which means they're always returned if `context.only` is not provided.

lsp/ts_ls.lua
thatsmydoing: I've removed the argument functionality.

PR #3784, opened 2025-04-26
feat(powershell_es): add vim.lsp.config support

ateoi:


lsp/powershell_es.lua
```diff
@@ -0,0 +1,59 @@
...
+return {
+  cmd = function(dispatchers)
+    local temp_path = vim.fn.stdpath('cache')
+    local bundle_path = vim.lsp.config.powershell_es.bundlePath
+
+    local shell
+    shell = vim.lsp.config.powershell_es.shell
+    if not shell then
+      shell = 'pwsh'
+    end
```
justinmk, line 48:
```suggestion
    local shell = vim.lsp.config.powershell_es.shell or 'pwsh'
```

PR #3796, opened 2025-04-27
feat(lua_ls): add rootPath to workspace.library

Mo-tmd:
This seems to make it possible to jump to any lua file under rootPath, regardless of directory structure.

I'm a noob in this area, please let me know if there's anything strange.

### DIFF
--- a/lsp/lua_ls.lua
+++ b/lsp/lua_ls.lua
@@ -65,4 +65,14 @@ return {
     'selene.yml',
     '.git',
   },
+  settings = {
+    Lua = {
+      workspace = {
+        library = {},
+      },
+    },
+  },
+  before_init = function(init_params, config)
+    table.insert(config.settings.Lua.workspace.library, init_params.rootPath)
+  end,
 }


antonk52:
I think this is better suited in the docs to make lua_ls more suitable for neovim configuration. This should not be in the default settings as not every lua project opened in neovim is a neovim configuration.

Mo-tmd:
> I think this is better suited in the docs to make lua_ls more suitable for neovim configuration. This should not be in the default settings as not every lua project opened in neovim is a neovim configuration.

Why do you think this is a neovim specific configuration?

antonk52:
Please ignore my comment. For some reason I thought it was injecting runtime path, which is clearly not the case

justinmk:
This seems interesting but in the Lua projects I tried, it doesn't seem to have any effect, the `library` is always `{}`.

Here is the output of `:lua =vim.lsp.config['lua_ls']` : 

```
{
  before_init = <function 1>,
  cmd = { "lua-language-server" },
  filetypes = { "lua" },
  name = "lua_ls",
  root_markers = { ".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml", "stylua.toml", "selene.toml", "selene.yml", ".git" },
  settings = {
    Lua = {
      workspace = {
        library = {}
      }
    }
  }
}
```

What is an example that you are seeing? And what does `:checkhealth vim.lsp` show for you?

Mo-tmd:
Hi, sorry for the late reply. `:lua =vim.lsp.config['lua_ls']` for me shows the same as you. But `:checkhealth vim.lsp` shows:
```lua
vim.lsp: Active Clients ~
- lua_ls (id: 1)
  - Version: 3.18.2-dev
  - Root directory: ~/dotfiles
  - Command: { "lua-language-server" }
  - Settings: {
      Lua = {
        codeLens = {
          enable = true
        },
        hint = {
          enable = true,
          semicolon = "Disable"
        },
        runtime = {
          path = { "lua/?.lua", "lua/?/init.lua" },
          version = "LuaJIT"
        },
        workspace = {
          checkThirdParty = false,
          library = { "/home/<USER>/dotfiles" }
        }
      }
    }
  - Attached buffers: 1, 6
```

PR #3800, opened 2025-04-28
fix(eslint): check nil before on_dir

Parsifa1:
i don't know if this fix should be used, but it surely fixed my problem.

if you dont check if it is `nil` before pass it to `on_dir`, the `root_dir`  became useless. or if we should check `nil` inside `on_dir` ? i don't know.

maybe there are more config didn't check `nil` in `root_dir` in `/lsp/*.lua`, but i dont have time to check them carefully.


maximtrp:
Would it be better just to require workspace here (like in biome and tailwind configs)? `workspace_required = true`

justinmk:
yeah

Issue #3062, opened 2024-03-10
Addition of oxlint.


niksingh710:
### Description

Addition of [oxlint](https://github.com/oxc-project/oxc)

emko:
have you gotten oxlint to work?

niksingh710:
> have you gotten oxlint to work?

nope not yet

IWANABETHATGUY:
I would like to have a try

diegoulloao:
any new on this?

DonIsaac:
Hi there 👋 Oxlint maintainer here. I'd be happy to help you guys get Oxlint set up if you're still struggling. I can check in on this issue, or you can DM me on Twitter (@Don_dev_) or on Discord.

justinmk:
send a pr

Lippiece:
> send a pr

Can you provide a link? I also don't see any support for oxlint in the code.

PR #3815, opened 2025-05-02
feat(oxlint): added oxlint lsp config

maximtrp:
This PR adds the missing Oxlint LSP config (#3062, #3705).

### DIFF
--- a/lsp/oxlint.lua
+++ b/lsp/oxlint.lua
@@ -1,0 +1,30 @@
+--- @brief
+---
+--- https://github.com/oxc-project/oxc
+---
+--- `oxc` is a linter / formatter for JavaScript / Typescript supporting over 500 rules from ESLint and its popular plugins
+--- It can be installed via `npm`:
+---
+--- ```sh
+--- npm i -g oxlint
+--- ```
+
+local util = require 'lspconfig.util'
+
+return {
+  cmd = { 'oxc_language_server' },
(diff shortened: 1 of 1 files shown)


lsp/oxlint.lua
```diff
@@ -0,0 +1,30 @@
...
+    'javascript.jsx',
+    'typescript',
+    'typescriptreact',
+    'typescript.tsx',
+  },
+  workspace_required = true,
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local root_markers = util.insert_package_json({ '.oxlintrc.json' }, 'oxlint', fname)
+    on_dir(vim.fs.dirname(vim.fs.find(root_markers, { path = fname, upward = true })[1]))
```
justinmk, line 28: need nil check, i think...
```suggestion
    local d = vim.fs.dirname(vim.fs.find(root_markers, { path = fname, upward = true })[1])
    if d then
      on_dir(d)
    end
```

maximtrp: No, this is handled by requiring workspace in this config. It simply won't attach if no root dir is found

justinmk: Ok, I'll assume that's correct and close https://github.com/neovim/neovim/issues/33787 then :)

PR #3816, opened 2025-05-02
docs(lua_ls): add field Lua.runtime.path to example config for Nvim Lua

brianhuster:


brianhuster:
This is off-topic, but I think the example for Nvim Lua is too complicated. Isn't it better to define that in a `.nvim.lua` instead?

justinmk:
involing `.nvim.lua` is a new concept that isn't common yet.

some parts of this example should find their way into the actual config, if possible. though that has been controversial in the past...

PR #3817, opened 2025-05-02
fix(omnisharp): update executable name

vincentlabelle:
Problem:
The executable name may be in lowercase (e.g., when installed with Mason on Linux) in which case it won't be found.

Solution:
Verify if the name is in PascalCase, and otherwise use lowercase.

### DIFF
--- a/lsp/omnisharp.lua
+++ b/lsp/omnisharp.lua
@@ -16,7 +16,7 @@
 
 return {
   cmd = {
-    'OmniSharp',
+    vim.fn.executable('OmniSharp') == 1 and 'OmniSharp' or 'omnisharp',
     '-z', -- https://github.com/OmniSharp/omnisharp-vscode/pull/4300
     '--hostPID',
     tostring(vim.fn.getpid()),


lsp/omnisharp.lua
```diff
@@ -14,9 +14,19 @@
 ---
 ---
 
+--- Get the OmniSharp executable name.
+---@return string
+local function get_exec()
+  local exec = 'OmniSharp'
+  if vim.fn.executable(exec) ~= 1 then
+    exec = 'omnisharp'
```
lithammer, line 22: Is the binary always lowercase on Linux? If so you can just use the lowercase version regardless since windows is case-insensitive anyway.

vincentlabelle: Thanks for the feedback! I've downloaded the release (i.e., [omnisharp-linux-x64-net6.0.tar.gz](https://github.com/OmniSharp/omnisharp-roslyn/releases/download/v1.39.13/omnisharp-linux-x64-net6.0.tar.gz)) from the official github repository to compare with Mason, and I can confirm the executable name is in PascalCase.

lsp/omnisharp.lua
```diff
@@ -14,9 +14,19 @@
...
+  end
+  return exec
+end
+
 return {
   cmd = {
-    'OmniSharp',
+    get_exec(),
```
justinmk, line 29:
```suggestion
    vim.fn.executable('OmniSharp') == 1 and 'OmniSharp' or 'omnisharp',
```

lsp/omnisharp.lua
```diff
@@ -14,9 +14,19 @@
 ---
 ---
 
+--- Get the OmniSharp executable name.
+---@return string
+local function get_exec()
```
justinmk, line 19: remove

vincentlabelle: Yes, I was cleaning up at the same time you sent this. I've rebased too to clean the commit history.

616b2f:
as a reference, I created an PR to fix this on the mason site, as there is no real need for it to be lowercase in the first place:

https://github.com/mason-org/mason-registry/pull/9738

NonlinearFruit:
Isn't this a breaking change? The old default was `OmniSharp`. With this change, the default is `omnisharp`.


I think this might be a more backwards friendly solution:
```lua
vim.fn.executable('omnisharp') == 1 and 'omnisharp' or 'OmniSharp',
```


Using this commit breaks my config since I was running lspconfig _before_ running mason. So `OmniSharp` is not on my path so the `or` executes and decides to use `omnisharp`

PR #3819, opened 2025-05-04
fix(taplo): add Taplo configuration files to root markers

NinjacksonXV:
According to the [Taplo documentation](https://taplo.tamasfe.dev/configuration/file.html#configuration-file):
> Taplo supports configuration via file, unsurprisingly it uses the TOML format.
> 
> By default, every tool looks for one in the working directory or the root of the workspace by the following names (in precedence order):
> - .taplo.toml
> - taplo.toml

This simply adds those two files to the root directory definition.

Note: the commit message/PR is potentially mischaracterized as `fix` instead of `feat`; I was unsure of which category this falls under.

### DIFF
--- a/lsp/taplo.lua
+++ b/lsp/taplo.lua
@@ -11,5 +11,5 @@
 return {
   cmd = { 'taplo', 'lsp', 'stdio' },
   filetypes = { 'toml' },
-  root_markers = { '.git', '.taplo.toml', 'taplo.toml' },
+  root_markers = { '.taplo.toml', 'taplo.toml', '.git'  },
 }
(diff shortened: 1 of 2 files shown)


lsp/taplo.lua
```diff
@@ -11,5 +11,5 @@
 return {
   cmd = { 'taplo', 'lsp', 'stdio' },
   filetypes = { 'toml' },
-  root_markers = { '.git' },
+  root_markers = { '.git', '.taplo.toml', 'taplo.toml' },
```
justinmk, line 14: these are ordered by priority. sure .git should be last ?

NinjacksonXV: Thank you, this was an oversight when moving the changes to `lsp/taplo.lua`.

NinjacksonXV:
Apologies for the haphazard formatting and commit history; I've been making minor edits in Codespaces using the VSCode Git frontend, which has caused some inconsistencies.

PR #3822, opened 2025-05-06
Fix the 'cmd' setting for the 'jsonls' LSP server

austindd:
I encountered a bug with the default `jsonls` LSP server config. Perhaps I was missing something, but this change fixed my issue.

Before, the default `cmd` for `jsonls` was this:

```lua
    cmd = { 'vscode-json-language-server', '--stdio' },
```
This caused a bug where the LSP server was not loaded. I changed it to this (removed a dash):

```lua
    cmd = { 'vscode-json-languageserver', '--stdio' },
```

This change fixed the problem and enabled the LSP server to be attached.

### DIFF
--- a/lua/lspconfig/configs/jsonls.lua
+++ b/lua/lspconfig/configs/jsonls.lua
@@ -1,6 +1,6 @@
 return {
   default_config = {
-    cmd = { 'vscode-json-language-server', '--stdio' },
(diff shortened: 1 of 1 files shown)


justinmk:
ok, but that doesn't match the docs.

@hrsh7th does this look right ?

lithammer:
How did you install it (os, platform, package manager etc)?

PR #3825, opened 2025-05-07
feat(jsonnet_ls): jsonnet_ls lsp

ilpianista:
Should remove `jsonnet_ls` from the list in #3705.

### DIFF
--- a/lsp/jsonnet_ls.lua
+++ b/lsp/jsonnet_ls.lua
@@ -1,0 +1,18 @@
+--- @brief
+---
+--- https://github.com/grafana/jsonnet-language-server
+---
+--- A Language Server Protocol (LSP) server for Jsonnet.
+---
+--- The language server can be installed with `go`:
+--- ```sh
+--- go install github.com/grafana/jsonnet-language-server@latest
+--- ```
+return {
+  cmd = { 'jsonnet-language-server' },
+  filetypes = {
+    'jsonnet',
+    'libsonnet',
+  },
+  root_markers = { 'jsonnetfile.json', '.git' },
+}


lsp/jsonnet_ls.lua
```diff
@@ -0,0 +1,24 @@
...
+return {
+  cmd = { 'jsonnet-language-server' },
+  filetypes = {
+    'jsonnet',
+    'libsonnet',
+  },
+  root_dir = function(fname)
+    return util.root_pattern 'jsonnetfile.json'(fname)
+      or vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
+  end,
```
lithammer, line 23: Looks like you should be able to replace this with `root_markers`?

ilpianista: I wasn't sure about that, is that as easy as my last commit?

lsp/jsonnet_ls.lua
```diff
@@ -0,0 +1,37 @@
...
+
+local root_dir = function(fname)
+  return util.root_pattern 'jsonnetfile.json'(fname)
+    or vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
+end
+
+return {
+  cmd = { 'jsonnet-language-server' },
+  cmd_env = {
+    JSONNET_PATH = vim.env.JSONNET_PATH or jsonnet_path(root_dir()),
```
lithammer, line 30: I'm not sure the fallback here is a good idea. It will be relative to cwd rather than the buffer. It's also calculated once in the beginning. So if you `:cd` around it will be incorrect as well. It will also be wrong when doing something like `nvim ../foo/bar.jsonnet` or just `nvim ~/some/other/dir/file.jsonnet`.

lsp/jsonnet_ls.lua
```diff
@@ -0,0 +1,37 @@
...
+  },
+  filetypes = {
+    'jsonnet',
+    'libsonnet',
+  },
+  root_markers = { '.git' },
```
lithammer, line 36: Shouldn't this include `jsonnetfile.json`? Note that the order of these are important.

lithammer (approved):
I guess this is one of those cases where `cmd_env` being function with the buffer being attached or `root_dir` as argument would be useful.

It also seems like the solution in `lua/lspconfig/configs/jsonnet_ls.lua` is incorrect for Windows since it's splitting on `:`. It should split on `;`.

PR #3835, opened 2025-05-10
feat(matlab_ls): vim.lsp.config support

haraldan:
Adding a vim.lsp.config option for matlab_ls, as discussed in #3705.

This also brings a couple of improvements compared to the old config:
- Outside of a .git repository, CWD will be passed to the language server as root_dir, replicating normal Matlab behaviour
- indexWorkspace is set to "true" by default

### DIFF
--- a/lsp/matlab_ls.lua
+++ b/lsp/matlab_ls.lua
@@ -1,0 +1,21 @@
+---@brief
+---
+--- https://github.com/mathworks/MATLAB-language-server
+---
+--- MATLAB® language server implements the Microsoft® Language Server Protocol for the MATLAB language.
+return {
+  cmd = { 'matlab-language-server', '--stdio' },
+  filetypes = { 'matlab' },
+  root_dir = function(bufnr, on_dir)
+    local root_dir = vim.fs.root(bufnr, '.git')
+    on_dir(root_dir or vim.fn.getcwd())
+  end,
+  settings = {
+    MATLAB = {
+      indexWorkspace = true,
+      installPath = '',
+      matlabConnectionTiming = 'onStart',
+      telemetry = true,
+    },
+  },
+}


lsp/matlab_ls.lua
```diff
@@ -0,0 +1,25 @@
...
+return {
+  cmd = { 'matlab-language-server', '--stdio' },
+  filetypes = { 'matlab' },
+  root_dir = function(_, callback)
+    local root_dir = vim.fs.root(0, '.git')
+    if root_dir then
+      callback(root_dir)
+    else
+      callback(vim.env.PWD)
+    end
```
justinmk, line 15:
```suggestion
  root_dir = function(_, on_dir)
    local root_dir = vim.fs.root(0, '.git')
    on_dir(root_dir and root_dir or vim.fn.getcwd())
```

lsp/matlab_ls.lua
```diff
@@ -0,0 +1,25 @@
+---@brief
+---
+--- https://github.com/mathworks/MATLAB-language-server
+---
+--- MATLAB® language server implements the Microsoft® Language Server Protocol for the MATLAB language.
+return {
+  cmd = { 'matlab-language-server', '--stdio' },
+  filetypes = { 'matlab' },
+  root_dir = function(_, callback)
+    local root_dir = vim.fs.root(0, '.git')
```
lithammer, line 10:
```suggestion
  root_dir = function(bufnr, callback)
    local root_dir = vim.fs.root(bufnr, '.git')
```

I'm not 100% sure about this. But I suspect things like `nvim foo.m bar.m` (open multiple files) will be incorrect since it will attempt to attach to `foo.m` twice (since `0` refers to the currently active buffer, not the buffer being attached).


Regardless, I think it's a good practice.

lsp/matlab_ls.lua
haraldan: What is the purpose of `root_dir and root_dir`? I'm probably not getting something about this in terms of lua syntax. Is this more efficient than the if-else check?

justinmk: it's like a ternary. not more efficient, just less verbose.

also, want to avoid the `callback` name to avoid other configs copying that name.

haraldan: @justinmk Certainly, that makes sense. Isn't `root_dir and root_dir or vim.fn.getcwd()` equivalent to `root_dir or vim.fn.getcwd()` though?

justinmk: yes

haraldan: Commited.

PR #3839, opened 2025-05-10
fix(gopls): prioritise `go.work` over `go.mod`

lithammer:
The purpose of a `go.work` file is to handle multi-module workspaces. A common such scenario is a monorepo.

This is also how the legacy config works because it's using `lsponfig.util.root_pattern()` which has similar semantics as `root_markers`.

### DIFF
--- a/lsp/gopls.lua
+++ b/lsp/gopls.lua
@@ -15,7 +15,7 @@ local function get_root(fname)
       return clients[#clients].config.root_dir
     end
   end
-  return vim.fs.root(fname, { 'go.work', 'go.mod', '.git' })
+  return vim.fs.root(fname, 'go.work') or vim.fs.root(fname, 'go.mod') or vim.fs.root(fname, '.git')
 end
 
 return {


lithammer:
Related upstream comment:

- https://github.com/neovim/neovim/pull/33485#issuecomment-2869116498

wayjam:
I'm encountering the same issue. The older version handled correctly

https://github.com/neovim/nvim-lspconfig/blob/4ea9083b6d3dff4ddc6da17c51334c3255b7eba5/lua/lspconfig/configs/gopls.lua#L25
https://github.com/neovim/nvim-lspconfig/blob/4ea9083b6d3dff4ddc6da17c51334c3255b7eba5/lua/lspconfig/util.lua#L99

Here are my solution

```
---@param path string
---@return string
local function strip_archive_subpath(path)
  -- Handle zipfile paths
  local zip_pattern = "zipfile://(.+)::[^\\].*$"
  local zip_match = path:match(zip_pattern)
  if zip_match then return zip_match end

  -- Handle tarfile paths
  local tar_pattern = "tarfile:(.+)::.*$"
  local tar_match = path:match(tar_pattern)
  if tar_match then return tar_match end

  return path
end

---@param fname string
---@return string?
local function get_root(fname)
  if mod_cache and fname:sub(1, #mod_cache) == mod_cache then
    local clients = vim.lsp.get_clients { name = "gopls" }
    if #clients > 0 then return clients[#clients].config.root_dir end
  end

  local start_path = strip_archive_subpath(fname)

  -- Use Neovim's built-in functions to find project root with fallback pattern
  return vim.fs.root(start_path, function(path)
    for _, pattern in ipairs { "go.work", "go.mod", ".git" } do
      if vim.loop.fs_stat(vim.fs.joinpath(path, pattern)) ~= nil then return true end
    end
    return false
  end)
end

```

lithammer:
Why do you need to handle zips and tars?

PR #3842, opened 2025-05-11
feat(ty): Add `ty` lsp config

hoel-bagard:
This PR adds the [ty lsp](https://github.com/astral-sh/ty) config. The project is still in early development but meets the criteria for a new config.

### DIFF
--- a/lsp/ty.lua
+++ b/lsp/ty.lua
@@ -1,0 +1,12 @@
+---@brief
+---
+--- https://github.com/astral-sh/ty
+---
+--- A Language Server Protocol implementation for ty, an extremely fast Python type checker and language server, written in Rust.
+---
+--- For installation instructions, please refer to the [ty documentation](https://github.com/astral-sh/ty/blob/main/README.md#getting-started).
+return {
+  cmd = { 'ty', 'server' },
+  filetypes = { 'python' },
+  root_markers = { 'ty.toml', 'pyproject.toml', '.git' },
+}


lsp/ty.lua
```diff
@@ -0,0 +1,32 @@
...
+  },
+  settings = {
+    python = {
+      analysis = {
+        autoSearchPaths = true,
+        useLibraryCodeForTypes = true,
+        diagnosticMode = 'openFilesOnly',
+      },
+    },
+  },
```
lithammer, line 31: Are you sure about these? They look very much like pyright settings.

lsp/ty.lua
```diff
@@ -0,0 +1,32 @@
...
+return {
+  cmd = { 'ty', 'server' },
+  filetypes = { 'python' },
+  root_markers = {
+    'pyproject.toml',
+    'setup.py',
+    'setup.cfg',
+    'requirements.txt',
+    'Pipfile',
+    'ty.toml',
```
lithammer, line 20: `ty.toml` should probably be top priority given that it takes precedence over `pyproject.toml` if both are present.

lsp/ty.lua
hoel-bagard: That's probably not correct indeed, I naively assumed they would be the same, I'm looking into it now.

pkos98:
This PR may need an additional file:

```
-- lua/lspconfig/configs/ty.lua

return {
  default_config = {
    cmd = { 'ty', 'server' },
    filetypes = { 'python' },
    root_dir = function(fname)
      return util.root_pattern(
        'ty.toml',
        'pyproject.toml',
        'setup.py',
        'setup.cfg',
        'requirements.txt',
        'Pipfile',
        '.git'
      )(fname)
    end,
  },
  docs = {
    description = 'https://github.com/astral-sh/ty',
  },
}
```

hoel-bagard:
@pkos98 Thank you!

lsp/ty.lua
```diff
@@ -0,0 +1,37 @@
...
+    end,
+  },
+  docs = {
+    description = [[
+https://github.com/astral-sh/ty
+`ty`, a static type checker and language server for python.
+The language server can be installed with `uv`: `uv tool install ty`
+]],
+  },
+}
```
lithammer, line 37: This isn't correct. This is what the legacy looks like.

```suggestion
return {
  cmd = { 'ty', 'server' },
  filetypes = { 'python' },
  root_markers = {
    'ty.toml',
    'pyproject.toml',
    'setup.py',
    'setup.cfg',
    'requirements.txt',
    'Pipfile',
    '.git'
  },
}
```

lithammer:
> This PR may need an additional file:
> 
> ```lua
> -- lua/lspconfig/configs/ty.lua
> 
> local util = require 'lspconfig.util'
> 
> return {
>   default_config = {
>     cmd = { 'ty', 'server' },
>     filetypes = { 'python' },
>     root_dir = function(fname)
>       return util.root_pattern(
>         'ty.toml',
>         'pyproject.toml',
>         'setup.py',
>         'setup.cfg',
>         'requirements.txt',
>         'Pipfile',
>         '.git'
>       )(fname)
>     end,
>   },
>   docs = {
>     description = [[
>     https://github.com/astral-sh/ty
>     `ty`, a static type checker and language server for python.
>     The language server can be installed with `uv`: `uv tool install ty`
> ]]
>   },
> }
> ```

I don't think new legacy configs are accepted anymore.

hoel-bagard:
@lithammer I had not added it to the PR at first since it did not appear in the python related configs I used as reference, but in my config I had to use `root_dir` to get the ty lsp to work.

Should I still revert the changes to use `root_markers` ?

(Btw, should I keep force pushing with every change to keep the commit name/commit history as is, or will the PR be squashed when merged ?)

lsp/ty.lua
```diff
@@ -0,0 +1,24 @@
+---@brief
+---
+--- https://github.com/astral-sh/ty
+---
+--- `ty`, a static type checker and language server for python.
```
dhruvmanila, line 5: It's just ty (no backticks) unless it references the executable `ty`.

```suggestion
--- A Language Server Protocol implementation for ty, an extremely fast Python type checker and language server, written in Rust.
```

lsp/ty.lua
```diff
@@ -0,0 +1,24 @@
...
+  filetypes = { 'python' },
+  root_markers = {
+    'ty.toml',
+    'pyproject.toml',
+    'setup.py',
+    'setup.cfg',
+    'requirements.txt',
+    'Pipfile',
+    '.git',
+  },
```
dhruvmanila, line 23: Just FYI, ty only looks for `pyproject.toml` and `ty.toml` so we could possibly start with just that and `.git`. We can add other root markers if the need arises.

lsp/ty.lua
```diff
@@ -0,0 +1,24 @@
+---@brief
+---
+--- https://github.com/astral-sh/ty
+---
+--- `ty`, a static type checker and language server for python.
+---
+--- The language server can be installed with `uv`:
+--- ```sh
+--- uv tool install ty
+--- ```
```
dhruvmanila, line 10: Let's instead link to the documentation: https://github.com/astral-sh/ty/blob/main/README.md#getting-started. Yes, for now the documentation is on GitHub :)

dhruvmanila (commented):
Disclaimer: I work at Astral

Thanks for creating a PR for the ty config! I've provided some suggestions

PR #3843, opened 2025-05-11
Rename `volar` to `vue_ls` and remove `vuels`

### LINKED ISSUE #3826: Rename `volar`
### Description

When [Vue's current language tooling](https://github.com/vuejs/language-tools) was new it used **Volar** to differentiate it from the old language tools (**[Vetur](https://github.com/vuejs/vetur)**). It has been over 2 years since the current tooling has completely replaced the old one and today **Volar** is not really used anymore in regards to Vue's language tooling, because it refers to the [underlying JS library](https://volarjs.dev/) instead.

With that in mind I propose two changes:
1. Rename `volar` within nvim-lspconfig
    - The npm package goes by `@vue/language-server` and the executable by `vue-language-server`, so in accordance with the contributing guidelines `vue_ls` would probably make the most sense
2. Remove `vuels` from nvim-lspconfig
    - it is no longer actively maintained
    - it creates confusion (see #3823)

I'm open to create the PR, but since these are breaking changes, I wanted to discuss them first.

Jak-Ch-ll:
Closes #3826 

A few questions:
1. Is there a good way with the new configs to alias an older name, so it stays available?
    - I found the [`aliases`](https://github.com/Jak-Ch-ll/nvim-lspconfig/blob/0230509108de896808ec6dc3b59c153f6e8a4385/lua/lspconfig.lua#L13) table, but this seems to only affect the old configs.
    - For the new configs I kept the `lsp/volar.lua` file, added a `vim.deprecate`, and returned `vim.lsp.config.vue_ls`
2. What is a good version to remove it later? I used `3.0.0` as placeholder

Additionally (even though not super necessary), I couldn't check the generated docs because `nvim -R -Es +'set rtp+=$PWD' +'luafile scripts/docgen.lua'` fails without message for me
- Running `luafile scripts/docgen.lua'` inside of Neovim (even after adding the path to rtp) fails with the error `scripts/docgen.lua:136: module 'lsp.ada_ls' not found`
- Do I need to setup anything locally?


Jak-Ch-ll:
Removed my changes from the old configs, as per lint rules

justinmk:
> * For the new configs I kept the `lsp/volar.lua` file, added a `vim.deprecate`, and returned `vim.lsp.config.vue_ls`

exactly 👍

justinmk:
> * Running `luafile scripts/docgen.lua'` inside of Neovim (even after adding the path to rtp) fails with the error `scripts/docgen.lua:136: module 'lsp.ada_ls' not found`
> * Do I need to setup anything locally?

that's strange. shouldn't happen

PR #3851, opened 2025-05-14
Update ts_ls.lua to include instructions for ember support

NullVoxPopuli:
idk if we just want instructions here? because we can do a lot better with the `on_new_config` hook.

Here is what I do in my own configs:
```lua
    if (serverName == 'ts_ls') then
      local filetypes = {
        'typescript',
        'javascript',
        'typescript.glimmer',
        'javascript.glimmer',
        'typescript.tsx',
        'javascript.jsx',
        'html.handlebars',
        'handlebars',
      }
      server.setup({
        -- filetypes MUST include our custom filetypes for all variants of TS that we might want to use
        -- else the server won't boot and on_new_config won't get called, so we can't conditionally
        -- configure the init_options based on our project
        filetypes = filetypes,
        root_dir = utils.is_ts_project,
        capabilities = capabilities,
        -- This allows us to switch types of TSServers based on the open file.
        -- We don't always need the @glint/tsserver-plugin -- for example, in backend projects.
          local info = utils.read_nearest_ts_config(new_root_dir)
          local glintPlugin = new_root_dir .. "node_modules/@glint/tsserver-plugin"

          if new_config.init_options then
            new_config.init_options.tsdk = get_typescript_server_path(new_root_dir)
            new_config.init_options.requestForwardingCommand = "forwardingTsRequest"


            if (info.isGlintPlugin) then
              new_config.init_options.plugins = {
                {
                  name = "@glint/tsserver-plugin",
                  location = glintPlugin,
                  languages = filetypes,
                  enableForWorkspaceTypeScriptVersions = true,
                  configNamespace = "typescript"
                }
              }
              print(glintPlugin)
            end
          end
        end,
        init_options = {
          tsserver = { logVerbosity = 'verbose', trace = "verbose" },
          preferences = {
            disableAutomaticTypingAcquisition = true,
            importModuleSpecifierPreference = "relative",
            importModuleSpecifierEnding = "minimal",
          },
          plugins = {}
        },
        settings = mySettings[serverName],
          keymap(bufnr)
          conditional_features(client, bufnr)
        end
      })
```

this allows automatic swapping of the tsplugins based on package.json, tsconfig.json, and/or node_modules contents

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -32,6 +32,34 @@
 --- Use the `:LspTypescriptSourceAction` command to see "whole file" ("source") code-actions such as:
 --- - organize imports
 --- - remove unused code
+--- 
+--- ### Ember support
+--- 
+--- ```lua
+--- vim.lsp.config('ts_ls',  {
+---   init_options = {
+---     plugins = {
+---       {
+---         name = "@glint/tsserver-plugin",
+---         location = "/your/path/to/@glint/tsserver-plugin",
+---         languages = {"javascript", "typescript", "typescript.glimmer", "javascript.glimmer"},
+---       },
+---     },
+---   },
+---   filetypes = {
+---     "javascript",
+---     "typescript",
+---     "javascript.glimmer",
+---     "typescript.glimmer"
+---   },
+--- })
+--- ```
+--- `location` MUST be defined. If the plugin is installed in `node_modules`,
+--- `location` can have any value.
+---
+--- `languages` must include `typescript.glimmer` and `javascript.glimmer` even if it is listed in `filetypes`.
+---
+--- `filetypes` is extended here to include the gjs and gts filetype names.
 ---
 --- ### Vue support
 ---
--- a/lua/lspconfig/configs/ts_ls.lua
+++ b/lua/lspconfig/configs/ts_ls.lua
@@ -70,43 +70,7 @@ require'lspconfig'.ts_ls.setup{
     "vue",
   },
 }
-
--- You must make sure volar is setup
--- e.g. require'lspconfig'.volar.setup{}
--- See volar's section for more information
 ```
-
-`location` MUST be defined. If the plugin is installed in `node_modules`,
-`location` can have any value.
-
-`languages` must include `vue` even if it is listed in `filetypes`.
-
-`filetypes` is extended here to include Vue SFC.
-
-### Ember support
-
-```lua
-require'lspconfig'.ts_ls.setup{
-  init_options = {
-    plugins = {
-      {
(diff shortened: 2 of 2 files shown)


justinmk:
stylua lint issue

justmejulian:
Isn't `@glint/tsserver-plugin` deprecated?
https://www.npmjs.com/package/@glint/tsserver-plugin

NullVoxPopuli:
No, we're working on an alpha which undeprecates it.

Latest alpha was released yesterday

NullVoxPopuli:
It's been undeprecated now

wagenet:
Note that npm still shows it as deprecated since the `@latest` is deprecated. The alpha version is _not_ deprecated.

justmejulian:
Isn't this very specific to ember? Does it make sense to bloat the `ts_ls` description with something that does not effect most typescript users?
Maybe it would make more sense add it to [glint.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/glint.lua).
We could list it as an alternative to using the `glint-language-server`.

NullVoxPopuli:
> Isn't this very specific to ember? Does it make sense to bloat the ts_ls description with something that does not effect most typescript users?

No moreso than the already existing vue/volar docs in here.

In reality, i'd like a way in these configs to automatically configure ts_ls plugins based on detected project config. That way, vue/volar, ember/glint, can 'just work', and not require additional config, unless for customization, like full LSes.

> glint.lua

Maybe, the hope is to declare this as legacy support when we release all our v2 LS/tsserver-plugin stuff

NullVoxPopuli:
@justmejulian -- would you be willing to receive a PR that adjusts the configs to automatically apply the configs listed here in the docs based on detected packages / config files? that way, we don't need a massive documentation block for every variation.

The main thing I'd bring over from my own dotfiles, is a function like this that does the branching logic / boolean creating: https://github.com/NullVoxPopuli/dotfiles/blob/main/home/.config/nvim/lua/plugin-config/lsp/utils.lua#L44

justmejulian:
I don’t think I have any say in the matter 🙆
I only came across the PR while setting up Glint in my personal config. After seeing that [@glint/tsserver-plugin](https://www.npmjs.com/package/@glint/tsserver-plugin) was deprecated, I ended up using the @glint/core LSP server instead.

My personal opinion would be to keep the configs as simple as possible, and split out things like Glint/Vue into their own configs, so only the people who need them are affected. Not sure how doable that is with plugins that rely on ts_ls, though.

Long story long: if you think it would be beneficial, go ahead and open a PR! From what I’ve seen in this repo, contributions are always appreciated.

PS: Thanks for all your work in the Ember community!

PR #3855, opened 2025-05-18
add missing lua/lspconfig/configs/ty.lua

fabridamicelli:
I see that configs under `lua/lspconfig/configs` are deprecated, but without it, we hit this error when trying to setup the server
```lua
require('lspconfig').ty.setup({})
```
image

This file is missing as mentioned [here](https://github.com/neovim/nvim-lspconfig/pull/3842#issuecomment-2875235725)

Adding this file fixes it.

Info:
image

Any idea on how to solve it if not adding the file?

maybe related #3705

### DIFF
--- a/lua/lspconfig/configs/ty.lua
+++ b/lua/lspconfig/configs/ty.lua
@@ -1,0 +1,20 @@
+-- lua/lspconfig/configs/ty.lua
+
+local util = require 'lspconfig.util'
+
+return {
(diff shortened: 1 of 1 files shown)


fabridamicelli:
I got the error. I should have used
```lua

vim.lsp.config('ty', {})
```

Everything works fine.
Closing this PR.

PR #3861, opened 2025-05-21
fix(nushell): avoid nil root_dir (reapply #3611)

Bahex:
Simplified `lsp/nushell.lua` unfortunately triggered the problem with the lsp server again.

### DIFF
--- a/lsp/nushell.lua
+++ b/lsp/nushell.lua
@@ -6,5 +6,7 @@
 return {
   cmd = { 'nu', '--lsp' },
   filetypes = { 'nu' },
-  root_markers = { '.git' },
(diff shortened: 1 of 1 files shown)


lsp/nushell.lua
```diff
@@ -6,5 +6,7 @@
 return {
   cmd = { 'nu', '--lsp' },
   filetypes = { 'nu' },
-  root_markers = { '.git' },
+  root_dir = function(bufnr, cb)
```
justinmk, line 9:
```suggestion
  root_dir = function(bufnr, on_dir)
```

justinmk:
In the future please give a clear explanation. I updated your description.

PR #3862, opened 2025-05-22
Fix(volar.lsp): Recent renaming of volar-lsp to vue_ls broke resolving of config on fresh installs

gx089:
Recent changes of renaming of volar-lsp to vue_ls-lsp broke config resolving on fresh installs. There is no vue_ls.lua config.
Fix is, to create a config with name vue_ls.lua.(Is a copy of the volar.lua)

Btw. The official name of the Vue-LSP is Volar. That's because, there was an much older/obsolete lsp called vetur for vue, that isn't used anymore. In my opinion the recent renaming of lsp and mason packages is confusing. but anyway...

neovim/nvim-lspconfig:
https://github.com/neovim/nvim-lspconfig/pull/3843

mason-registry: 
https://github.com/mason-org/mason-registry/pull/10150

### DIFF
--- a/lua/lspconfig/configs/vue_ls.lua
+++ b/lua/lspconfig/configs/vue_ls.lua
@@ -1,0 +1,102 @@
+local util = require 'lspconfig.util'
+
+-- https://github.com/vuejs/language-tools/blob/master/packages/language-server/lib/types.ts
+local volar_init_options = {
+  typescript = {
+    tsdk = '',
+  },
+}
+
+return {
+  default_config = {
+    cmd = { 'vue-language-server', '--stdio' },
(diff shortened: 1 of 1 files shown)


justinmk:
? https://github.com/neovim/nvim-lspconfig/pull/3843 did not change anything in `http://lua/lspconfig/configs/`.


cc @Jak-Ch-ll

justinmk:
Again `lua/lspconfig/configs/` is irrelevant; those are the old, frozen configs for Nvim 0.10. 

If there's a change needed in `lsp/`, propose one. I don't care about any of this because I don't use these configs.

PR #3867, opened 2025-05-22
feat: add kotlin_lsp from jetbrains

hbjydev:
This PR adds support for the new official (though pre-alpha) [Kotlin LSP](https://github.com/kotlin/kotlin-lsp).

I only added files to the root pattern that the [official docs on using Neovim with it](https://github.com/Kotlin/kotlin-lsp/blob/main/scripts/neovim.md) suggest, but I've tested it locally and it seems to work fine :)

### DIFF
--- a/lua/lspconfig/configs/kotlin_lsp.lua
+++ b/lua/lspconfig/configs/kotlin_lsp.lua
@@ -1,0 +1,31 @@
+local util = require 'lspconfig.util'
+
+local bin_name = 'kotlin-lsp'
+
+--- The presence of one of these files indicates a project root directory
+--
+--  These are configuration files for the various build systems supported by
+--  Kotlin.
+local root_files = {
+  'settings.gradle', -- Gradle (multi-project)
+  'settings.gradle.kts', -- Gradle (multi-project)
+  'pom.xml', -- Maven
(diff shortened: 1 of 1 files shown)


amgdev9:
I think `workspace.json` should be added as well as a marker, it is used to integrate your own build system

lsp/kotlin_lsp.lua
```diff
@@ -0,0 +1,22 @@
...
+  'settings.gradle.kts', -- Gradle (multi-project)
+  'pom.xml', -- Maven
+  'build.gradle', -- Gradle
+  'build.gradle.kts', -- Gradle
+}
+
+return {
+  filetypes = { 'kotlin' },
+  cmd = { 'kotlin-lsp', '--stdio' },
+  root_markers = root_files,
```
justinmk, line 21: just inline the files here, makes it easier for doc generation.
```suggestion
  root_markers = {
    'settings.gradle', -- Gradle (multi-project)
    'settings.gradle.kts', -- Gradle (multi-project)
    'pom.xml', -- Maven
    'build.gradle', -- Gradle
    'build.gradle.kts', -- Gradle
  },
```

hbjydev:
Thanks for the merge! 👍🏻

PR #3869, opened 2025-05-24
fix(fish_lsp): apply author suggestions

unlimitedsola:
Apply suggestions brought up by the `fish_lsp` author: 
https://github.com/mason-org/mason-registry/pull/8609#issuecomment-2904347943


unlimitedsola:
> lint.sh: Configs in `lua/lspconfig/configs/*` are deprecated. Add or update configs in `lsp/*` instead.

I'm not sure if `lua/lspconfig/configs/*` are still in use, if they are (for legacy users), I think it makes sense to have them updated as well.

justinmk:
Nice improvements, thanks.

> I'm not sure if `lua/lspconfig/configs/*` are still in use, if they are (for legacy users), I think it makes sense to have them updated as well.

Yeah, they're frozen and will be removed, but small changes like this are fine.

PR #3870, opened 2025-05-24
feat: tombi

yassun7010:
TOML language server with a distinctive auto-sorting feature that references the JSON schema store (https://github.com/tombi-toml/tombi)

### DIFF
--- a/lsp/tombi.lua
+++ b/lsp/tombi.lua
@@ -7,5 +7,5 @@
 return {
   cmd = { 'tombi', 'lsp' },
   filetypes = { 'toml' },
-  root_markers = { 'tombi.toml', '.git' },
+  root_markers = { 'tombi.toml', 'pyproject.toml', '.git' },
 }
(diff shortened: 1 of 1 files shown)


lsp/tombi.lua
```diff
@@ -0,0 +1,11 @@
+---@brief
+---
+--- https://tombi-toml.github.io/tombi/
+---
+--- Language server for Tombi, a TOML toolkit.
+---
+return {
+  cmd = { 'tombi', 'lsp' },
+  filetypes = { 'toml' },
+  root_markers = { 'tombi.toml', '.git' },
```
lithammer, line 10: I think you want to add `pyproject.toml` to that list as well. See https://tombi-toml.github.io/tombi/docs/configuration.

```suggestion
  root_markers = { 'tombi.toml', 'pyproject.toml', '.git' },
```

yassun7010: ddcfe0643f0a14f0accf93f0385f18047a86bd8d

Issue #3837, opened 2025-05-10
"E492: Not an editor command: EslintFixAll" when using the new vim.lsp.config


gjf7:
### Description

I use [mason.nvim](https://github.com/williamboman/mason.nvim) and [mason-lspconfig.nvim](https://github.com/williamboman/mason-lspconfig.nvim) to manage my language servers.

When using the new vim.lsp.config to set up the EslintFixAll command, I encountered the error E492: Not an editor command: EslintFixAll. However, switching back to the old-style lspconfig.eslint.setup worked correctly.

```lua
  require("mason").setup()
  require("mason-lspconfig").setup({
    ensure_installed = { "lua_ls", "clangd", "vtsls", "pyright", "eslint", "cssls", "typos_lsp" }
  })
  
  -- this doesn't work
  vim.lsp.config('eslint', {
    on_attach = function(_, bufnr)
      vim.api.nvim_create_autocmd("BufWritePre", {
        buffer = bufnr,
        command = "EslintFixAll",
      })
    end,
  })
  

  -- this works
  local lspconfig = require("lspconfig")
  lspconfig.eslint.setup({
    on_attach = function(_, bufnr)
      vim.api.nvim_create_autocmd("BufWritePre", {
        buffer = bufnr,
        command = "EslintFixAll",
      })
    end,
  })

```

justinmk:
all commands now have the `Lsp` prefix, e.g.  `LspEslintFixAll` https://github.com/neovim/nvim-lspconfig/blob/af4a9f5ce58763365999bfc558ddecd20d41f2b7/lsp/eslint.lua#L49

need to update the docs https://github.com/neovim/nvim-lspconfig/blob/af4a9f5ce58763365999bfc558ddecd20d41f2b7/doc/configs.txt#L2541

eyestorefivemstore:
The documentation in this repository is excellent! Very clear and comprehensive.

eyestoreoye:
The documentation in this repository is excellent! Very clear and comprehensive.

gjf7:
Even after changing `EslintFixAll` to `LspEslintFixAll`, it still doesn’t work. It seems the user’s on_attach function is overriding the default on_attach from lspconfig.

justinmk:
https://github.com/neovim/nvim-lspconfig/pull/3844

oedotme:
[...]

PR #3876, opened 2025-05-28
fix(eslint): LspEslintFixAll not running sync

pearofducks:
This fixes the problem described [here](https://github.com/neovim/nvim-lspconfig/issues/3837#issuecomment-2901107052). The new `LspEslintFixAll` writes the changes after save.

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -48,12 +48,9 @@ return {
     'astro',
   },
   workspace_required = true,
-  on_attach = function(client)
+  on_attach = function(client, bufnr)
     vim.api.nvim_buf_create_user_command(0, 'LspEslintFixAll', function()
-      local bufnr = vim.api.nvim_get_current_buf()
-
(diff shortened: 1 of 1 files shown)


justinmk:
Why close this? Was there a problem with it?

pearofducks:
The linter was upset with the use of `workspace/executeCommand`, but `client:exec_cmd` doesn't support sync operations that I can see.

```
lint.sh: Use client:exec_cmd() instead of calling request("workspace/executeCommand") directly. Example: lsp/pyright.lua
```

I'm unsure if `vim.lsp.buf.format` and running `eslint.applyAllFixes` would produce the same results (and don't have time to dive into that unfortunately).

justinmk:
special-cased in the linter until we have a better answer: https://github.com/neovim/nvim-lspconfig/commit/8475898082ebe4ce40594a0b572be86bcc11ac45

PR #3877, opened 2025-05-28
feat: cspell language server

Yuki-bun:
Add [cspell language server](https://github.com/vlabo/cspell-lsp).


justinmk:
What are the advantages vs Nvim's builtin spellcheck ?

PR #3886, opened 2025-06-03
feat(bicep): add `bicep-params` filetype for bicep lsp

### LINKED ISSUE #3885: Bicep config doesn't provide support for parameters
### Description

The Bicep integration doesn't work with bicepparam files. They are their own file type and need to be registered as such. If you are setting the bicepparam file type to bicep, it would result in unrelated errors, as the validation logic for bicepparam files is different from the normal bicep file logic.

damienpontifex:
- Support for .bicepparam files
- `bicep-params` filetype identifier from discussion at https://github.com/Azure/bicep/issues/17230

Resolves #3885

### DIFF
--- a/lsp/bicep.lua
+++ b/lsp/bicep.lua
@@ -32,7 +32,7 @@
 ---     && unzip -d /usr/local/bin/bicep-langserver bicep-langserver.zip)
 --- ```
 return {
-  filetypes = { 'bicep' },
+  filetypes = { 'bicep', 'bicep-params' },
   root_markers = { '.git' },
   init_options = {},
 }
--- a/lua/lspconfig/configs/bicep.lua
+++ b/lua/lspconfig/configs/bicep.lua
@@ -1,6 +1,6 @@
 return {
   default_config = {
-    filetypes = { 'bicep' },
+    filetypes = { 'bicep', 'bicep-params' },
     root_dir = function(fname)
       return vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
     end,


damienpontifex:
Updated title with `feat(bicep)` I _think_ is the suggestion from the "lint / commit" job. 

I also updated the `lua/lspconfig/configs/` file for completeness, but this triggers failure on the "lint / lint". Let me know if this should not be included?

ddbhm:
Adding some info if that's ok after it's closed. I came here from google trying to get bicepparams working after succesfully getting regular bicep to work through the config from this repo. For some reason it would not work for me until adding the language_id manually like so:

```
  get_language_id = function(_, filetype)
    if filetype == "bicepparam" then
      return "bicep-params"
    else
      return "bicep"
    end
  end,
 ```
 
It's very much possible that this is due to misconfiguration on my part, but perhaps it will help someone else who ends up here.

justinmk:
That seems like a useful change to include in `lsp/bicep.lua`

damienpontifex:
I have this in my config for it to work

vim.filetype.add({
  extension = {
    bicep = 'bicep',
    bicepparam = 'bicep-params',
  }
})

ddbhm:
For me that is not enough, I need to set the language id. I'm using LazyVim. I'm not a super experienced Nvim user though so I'm still not sure whether this is a me thing or a general thing with the LazyVim setup and bicep LSP. I would't mind putting in a PR but if it works for everybody else without this change it might be superfluous.

justinmk:
A PR that adds the change from https://github.com/neovim/nvim-lspconfig/pull/3886#issuecomment-3107828526 seems fine.

PR #3887, opened 2025-06-04
fix(clangd): set offset_encoding correctly

someoneinjd:
Problem:
Neovim's native `Client:initialize` implementation does not support clangd's [utf8 offsets extension](https://clangd.llvm.org/extensions#utf-8-offsets), resulting in an inconsistency in `offsetEncoding` between clangd and Neovim when using `vim.lsp.config`.

Solution:
Add a custom `on_init` to set the `offset_encoding` correctly.

Fix https://github.com/neovim/neovim/issues/34203

### DIFF
--- a/lsp/clangd.lua
+++ b/lsp/clangd.lua
@@ -57,6 +57,9 @@ local function symbol_info()
   end, bufnr)
 end
 
+---@class ClangdInitializeResult: lsp.InitializeResult
+---@field offsetEncoding? string
+
 return {
   cmd = { 'clangd' },
   filetypes = { 'c', 'cpp', 'objc', 'objcpp', 'cuda', 'proto' },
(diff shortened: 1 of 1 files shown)


lsp/clangd.lua
```diff
@@ -77,6 +77,11 @@ return {
     },
     offsetEncoding = { 'utf-8', 'utf-16' },
   },
+  on_init = function(client, init_result)
+    if init_result.offsetEncoding then
+      client.offset_encoding = init_result.offsetEncoding
+    end
+  end,
```
lithammer, line 89: I would consider something like this to improve the type hints. Otherwise the lua-language-server will complain about `undefined-field`.

```lua
---@class ClangdInitializeResult: lsp.InitializeResult
---@field offsetEncoding? string

return {
  ...
  ---@param init_result ClangdInitializeResult
  on_init = function(client, init_result)
```

someoneinjd: Added

PR #3890, opened 2025-06-06
fix: restore :LspStart/LspStop behaviour without arguments

### LINKED ISSUE #3892: `LspStart` and `LspRestart` need an argument to run
### Description

With NeoVim 11.2, the new implementations for `LspStart` and `LspRestart` require an argument instead of starting every language servers available to the buffer. Is it possible to get them to work the same as they do on 11.1 and before?

toupeira:
### Description

After the refactoring in e4d1c8b for Neovim 0.11.2 these commands now require an argument.

This restores the previous behaviour where `:LspStart` defaults to enabling all servers matching the filetype of the current buffer, and `:LspStop` defaults to disabling all servers attached to the current buffer.

### Context

- https://github.com/neovim/nvim-lspconfig/pull/3734
- https://github.com/neovim/neovim/discussions/33978
- https://github.com/neovim/neovim/issues/28479#issuecomment-2941615809

fix #3892

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -112,9 +112,7 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
       end
     end
 
-    for _, name in ipairs(servers) do
-      vim.lsp.enable(name)
-    end
+    vim.lsp.enable(servers)
   end, {
     desc = 'Enable and launch a language server',
     nargs = '?',


plugin/lspconfig.lua
```diff
@@ -133,16 +145,28 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
...
-        vim.notify(("Invalid server name '%s'"):format(info.args))
+        vim.notify(("Invalid server name '%s'"):format(name))
       else
         vim.lsp.enable(name, false)
       end
     end
   end, {
     desc = 'Disable and stop the given client(s)',
-    nargs = '+',
+    nargs = '*',
```
toupeira, line 167: Noticed some inconsistencies here:

- We should probably use "servers" instead of "clients" in all user-facing descriptions etc.
- `:LspStop` takes multiple arguments but `:LspStart` only one (or zero now).

Happy to change these too in this PR if nobody objects.

plugin/lspconfig.lua
```diff
@@ -97,12 +97,24 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
...
+        if filetypes and vim.tbl_contains(filetypes, filetype) then
+          table.insert(servers, name)
+        end
+      end
     end
 
-    vim.lsp.enable(info.args)
+    for _, name in ipairs(servers) do
+      vim.lsp.enable(name)
+    end
```
justinmk: enable() accepts a list. this is a case where "configs" might be less confusing than "servers". we are enabling configs, which may eventually start a client and a server. we aren't really enabling "servers"

```suggestion
    vim.lsp.enable(servers)
```

justinmk (approved):

plugin/lspconfig.lua
justinmk: > We should probably use "servers" instead of "clients" in all user-facing descriptions etc

Only for cases where "server" is the least confusing name. Doesn't look like the case here.

In old versions of this repo "servers" (as opposed to "configs" or "clients") was used everywhere in a nonsensical way.

justinmk: >  `:LspStop` takes multiple arguments but `:LspStart` only one (or zero now).

let's finish this pr first

justinmk:
Thanks! Is this needed for :LspRestart too? #3892

plugin/lspconfig.lua
toupeira: > Only for cases where "server" is the least confusing name. Doesn't look like the case here.
> 
> In old versions of this repo "servers" (as opposed to "configs" or "clients") was used everywhere in a nonsensical way.

@justinmk okay I figured maybe for users "servers" would make the most sense (and I did just mean changing the user-facing parts, not internal variable names), but you definitely have more insight here :grinning: 

>  :LspStop takes multiple arguments but :LspStart only one (or zero now).

I pushed https://github.com/neovim/nvim-lspconfig/pull/3896 for that now.

Issue #3768, opened 2025-04-23
upstream `root_pattern` , `insert_package_json`, `strip_archive_subpath` ?


justinmk:
### Description

These are the last 3 util functions that haven't been deprecated:

https://github.com/neovim/nvim-lspconfig/blob/79c12bc6e789dc3774a55289cc8fdedbd3b8ca38/lua/lspconfig/util.lua#L41

https://github.com/neovim/nvim-lspconfig/blob/79c12bc6e789dc3774a55289cc8fdedbd3b8ca38/lua/lspconfig/util.lua#L61

https://github.com/neovim/nvim-lspconfig/blob/79c12bc6e789dc3774a55289cc8fdedbd3b8ca38/lua/lspconfig/util.lua#L80

Need to decide if they should be upstremed to e.g. `vim.lsp.util` or `vim.fs`, in some form. They are too cumbersome to inline in each config.

justinmk:
Many `root_pattern()` usages can be eliminated after:

- https://github.com/neovim/neovim/pull/33771

PR #3891, opened 2025-06-06
feat(eslint): do not use lspconfig.util

amonraRivers:
The lspconfig.util library is no longer used. This fixes issues importing lspconfig.util while configuring eslint.


justinmk:
Thanks, but inlining the code from util isn't part of the plan, that just causes a different problem (duplicated code in numerous different configs). 

The plan is outlined in https://github.com/neovim/nvim-lspconfig/issues/3768

amonraRivers:
Thank you!

PR #3893, opened 2025-06-07
fix(helm_ls,yamlls): support custom filetype for helm-ls

qvalentin:
also in legacy config, fixup for https://github.com/neovim/nvim-lspconfig/pull/3881


justinmk:
these configs are frozen and will be deleted.

PR #3895, opened 2025-06-09
fix: support :LspRestart without arguments

toupeira:
### Description

Problem:
After the refactoring in e4d1c8b for Neovim 0.11.2 this command now requires an argument.

Solution:
Restore the previous behaviour where `:LspRestart` defaults to restarting all active servers.

### Context

- Follow-up to https://github.com/neovim/nvim-lspconfig/pull/3890
- Fixes https://github.com/neovim/nvim-lspconfig/issues/3892

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Most of the time, the reason for failure is present in the logs.
 * `:LspInfo` (alias to `:checkhealth vim.lsp`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config.
 * `:LspStop [<client_id_or_name> ...]` Stops the given server(s). Defaults to stopping all servers active on the current buffer. To force stop add `++force`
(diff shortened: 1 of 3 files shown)


toupeira:
@justinmk @haunt98 sorry I missed the `LspRestart` command in https://github.com/neovim/nvim-lspconfig/pull/3890 :grinning: 

I wasn't using this command so not 100% sure on the behaviour, but if I'm reading the old code correctly it restarted all active servers, not just those attached to the current buffer:

- https://github.com/neovim/nvim-lspconfig/blob/a182334ba933e58240c2c45e6ae2d9c7ae313e00/plugin/lspconfig.lua#L196
- https://github.com/neovim/nvim-lspconfig/blob/a182334ba933e58240c2c45e6ae2d9c7ae313e00/lua/lspconfig/util.lua#L380

Is that correct?

haunt98:
Hi @toupeira , my neovim config only has golsp (lsp for go) so my usecase for LspRestart is only restart golsp quickly.

justinmk:
> but if I'm reading the old code correctly it restarted all active servers, not just those attached to the current buffer:

I guess so. If people want granularity they can pass the config name.

PR #3896, opened 2025-06-09
chore: don't accept multiple arguments in :LspStop/LspRestart

toupeira:
### Description

We don't want to support multiple arguments in the upcoming `:Lsp` multi-command, so remove this now from `:LspStop` and `:LspRestart`.

### Context

- Follow-up from https://github.com/neovim/nvim-lspconfig/pull/3890#discussion_r2132402506

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,8 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (alias to `:checkhealth vim.lsp`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config.
-* `:LspStop [<client_id_or_name> ...]` Stops the given server(s). Defaults to stopping all servers active on the current buffer. To force stop add `++force`
-* `:LspRestart [<client_id_or_name> ...]` Restarts the given client(s), and attempts to reattach to all previously attached buffers. Defaults to restarting all active servers.
+* `:LspStop [<client_id_or_name>]` Stops the given server. Defaults to stopping all servers active on the current buffer. To force stop add `++force`
+* `:LspRestart [<client_id_or_name>]` Restarts the given client, and attempts to reattach to all previously attached buffers. Defaults to restarting all active servers.
 
 ## Contributions
 
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -149,8 +149,8 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
       end
     end)
   end, {
-    desc = 'Restart the given client(s)',
-    nargs = '*',
+    desc = 'Restart the given client',
+    nargs = '?',
     complete = complete_client,
   })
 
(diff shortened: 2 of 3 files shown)


justinmk:
I would prefer that neither of them supported multiple args. What is the use-case? These commands should not be used in scripts. Interactive usage is typically not going to specify multiple names. 

Scripts should use `vim.lsp`, not these commands. Commands are for interactive use.

toupeira:
@justinmk that sounds good to me too, I just added this for consistency with the `LspStop` and `LspRestart` commands.

I personally don't have a use case, but I'm also not sure if we should change the behaviour of `LspStop` / `LspRestart` now since I guess that would technically be a breaking change. Maybe I should leave this alone, and then any new behaviour can be implemented in the upstreamed `:Lsp` multi-command?

(if you agree feel free to close this PR :grinning:)

justinmk:
> but I'm also not sure if we should change the behaviour of `LspStop` / `LspRestart` now since I guess that would technically be a breaking change.

We should make that change. The goal is to get the code in shape and then upstream it to core Nvim.

toupeira:
@justinmk ok I changed this now! I hope nobody's relying on this, but I guess all these commands will get deprecated soon anyway :grinning: 

I noticed that with `nargs = '?'` and passing multiple arguments, `info.fargs` then just contains a single string (e.g. `foo bar`). So  I didn't add any special error handling and users will just see the notification `Invalid server name 'foo bar'`, which I guess is good enough?

Let me know if there's an easy way to avoid this, i.e still getting the split arguments in `fargs`.

README.md
```diff
@@ -153,8 +153,8 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (alias to `:checkhealth vim.lsp`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config.
-* `:LspStop [<client_id_or_name> ...]` Stops the given server(s). Defaults to stopping all servers active on the current buffer. To force stop add `++force`
-* `:LspRestart [<client_id_or_name> ...]` Restarts the given client(s), and attempts to reattach to all previously attached buffers. Defaults to restarting all active servers.
+* `:LspStop [<client_id_or_name>]` Stops the given server. Defaults to stopping all servers active on the current buffer. To force stop add `++force`
```
justinmk, line 156: off-topic/future: why didn't we use `!` instead of `++force` to "force" ...?

plugin/lspconfig.lua
```diff
@@ -175,8 +175,8 @@ if vim.version.ge(vim.version(), { 0, 11, 2 }) then
       end
     end
   end, {
-    desc = 'Disable and stop the given client(s)',
-    nargs = '*',
+    desc = 'Disable and stop the given client',
+    nargs = '?',
```
justinmk, line 179: Oh, I thought it would simplify the implementation too :) 

well, let's try it for now and see how it goes.

toupeira: Yeah we still need to handle multiple servers in the default case with no arguments :grinning: 

Thanks for the merge!

PR #3899, opened 2025-06-10
fix(svelte): only attach to existing files

igorlfs:
Similar to #3875, but this time I [checked](https://github.com/sveltejs/language-tools/issues/2777), and the server actually doesn't support schemas other than `file://`.

### DIFF
--- a/lsp/svelte.lua
+++ b/lsp/svelte.lua
@@ -12,7 +12,14 @@
 return {
   cmd = { 'svelteserver', '--stdio' },
   filetypes = { 'svelte' },
-  root_markers = { 'package.json', '.git' },
+  root_dir = function(bufnr, on_dir)
+    local root_files = { 'package.json', '.git' }
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    -- Svelte LSP only supports file:// schema. https://github.com/sveltejs/language-tools/issues/2777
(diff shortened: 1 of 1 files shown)


justinmk:
Does `requires_workspace=true` fix the issue? Even if the server claims to not require a workspace, if it requires a file on-disk then we will just treat it as requiring a workspace. The distinction is just silly and having to add special-case code like this as a workaround is not worth the maintenance cost. 

Is this a separate issue from https://github.com/neovim/neovim/issues/34208#issuecomment-2949131383 , i.e.:

> Would it help if root() returns an absolute path instead of `''` or `'.'` ?

igorlfs:
> Does requires_workspace=true fix the issue?

No, it doesn't.

> Is this a separate issue from https://github.com/neovim/neovim/issues/34208#issuecomment-2949131383 , i.e.:

Unfortunately, it is. At first, I thought they were the same, too. But the Svelte LSP doesn't care about the `root_dir` at all, it just won't work with virtual files.

>  and having to add special-case code like this as a workaround is not worth the maintenance cost.

I'm fine with keeping this to my own config if you consider it's not worth it adding here. Let me know 👍🏼

justinmk:
Need to think about a more maintainable approach, but this is helpful meanwhile. Thanks

PR #3900, opened 2025-06-11
fix(eslint): detect Yarn PnP in function form cmd

### LINKED ISSUE #3858: Unable to find ESLint library when using Yarn2 (PnP)
### Description

When working in a project which uses Yarn2 (also known as PnP or Plug and Play) and then opening a for example a `.js` file, the ESLint LSP tries to start but can't, printing this warning:
```
[lspconfig] Unable to find ESLint library.
```

When using Yarn2, the eslint executable is not `node_modules` but instead in `.yarn/sdks`. 

This can be fixed by hardcoding the following setting:
```lua
vim.lsp.config('eslint', {
  settings = { nodePath = '.yarn/sdks' },
})
```

That is an okay solution, but not good if working in a multiple projects where some use Yarn2 and some use npm. It would be great if the config could detect if Yarn2 or something else is used and set the `nodePath` according to that.

When looking in the [eslint.lua](https://github.com/neovim/nvim-lspconfig/blob/943a0556638fb809d3f72f1f091454bd56b23021/lsp/eslint.lua#L151) in this repo, it seems there is a solution for that, but it's not working (at least not for me). 

I tried the following but that did not work either.
```lua    
vim.lsp.config('eslint', {
  settings = { nodePath = 'node_modules' },
  before_init = function(_params, config)
    local yarn_lib = vim.fs.find('.yarn/sdks/eslint', { path = config.root_dir, upward = true })[1]
    if yarn_lib then
      config.settings.nodePath = '.yarn/sdks'
    end
  end
})
```

Does anyone have an idea on another solution for this? Could this be done?

### LINKED ISSUE #3889: `:LspRestart eslint` results in 'Unable to find ESLint library' when using Yarn PnP
### Description

(I'm using LazyVim + Mason pinned to v1. IIRC there are still some unresolved issues with the new vim.lsp ecosystem, so first, sorry if this issue doesn't belong here.)

lspconfig correctly adds `yarn exec` upon first start https://github.com/neovim/nvim-lspconfig/blob/036885e8e5456d3907626b634693234f628afef6/lua/lspconfig/configs/eslint.lua#L129:
```
> vim.lsp.get_clients({ name = 'eslint' })[1].config.cmd
< { "yarn", "exec", "vscode-eslint-language-server", "--stdio" }
> vim.lsp._enabled_configs
< {}
```
And I could confirm the command with `ps aux`.

However after executing `:LspRestart eslint`,
```
> vim.lsp.get_clients({ name = 'eslint' })[1].config.cmd
< { "yarn", "exec", "vscode-eslint-language-server", "--stdio" }
> vim.lsp._enabled_configs.eslint.resolved_config.cmd
< { "vscode-eslint-language-server", "--stdio" }
```
`vim.lsp._enabled_configs` somehow got populated with an `eslint.resolved_config`, where `cmd` didn't have the `yarn exec` prefix. I'd see an '[lspconfig] Unable to find ESLint library.' notification, and `ps aux` also said the vscode-eslint-language-server was executed with `node` directly.

I understand 'The configs in this repo are UNSUPPORTED and provided only as a starting point', but this might be some more generic issue with built-in LSP and config modifications in `on_new_config`?

I'm also aware of https://github.com/neovim/nvim-lspconfig/issues/3705 and https://github.com/neovim/neovim/issues/32287, but I'm not entirely sure if it's something that'll resolve this problem and whether there's anything we can do in the interim?

Frederick888:
---
name: Pull Request
about: Submit a pull request
title: fix(eslint): detect Yarn PnP in function form cmd
---
  

### [`c86804a3`](https://github.com/neovim/nvim-lspconfig/pull/3900/commits/c86804a3d197274fad9a75bbf2ec4297a3a5a879) fix(eslint): detect Yarn PnP in function form cmd

Problem:
In the new lsp/eslint.lua, the PnP detection in before_init did not work
at all as before_init cannot modify config.cmd [1].

Solution:
The code is moved to the new cmd function, which detects PnP by checking
.pnp.[c]js and prepends cmd with 'yarn exec' so that Node.js can load a
PnP project's ESLint library zip file for ESLint LS.

In the old lua/lspconfig/configs/eslint.lua, on_new_config still works,
though since [2] the new :Lsp* commands no longer utilise the old files.

Note since the cmd function does not have access to either bufnr or the
resolved root_dir [3], it only looks for .pnp.[c]js in current working
directory.

[1] https://github.com/neovim/neovim/blob/93925fe0/runtime/lua/vim/lsp/client.lua#L456-L466
[2] https://github.com/neovim/nvim-lspconfig/commit/e4d1c8bf9077ecf617ce18f27738658a7d76ac95
[3] https://github.com/neovim/neovim/blob/93925fe0/runtime/lua/vim/lsp.lua#L558

---
Closes #3858

Closes #3889

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -35,7 +35,20 @@ local util = require 'lspconfig.util'
 local lsp = vim.lsp
 
 return {
-  cmd = { 'vscode-eslint-language-server', '--stdio' },
+  cmd = function(dispatchers)
+    local eslint_cmd = { 'vscode-eslint-language-server', '--stdio' }
+    -- No bufnr or root_dir here. Always use cwd() for now.
+    local root_dir = vim.uv.cwd()
+
+    -- Support Yarn2 (PnP) projects
+    local pnp_cjs = root_dir .. '/.pnp.cjs'
+    local pnp_js = root_dir .. '/.pnp.js'
+    if vim.uv.fs_stat(pnp_cjs) or vim.uv.fs_stat(pnp_js) then
+      eslint_cmd = { 'yarn', 'exec', unpack(eslint_cmd) }
+    end
+
+    return vim.lsp.rpc.start(eslint_cmd, dispatchers)
+  end,
   filetypes = {
     'javascript',
     'javascriptreact',
@@ -146,14 +159,6 @@ return {
           config.settings.experimental.useFlatConfig = true
           break
         end
-      end
-
-      -- Support Yarn2 (PnP) projects
-      local pnp_cjs = root_dir .. '/.pnp.cjs'
-      local pnp_js = root_dir .. '/.pnp.js'
-      if vim.uv.fs_stat(pnp_cjs) or vim.uv.fs_stat(pnp_js) then
-        local cmd = config.cmd
-        config.cmd = vim.list_extend({ 'yarn', 'exec' }, cmd)
       end
     end
   end,


justinmk:
> Note since the cmd function does not have access to either bufnr or the  resolved root_dir [3], it only looks for .pnp.[c]js in current working directory.

That is a confusing thing to mention. Even if `cmd` had access to bufnr/root-dir, it is starting the LS. Obviously, you wouldn't want to start a new LS for every buffer, so what good would it do for `cmd` to know the bufnr?

I still would like to confirm my question from https://github.com/neovim/nvim-lspconfig/issues/3889#issuecomment-2959410872 : is this decision required as a CLI arg? Then if so, it makes sense to define `cmd` as you have done here. Please mention that in a one-line comment. And that also means there is no reason to mention that "cmd does not know bufnr", because that's irrelevant.

justinmk:
> Obviously, you wouldn't want to start a new LS for every buffer, so what good would it do for `cmd` to know the bufnr?

I think you may want to define a `reuse_client` function. Then `cmd` can use `nvim_get_current_buf()` instead of CWD ?

Frederick888:
> That is a confusing thing to mention. Even if `cmd` had access to bufnr/root-dir, it is starting the LS. Obviously, you wouldn't want to start a new LS for every buffer, so what good would it do for `cmd` to know the bufnr?

I agree it's not quite intuitive. Basically having bufnr/root_dir allows it to check if `root_dir/.pnp.[c]js` exists, so that users can e.g. open a file from any sub-directory in a project.

If you take my playground project in #3889, with this PR + the new `vim.lsp.enable('eslint')`, `nvim -u repro.lua packages/shared/src/index.ts` works while `cd packages/shared/src; nvim -u ../../../repro.lua ./index.ts` still has the 'Unable to find ESLint library' issue. (On the other hand with the old `require('lspconfig').eslint.setup({})`, both can work.)

> I still would like to confirm my question from [#3889 (comment)](https://github.com/neovim/nvim-lspconfig/issues/3889#issuecomment-2959410872) : is this decision required as a CLI arg?

Yes, somewhat. It's not an arg of the eslint LS itself but some args of `node` that `yarn exec` adds.

```sh
# still in my playground project
$ node
Welcome to Node.js v24.2.0.
Type ".help" for more information.
> process.env.NODE_OPTIONS
undefined

$ yarn exec node
Welcome to Node.js v24.2.0.
Type ".help" for more information.
> process.env.NODE_OPTIONS
'--require /home/frederick/Programming/Others/playground/.pnp.cjs --experimental-loader file:///home/frederick/Programming/Others/playground/.pnp.loader.mjs'
```

AFAIU `yarn exec` instructs `node` to load the `.pnp.[c]js` file upon start, which monkey patches `require`/`import` to allow loading dependency zip files, which replaces `node_modules/` in Yarn PnP. So eslint LS itself probably just always calls `const eslintLib = require('eslint')` or something, but without these additional node args, node only looks for `node_modules/`, which is nowhere to be found in a Yarn PnP project.

> I think you may want to define a `reuse_client` function. Then `cmd` can use `nvim_get_current_buf()` instead of CWD ?

I'm not familiar with this function but after having a glace at the docs, I assume it's something that gets called every time a new buffer of a related filetype (.js / .ts) is opened. So I assume you were talking about that in this `fun(client: vim.lsp.Client, config: vim.lsp.ClientConfig): boolean`, I can modify `config: vim.lsp.ClientConfig` and return `false` ~~once (be careful of >1 existing eslint LS client)~~, and here `config.cmd` modifications are actually effective since it happens before `Client.create(config)`?

If I'm not mistaken, unless `reuse_client` gets called when starting the first client too with `(nil, config)`, it only helps when opening new files? Plus if a user does something like `nvim package.json packages/shared/src/index.ts` then `:tab all`, `nvim_get_current_buf()` may not return the buffer we want?

justinmk:
> > I still would like to confirm my question from [#3889 (comment)](https://github.com/neovim/nvim-lspconfig/issues/3889#issuecomment-2959410872) : is this decision required as a CLI arg?
> 
> Yes, somewhat. It's not an arg of the eslint LS itself but some args of `node` that `yarn exec` adds.

Nevertheless, a one-line comment should mention that.


> unless `reuse_client` gets called when starting the first client too with `(nil, config)`, it only helps when opening new files?

`reuse_client` is not used for the first client because `cmd` of course will always be invoked for the first client. Thereafter, `reuse_client` decides whether to re-invoke `cmd`.

Frederick888:
> Nevertheless, a one-line comment should mention that.

Done.

> > unless `reuse_client` gets called when starting the first client too with `(nil, config)`, it only helps when opening new files?
> 
> `reuse_client` is not used for the first client because `cmd` of course will always be invoked for the first client. Thereafter, `reuse_client` decides whether to re-invoke `cmd`.

Then I think really it'd be nice if something simpler can come out of https://github.com/neovim/neovim/issues/32287 that allows constructing cmd according to root_dir. Then the default behaviour of reuse_client, i.e. reuse if root_dir is the same, should just work nicely too.

justinmk:
> it'd be nice if something simpler can come out of [neovim/neovim#32287](https://github.com/neovim/neovim/issues/32287) that allows constructing cmd according to root_dir. Then the default behaviour of reuse_client, i.e. reuse if root_dir is the same, should just work nicely too.

can you show a concrete example

meanwhile, why not use `reuse_client` in this config? Otherwise the CWD gets saved in `cmd` for all future buffers.

jsit:
FWIW, I tried this in my `init.lua` and am getting this error:

`LSP[eslint]: Error BEFORE_INIT_CALLBACK_ERROR: "vim/shared.lua:0: src: expected table, got function"`

```lua
vim.lsp.config("eslint", {
  cmd = function(dispatchers)
    local eslint_cmd = { 'vscode-eslint-language-server', '--stdio' }
    -- No bufnr or root_dir here. Always use cwd() for now.
    local root_dir = vim.uv.cwd()

    -- Support Yarn2 (PnP) projects
    local pnp_cjs = root_dir .. '/.pnp.cjs'
    local pnp_js = root_dir .. '/.pnp.js'
    if vim.uv.fs_stat(pnp_cjs) or vim.uv.fs_stat(pnp_js) then
      eslint_cmd = { 'yarn', 'exec', unpack(eslint_cmd) }
    end

    return vim.lsp.rpc.start(eslint_cmd, dispatchers)
  end,
})
```

justinmk:
> `LSP[eslint]: Error BEFORE_INIT_CALLBACK_ERROR: "vim/shared.lua:0: src: expected table, got function"`

Hmm, this pr removes some part of `before_init` so may need to figure that out before merging this.

Frederick888:
Sorry for the delay. I gave `reuse_client` a shot but wasn't able to get it working. And after encountering [yet another annoying PnP quirk](https://github.com/yarnpkg/berry/issues/6823), I now plan to just disable PnP and stop investing more time in it (not only this PR but also a few TODOs in my project). In theory it's a novel idea, in practice it's just of too much friction...

I'm afraid I'm gonna close this PR. Anyone else who's still willing to wrestle with the PnP dragon, feel free to take any of the code here :)

jsit:
I'm finding myself having to make this change locally to get ESLint working with Yarn2, as well. @Frederick888 Would you mind re-opening this PR, or should I create a new one?

Frederick888:
@jsit Please open a new PR. Also feel free reuse any of the code :)

PR #3903, opened 2025-06-12
Fix millet install target to millet-ls (#3902)

### LINKED ISSUE #3902: Millet instructions no target
### Description

Also referenced here https://github.com/mason-org/mason.nvim/issues/1097

https://github.com/neovim/nvim-lspconfig/blob/77d3fdfb3554632c7a3b101ded643d422de7626f/lua/lspconfig/configs/millet.lua#L20-L21

Cargo

error: no bin target named `lang-srv` in default-run packages
help: available bin targets:
    millet-cli
    millet-ls
    xtask

I think it should be millet-ls.

jxu:
Fixes #3902

### DIFF
--- a/lsp/millet.lua
+++ b/lsp/millet.lua
@@ -8,8 +8,8 @@
 ---
 --- 1. Install a Rust toolchain: https://rustup.rs
 --- 2. Clone the repo
---- 3. Run `cargo build --release --bin lang-srv`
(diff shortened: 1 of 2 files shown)


jxu:
Btw even after I have it installed, I can't get it to do anything. Maybe it is only designed to work with VS Code.

PR #3911, opened 2025-06-17
feat(glint): add vim.lsp.config support

justmejulian:
Saw this one was still missing in the list in #3705.

I tested it by creating a new ember ts project -> `ember new my-typescript-app --typescript`
then adding glint to that project -> `npm install --save-dev @glint/core`

Per default we use the project local glint-language-server, but the user can also choose to use a globally installed version.

Based on [lua/lspconfig/configs/glint.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/glint.lua)

### DIFF
--- a/lsp/glint.lua
+++ b/lsp/glint.lua
@@ -1,0 +1,63 @@
+---@brief
+---
+--- https://github.com/typed-ember/glint
+--- https://typed-ember.gitbook.io/glint/
+--- `glint-language-server` is installed when adding `@glint/core` to your project's devDependencies:
+---
+--- ```sh
+--- npm install @glint/core --save-dev
+--- yarn add -D @glint/core
+---
+--- This configuration uses the local installation of `glint-language-server`
+--- (found in the `node_modules` directory of your project).
+---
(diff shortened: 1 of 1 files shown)


lsp/glint.lua
```diff
@@ -0,0 +1,63 @@
...
+      return
+    end
+
+    local root_dir = config.root_dir
+    if not root_dir then
+      error('No root directory found for glint')
+    end
+    config.cmd = {
+      'exec',
+      root_dir .. '/node_modules/.bin/glint-language-server',
```
justinmk, line 43: I don't think this will do what is expected. `before_init` happens after the server was starting.

If `cmd` needs to be dynamic it must be defined as a function.

justinmk:
Might need an improvement as mentioned above, but can be a followup if you find the inspiration :)

lsp/glint.lua
justmejulian: Ah you are right. Looks like it just changed the config, but it did not have an effect.

PR #3912, opened 2025-06-17
feat(glint): dynamic cmd

justmejulian:
Follow up to #3911
@justinmk pointed out that it might not work as expected.

https://github.com/neovim/nvim-lspconfig/pull/3911#discussion_r2151857484

### DIFF
--- a/lsp/glint.lua
+++ b/lsp/glint.lua
@@ -22,7 +22,7 @@
 ---   },
 --- })
 
-function getCmd()
+function get_cmd()
   local useGlobal = vim.lsp.config.glint.init_options.glint.useGlobal
   if useGlobal then
     return { 'glint-language-server' }
@@ -45,7 +45,7 @@ end
 
 return {
   cmd = function(dispatchers)
-    local cmd = getCmd()
+    local cmd = get_cmd()
     return vim.lsp.rpc.start(cmd, dispatchers)
   end,
   init_options = {
--- a/lsp/glint.lua
+++ b/lsp/glint.lua
@@ -29,16 +29,7 @@ function get_cmd()
   end
 
   local root_markers = vim.lsp.config.glint.root_markers
-  if not root_markers or #root_markers == 0 then
-    vim.notify('No root markers defined for glint', vim.log.levels.ERROR)
-    return nil
-  end
-
-  local root_dir = vim.fs.dirname(vim.fs.find(root_markers, { upward = true })[1])
-  if not root_dir then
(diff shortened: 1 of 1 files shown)


lsp/glint.lua
```diff
@@ -22,27 +22,37 @@
 ---   },
 --- })
 
+function getCmd()
```
justinmk:
```suggestion
function get_cmd()
```

lsp/glint.lua
```diff
@@ -22,27 +22,37 @@
...
+  if not root_markers or #root_markers == 0 then
+    vim.notify('No root markers defined for glint', vim.log.levels.ERROR)
+    return nil
+  end
+
+  local root_dir = vim.fs.dirname(vim.fs.find(root_markers, { upward = true })[1])
+  if not root_dir then
+    vim.notify('Could not find root directory for glint', vim.log.levels.ERROR)
+    return nil
+  end
```
justinmk: this is kind of strange. `cmd()` is not going to be invoked unless the config was activated. Nvim already checks `root_markers` and `root_dir` to decide whether to activate a client for a given buffer.

when would this not be the case ? we should avoid reinventing the vim.lsp framework here.

justmejulian: oke so then I guess the checks are not needed.

Is there a way for me to access the client `root_dir` from within `get_cmd()`?
I could not find in in `vim.lsp`. 

Or would you recommend using lspconfig.util `root_pattern`?

I would prefer not having to find it again myself. / Would like it to match the client root_dir

justmejulian: Ah I found  `vim.fs.root`. Let me give that a try.

justmejulian: Oke I changed it to `vim.fs.root`. I works, but I would still prefer not having to find the root_dir myself.

lsp/glint.lua
```diff
@@ -22,27 +22,28 @@
...
 --- })
 
+function get_cmd()
+  local useGlobal = vim.lsp.config.glint.init_options.glint.useGlobal
+  if useGlobal then
+    return { 'glint-language-server' }
+  end
+
+  local root_markers = vim.lsp.config.glint.root_markers
+  local root_dir = vim.fs.root(0, root_markers)
```
justinmk, line 32: This still doesn't handle the case where a user might define `root_dir()`.

I have created https://github.com/neovim/neovim/issues/34547 to propose a potential solution. Feel free to send a PR for that if you feel like it.

PR #3914, opened 2025-06-18
fix(powershell_es): update LogLevel to Information

wenijinew:
LogLevel `Normal` of PowerShellEditorServices has been removed by the commit https://github.com/PowerShell/PowerShellEditorServices/commit/d9de5bd8b1ed624a687d701fa2616e5991db2e2c
The corresponding value after the change is `Information = 2`.


justinmk:
https://github.com/neovim/nvim-lspconfig/pull/3920 reports that this broke the config

PR #3916, opened 2025-06-19
fix: some `on_attach` callbacks may create commands in a wrong buffer

dmitmel:
This is a super simple fix, which addresses the bug that if a server is attached to a buffer other than the current one by means of `vim.lsp.buf_attach_client`, then the Language Server-related commands may be created in the current buffer, and not in the one the server has been attached to.

P. S. Does it need to be documented somewhere explicitly that the `on_attach` callback may not always be run within the context of the target buffer? There is an example of writing such a callback in `CONTRIBUTING.md` that correctly uses the `bufnr` passed in as the 2nd parameter, is this enough?

### DIFF
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -50,7 +50,7 @@ return {
       desc = 'Organize Imports',
     })
 
-    vim.api.nvim_buf_create_user_command(0, 'LspPyrightSetPythonPath', set_python_path, {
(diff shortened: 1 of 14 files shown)


justinmk:
Nice, thanks!

> P. S. Does it need to be documented somewhere explicitly that the `on_attach` callback may not always be run within the context of the target buffer? There is an example of writing such a callback in `CONTRIBUTING.md` that correctly uses the `bufnr`

Sure, adding a brief note there is worthwhile.

PR #3919, opened 2025-06-20
fix(omnisharp): missing root_dir patterns

horriblename:
---
name: Pull Request
about: Submit a pull request
title: ''
---

Adds missing root file patterns copied from csharp-ls: `*.csproj` and `*.sln`


lsp/omnisharp.lua
```diff
@@ -27,6 +29,15 @@ return {
   },
   filetypes = { 'cs', 'vb' },
   root_markers = { '.sln', '.csproj', 'omnisharp.json', 'function.json' },
+  root_dir = function(bufnr, on_dir)
```
justinmk, line 31: if root_dir is given then remove `root_markers`

lsp/omnisharp.lua
```diff
@@ -27,6 +29,15 @@ return {
...
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    on_dir(
+      util.root_pattern '*.sln'(fname)
+        or util.root_pattern '*.csproj'(fname)
+        or util.root_pattern 'omnisharp.json'(fname)
+        or util.root_pattern 'function.json'(fname)
```
justinmk, line 37: should omnisharp.json be highest precedence ?

horriblename: I couldn't find any information on this, but since .csproj files are the build files, I assume it would be more important?

Issue #3918, opened 2025-06-20
powershell_es wrong loglevel


bcarnazzi:
### Description

I've got this into my lsp.log With PowershellEditorServices 4.3.0: 

`[ERROR][2025-06-20 18:08:09] ...p/_transport.lua:36	"rpc"	"pwsh"	"stderr"	"\27[31;1mStart-EditorServices.ps1: \27[31;1mCannot validate argument on parameter 'LogLevel'. The argument \"Information\" does not belong to the set \"Diagnostic;Verbose;Normal;Warning;Error\" specified by the ValidateSet attribute. Supply an argument that is in the set and then try the command again.\27[0m\n"`

init.lua workaround (manually setting LogLevel to Normal): 

```
	local bundle_path = vim.fn.stdpath("data") .. "/PowerShellEditorServices"
	local temp_path = vim.fn.stdpath("cache")
	local command_fmt =
		[[& '%s/PowerShellEditorServices/Start-EditorServices.ps1' -BundledModulesPath '%s' -LogPath '%s/powershell_es.log' -SessionDetailsPath '%s/powershell_es.session.json' -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
	local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
	vim.lsp.config("powershell_es", {
		bundle_path = bundle_path,
		cmd = { "pwsh", "-NoLogo", "-NoProfile", "-Command", command },
	})
```

justinmk:
see discussion in https://github.com/neovim/nvim-lspconfig/pull/3920

PR #3920, opened 2025-06-20
fix(powershell_es): update LogLevel to recognized value

seejaysea-dev:
Problem:
Powershell Editor Services does not recognize `-LogLevel Information` and this prevents the LSP from starting

Solution:
Revert update to `command_fmt` to use a recognized log level from [Start-EditorServices.ps1](https://github.com/PowerShell/PowerShellEditorServices/blob/3a429eac5544fd7613c281c0814f94ac03fad551/module/PowerShellEditorServices/Start-EditorServices.ps1)

### DIFF
--- a/lsp/powershell_es.lua
+++ b/lsp/powershell_es.lua
@@ -44,7 +44,7 @@ return {
     local shell = vim.lsp.config.powershell_es.shell or 'pwsh'
 
     local command_fmt =
-      [[& '%s/PowerShellEditorServices/Start-EditorServices.ps1' -BundledModulesPath '%s' -LogPath '%s/powershell_es.log' -SessionDetailsPath '%s/powershell_es.session.json' -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Information]]
+      [[& '%s/PowerShellEditorServices/Start-EditorServices.ps1' -BundledModulesPath '%s' -LogPath '%s/powershell_es.log' -SessionDetailsPath '%s/powershell_es.session.json' -FeatureFlags @() -AdditionalModules @() -HostName nvim -HostProfileId 0 -HostVersion 1.0.0 -Stdio -LogLevel Normal]]
     local command = command_fmt:format(bundle_path, bundle_path, temp_path, temp_path)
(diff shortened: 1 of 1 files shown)


seejaysea-dev:
fixes bug #3918

sharpchen:
I think we should revert https://github.com/neovim/nvim-lspconfig/pull/3914

seejaysea-dev:
Ah, I didn't see that PR

seejaysea-dev:
Updated to revert commit

justinmk:
https://github.com/neovim/nvim-lspconfig/pull/3914 notes that `Normal` was removed in https://github.com/PowerShell/PowerShellEditorServices/commit/d9de5bd8b1ed624a687d701fa2616e5991db2e2c , do we need to decide this based on the LS version ?

sharpchen:
@justinmk The issue is the wrapper script `Start-EditorServices.ps1` has an outdated validation that doesn't allow new log levels, which caused the error. `Normal` will be deprecated in the future but currently still available, so we don't have to decide on ls version, once the pwsh_es updated the log level validation we can change it to `Information` again.

https://github.com/PowerShell/PowerShellEditorServices/blob/3a429eac5544fd7613c281c0814f94ac03fad551/module/PowerShellEditorServices/Start-EditorServices.ps1#L49

eitamal:
I've opened with [PowerShell/PowerShellEditorServices](https://github.com/PowerShell/PowerShellEditorServices/):
- **Issue:** PowerShell/PowerShellEditorServices#2235 
- **PR fixing the issue:** PowerShell/PowerShellEditorServices#2236

So, hopefully it'll be fixed soon :)

Maybe worth noting, they've added some conversion code for backwards compatibility: [StartEditorServicesCommand.cs](https://github.com/PowerShell/PowerShellEditorServices/blob/3a429eac5544fd7613c281c0814f94ac03fad551/src/PowerShellEditorServices.Hosting/Commands/StartEditorServicesCommand.cs#L268-L277) which covers `Normal` -> `Information` - so there shouldn't be any urgency in getting the change back to `Information` out (in case there was any).

seejaysea-dev:
Thanks @eitamal. I was hoping to get the time to do almost this exact PR today in PES

PR #3923, opened 2025-06-21
fix(eslint): only call `on_dir` when root directory is found

### LINKED ISSUE #3922: ⁠ESLint: "Unable to load ESLint library" error in JS projects without eslint configuration
## Summary

The ESLint LSP configuration in lspconfig incorrectly calls `on_dir` unconditionally, causing "Unable to load ESLint library." errors in JavaScript projects that don't use ESLint.

## Expected Behavior

Per neovim documentation, `on_dir` should only be called conditionally to start the LSP when a configuration file has been found.

> Example: To dynamically decide whether LSP is activated, define a [lsp-root_dir()](https://neovim.io/doc/user/lsp.html#lsp-root_dir()) function which calls on_dir() only when you want that config to activate:
> ```lua
> vim.lsp.config('lua_ls', {
>  root_dir = function(bufnr, on_dir)
>     if not vim.fn.bufname(bufnr):match('%.txt$') then
>       on_dir(vim.fn.getcwd())
>     end
>   end
> })
> ```

## Current Behavior

The ESLint configuration in lspconfig calls `on_dir` regardless of whether any configuration files are found:

```lua
root_files = util.insert_package_json(root_files, "eslintConfig", fname)
on_dir(vim.fs.dirname(vim.fs.find(root_files, { path = fname, upward = true })[1]))
```

This results in:
- The ESLint LSP server starting in JavaScript projects even when no ESLint configuration exists
- "Unable to load ESLint library." errors appearing when editing JavaScript files in projects that don't use ESLint

## Proposed Solution

The `on_dir` call should be wrapped in a conditional check to ensure configuration files exist before starting the LSP server.

danteissaias:
Closes #3922 

This fixes "Unable to load ESLint library" errors in JavaScript projects without eslint.

The eslint config was calling `on_dir` unconditionally, but per neovim docs it should only be called when configuration files exist:

> To dynamically decide whether LSP is activated, define a [[lsp-root_dir()](https://neovim.io/doc/user/lsp.html#lsp-root_dir())](https://neovim.io/doc/user/lsp.html#lsp-root_dir()) function which calls on_dir() only when you want that config to activate

Now `on_dir` is only called when a valid root directory is found.

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -80,7 +80,11 @@ return {
 
     local fname = vim.api.nvim_buf_get_name(bufnr)
     root_file_patterns = util.insert_package_json(root_file_patterns, 'eslintConfig', fname)
-    on_dir(vim.fs.dirname(vim.fs.find(root_file_patterns, { path = fname, upward = true })[1]))
+    local root_dir = vim.fs.dirname(vim.fs.find(root_file_patterns, { path = fname, upward = true })[1])
+    if not root_dir then
+      return
+    end
+    on_dir(root_dir)
   end,
   -- Refer to https://github.com/Microsoft/vscode-eslint#settings-options for documentation.
   settings = {


danteissaias:
Actually seems like this isn't needed since `workspace_required` was added (See: https://github.com/neovim/nvim-lspconfig/pull/3805)

PR #3925, opened 2025-06-21
feat: `herb_ls`

marcoroth:
Hey there, thanks for maintaining this useful project!

I started working on a set of language tools for HTML+ERB files which are being worked on the `herb` umbrella project:
https://github.com/marcoroth/herb

Let me know if the docstring has too much context/information, happy to cut it down if the given context is too much!

Thank you!

### DIFF
--- a/lsp/herb_ls.lua
+++ b/lsp/herb_ls.lua
@@ -1,0 +1,27 @@
+---@brief
+---
+--- https://www.npmjs.com/package/@herb-tools/language-server
+--- https://github.com/marcoroth/herb
+---
+--- HTML+ERB (HTML + Embedded Ruby)
+--- Powerful and seamless HTML-aware ERB parsing and tooling.
+---
+--- Herb is designed from the ground up to deeply understand `.html.erb` files,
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks for the well-formed contribution. Does this meet https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md#criteria

calebhearth:
I took a quick look, and I would say that it meets the criteria. The project has over 300 stars at present, and as the targeted language, ERB, is part of both Ruby and Rails, I would say that it has a very active user base.

marcoroth:
Thanks for merging, @justinmk!

albandiguer:
Thanks @marcoroth for the great LSP. Just a quick one, the filetypes include `ruby` and `html`. 
Most formatting tools would delegate the formatting to the lsp when available unless specified otherwise. 
By default this would cause the ruby files to be formatted by herb and break the file (ruby ends up in a one-liner). 
Was the intent to have this active only on eruby/erb files maybe ?

PR #3926, opened 2025-06-23
docs: Correct install command for pug lsp

glektarssza:
Problem:
Installing the pug lsp using the suggested method (`go get github.com/opa-oz/pug-lsp`) does not work in the latest versions of `go` and instead produces a message the "'go get' is no longer supported outside a module".

Solution:
Change the install command for the pug lsp to the correct command which is `go install github.com/opa-oz/pug-lsp@latest`.


glektarssza:
I am aware that `lua/lspconfig/configs/*` is deprecated but I changed the command in both locations for internal consistency.

PR #3927, opened 2025-06-24
feat(pyright): added bypass for marker check for small single python files

SunnyTamang:
I was constantly running into issues where if I have a folder with a single python file for a quick test, and the pyright lsp starts searching for root_marker files.

It spikes the memory usage upto ~4gigs and throws the error as below.
```[ERROR][2025-06-23 13:35:08] ...p/_transport.lua:36 "rpc" "pyright-langserver" "stderr" "FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory\n----- Native stack trace -----\n\n"```

And it was a small python file called test.py with just a print statment.
Hence created a small check where it checks for marker and if not found use the current directory as root directory instead of ~ .

### DIFF
--- a/lua/lspconfig/configs/pyright.lua
+++ b/lua/lspconfig/configs/pyright.lua
@@ -45,7 +45,12 @@ return {
     cmd = { 'pyright-langserver', '--stdio' },
     filetypes = { 'python' },
     root_dir = function(fname)
-      return util.root_pattern(unpack(root_files))(fname)
+      local home = vim.loop.os_homedir()
+      local root = util.root_pattern(unpack(root_files))(fname)
+      if root == nil or (root == home and vim.fn.fnamemodify(fname, ':h') ~= home) then
+        root = vim.fn.fnamemodify(fname, ':h')
+      end
+      return root
     end,
     single_file_support = true,
     settings = {

### CODE REVIEW

**Comment by justinmk** [comment]
> use the current directory as root directory instead of ~ .

well, both of those guesses are wrong. where in the old config was it using "~"?

please try the lsp/ config instead of this old lua/lspconfig/ config, using Nvim 0.11.3 or newer with `vim.lsp.config`. see readme for details. we don't support the old configs.

**Comment by SunnyTamang** [comment]
> > use the current directory as root directory instead of ~ .
> 
> well, both of those guesses are wrong. where in the old config was it using "~"?
> 
> please try the lsp/ config instead of this old lua/lspconfig/ config, using Nvim 0.11.3 or newer with `vim.lsp.config`. see readme for details. we don't support the old configs.


You can see the screenshot below. Though i was in a different directory.
Screenshot 2025-06-24 at 9 05 35 PM

If I have the marker than its works fine but if marker not present, it fails as memory swells up.

PS:

I am using 0.11.2 version of neovim. 0.11.3 is out??

**Comment by SunnyTamang** [comment]
Also this is my configuration of nvim-lspconfig

```{
		"neovim/nvim-lspconfig",
		lazy = false,
		config = function()
			local capabilities = require("cmp_nvim_lsp").default_capabilities()

			lspconfig.pyright.setup({
				capabilities = capabilities,})
		
			vim.keymap.set("n", "K", vim.lsp.buf.hover, {})
			vim.keymap.set("n", "gd", vim.lsp.buf.definition, {})
			vim.keymap.set("n", "gt", vim.lsp.buf.references, {})
			vim.keymap.set({ "n", "v" }, "<leader>ca", vim.lsp.buf.code_action, {})
		end,
	},
```

So what you are saying that having a simple ```-- vim.lsp.enable('pyright')``` in init.lua will work??

We dont have to set nvim-lspconfig like above??

justinmk:
any 0.11.x Nvim supports `vim.lsp.config`

> So what you are saying that having a simple `-- vim.lsp.enable('pyright')` in init.lua will work??

I'm saying to try it, and if there is a problem then fix the problem in the new config: https://github.com/neovim/nvim-lspconfig/blob/master/lsp/pyright.lua

change your config to:

```
vim.lsp.config('pyright', { capabilities = capabilities })
```

SunnyTamang:
> any 0.11.x Nvim supports `vim.lsp.config`
> 
> > So what you are saying that having a simple `-- vim.lsp.enable('pyright')` in init.lua will work??
> 
> I'm saying to try it, and if there is a problem then fix the problem in the new config: https://github.com/neovim/nvim-lspconfig/blob/master/lsp/pyright.lua
> 
> change your config to:
> 
> ```
> vim.lsp.config('pyright', { capabilities = capabilities })
> ```

I tried using capabilites from ```local capabilities = require("cmp_nvim_lsp").default_capabilities(),``` and passing it as 
```vim.lsp.config('pyright', { capabilities = capabilities })``` but again same problem it starts from ~ direcrtory and goes heap out of memoory

```[ERROR][2025-06-25 16:24:34] ...p/_transport.lua:36	"rpc"	"pyright-langserver"	"stderr"	"FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory\n----- Native stack trace -----\n\n"
[ERROR][2025-06-25 16:24:34] ...p/_transport.lua:36	"rpc"	"pyright-langserver"	"stderr"	" 1: 0x1046859b4 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [/usr/local/bin/node]\n"
[ERROR][2025-06-25 16:24:34] ...p/_transport.lua:36	"rpc"	"pyright-langserver"	"stderr"	" 2: 0x10485405c v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [/usr/local/bin/node]\n"
```

screnrecord:

https://drive.google.com/file/d/1vTREMqX4NqbV7oRiBhLeplL7_8KHVDWj/view?usp=sharing


However, if I use the below config, it works fine as the root_markers is blank and picks the cwd as well.

```
vim.lsp.config("pyright", {
			cmd = { "pyright-langserver", "--stdio" },
			filetypes = { "python" },
			root_markers = {},
			settings = {
				python = {
					analysis = {
						autoSearchPaths = true,
						useLibraryCodeForTypes = true,
						diagnosticMode = "openFilesOnly",
					},
				},
			},
		}),
```

justinmk:
> it works fine as the root_markers is blank and picks the cwd

well, are any of these root markers in your home dir...? https://github.com/neovim/nvim-lspconfig/blob/6bba673aa8993eceec233be17b42ddfb9540794b/lsp/pyright.lua#L25C3-L33

> if I have a folder with a single python file for a quick test, and the pyright lsp starts searching for root_marker files.

So the situation is that your home dir (but not your workspace) happens to have a matching root marker.

This situation could apply to any LS, not just "pyright", so adding a workaround in a random config isn't the right approach.

SunnyTamang:
> well, are any of these root markers in your home dir...? https://github.com/neovim/nvim-lspconfig/blob/6bba673aa8993eceec233be17b42ddfb9540794b/lsp/pyright.lua#L25C3-L33
> 
> >

Oh yeah one marker was present in my home dir, requirements.txt
Ok understood why this behaviour.

But what is the purpose having those markers in the first place??
Correct me if i am wrong here, my understanding was wherever the marker is found it will use that dir as a working dir/workspace.

>This situation could apply to any LS, not just "pyright", so adding a workaround in a random config isn't the right approach.
>
>>
I was thinking of some fallback mechanism where if the marker is not found in the current dir and the marker happens to be present in the home dir, and there was some discrepancy in the home dir and current workspace dir and hence consider using the current working dir as the current workspace. ( that was my whole motive for the workaround)

justinmk:
> my understanding was wherever the marker is found it will use that dir as a working dir/workspace.

yes. how else can a workspace be determined for a given buffer ?

> I was thinking of some fallback mechanism ...

but let's make that change in Nvim core lsp framework, not in this random config.

SunnyTamang:
> but let's make that change in Nvim core lsp framework, not in this random config.
>
>>
I was not aware that it is deprecated or may be I have missed that in README. Hence I made the changes in the old lspconfig folder.

Just for confirmation, anything inside lua/lspconfig dir is deprecated right?

justinmk:
> Just for confirmation, anything inside lua/lspconfig dir is deprecated right?

yes, but that is unrelated to my suggestion to make this decision in Nvim *core*, not a specific *config*.

PR #3929, opened 2025-06-25
update config example in readme

RileyMathews:
I think this example in the readme of the new neovim 0.11 syntax is wrong. When I went to migrate my config to the new syntax I first tried starting with this example which didn't work. I then looked at the pyright config further down in the readme and saw that example didn't have the nested `settings = {server_name = ...` table in it. When I tried making my configs look like that things started working. Please let me know if I am just misunderstanding something about this way of configuration though!


RileyMathews:
The force-push was just to push an amended commit message that passed the lint check

sharpchen:
`settings` field is for communicating with language servers with your custom setting, it can be any shape that language server specified, it's generally optional. So there's no mistake here.

RileyMathews:
Gotcha. I must have had something else going on with my config then when I was having issues. Sorry for the bother!

Issue #3930, opened 2025-06-27
Intelephense LSP config update broke the support for multiple composer.json files


pimpeters:
### Description

Having multiple composer.json files within a project seems to misbehave after the following change: fa91d1c90433aaf68c6a2774d5ffce7319effa09

i.e. having a `composer.json` file in the root and one in `Test/composer.json` will now not include both `composer.json` files for LSP usage, but just the closest one.
Files within the `Test/` dir will use the `Test/composer.json` and the ones outside `Test/` will use the root one.

This isn't the way the LSP should be behaving when using PHP, seems like this change was done without really understanding how modern PHP apps work nowadays, seeing https://github.com/neovim/nvim-lspconfig/issues/3901#issuecomment-2965522593

It was behaving correctly before this commit, but updating this plugin (sadly) broke the entire workflow.

justinmk:
> will now not include both `composer.json` files for LSP usage

what does this mean? how did the [old code](https://github.com/neovim/nvim-lspconfig/commit/fa91d1c90433aaf68c6a2774d5ffce7319effa09) "include both files"? 

and if this is something required bt php lsp, it should be explained in the config (as a code comment), because it wasn't clear at all.

i'm not quite sure what you actually are saying though. do you have a docs reference ?

pimpeters:
After further investigation, I realized only one root directory is being used.
Both the old `root_dir` function and the new `root_markers` setup resolve to the same path. 
So, the behavior difference likely isn’t due to the root detection logic, but possibly something internal to Intelephense or specific to our PHP project setup. 
I also tested Intelephense in VS Code (default config) and it behaves the same as before the recent Intelphense config change. I’ll keep investigating whether this is a broader issue or unique to our environment.

pimpeters:
Did some investigation, looks like [vscode-intelephense](https://github.com/bmewburn/vscode-intelephense) is always using the workspace root as base for the LSP.
There isn't any docs about in which dir Intelphense should be used as the [docs](https://github.com/bmewburn/intelephense-docs) are mainly targeted towards VS Code usage, but maybe @bmewburn can enlighten us on this question.

Did some tests and removing `composer.json` from the `root_markers` (as it will always use the "git" root this way) and it does result in my situation to intended behavior.

I'm wondering if @pogyomo (https://github.com/neovim/nvim-lspconfig/issues/3901#issuecomment-2965522593) is having a similar situation when remove `composer.json` from these root markers.

[...]

PR #3933, opened 2025-06-30
fix(intelephense): change root markers order

pimpeters:
See discussion: https://github.com/neovim/nvim-lspconfig/issues/3930#issuecomment-3018465294

Prioritize `.git` above the `composer.json`, as in PHP projects sometimes multiple composer.json files are used. 
Having `composer.json` above `.git` will prioritize the closest `composer.json`, leading into some unintended behavior when having multiple `composer.json` files within your project. 
(i.e. when using module strategy which each their own composer file)
The common strategy is to autoload these files in the root composer.json.

This will also resolve problems similar to this: 477f412973484b4de20ae70c142e28a5790936a5

This solve a bug introduced in fa91d1c

**References:**
Basic usage/placement of a composer.json file:
https://getcomposer.org/doc/01-basic-usage.md#composer-json-project-setup 
> To start using Composer in your project, all you need is a composer.json file. This file describes the dependencies of your project and may contain other metadata as well. It typically should go in the **top-most directory** of your project/VCS repository.

Autoloading strategy:
https://getcomposer.org/doc/01-basic-usage.md#autoloading

[phpactor](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/phpactor.lua) using the same approach for similar reason

### DIFF
--- a/lsp/intelephense.lua
+++ b/lsp/intelephense.lua
@@ -28,5 +28,5 @@
 return {
   cmd = { 'intelephense', '--stdio' },
   filetypes = { 'php' },
-  root_markers = { 'composer.json', '.git' },
+  root_markers = { '.git', 'composer.json' },
 }


justinmk:
> This will also resolve problems similar to this: [477f412](https://github.com/neovim/nvim-lspconfig/commit/477f412973484b4de20ae70c142e28a5790936a5)

That's a good reference. It makes sense that the PHP configs follow the same pattern.

PR #3935, opened 2025-07-03
feat(tailwindcss): tailwind root_dir rails phoenix

studious:
The location of the config in Rails and Phoenix does not represent the root of the project. We can look inside mix.lock and Gemfile.lock for the existence of tailwind, which guarantees that it is the root.

### DIFF
--- a/lsp/tailwindcss.lua
+++ b/lsp/tailwindcss.lua
@@ -117,24 +117,16 @@ return {
       'postcss.config.cjs',
       'postcss.config.mjs',
       'postcss.config.ts',
-      -- Phoenix
-      'assets/tailwind.config.js',
-      'assets/tailwind.config.cjs',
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks for the clear explanation.

lesm:
Thanks guys, after these changes tailwindcss-lsp is working again (in rails views)

But only works for classes like: `class="flex flex"` but not for `class: 'flex flex'`

If someone is having the same problem we can update the settings like:

```
vim.lsp.config("tailwindcss", {
  settings = {
    tailwindCSS = {
      experimental = {
        classRegex = {
	  "class:? [\\\"\\']([^\\\"\\']*)[\\\"\\']", -- Dynamic class strings
	},
      },
    },
  },
})

```

PR #3938, opened 2025-07-05
feat(ts_ls): implement code lens support

sybereal:
Implement the `editor.action.showReferences` client command, which is used by the references and implementation code lenses that TypeScript Language Server provides.[^1]
I've mostly based the implementation on the existing code for `vim.lsp.buf.references()`. The main visible difference is that the latter includes the item being referenced at the top of the list. Although, if desired, this could be emulated by manually inserting the position passed along in the command arguments at the top of the list.

[^1]: https://github.com/typescript-language-server/typescript-language-server?tab=readme-ov-file#code-lenses-textdocumentcodelens

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -97,6 +97,32 @@ return {
       return vim.NIL
     end,
   },
+  commands = {
+    ['editor.action.showReferences'] = function(command, ctx)
+      local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
+      local file_uri, position, references = unpack(command.arguments)
+
+      local quickfix_items = vim.lsp.util.locations_to_items(references, client.offset_encoding)
+      vim.fn.setqflist({}, ' ', {
+        title = command.title,
+        items = quickfix_items,
+        context = {
+          command = command,
+          bufnr = ctx.bufnr,
+        },
+      })
+
+      vim.lsp.util.show_document({
+        uri = file_uri,
+        range = {
+          start = position,
+          ['end'] = position,
+        },
+      }, client.offset_encoding)
+
+      vim.cmd('botright copen')
+    end,
+  },
   on_attach = function(client, bufnr)
     -- ts_ls provides `source.*` code actions that apply to the whole file. These only appear in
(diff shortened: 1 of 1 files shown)


justinmk:
Looks like the right approach. How does one use this? Does the default `gra` mapping (`vim.lsp.buf.code_action()`) include extra items after this? I'm guessing not, but it probably should. (Not a blocker for this PR, of course.)

sybereal:
This does not interact with code actions, but rather code _lenses_.

The TypeScript language server provides code lenses that look like "N references M implementations". Triggering these via `vim.lsp.codelens.run()` invokes a LSP command named `editor.action.showReferences`, as described in the link. This is a command that has to be implemented on the LSP client side, which is what this PR does.

Compared to `vim.lsp.buf.references()`, it avoids a server roundtrip, as the reference data is already part of the code lens object.

justinmk:
> This does not interact with code actions, but rather code _lenses_.

Yes, but it might make sense to include codelenses in the "code actions" menu.

sybereal:
Is that how code lenses usually handled? I'm sorry, I wasn't aware. I just have a mapping for `vim.lsp.codelens.run()` and an autocmd to regularly refresh them, like the help suggests.

What would that action item do then? Would it just be an alias to `vim.lsp.codelens.run()` as well?

justinmk:
> I just have a mapping for `vim.lsp.codelens.run()` and an autocmd to regularly refresh them, like the help suggests.

Yeah that's not very discoverable, I doubt most users are aware of that :)

> What would that action item do then? Would it just be an alias to `vim.lsp.codelens.run()` as well?

I assume it would run the codelens that the user chooses from the `gra` menu. https://github.com/neovim/neovim/issues/34813

sybereal:
> Yeah that's not very discoverable, I doubt most users are aware of that :)

Very good point. I did create a mapping for `vim.lsp.codelens.run()` when I first switched to native LSP, because I already had it previously, but it took me until recently to dig through the docs and finally figure out that I also need an autocmd running `vim.lsp.codelens.refresh()` to make code lenses actually show up.

> I assume it would run the codelens that the user chooses from the gra menu.

That makes sense. Would be nice to have, yeah.

Issue #3940, opened 2025-07-07
[gopls]: a new client is started when opening files in the std lib path


balpert89:
### Description

Hi there,

I stumbled upon a weird issue with `gopls` and how it handles / registers files outside of the configured root dir.

If a file resides outside of the root_dir (commonly files in mod cache *or the standard library files found in GOROOT*) using vanilla lsp configuration, I can see a started client for each directory of my current project, mod cache and GOROOT. This results in weird behavior how diagnostics are shown, false-positive messages where variables could not be found and so on:

<img width="1650" height="117" alt="Image" src="https://github.com/user-attachments/assets/ae8d9d1f-f225-4704-92e0-48996254ede5" />

Using the configuration at https://github.com/neovim/nvim-lspconfig/blob/7fac9025a967a4d0846660f751cd392fac6bb788/lsp/gopls.lua this is solved partially as the `get_root` function checks for mod cache files if there already is a client for `gopls`, otherwise it creates one. What is missing is the same check for std lib files. I have implemented a fix [in my own configuration](https://www.reddit.com/r/neovim/comments/1lqzkur/comment/n1b39aq/).

Question: does it make sense to open a PR to upstream this?

justinmk:
yes

vivindeena:
@balpert89 can you share how you set the root_dir variable in your config

EmilSodergren:
@vivindeena Not sure exactly what configuration framework you use but I struggled with this for some weeks before I found this thread and finally got it working again. 
This solved my issues in my Lazy.nvim config, best of luck!
```lua
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        golangci_lint_ls = {
          cmd = { os.getenv("HOME") .. "/go/bin/golangci-lint-langserver" },
          root_dir = function(bufnr, ondir)
            local fname = vim.api.nvim_buf_get_name(bufnr)
            ondir(require("lspconfig.util").root_pattern("go.mod", "go.work", ".git")(fname))
          end,
          init_options = {
            command = { "golangci-lint", "run", "--output.json.path=stdout", "--show-stats=false" },
          },
        },
      },
    },
  },
```

vivindeena:
thanks a lot will look it to it

jgonmor16:
Doesn't reproduce on current master (43ed3797).

Fixed by 592916db (#3942, @balpert89), which added the GOROOT check to
[...]

PR #3942, opened 2025-07-07
feat(gopls): refactor root_dir identification #3940

balpert89:
Refactor the logic on how to identify the root_dir for gopls. 

This will provide a viable solution to the issue where multiple gopls lsp client instances are started, causing adverse effects in certain areas like diagnostics, lsp functionality (such as `go to definition`, workspace-enabled projects).

The identification for required directories are based on golang-specific environment variables which are set by default.


lsp/gopls.lua
```diff
@@ -4,13 +4,82 @@
...
 local mod_cache = nil
+local std_lib = nil
+
+---@param custom_args go_dir_custom_args
+---@param on_complete fun(dir: string | nil)
+local function identify_go_dir(custom_args, on_complete)
+  local cmd = { 'go', 'env', custom_args.envvar_id }
+  vim.system(cmd, { text = true }, function(output)
+    if output.code == 0 then
+      if output.stdout then
```
justinmk: what happens if `output.stdout` is nil? should it be like this, so that an error is shown ?  
```suggestion
    if output.code == 0 and output.stdout then
```

justinmk:
thanks! ideally this would be much less code, but i don't see any obvious ways to do that.

lsp/gopls.lua
balpert89: yes, the nested if can be put together. `GOMODCACHE` and `GOROOT` will always be set in a system where go is installed.

So the only situation I can think of where `output.stdout` is empty would be in a system without go installed.

Though, on my system `output.stdout` is never `nil`, its just an empty string when I pass in a non-existing env var, e.g. `GOBLABLA`. I suggest to rewrite to

```
    if output.code == 0 and output.stdout and output.stdout ~= "" then
```

It is a bit verbose but catches all non-happy paths.

balpert89: applied.

balpert89:
> thanks! ideally this would be much less code, but i don't see any obvious ways to do that.

I agree, I tried to de-duplicate as much as possible / where it makes sense. Hopefully it is not a problem.

lsp/gopls.lua
```diff
@@ -4,13 +4,79 @@
...
+  local cmd = { 'go', 'env', custom_args.envvar_id }
+  vim.system(cmd, { text = true }, function(output)
+    if output.code == 0 and output.stdout and vim.trim(output.stdout) ~= '' then
+      local res = vim.trim(output.stdout)
+      if custom_args.custom_subdir and custom_args.custom_subdir ~= '' then
+        res = res .. custom_args.custom_subdir
+      end
+      on_complete(res)
+    else
+      vim.schedule(function() end)
```
justinmk: what's this for ?

lsp/gopls.lua
```diff
@@ -4,13 +4,79 @@
...
 local mod_cache = nil
+local std_lib = nil
+
+---@param custom_args go_dir_custom_args
+---@param on_complete fun(dir: string | nil)
+local function identify_go_dir(custom_args, on_complete)
+  local cmd = { 'go', 'env', custom_args.envvar_id }
+  vim.system(cmd, { text = true }, function(output)
+    if output.code == 0 and output.stdout and vim.trim(output.stdout) ~= '' then
+      local res = vim.trim(output.stdout)
```
justinmk: nit
```suggestion
    local res = vim.trim(output.stdout or '')
    if output.code == 0 and res ~= '' then
```

lsp/gopls.lua
balpert89: sorry, the notify should have been wrapped by the schedule function. Fixing this.

balpert89: applied.

lsp/gopls.lua
balpert89: applied.

lsp/gopls.lua
```diff
@@ -23,23 +90,9 @@ return {
...
-          mod_cache = vim.trim(output.stdout)
-        end
-        on_dir(get_root(fname))
-      else
-        vim.schedule(function()
-          vim.notify(('[gopls] cmd failed with code %d: %s\n%s'):format(output.code, cmd, output.stderr))
-        end)
-      end
-    end)
+    on_dir(get_root_dir(fname))
```
justinmk, line 96: FYI if a dir is NOT found, it's generally best to NOT call `on_dir`. Otherwise the config will be activated even if `workspace_required=true`.

lsp/gopls.lua
```diff
@@ -4,13 +4,80 @@
...
+      end)
+      on_complete(nil)
+    end
+  end)
+end
+
+---@return string?
+local function get_std_lib_dir()
+  if std_lib and std_lib ~= '' then
+    return std_lib
```
justinmk, line 45: this is a race, but I guess the main "risk" is that `identify_go_dir` could run multiple times, which seems acceptable. though, that means it could show redundant errors...

PR #3944, opened 2025-07-10
fix(docs): "commands" item breaks docgen

Fidelxyz:
Problem:
`docgen.lua` is broken since the `description` field for `commands` has
been removed.

Solution:
Remove the description output for `commands` in `docgen.lua`.

### DIFF
--- a/scripts/docgen.lua
+++ b/scripts/docgen.lua
@@ -149,10 +149,7 @@ local function make_lsp_section(config_sections, config_name, config_file, is_ma
         return
       end
       return ('\nCommands:\n%s\n'):format(make_section(0, '\n', {
-        map_sorted(config.commands, function(name, def)
-          if def.description then
-            return string.format('- %s: %s', name, def.description)
-          end
+        map_sorted(config.commands, function(name, _)
           return string.format('- %s', name)
         end),
       }))


lsp/ts_ls.lua
```diff
@@ -98,30 +98,33 @@ return {
     end,
   },
   commands = {
-    ['editor.action.showReferences'] = function(command, ctx)
-      local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
-      local file_uri, position, references = unpack(command.arguments)
+    ['editor.action.showReferences'] = {
+      function(command, ctx)
```
justinmk: This doesn't seem right. The type and also usage in Nvim expects a function, not a table? Maybe you are thinking of the old nvim-lspconfig `commands` which had a different format?

https://github.com/neovim/neovim/blob/ce8104ef07c87f9b3e50e76cb43c38d2c0eca85e/runtime/lua/vim/lsp/client.lua#L70

https://github.com/neovim/neovim/blob/ce8104ef07c87f9b3e50e76cb43c38d2c0eca85e/runtime/lua/vim/lsp/client.lua#L1000

justinmk:
Thanks for looking at this! docgen is indeed broken, maybe the docgen.lua script needs a fix?

lsp/ts_ls.lua
Fidelxyz: You're right. I should edit the docgen script to fix that. Thanks!

scripts/docgen.lua
```diff
@@ -149,10 +149,7 @@ local function make_lsp_section(config_sections, config_name, config_file, is_ma
         return
       end
       return ('\nCommands:\n%s\n'):format(make_section(0, '\n', {
-        map_sorted(config.commands, function(name, def)
-          if def.description then
-            return string.format('- %s: %s', name, def.description)
-          end
+        map_sorted(config.commands, function(name, _)
           return string.format('- %s', name)
```
justinmk, line 153: yeah this looks good, the old code was for the old configs and I forgot to update this. thanks!

PR #3946, opened 2025-07-11
fix(vue_ls): missing config table, suspress message

gx089:
Update vue_ls.lua
Fix nil table error/message:
Added init_options table to suppress error message. So no additional config override is necessary. 

```lua
LSP[vue_ls]: Error BEFORE_INIT_CALLBACK_ERROR: ".../mason-lspconfig.nvim/lua/mason-lspconfig/lsp/vue_ls.lua:8: attempt to index field 'init_options' (a nil value)"
```

### DIFF
--- a/lsp/vue_ls.lua
+++ b/lsp/vue_ls.lua
@@ -22,6 +22,11 @@ return {
   cmd = { 'vue-language-server', '--stdio' },
   filetypes = { 'vue' },
   root_markers = { 'package.json' },
+  init_options = {
+    typescript = {
+      tsdk = '',
+    }
+  },
   on_init = function(client)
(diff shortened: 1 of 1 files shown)


justinmk:
Why is mason not checking if the thing exists? Seems like a mason bug.

Your workaround here may have implications for users that set their own `init_options`, so we should avoid it if possible.

RayGuo-ergou:
I do agree this should be fixed on mason side.

RayGuo-ergou:
created a PR on mason-lspconfig: https://github.com/mason-org/mason-lspconfig.nvim/pull/588

gx089:
sounds reasonable. thx @RayGuo-ergou

PR #3947, opened 2025-07-13
feat: swap pkgbuild_language_server with termux_language_server

an0nfunc:
`pkgbuild_language_server` was replaced by `termux_language_server`. This PR tries to reflect that.


lsp/termux_language_server.lua
```diff
@@ -0,0 +1,10 @@
+---@brief
+---
+--- https://github.com/termux/termux-language-server
+---
+--- Language server for some specific bash scripts.
```
justinmk: ? need a more meaningful description. presumably this is for termux users? or is it a general PKGBUILD ls?
```suggestion
--- Language server for various bash scripts such as PKGBUILD, gentoo ebuild, termux <something>, etc.
```

TomJo2000: Hi, Termux package maintainer here.
The appropriate term would be Termux build.sh
```suggestion
--- Language server for various bash scripts such as Arch PKGBUILD, Gentoo ebuild, Termux build.sh, etc.
```

termux-language-server is a direct continuation project of `pkgbuild-language-server`, and retains support for the build script formats used by Arch (`PKGBUILD`) and Gentoo (`.ebuild`), in addition to the Termux build script format (`build.sh`)

an0nfunc: ~~I was unsure how to formulate that, since `termux_language_server` [does a a lot of things](https://github.com/termux/termux-language-server).~~

~~How about:~~

> Language server for Termux build scripts, Arch/Gentoo PKGBUILDs, and other distro-specific Bash packaging files.

~~This may get outdated as soon as new features are added ofc. Your suggestion would also be fine with me.~~

Adopted the suggestion from @TomJo2000

TomJo2000: > I was unsure how to formulate that, since `termux_language_server` does a a lot of things.
> 
> How about:
> 
> > Language server for Termux build scripts, Arch/Gentoo PKGBUILDs, and other distro-specific Bash packaging files.

That feels a bit imprecise.
To me that would imply that Arch and Gentoo share a build script format, which they do not.

Something like;
`Language server for writing build script files for Arch, Gentoo and Termux packages.`
would be fine, I think the important thing is keeping it clear that they're distinct formats.

> This may get outdated as soon as new features are added ofc. Your suggestion would also be fine with me.

I'm not sure to what extent @Freed-Wu plans to extend the scope of this LSP project.

TomJo2000:
I'm finally getting around to setting this up in my own config and I'm a bit confused how the default config is actually supposed to work?
It's got `filetypes = { 'PKGBUILD' },` set.
There is no such filetype that ships with Neovim out of the box.

The upstream documentation suggests using this.
```lua
vim.api.nvim_create_autocmd({ "BufEnter" }, {
  pattern = { "build.sh", "*.subpackage.sh", "PKGBUILD", "*.install",
    "makepkg.conf", "*.ebuild", "*.eclass", "color.map", "make.conf" },
  callback = function()
    vim.lsp.start({
      name = "termux",
      cmd = { "termux-language-server" }
    })
  end,
})
```
Which I'm not a huge fan of.

But I can't figure out what the best way to enable an LSP for a set of filenames would be with just a `vim.lsp.Config` table.
`root_dir` can take a function and the documentation states quote
```console
Thus a `root_dir()` function can dynamically decide
per-buffer whether to activate (or skip) LSP.
See example at |vim.lsp.enable()|.
```
But I'm not sure I'm figuring this out at 2am in the morning, and this is probably also something that should be fixed in the default config provided in `nvim-lspconfig`.

<h1><em></em></h1> 

Edit:
tl;dr I think the `filetypes` should probably be changed to `{ 'bash' }`.
And this probably needs a `root_dir` function to handle the filenames.

justinmk:
> It's got `filetypes = { 'PKGBUILD' },` set. There is no such filetype that ships with Neovim out of the box.

yes that's a mistake, thanks for mentioning that.

> tl;dr I think the `filetypes` should probably be changed to `{ 'bash' }`.

yes 👍 

> And this probably needs a `root_dir` function to handle the filenames.

there is a better way, documented at `:help vim.lsp.Config` (not sure if these docs are in Nvim 0.11 or only 0.12+):

      • {filetypes}?     (`string[]`) Filetypes the client will attach to, or
                         `nil` for ALL filetypes. To match files by name,
                         pattern, or contents, you can define a custom
                         filetype using |vim.filetype.add()|: >lua
                             vim.filetype.add({
                               filename = {
                                 ['my_filename'] = 'my_filetype1',
                               },
                               pattern = {
                                 ['.*/etc/my_file_pattern/.*'] = 'my_filetype2',
                               },
                             })
                             vim.lsp.config('…', {
                               filetypes = { 'my_filetype1', 'my_filetype2' },
                             }

TomJo2000:
> there is a better way, documented at `:help vim.lsp.Config` (not sure if these docs are in Nvim 0.11 or only 0.12+):
> ```
>   • {filetypes}?     (`string[]`) Filetypes the client will attach to, or
>                      `nil` for ALL filetypes. To match files by name,
>                      pattern, or contents, you can define a custom
>                      filetype using |vim.filetype.add()|: >lua
>                          vim.filetype.add({
>                            filename = {
>                              ['my_filename'] = 'my_filetype1',
>                            },
>                            pattern = {
>                              ['.*/etc/my_file_pattern/.*'] = 'my_filetype2',
>                            },
>                          })
>                          vim.lsp.config('…', {
>                            filetypes = { 'my_filetype1', 'my_filetype2' },
>                          }
> ```
 
Thanks for pointing that out, that isn't yet documented in 0.11.4, but that is an option I hadn't considered before.

That's not quite as simple as just dropping in a `after/lsp/termux_language_server.lua` file, but it does at least handle the file pattern

TomJo2000:
Fwiw, this is what I came up with using `root_dir`.
```lua
---@type vim.lsp.Config
return {
  filetypes = { 'bash' },
    local util = require('lspconfig').util
    local fname = vim.api.nvim_buf_get_name(bufnr or 0)
    -- stylua: ignore
    local pattern = {
      'build.sh', '*.subpackage.sh', -- Termux
      'PKGBUILD', '*.install', 'makepkg.conf', -- Arch/MSYS2
      '*.ebuild', '*.eclass', 'color.map', 'make.conf', -- Gentoo
    }
    local match = util.root_pattern(pattern)(fname)
    if match then
      on_dir(vim.fs.root(bufnr, '.git') or match)
    end
  end,
}
```
Don't think this quite works yet.

<h1><em></em></h1>  

Edit: Okay so `build.sh` files get picked up as `ft=sh`, not `ft=bash`.
So the reason why I haven't been able to get this cooperating is because I told it to only attach to `ft=bash`.
Time is a flat circle.

justinmk:
can someone send a PR to fix the `filetypes` in this config

TomJo2000:
I'll get right on that after I grab something to eat.

PR #3950, opened 2025-07-15
[lsp] devsense-php-ls PHP language server

jakubmisek:
This pull request adds information about using [devsense-php-ls](https://www.npmjs.com/package/devsense-php-ls) in Neovim. The language server features 

- Laravel IDE features
- Support for PhpStan, Psalm, and other PHPDoc annotations
- Code completion
- Inlay hints
- Code diagnostics
- Code navigation
- Code folding
- And more

Provided language server is FREE with option to activate premium features, as described on https://www.npmjs.com/package/devsense-php-ls #activation

We already tested the integration, and we have positive feedback from other users.


jakubmisek:
will create a new PR with correct commit message and without bad changes according to the docs

PR #3954, opened 2025-07-17
docs(rescriptls): add didChangeWatchedFiles example

dkirchhof:


justinmk:
> When it will be released, we should add the capabilities to the default config.

does it break anything if we enable it now ?

PR #3957, opened 2025-07-17
feat(biome): use project local bin if available

igorlfs:
Problem: users may have a different version of biome installed in a project, compared to the global installation.

Solution: use project local binary if available.

Note: I used `vim.system({ ... }):wait()` to check if the binary is executable. LMK if using `vim.fn.executable` is preferred.

This PR is inspired by the release of biome v2, which introduced some breaking changes.

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -10,7 +10,14 @@
 local util = require 'lspconfig.util'
 
 return {
-  cmd = { 'biome', 'lsp-proxy' },
+  cmd = function(dispatchers, config)
+    local cmd = 'biome'
+    local local_cmd = config.root_dir and config.root_dir .. '/node_modules/.bin/biome'
+    if local_cmd and vim.fn.executable(local_cmd) == 1 then
+      cmd = local_cmd
+    end
+    return vim.lsp.rpc.start({ cmd, 'lsp-proxy' }, dispatchers)
+  end,
   filetypes = {
     'astro',
     'css',


justinmk:
cool to see people using `cmd(..,config)`

> LMK if using `vim.fn.executable` is preferred.

yes

lsp/biome.lua
```diff
@@ -9,8 +9,24 @@
 
 local util = require 'lspconfig.util'
 
+---@param config vim.lsp.ClientConfig
+local get_cmd = function(config)
```
justinmk:
```suggestion
local function get_cmd(config)
```

lsp/biome.lua
```diff
@@ -9,8 +9,24 @@
...
+  local root_dir = config.root_dir
```
justinmk: unnecessary alias

igorlfs:
Should be good to go

lsp/biome.lua
```diff
@@ -9,8 +9,20 @@
...
+    return 'biome'
+  end
+  local local_cmd = config.root_dir .. '/node_modules/.bin/biome'
+  return vim.fn.executable(local_cmd) == 1 and local_cmd or 'biome'
+end
+
 return {
-  cmd = { 'biome', 'lsp-proxy' },
+  cmd = function(dispatchers, config)
+    local command = get_cmd(config)
```
justinmk: would you mind putting the logic directly here, rather than in a get_cmd function. i don't want people copying this pattern, it leads to over-engineered functions. these functions should be simple and obvious.

igorlfs: Inlined! I got the inspiration from [this](https://github.com/neovim/nvim-lspconfig/blob/f6da6f19b7d83a5133745adc41042170bdff835a/lsp/glint.lua#L39) file, which uses a similar logic.

justinmk:
I wonder if we should do this kind of thing for python venv, ruby gems, etc.

igorlfs:
> I wonder if we should do this kind of thing for python venv, ruby gems, etc.

Now that it's possible to use config to dynamically set the cmd, I'd say it's just a matter a time for the community to adjust.

PR #3958, opened 2025-07-18
feat(svelte): notify LSP of changes in JS/TS files

igorlfs:
The Svelte LSP does not attach to regular JS / TS files. To get notified about changes in these files, it "needs" to send a custom message via `notify`.

The VS Code extension uses the same workaround[^1] and it has been discussed to death how to port it to neovim[^2]. Instead of relying on users finding out about said discussion, it should be included in the default configuration.

[^1]: https://github.com/sveltejs/language-tools/blob/2e2c6c3624f2dd21cbda4d8b4b63e0d4df93152e/packages/svelte-vscode/src/extension.ts#L346
[^2]: https://github.com/sveltejs/language-tools/issues/2008

### DIFF
--- a/lsp/svelte.lua
+++ b/lsp/svelte.lua
@@ -21,6 +21,16 @@ return {
     end
   end,
   on_attach = function(client, bufnr)
+    -- Workaround to trigger reloading JS/TS files
+    -- See https://github.com/sveltejs/language-tools/issues/2008
+    vim.api.nvim_create_autocmd('BufWritePost', {
+      pattern = { '*.js', '*.ts' },
+      buffer = bufnr,
+      callback = function(ctx)
+        -- internal API to sync changes that have not yet been saved to the file system
+        client:notify('$/onDidChangeTsOrJsFile', { uri = ctx.match })
+      end,
+    })
     vim.api.nvim_buf_create_user_command(bufnr, 'LspMigrateToSvelte5', function()
       client:exec_cmd({
         command = 'migrate_to_svelte_5',


jasonlyu123:
To clarify, as I have mentioned many times in the linked issue, the `$/onDidChangeTsOrJsFile` request is an internal API designed to sync changes that have not yet been saved to the file system. Neovim user *does not* "need to send a custom message via notify". It is more of a bonus. So this shouldn't be considered a "solution" for the linked issue. 

Also, in the latest version of svelte-language-server, we have tweaked the detection for enabling a fallback watcher, which should improve the issue.

igorlfs:
> Neovim user does not "need to send a custom message via notify". It is more of a bonus.

Yeah, that's why I used quotation marks!

> So this shouldn't be considered a "solution" for the linked issue. Also, in the latest version of svelte-language-server, we have tweaked the detection for enabling a fallback watcher, which should improve the issue.

I feel that, to some extent, it should be considered a solution. `didChangeWatchedFiles` is only disabled by default on Linux (which is actually the main issue, AFAIK), but even then, users can enable it anyway. I'm one of them, I'm using a custom backend for the file watcher. For my experience to be "complete" file-watching wise, the only missing piece is enabling the `$/onDidChangeTsOrJsFile` notification (which I did, for my own config).

The intent of this PR is just to provide a better OOTB experience, especially considering users that are already using the file watcher from the client. I'm sorry if it came off as something different!

lsp/svelte.lua
```diff
@@ -21,6 +21,14 @@ return {
     end
   end,
   on_attach = function(client, bufnr)
+    -- Workaround to trigger reloading JS/TS files
+    -- See https://github.com/sveltejs/language-tools/issues/2008
+    vim.api.nvim_create_autocmd('BufWritePost', {
```
justinmk, line 26: this will create a new global BufWritePost handler every time `on_attach` is called. probably this should specify `buffer = bufnr` to make the handler buffer-local.

lsp/svelte.lua
```diff
@@ -21,6 +21,14 @@ return {
...
+      pattern = { '*.js', '*.ts' },
+      callback = function(ctx)
+        client:notify('$/onDidChangeTsOrJsFile', { uri = ctx.match })
```
justinmk, line 31: the context that @jasonlyu123 provided would be useful here, or some other reference
```suggestion
        -- internal API to sync changes that have not yet been saved to the file system
        client:notify('$/onDidChangeTsOrJsFile', { uri = ctx.match })
```

lsp/svelte.lua
igorlfs: Nice catch

PR #3960, opened 2025-07-19
refactor(glint): use `cmd(…, config)`

justinmk:
Problem:
The glint config `cmd()` gets its config from the currently resolved global config, which may be different than the one for the current client.

Solution:
Since Nvim 0.11.3 https://github.com/neovim/neovim/issues/34547 `cmd()` receives a `config` arg, so use that instead.

Followup to https://github.com/neovim/nvim-lspconfig/pull/3912 @justmejulian 

Note that this requires Nvim 0.11.3

### DIFF
--- a/lsp/glint.lua
+++ b/lsp/glint.lua
@@ -22,21 +22,10 @@
 ---   },
 --- })
 
-function get_cmd()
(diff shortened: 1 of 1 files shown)


justmejulian:
Cool. Thanks!

PR #3962, opened 2025-07-19
fix(biome): error when using non-local biome binary

chrisgrieser:
#3957 introduced using local binaries of biome when available, but results in an error on lsp startup when there is no local installation due to a missing nil-check:

```
...eser/.local/share/nvim/lazy/nvim-lspconfig/lsp/biome.lua:15: attempt to index local 'config' (a nil value)
```

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -12,7 +12,7 @@ local util = require 'lspconfig.util'
 return {
   cmd = function(dispatchers, config)
     local cmd = 'biome'
-    local local_cmd = config and config.root_dir and config.root_dir .. '/node_modules/.bin/biome'
+    local local_cmd = (config or {}).root_dir and config.root_dir .. '/node_modules/.bin/biome'
     if local_cmd and vim.fn.executable(local_cmd) == 1 then
       cmd = local_cmd
     end


igorlfs:
Hey, would you happen to be using a version prior to 0.11.3? `config` is a required field from this version onwards.

chrisgrieser:
Yes, I am on 0.11.2. 

Nonetheless, it's just an extra nil check to make it work for people who did not upgrade to 0.11.3 yet

lsp/biome.lua
```diff
@@ -12,7 +12,7 @@ local util = require 'lspconfig.util'
 return {
   cmd = function(dispatchers, config)
     local cmd = 'biome'
-    local local_cmd = config.root_dir and config.root_dir .. '/node_modules/.bin/biome'
+    local local_cmd = config and config.root_dir and config.root_dir .. '/node_modules/.bin/biome'
```
justinmk:
```suggestion
    local local_cmd = (config or {}).root_dir and config.root_dir .. '/node_modules/.bin/biome'
```

justinmk:
Nvim 0.11.3 is required, but this is simple enough so ok.

PR #3963, opened 2025-07-19
fix(roslyn_ls): replace deprecated function call

ateoi:
vim.lsp.util._refresh() was removed by https://github.com/neovim/neovim/pull/33903


justinmk:
Thanks. A private api should never have been used here...

PR #3964, opened 2025-07-19
fix(svelte): use augroup to avoid creating multiple autocmds 

igorlfs:
Follow up: https://github.com/neovim/nvim-lspconfig/pull/3958

As it turns out, we can't use both `buffer` and `pattern` in the same autocmd. And even if we could set the buffer, it is not necessarily the one the server is attaching to.

By removing the `buffer` we are creating a new autocmd every time the buffer attaches. The ideal solution would be to guard against that, but I'm not sure if that would be the best approach (I imagine there are some augroup shenanigans?).

### DIFF
--- a/lsp/svelte.lua
+++ b/lsp/svelte.lua
@@ -25,7 +25,7 @@ return {
     -- See https://github.com/sveltejs/language-tools/issues/2008
     vim.api.nvim_create_autocmd('BufWritePost', {
       pattern = { '*.js', '*.ts' },
-      buffer = bufnr,
+      group = vim.api.nvim_create_augroup('svelte_js_ts_file_watch', {}),
       callback = function(ctx)
         -- internal API to sync changes that have not yet been saved to the file system
         client:notify('$/onDidChangeTsOrJsFile', { uri = ctx.match })


justinmk:
> if we could set the buffer, it is not necessarily the one the server is attaching to.

? the `bufnr` passed to on_attach is the buffer being attached.

Removing the pattern seems like the correct approach. A pattern isn't needed, on_attach already decided the "pattern" by matching against `filetypes`.

SebasF1349:
The lsp attaches to svelte files, but the autocmd is for js/ts files, so the buffers and filetypes will be different. I have this autocmd in my config (with pattern and no buffer) and it worked last time I used svelte. I would only add a augroup.

justinmk:
oops, yeah then an augroup should be used.

this is why a code comment helps with this kind of thing. especially when multiple configs are working together, that isn't obvious.

igorlfs:
> oops, yeah then an augroup should be used.

Done! I also messed up, should have tested after setting the buffer 🤦🏼‍♂️ 

> this is why a code comment helps with this kind of thing. especially when multiple configs are working together, that isn't obvious.

LMK if any comments would be useful now

PR #3971, opened 2025-07-24
fix(pyright): command :LspPyrightOrganizeImports fails

michelesr:
Fix error raised when running `:LspPyrightOrganizeImports` :

```
Language server `pyright` does not support command `pyright.organizeimports`. This command may require a client extension.
```

### DIFF
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -48,7 +48,9 @@ return {
         arguments = { vim.uri_from_bufnr(bufnr) },
       }
 
-      -- cannot use client:exec_cmd() as it will fail due to the LSP command not being advertised by the server
+      -- Using client.request(() directly because this is a private command (not
+      -- advertised via capabilities), which client:exec_cmd() refuses to call.
+      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
       client.request('workspace/executeCommand', params, nil, bufnr)
     end, {
       desc = 'Organize Imports',
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -43,10 +43,13 @@ return {
   },
   on_attach = function(client, bufnr)
     vim.api.nvim_buf_create_user_command(bufnr, 'LspPyrightOrganizeImports', function()
-      client:exec_cmd({
+      local params = {
         command = 'basedpyright.organizeimports',
         arguments = { vim.uri_from_bufnr(bufnr) },
-      })
+      }
+
+      -- cannot use client:exec_cmd() as it will fail due to the LSP command not being advertised by the server
+      client.request('workspace/executeCommand', params, nil, bufnr)
     end, {
       desc = 'Organize Imports',
     })
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -48,8 +48,8 @@ return {
         arguments = { vim.uri_from_bufnr(bufnr) },
       }
 
-      -- Using client.request(() directly because this is a private command (not
-      -- advertised via capabilities), which client:exec_cmd() refuses to call.
+      -- Using client.request() directly because "basedpyright.organizeimports" is private
+      -- (not advertised via capabilities), which client:exec_cmd() refuses to call.
       -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
       client.request('workspace/executeCommand', params, nil, bufnr)
     end, {
--- a/lsp/pyright.lua
+++ b/lsp/pyright.lua
@@ -48,7 +48,9 @@ return {
         arguments = { vim.uri_from_bufnr(bufnr) },
       }
 
-      -- cannot use client:exec_cmd() as it will fail due to the LSP command not being advertised by the server
+      -- Using client.request() directly because "pyright.organizeimports" is private
+      -- (not advertised via capabilities), which client:exec_cmd() refuses to call.
+      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
       client.request('workspace/executeCommand', params, nil, bufnr)
     end, {
       desc = 'Organize Imports',
--- a/lsp/pyright.lua
+++ b/lsp/pyright.lua
@@ -43,10 +43,13 @@ return {
   },
   on_attach = function(client, bufnr)
     vim.api.nvim_buf_create_user_command(bufnr, 'LspPyrightOrganizeImports', function()
-      client:exec_cmd({
+      local params = {
         command = 'pyright.organizeimports',
         arguments = { vim.uri_from_bufnr(bufnr) },
-      })
+      }
+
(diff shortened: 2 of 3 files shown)


lsp/basedpyright.lua
```diff
@@ -42,10 +42,12 @@ return {
   },
   on_attach = function(client, bufnr)
     vim.api.nvim_buf_create_user_command(bufnr, 'LspPyrightOrganizeImports', function()
-      client:exec_cmd({
+      local params = {
         command = 'basedpyright.organizeimports',
         arguments = { vim.uri_from_bufnr(bufnr) },
-      })
+      }
```
justinmk, line 49: not sure why you changed this (always need to mention that in the pr), but if your intention is to pass bufnr, that can be done as the 2nd arg to exec_cmd

    client:exec_cmd({ ... }, { bufnr = bufnr })

michelesr: @justinmk It's explained in the PR description and commit msg: `client:exec_cmd()` was raising an error about that command not being supported by the server and aborting, so I got the old version from [here](https://github.com/neovim/nvim-lspconfig/blob/f47cd681d7cb6048876a2e908b6d8ba1e530d152/lua/lspconfig/configs/pyright.lua#L24) but I assumed you needed to pass `bufnr` instead of `0`; it works with `0` too, but it doesn't work if you change it back to `client:exec_cmd()`.

justinmk: > It's explained in the PR description and commit msg: `client:exec_cmd()` was raising an error about that command not being supported by the server and aborting,

That doesn't explain anything. `client:exec_cmd()` should work.



> it doesn't work if you change it back to `client:exec_cmd()`.

Did you pass bufnr in the 2nd dict arg as I suggested?

michelesr: > Did you pass bufnr in the 2nd dict arg as I suggested?

```lua
      client:exec_cmd({
        command = 'pyright.organizeimports',
        arguments = { vim.uri_from_bufnr(bufnr) },
      }, {bufnr = bufnr})
```

This raises the same error.

michelesr: @justinmk I think it's because of https://github.com/microsoft/pyright/issues/10406 ?

justinmk: ? `client:exec_cmd` performs `"workspace/executeCommand"`. The only difference is that  `exec_cmd` currently is [always asynchronous](https://github.com/neovim/nvim-lspconfig/pull/3876) (but so is `client.request`).

So, there is some other problem, and the change you've made here only accidentally "fixes" a symptom. Need to find out what the actual problem is.

I wonder if `client.request` just doesn't show the error?

Based on https://github.com/microsoft/pyright/issues/10406 , it seems like the server is literally telling us not to use this feature. So if your intent is to ignore that and use it anyway, that sure is worth mentioning in the PR description (and commit message) and a code comment.

Here's an outline of an improved approach:

1. use `pcall` or ignore the error in some other explicit manner
2. put a 1-line code comment mentioning that we are intentionally ignoring the server error and this could break in the future. also link to the upstream issue 
3. actually mention this in the PR description and commit message

michelesr: >  So if your intent is to ignore that and use it anyway, that sure is worth mentioning in the PR description (and commit message) and a code comment.

Not really, I just found out about this, because I moved my config to use `vim.lsp.config()`, and noticed the difference between the code in `lua/` and the one in `lsp/`. I don't rely on this feature and there are better alternatives (like using `!ruff check --fix --select I %` for example).

I'm not sure what's the best decision to take here, if the feature is discouraged from the pyright maintainers then is probably better to just remove the vim command altoghether rather than leaving it broken?

I agree that wrapping `exec_cmd` with `pcall()` seems an hack, and I've tried to do so but it still prints the message and doesn't perform the required operation.

justinmk: Let's wait a bit and see if others have comments.

justinmk:
Ok, I did not realize that `exec_cmd()` is actively doing a check: https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1027

and that fails because "basedpyright.organizeimports" is a private command (not advertised via capabilities). 

Whereas `client.request()` just doesn't do that check.

So, we can merge this after:

- add a comment above `client.request()` with the above information
- rebase

michelesr:
@justinmk what about the failing linting checks, is there a way to disable them for the relevant LoC? 

```
lint.sh: Use client:exec_cmd() instead of calling request("workspace/executeCommand") directly. Example: lsp/pyright.lua
```

justinmk:
To pass the lint you can change this line https://github.com/neovim/nvim-lspconfig/blob/408cf07b97535825cca6f1afa908d98348712ba6/.github/ci/lint.sh#L56

to:

    local exclude='eslint\|pyright\|basedpyright'

i'll think about how to make it more granular later.

michelesr:
@justinmk rebased and added the lint exclusion.

P.S: added the comments.

lsp/basedpyright.lua
```diff
@@ -43,10 +43,13 @@ return {
...
+
+      -- cannot use client:exec_cmd() as it will fail due to the LSP command not being advertised by the server
```
justinmk:
```suggestion
      -- Using client.request(() directly because this is a private command (not
      -- advertised via capabilities), which client:exec_cmd() refuses to call.
      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
```

lsp/pyright.lua
```diff
@@ -43,10 +43,13 @@ return {
...
   on_attach = function(client, bufnr)
     vim.api.nvim_buf_create_user_command(bufnr, 'LspPyrightOrganizeImports', function()
-      client:exec_cmd({
+      local params = {
         command = 'pyright.organizeimports',
         arguments = { vim.uri_from_bufnr(bufnr) },
-      })
+      }
+
+      -- cannot use client:exec_cmd() as it will fail due to the LSP command not being advertised by the server
```
justinmk:
```suggestion
      -- Using client.request() directly because "pyright.organizeimports" is private
      -- (not advertised via capabilities), which client:exec_cmd() refuses to call.
      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
```

lsp/basedpyright.lua
```diff
@@ -43,10 +43,15 @@ return {
...
-      client:exec_cmd({
+      local params = {
         command = 'basedpyright.organizeimports',
         arguments = { vim.uri_from_bufnr(bufnr) },
-      })
+      }
+
+      -- Using client.request(() directly because this is a private command (not
+      -- advertised via capabilities), which client:exec_cmd() refuses to call.
+      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
```
justinmk:
```suggestion
      -- Using client.request() directly because "basedpyright.organizeimports" is private
      -- (not advertised via capabilities), which client:exec_cmd() refuses to call.
      -- https://github.com/neovim/neovim/blob/c333d64663d3b6e0dd9aa440e433d346af4a3d81/runtime/lua/vim/lsp/client.lua#L1024-L1030
```

justinmk:
Thanks for your help. I updated the description to be 100% clear, please consider that next time.

PR #3973, opened 2025-07-24
docs: mention a way to explicitly set `root_dir` in README.md

faergeek:
Continued from https://github.com/neovim/nvim-lspconfig/issues/3910#issuecomment-2982698167

Even though in nvim before 0.12 `debug.getinfo(1, 'S').source:sub(2)` inside `.nvim.lua` doesn't result in a proper filename and instead returns `'<nvim>'`, it still works fine and there's no need to adjust the code between versions.

nvim 0.12 only returns full path if run from a directory below. If it's run from cwd it returns just `'.nvim.lua'` (filename only). nvim pre-0.12 returns `'<nvim>'`, which might still be seen as a filename in the current dir.

Using `fnamemodify` this way results in a proper absolute path to a dir in all of those cases. So the only thing about nvim version I mentioned is that `exrc` pre-0.12 is only being run when nvim started from cwd.


justinmk:
Nice, thanks. I've been recently leveraging `.nvim.lua` + 'exrc' myself, and I'm kind of annoyed that we waited this long to invest in it 😅 . It's clearly a big improvement for working on team projects. We need to surface it much more clearly in Nvim's own docs and user materials...

PR #3976, opened 2025-07-28
feat: `herb_ls` config

hschne:
This is a follow up to #3925 and adds a config for `herb_ls`.


igorlfs:
Hello,

Configs in `lua/lspconfig/` are discontinued. From the README:

> The configs in lua/lspconfig/ are deprecated and will be removed.

PR #3980, opened 2025-07-29
fix(tailwindcss): add fallback for tailwind v4

### LINKED ISSUE #3979: Incomplete tailwindcss `root_dir` configuration
### Description

Since tailwind v4 it is not required to have a `tailwind.config.*` in the project [^1]:

> One of the biggest changes in Tailwind CSS v4.0 is the shift from configuring your project in JavaScript to configuring it in CSS.
>
> Instead of a tailwind.config.js file, you can configure all of your customizations directly in the CSS file where you import Tailwind, giving you one less file to worry about in your project: (...)

The current configuration is preventing the LSP to attach to the configured file types in cases when no `tailwind.config.*` or none of the other markers (`package.json` etc.) is present. This is the case when you integrate tailwind through a CDN like `<script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>`.

**Proposal:**

1. Either use `.git` as a fallback similar to how it is done in other lsp configs and/or
2. properly document this limitation so that users can create a placeholder `tailwind.config.js` as a workaround

It took me quite some time to understand why the LSP is not attached, after checking file types and debugging my LSP config. So to save others some time, I suggest 

[^1]: https://tailwindcss.com/blog/tailwindcss-v4#css-first-configuration

jsattler:
Since tailwind v4 it is not required to have a tailwind.config.* in the project 1. The current configuration is preventing the LSP to attach to the configured file types in cases when no tailwind.config.* or none of the other markers (package.json etc.) is present. This commit fixes this limitation by providing .git as a fallback and by documenting this limitation.

fixes #3979

### DIFF
--- a/lsp/tailwindcss.lua
+++ b/lsp/tailwindcss.lua
@@ -123,6 +123,8 @@ return {
       'theme/static_src/tailwind.config.mjs',
       'theme/static_src/tailwind.config.ts',
       'theme/static_src/postcss.config.js',
+      -- Fallback for tailwind v4, where tailwind.config.* is not required anymore
+      '.git',
     }
     local fname = vim.api.nvim_buf_get_name(bufnr)
     root_files = util.insert_package_json(root_files, 'tailwindcss', fname)


igorlfs:
Note: this change has the side effect of attaching tailwind in pretty much every directory. That's especially the case given the extensive list of supported filetypes. Not much of a concern, other than a slight increase in memory usage (about a couple hundred megabytes on my machine)

justinmk:
Yes, because tailwind LS "supports" a lot of filetypes, declaring `.git` as a root marker effectively makes it attach to "everything".

I have been thinking that for cases like this, the best approach is probably for users to decide when they want an "omni LS" such as tailwind (which claims to support "basically everything"), by explicitly running `:LspStart` (or something). That avoids the paradox.

PR #3991, opened 2025-08-09
fix(texlab): adapt commands to fit 0.11+

DimitrisDimitropoulos:
This is a much needed cleanup to the texlab configuration. In essence this PR updates the config to fit the 0.11+ way of neovim. The changes can be summed up to:

- Drop support for <0.11 in vim.lsp.config version
- Use vim.ui.input instead of vim.fn.input
- Replace explicit border setting in favor to winborder option
- Call request as method to avoid deprecation warning
- Simplify command setup via for loop

### DIFF
--- a/lsp/texlab.lua
+++ b/lsp/texlab.lua
@@ -20,7 +20,7 @@ end
 local function buf_build(client, bufnr)
   local win = vim.api.nvim_get_current_win()
   local params = vim.lsp.util.make_position_params(win, client.offset_encoding)
-  client.request('textDocument/build', params, function(err, result)
+  client:request('textDocument/build', params, function(err, result)
     if err then
       error(tostring(err))
     end
@@ -37,7 +37,7 @@ end
 local function buf_search(client, bufnr)
   local win = vim.api.nvim_get_current_win()
   local params = vim.lsp.util.make_position_params(win, client.offset_encoding)
-  client.request('textDocument/forwardSearch', params, function(err, result)
+  client:request('textDocument/forwardSearch', params, function(err, result)
     if err then
       error(tostring(err))
     end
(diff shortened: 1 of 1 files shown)


DimitrisDimitropoulos:
The failed test seems to be a false positive, because of the way the commands are setup. If necessary I can revert it back to the original way.

lsp/texlab.lua
```diff
@@ -193,29 +169,17 @@ return {
...
+      { name = 'LspTexlabBuild', fn = buf_build, desc = 'Build the current buffer' },
+      { name = 'LspTexlabForward', fn = buf_search, desc = 'Forward search from current position' },
+      { name = 'LspTexlabCancelBuild', fn = buf_cancel_build, desc = 'Cancel the current build' },
+      { name = 'LspTexlabDependencyGraph', fn = dependency_graph, desc = 'Show the dependency graph' },
+      { name = 'LspTexlabCleanArtifacts', fn = command_factory('Artifacts'), desc = 'Clean the artifacts' },
+      { name = 'LspTexlabCleanAuxiliary', fn = command_factory('Auxiliary'), desc = 'Clean the auxiliary files' },
+      { name = 'LspTexlabFindEnvironments', fn = buf_find_envs, desc = 'Find the environments at current position' },
+      { name = 'LspTexlabChangeEnvironment', fn = buf_change_env, desc = 'Change the environment at current position' },
+    }) do
+      vim.api.nvim_buf_create_user_command(buf, command.name, client_with_fn(command.fn), { desc = command.desc })
```
justinmk: To avoid the (naive :) lint-rule could you do something like:


```suggestion
      vim.api.nvim_buf_create_user_command(buf, 'Lsp' .. command.name, client_with_fn(command.fn), { desc = command.desc })
```

justinmk:
Thanks! This is very helpful. Will merge after lint fix

PR #3998, opened 2025-08-14
fix(clangd): old config style still including proto filetype

amarshall:
Missing change from cc6b21ea16cc12df290152f02513a8bd0ed9d932.


justinmk:
Reminder that the old configs are not supported and will be deleted very soon.

Issue #4000, opened 2025-08-16
Typo in filename for `kotlin-lsp` config


jithu7432:
### Description

The configuration for [kotlin-lsp](https://github.com/Kotlin/kotlin-lsp) is in [lsp/kotlin_lsp.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/kotlin_lsp.lua) which raises this error when enabled.

```
- ⚠️ WARNING 'kotlin-lsp' config not found. Ensure that vim.lsp.config('kotlin-lsp') was called.
```

justinmk:
This is an error in your config as explained in https://github.com/neovim/nvim-lspconfig/pull/4001#issuecomment-3193825441

LS program names can be any random thing, and is unrelated to the config name, which often can't be exactly the same as the LS program name, for various reasons.

jithu7432:
Right, and the `configs.md` correctly mentions the lsp as `kotlin_lsp`. Thanks

PR #4001, opened 2025-08-16
fix(kotlin-lsp): typo in config filename

jithu7432:
- Fixes https://github.com/neovim/nvim-lspconfig/issues/4000


sharpchen:
I think using _ is just a convention

jithu7432:
I see, but this manages to fix the issues in `:checkhealth lsp`
```
- ⚠️ WARNING 'kotlin-lsp' config not found. Ensure that vim.lsp.config('kotlin-lsp') was called.
- kotlin-lsp:
```

sharpchen:
I don't think filename matters, if it's `kotlin_lsp` you should say `vim.lsp.enable('kotlin_lsp')` instead of `vim.lsp.enable('kotlin-lsp')`. The warning appears when you enabled with a mdoule name that nvim can't find in all `lsp/` runtime path. If you do `vim.lsp.enable("foo")` you'll see the same warning

PR #4003, opened 2025-08-17
feat(yamlls): Added document formatting support

ChrisPJohnstone:


lsp/yamlls.lua
```diff
@@ -66,4 +66,8 @@ return {
     -- https://github.com/redhat-developer/vscode-redhat-telemetry#how-to-disable-telemetry-reporting
     redhat = { telemetry = { enabled = false } },
   },
+  on_init = function(client)
+    -- https://github.com/redhat-developer/yaml-language-server/issues/486
+    client.server_capabilities.documentFormattingProvider = true
```
justinmk, line 71: Normally the client should not have to do this manually. I skimmed the issue and I guess my conclusion is supposed to be that redhat isn't doing this for whatever reason and so this workaround is needed. In the future, your PR should explain that concisely.

liskin: Your hunch is correct, the client doesn't have to do this manually. It's simply that formatting is a dynamically registered capability in yamlls, and it's disabled by default:

https://github.com/redhat-developer/yaml-language-server/blob/3821411ee8c92e5b7e5ca88f84ab443ae3b2791a/src/yamlServerInit.ts#L128
https://github.com/redhat-developer/yaml-language-server/blob/3821411ee8c92e5b7e5ca88f84ab443ae3b2791a/src/languageserver/handlers/settingsHandlers.ts#L159-L174

A better fix would be to default yamlls settings to the equivalent of:
```lua
vim.lsp.config('yamlls', {
	settings = {
		yaml = {
			format = {
				enable = true,
			},
		},
	},
})
```

(Or to tell users to do so, although I wonder why would anyone want that disabled? Just to keep `gq` doing what plain vim does?)

justinmk: Thanks! I do think that enabling low-cost LSP features by default is a good general practice.

liskin: https://github.com/neovim/nvim-lspconfig/pull/4012 fixes this

ChrisPJohnstone: Sorry for not explaining properly, I raised another PR with a proper explanation https://github.com/neovim/nvim-lspconfig/pull/4016

PR #4004, opened 2025-08-17
fix(eslint): fix performance regression on configuration lookup

fuegoio:
I just found out that I introduced a pretty big performance bottleneck with #3955 on `eslint` LSP. I thought using `vim.fs.find` would be much faster, but in fact it's taking a solid 3-4 seconds on a project of mine, compared to around 100-200ms with the previous implementation I had with `vim.fn.globpath`.

I'm pretty sorry about having put this upstream, I thought it was my local setup that had an issue and did not think about this.

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -172,11 +172,17 @@ return {
       end, eslint_config_files)
 
       for _, file in ipairs(flat_config_files) do
-        local found_files = vim.fs.find(function(name, path)
-          return name == file and not path:match('[/\\]node_modules[/\\]')
-        end, { path = root_dir, type = 'file', limit = 1 })
+        local found_files = vim.fn.globpath(root_dir, file, true, true)
 
-        if #found_files > 0 then
+        -- Filter out files inside node_modules
+        local filtered_files = {}
+        for _, found_file in ipairs(found_files) do
+          if string.find(found_file, '[/\\]node_modules[/\\]') == nil then
+            table.insert(filtered_files, found_file)
+          end
+        end
+
+        if #filtered_files > 0 then
           config.settings.experimental = config.settings.experimental or {}
           config.settings.experimental.useFlatConfig = true
           break


lsp/eslint.lua
```diff
@@ -172,11 +172,17 @@ return {
...
-        local found_files = vim.fs.find(function(name, path)
-          return name == file and not path:match('[/\\]node_modules[/\\]')
-        end, { path = root_dir, type = 'file', limit = 1 })
+        local found_files = vim.fn.globpath(root_dir, file, true, true)
 
-        if #found_files > 0 then
+        -- Filter out files inside node_modules
+        local filtered_files = {}
+        for _, found_file in ipairs(found_files) do
+          if string.find(found_file, '/node_modules/') == nil then
```
justinmk: globpath might return `\` slashes on windows?

lsp/eslint.lua
```diff
@@ -172,11 +172,17 @@ return {
       end, eslint_config_files)
 
       for _, file in ipairs(flat_config_files) do
-        local found_files = vim.fs.find(function(name, path)
-          return name == file and not path:match('[/\\]node_modules[/\\]')
-        end, { path = root_dir, type = 'file', limit = 1 })
```
justinmk, line 177: here you had limit=1 but  the new code collects all matches?

lsp/eslint.lua
fuegoio: Argh something I'm not super familiar with :/ I guess in this case I should match with a regex? Or maybe just `node_modules` would be sufficient, it's a private dir name in Node for sure

lsp/eslint.lua
fuegoio: Yes as the filtering happens after (to filter out config files in node_modules)

lsp/eslint.lua
```diff
@@ -172,11 +172,17 @@ return {
       end, eslint_config_files)
 
       for _, file in ipairs(flat_config_files) do
-        local found_files = vim.fs.find(function(name, path)
-          return name == file and not path:match('[/\\]node_modules[/\\]')
```
justinmk, line 176: try using `:find()` , it's likely faster than `:match()` , though they both accept patterns, so you can still use the `'[/\\]...` pattern.

fuegoio: It's still extremely slow :/ I think it's because `vim.fs.find` is implemented in lua but the globpath is unix a binary directly on the system no?

lsp/eslint.lua
fuegoio: Just added the same pattern as before in fact as `:find` works as well

justinmk:
I think the main difference is that `vim.fs.find()` is recursive, whereas `globpath()` is not (unless a `**` pattern is given).

With the logic the way it is, `vim.fs.find()` doesn't fit I guess. Though maybe the logic could be revisited, but can merge this meanwhile. https://github.com/neovim/neovim/pull/33771 could possibly help, idk.

PR #4008, opened 2025-08-18
fix(ts/js): servers do not start when `bun.lock` exists

Methapon2001:
This PR add `bun.lock` to `root_markers` to multiple server related to TypeScript/JavaScript.

> Bun v1.2 changed the default lockfile format to the text-based bun.lock. Existing binary 
> bun.lockb lockfiles can be migrated to the new format by running bun install --save-text-
> lockfile --frozen-lockfile --lockfile-only and deleting bun.lockb.

[reference](https://bun.com/docs/install/lockfile)


fuegoio:
Thanks a lot, I missed those ones.

PR #4010, opened 2025-08-19
fix: type annotations

igorlfs:
Adds `---@type vim.lsp.Config` for configs and fixes some type related issues (and some minor inconsistencies). 

To make it easier for review, it's split into 2 commits.

TODO: add lint rule

### DIFF
--- a/lsp/anakin_language_server.lua
+++ b/lsp/anakin_language_server.lua
@@ -8,6 +8,7 @@
 ---
 --- * Initialization: https://github.com/muffinmad/anakin-language-server#initialization-option
 --- * Configuration: https://github.com/muffinmad/anakin-language-server#configuration-options
+---@type vim.lsp.Config
 return {
   cmd = { 'anakinls' },
   filetypes = { 'python' },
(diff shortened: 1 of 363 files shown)


justinmk:
Nice, this found some actual bugs. Thanks!

lsp/anakin_language_server.lua
```diff
@@ -8,6 +8,7 @@
 ---
 --- * Initialization: https://github.com/muffinmad/anakin-language-server#initialization-option
 --- * Configuration: https://github.com/muffinmad/anakin-language-server#configuration-options
+---@type vim.lsp.Config
```
justinmk, line 11: Need a blank line above this. Else the `---@type...` is part of the generated doc. Plus that's more "correct" anyway since a `@brief` doc describes the module, not any particular code in the module.

igorlfs: Oopsie! I swear, this must be a curse: every small contribution I make here comes with a subtle bug 😭 

Follow-up: https://github.com/neovim/nvim-lspconfig/pull/4017

PR #4011, opened 2025-08-19
fix(ts_ls): server not starting if there is no typescript

ViniWei:
This fix a bug that makes ts_ls not start if project is not using typescript.

This is my first time contributing for open source so let me know if i'm doing something wrong.

### DIFF
--- a/doc/configs.md
+++ b/doc/configs.md
@@ -13804,7 +13804,6 @@ Default config:
   ```lua
   { "yaml", "yaml.docker-compose", "yaml.gitlab", "yaml.helm-values" }
   ```
-- `on_init`: [../lsp/yamlls.lua:62](../lsp/yamlls.lua#L62)
 - `root_markers` :
   ```lua
   { ".git" }
@@ -13815,6 +13814,11 @@ Default config:
     redhat = {
       telemetry = {
         enabled = false
+      }
+    },
+    yaml = {
+      format = {
+        enable = true
       }
     }
   }
--- a/doc/configs.txt
+++ b/doc/configs.txt
@@ -10669,7 +10669,6 @@ Default config:
   { "yaml-language-server", "--stdio" }
 - filetypes: >lua
   { "yaml", "yaml.docker-compose", "yaml.gitlab", "yaml.helm-values" }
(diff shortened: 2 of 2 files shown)


justinmk:
Why would you want ts_ls to start if not using typescript?



> This is my first time contributing for open source so let me know if i'm doing something wrong.

Please review https://github.com/neovim/nvim-lspconfig/blob/master/CONTRIBUTING.md .

Especially important to say what the Problem is, and then say what your Solution is, i.e. explain why you are doing this.

ViniWei:
Sorry for that.

The typescript-language-server can be use for javascript only projects, i'm been using it for js only projects and vue projects for over a year, but a recent update broke the compatibility for javascript projects.

justinmk:
Oh, right. I actually don't know why that check was added. @fuegoio what is the purpose of this check?

fuegoio:
> Oh, right. I actually don't know why that check was added. @fuegoio what is the purpose of this check?

Maybe that's my mistake, I thought it was what we wanted. In the previous version, the root markers were `tsconfig.json` and `jsconfig.json`, but also `package.json` and `.git`, so I guess it was always starting even if Typescript was not "configured" for the project.

But I agree we can start the TS LSP on JS pure projects. It will assume a default configuration without any `tsconfig.json` or `jsconfig.json`.

@ViniWei There are also the `tsgo` and `vtsls` LSP to adapt in this case.

ViniWei:
Thank you for pointing that out, @fuegoio . I've now made the same change to vtsls.

I couldn’t get tsgo to work for me to test it, so I haven’t made any changes to it.

yadex:
previous version can start ts_ls without any jsconfig.json or tsconfig, and fast. now it slow speedup

justinmk:
there are merge conflicts, can you rebase

> I couldn’t get tsgo to work for me to test it, so I haven’t made any changes to it.

if it's just deleting the same block, go ahead and do so. better to do it now to avoid confusion.

ViniWei:
Ok @justinmk, thank you for your help.

I think everything is correct now.

PR #4012, opened 2025-08-19
refactor(yamlls): Non-hacky way to enable formatting

liskin:
Faking capabilities isn't necessary to enable formatting support. It's a dynamically registered capability that yamlls announces whenever yaml formatting is enabled in settings, which it isn't by default.

See:
https://github.com/redhat-developer/yaml-language-server/blob/3821411ee8c92e5b7e5ca88f84ab443ae3b2791a/src/yamlServerInit.ts#L128
https://github.com/redhat-developer/yaml-language-server/blob/3821411ee8c92e5b7e5ca88f84ab443ae3b2791a/src/languageserver/handlers/settingsHandlers.ts#L159-L174

Fixes: 63a016437e44 ("feat(yamlls): document formatting support #4003")

### DIFF
--- a/lsp/yamlls.lua
+++ b/lsp/yamlls.lua
@@ -66,9 +66,7 @@ return {
   settings = {
     -- https://github.com/redhat-developer/vscode-redhat-telemetry#how-to-disable-telemetry-reporting
     redhat = { telemetry = { enabled = false } },
+    -- formatting disabled by default in yaml-language-server; enable it
+    yaml = { format = { enable = true } },
   },
-  on_init = function(client)
-    -- https://github.com/redhat-developer/yaml-language-server/issues/486
-    client.server_capabilities.documentFormattingProvider = true
-  end,
 }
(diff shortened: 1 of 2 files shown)


liskin:
@dhruvmanila ^

liskin:
wtf is commitlint on about? https://github.com/conventional-changelog/commitlint/issues/4101 hits hard :-(

Anyway let me know how hard do I need to try to satisfy these lints. Neither seems super relevant:
* commitlint is straight hallucinating about the footer
* lint.sh tells me not to touch `lua/lspconfig/configs/*` but someone touched them last Thursday (4da7247b2b348b4f6cade30a7a7fcb299879d275) so presumably it's okay actually?

ChrisPJohnstone:
I recognise this is my fault for not explaining properly on #4003, apologies, hopefully this clears things up.

So enabling this way let you do formatting manually `:lua vim.lsp.buf.format()` but the lsp doesn't have capabilities registered correctly so if you use auto formatting like outlined in [help pages](https://neovim.io/doc/user/lsp.html#lsp-attach)
```lua
    if not client:supports_method('textDocument/willSaveWaitUntil')
        and client:supports_method('textDocument/formatting') then
        group = vim.api.nvim_create_augroup('my.lsp', {clear=false}),
        buffer = args.buf,
        callback = function()
          vim.lsp.buf.format({ bufnr = args.buf, id = client.id, timeout_ms = 1000 })
        end,
      })
    end
```

It won't enable for the yaml lsp because the capability is registered as false. That's why the hacky solution

I'm not sure if there's a better way to override the LSP capabilities? Imo this should be addressed in the default config though

PR #4016, opened 2025-08-19
feat(yamlls): added lsp capability override

ChrisPJohnstone:
Some back and forth on this, I didn't explain things properly on #4003 and this section got removed in #4012.

The current version of the config will let you do formatting manually `:lua vim.lsp.buf.format()` but the lsp doesn't have capabilities registered correctly so if you use auto formatting like outlined in [help pages](https://neovim.io/doc/user/lsp.html#lsp-attach)
```lua
    if not client:supports_method('textDocument/willSaveWaitUntil')
        and client:supports_method('textDocument/formatting') then
        group = vim.api.nvim_create_augroup('my.lsp', {clear=false}),
        buffer = args.buf,
        callback = function()
          vim.lsp.buf.format({ bufnr = args.buf, id = client.id, timeout_ms = 1000 })
        end,
      })
    end
```
It won't enable for the yaml lsp because `client:supports_method('textDocument/formatting')` returns false.

The hook, while hacky, will mark the capability as true which is worthwhile in my opinion.

@justinmk @liskin

### DIFF
--- a/lsp/yamlls.lua
+++ b/lsp/yamlls.lua
@@ -69,4 +69,8 @@ return {
     -- formatting disabled by default in yaml-language-server; enable it
     yaml = { format = { enable = true } },
   },
+  on_init = function(client)
+    -- Fix for https://github.com/redhat-developer/yaml-language-server/issues/486
+    client.server_capabilities.documentFormattingProvider = true
+  end,
 }
--- a/lsp/yamlls.lua
+++ b/lsp/yamlls.lua
@@ -70,7 +70,10 @@ return {
     yaml = { format = { enable = true } },
   },
   on_init = function(client)
-    -- Fix for https://github.com/redhat-developer/yaml-language-server/issues/486
+    --- https://github.com/neovim/nvim-lspconfig/pull/4016
+    --- Since formatting is disabled by default if you check `client:supports_method('textDocument/formatting')`
+    --- during `LspAttach` it will return `false`. This hack sets the capability to `true` to facilitate
+    --- autocmd's which check this capability
     client.server_capabilities.documentFormattingProvider = true
   end,
 }


justinmk:
waiting for @liskin to comment/approve

ChrisPJohnstone:
I saw the comment about `gq` (not something I was aware of so I wasn't testing it) on https://github.com/neovim/nvim-lspconfig/pull/4003#discussion_r2282742089

I don't currently have a solution but neither of the fixes (#4012 or this one) make `gq` work as intended

justinmk:
`gq` should use LSP formatting, if possible. Don't care at all about overriding the default behavior, if that's what you mean.

liskin:
> The current version of the config will let you do formatting manually `:lua vim.lsp.buf.format()` but the lsp doesn't have capabilities registered correctly so if you use auto formatting like outlined in [help pages](https://neovim.io/doc/user/lsp.html#lsp-attach)
> […]
> It won't enable for the yaml lsp because `client:supports_method('textDocument/formatting')` returns false.

Oh, that's an interesting use-case I hadn't considered. I happen to use https://github.com/lukas-reineke/lsp-format.nvim. That works fine because while `client:supports_method('textDocument/formatting')` is indeed `false` during `LspAttach`, it becomes `true` shortly after. By the time the plugin actually tries to do any formatting, the (dynamic) capability is available/supported. I reckon https://github.com/stevearc/conform.nvim probably behaves the same.

I'm not against putting the hack back, as I recognise the value of being able to use plain neovim without extra plugins. Can we at least improve the comment to explain the actual reason for the hack cf. enabling the formatter in settings, rather than pointing to an old issue full of comments from confused people? :-)

---

That being said, I'd recommend you install https://github.com/stevearc/conform.nvim anyway, as it'll make `gq` work. 🙂
yamlls only provides document formatting rather than range formatting, but conform.nvim does some smart trickery that formats the whole doc, computes the minimal diff, and then applies parts of it. I'm going to switch whenever I get to it. And also it probably does the formatting async or something.

Oh and my earlier comment about `gq` was misguided, because yeah it does nothing by default with yamlls.
What I referred to is that with plain neovim plus nvim-lspconfig and nothing else, the primary way to observe whether textDocument/rangeFormatting is supported is to press `gq`. When textDocument/rangeFormatting is supported, `formatexpr` is set to use it, and thus `gq` will reformat the lines using LSP, otherwise it'll do whatever old vi(m) does. But this is rangeFormatting, not formatting, so nevermind — that's the mistake I made.

justinmk:
> Can we at least improve the comment to explain the actual reason for the hack cf. enabling the formatter in settings, rather than pointing to an old issue full of comments from confused people? :-)

👍 


> When textDocument/rangeFormatting is supported, `formatexpr` is set to use it, and thus `gq` will reformat the lines using LSP

👍

ChrisPJohnstone:
Comment updated with a quick explanation and link to this PR

liskin (approved):

PR #4021, opened 2025-08-20
fix(ts/js): adjust `vim.fs.root` args on nvim < 0.12

BarrensZeppelin:
Prioritization through nested tables is a v0.12 feature.


justinmk:
It was shipped in Nvim 0.11.3 https://github.com/neovim/neovim/commit/5d0766ddceecca32e3764206558eba54850cf64c , which is required by this repo.

There is no reason anyone should be using Nvim 0.11.0-0.11.2.

BarrensZeppelin:
Ah, nice. My bad.

Issue #4015, opened 2025-08-19
js/ts language servers (eslint/biome/svelte/ts_ls/vtsls/tsgo) attach to files that are only in workspace directory (ex. directory contains lock file)


krukroman:
### Description
[Commit 8ad2d8d](https://github.com/neovim/nvim-lspconfig/commit/8ad2d8d8991d868ff8e7c21a7db624f32b882531) introduced changes that restrict mentioned language servers from attaching to files outside the workspace directory. While this behavior makes sense for some servers, it creates inconveniences for JavaScript/TypeScript development.

#### Problem:

* General JS/TS files: Developers often edit standalone JS/TS files not bound to any workspace or project. The current restriction prevents language servers from attaching to these files, reducing functionality (e.g., linting, autocompletion, and diagnostics).
* ESLint/Biome: These servers do not support single files and should only attach when configuration files (e.g., .eslintrc, biome.json) are present in the directory, even if the directory is not a formal workspace.
* Svelte: Unlike JS/TS, the Svelte language server should likely only attach to files within a Svelte project, as standalone Svelte files are rare and less practical.

#### Expected Behavior:

* JS/TS Language Servers: Should attach to any JS/TS file, regardless of workspace context, to support standalone file editing.
* ESLint/Biome: Should attach to files when configuration files are present in the directory, even if it’s not a workspace.
* Svelte Language Server: Should only attach to files within a Svelte project.

#### Current Behavior:

* All mentioned servers are restricted from attaching to files outside the workspace directory, 

<details>
<summary>Previous description</summary>

### Summary

ts_ls server don't attach to javascript or typescript files when there are no root marker files in directory (ex. package-lock.json)

#### Steps to reproduce

1. create minimal neovim config file.

<details><summary>Minimal init.lua</summary>

```
vim.o.clipboard = 'unnamedplus'
vim.o.softtabstop = 2
vim.o.shiftwidth = 2
vim.o.expandtab = true


-- Bootstrap lazy.nvim
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  local lazyrepo = "https://github.com/folke/lazy.nvim.git"
  local out = vim.fn.system({ "git", "clone", "--filter=blob:none", "--branch=stable", lazyrepo, lazypath })
  if vim.v.shell_error ~= 0 then
    vim.api.nvim_echo({
[...]

PR #4022, opened 2025-08-20
fix(ts/js): fixed behavior of root_dir funcion (issue #4015)

krukroman:
This pr is intended to fix issue #4015 
### Problem:

ts_ls, vtsls, tsgo LSP do not start for standalone files
eslint/biome LSP do not start when there is config file but lock file is absent in workspace directory

### Solution:

ts_ls, vtsls, tsgo: Remove redundant truthy check when calling `on_dir(project_root)`. If `project_root` is provided, it will be used as `root_dir`; otherwise, `root_dir` will be `nil`, allowing the LSP to start for standalone files.

eslint/biome: added check for config root directory when project root directory not found

Please, consider that i am a beginner, so I will be grateful for any suggestions for improvements.

### DIFF
--- a/lsp/vtsls.lua
+++ b/lsp/vtsls.lua
@@ -82,12 +82,12 @@ return {
     -- We select then from the project root, which is identified by the presence of a package
     -- manager lock file.
     local project_root_markers = { 'package-lock.json', 'yarn.lock', 'pnpm-lock.yaml', 'bun.lockb', 'bun.lock' }
+
     -- Give the root markers equal priority by wrapping them in a table
     local project_root = vim.fs.root(bufnr, { project_root_markers })
-    if not project_root then
-      return
-    end
 
+    -- If project root not found `project_root` will be nil and on_dir function will be called with nil value
+    -- to allow LSP to start for standalone TS/JS files.
     on_dir(project_root)
   end,
 }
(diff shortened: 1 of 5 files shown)


lsp/vtsls.lua
```diff
@@ -82,12 +82,12 @@ return {
...
     local project_root_markers = { 'package-lock.json', 'yarn.lock', 'pnpm-lock.yaml', 'bun.lockb', 'bun.lock' }
+
     -- Give the root markers equal priority by wrapping them in a table
     local project_root = vim.fs.root(bufnr, { project_root_markers })
-    if not project_root then
-      return
-    end
 
+    -- If project root not found `project_root` will be nil and on_dir function will be called with nil value
+    -- to allow LSP to start for standalone TS/JS files.
```
justinmk, line 90: This can be less verbose. Same for other cases where applicable.

```suggestion
    -- If `project_root=nil`, we still want on_dir() because this server supports "standalone" mode (workspace_required=false).
```

justinmk:
If lsp/eslint.lua and lsp/biome.lua have chosen `workspace_required=true` then we don't need to workaround that. I suggest dropping those changes for now. Looks complicated. Should be a separate PR if at all.

krukroman:
@justinmk , I’ve decided to close this PR, as it only addresses the symptoms of the larger issue that PR #3955 attempted to resolve, but encountered edge cases. At this point, fixing this problem is beyond my current experience, since I believe it is primarily a design issue that requires careful discussion and consideration.

PR #4025, opened 2025-08-21
fix(gitlab_ls): root dir ordering

alesbrelih:
Problem:

If project had a nested child gitlab file named: .gitlab-ci.yml it would take it as root.

Solution:

If inside GIT repository just use git repository root as root and still keep the option for .gitlab* as fallback if there is no git repo yet.


justinmk:
I see you are the original author, so don't need to wonder about the original intent :)

PR #4026, opened 2025-08-22
fix(vue_ls): make tsserver handler more resilient 

Jak-Ch-ll:
## Problem

After updating the Vue Language Server to version 3 I ran into the error "Could not find `ts_ls` or `vtsls` lsp client, required by `vue_ls`." at startup. It did not happen every time but most of the time. After playing around with as part of [this discussion](https://github.com/vuejs/language-tools/discussions/5603), I found out that it only happens when I have the lsp configurations in `/after/lsp`. If I configure everything in the same file the issue does not seem to occur.

I presume that there is some kind of race condition going on which either does not or is far less likely to happen with a single configuration file.

## Solution

In any case since I'm probably not the only one using the `/after/lsp` style of configuration, I think it makes sense to have some kind of solution within lspconfig.

I added a retry into the handler that allows some time for the typescript ls to attach.

## Additional notes

- If there is clean way of having one language server wait for another I would be all ears
- I also made a small refactor which I can split into its own PR if desired

### DIFF
--- a/lsp/vue_ls.lua
+++ b/lsp/vue_ls.lua
@@ -24,12 +24,25 @@ return {
   filetypes = { 'vue' },
   root_markers = { 'package.json' },
   on_init = function(client)
-    client.handlers['tsserver/request'] = function(_, result, context)
+    local retries = 0
+
+    ---@param _ lsp.ResponseError
+    ---@param result any
+    ---@param context lsp.HandlerContext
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks. configs really should not be implemented bespoke "retry" logic, hopefully this can be simplified after `vim.async` is available.

PR #4027, opened 2025-08-22
fix(ts/js): update root_dir function behavior (#4015)

krukroman:
Problem:
js/ts language servers (eslint/biome/ts_ls/vtsls/tsgo) attach to files that are only in workspace directory (ex. directory contains lock file)

Solution:
Drop lock files as root markers.

Search for the project root by moving upwards from the current working directory (cwd). If no folder contains a root marker, the file path becomes the new starting point.
Before the next search, we check if the file is within a nested directory of the cwd. If it is, we search upwards from the file path to the cwd. If a folder with a root marker is found, we set the project root to the cwd to avoid the Language Server Protocol (LSP) from creating multiple clients for nested directories.
If the file is not nested, it means it’s located elsewhere in the file system. In this case, we search upwards from the file path. If a folder with a root marker is found, we set that folder as the project root.


lsp/eslint.lua
```diff
@@ -39,9 +39,95 @@
...
+end
+
+--- @param root_markers string[] List of root-marker files to append to.
+--- @param field string Field to search for in the given `new_names` files.
+--- @param path string Full path to start searching upwards from.
+--- @param stop_path? string Full path to stop searching there.
+--- @return string[]
+local insert_package_json = function(root_markers, field, path, stop_path)
+  return root_markers_with_filed(root_markers, { 'package.json', 'package.json5' }, field, path, stop_path)
+end
```
krukroman: These are modified functions from lspconfig.util which allow to stop search if stop_path provided.

lsp/biome.lua
```diff
@@ -11,7 +11,94 @@
...
+end
+
+--- @param root_markers string[] List of root-marker files to append to.
+--- @param field string Field to search for in the given `new_names` files.
+--- @param path string Full path to start searching upwards from.
+--- @param stop_path? string Full path to stop searching there.
+--- @return string[]
+local insert_package_json = function(root_markers, field, path, stop_path)
+  return root_markers_with_filed(root_markers, { 'package.json', 'package.json5' }, field, path, stop_path)
+end
```
krukroman: These are modified functions from lspconfig.util which allow to stop search if stop_path provided.

justinmk:
Thanks for the effort and thanks for avoiding `util.lua`. However in this case, assuming this logic is actually needed, let's make an exception and place it in `util.lua` until we figure out how to upstream the core idea.

Configs shouldn't require tons of special logic. We need to find what patterns are being used in configs and then upstream that logic into core Nvim, as we did for e.g. `vim.fs.root()`. Meanwhile, `util.lua` is where things can live while we ponder how to upstream them. But also need a clear way to signal that no one outside of this repo should take a dependency on `util.lua` ...

krukroman:
I’ve moved all the complex logic into util.lua to keep things cleaner. I totally get that this logic is a bit involved, so if this approach doesn’t align with the current plans, feel free to close this PR, no worries at all!
That said, if you’d like me to keep iterating on this, I’m happy to refine it further. Just let me know your thoughts or any suggestions you have, I’m all ears and ready to help!

justinmk:
I do think the problem needs to be solved, but the logic here does look too complicated currently.

PR #4033, opened 2025-08-24
fix(docker_language_server): missing config file

guillermodotn:


justinmk:
read the readme. everything in lspconfig/ will be deleted and is not supported. use `lsp/` (`vim.lsp.enable()` in Nvim 0.11.x) and do not use `require'lspconfig'`

PR #4040, opened 2025-08-28
fix(herb): remove ruby filetype

hschne:
As outlined in [this issue on herb](https://github.com/marcoroth/herb/issues/309#issuecomment-3130071214), herb should not actually run on Ruby files. This PR removes the offending filetype.

> Yeah, I think Herb shouldn't be run on .rb files in general. So it would probably be best to remove the ruby file type in the Language Server configuration.

@marcoroth FYI


marcoroth:
Thanks @hschne and @justinmk! 🙏🏼

PR #4046, opened 2025-08-31
feat(gh_actions_ls): add handler for actions/readFile request

jugarpeupv:
Implement a custom handler for the "actions/readFile" request in the gh_actions_ls config. This handler reads the requested file from disk and returns its contents if the file is readable. This improves integration with the GitHub Actions language server by supporting file content requests.

Here is how vscode github actions extension is setting the handler
https://github.com/github/vscode-github-actions/blob/main/src/workflow/languageServer.ts#L68

Solves:
https://github.com/lttb/gh-actions-language-server/issues/5

We could also provide an implementation for populating the init_options properly, now in the docs it is documented as an empty table

But in order for the lsp to work properly, it should be populated with this "shape", i dont know whats your opinion on providing a default implementation for this

```lua

				init_options = {
					sessionToken = session_token,
					repos = {
						{
							id = 1008200293,
							owner = org_name,
							name = repo_name,
							workspaceUri = "file://" .. vim.fn.getcwd(),
							organizationOwned = true,
						},
					},
				},
```

### DIFF
--- a/lsp/gh_actions_ls.lua
+++ b/lsp/gh_actions_ls.lua
@@ -37,8 +37,10 @@ return {
       end
       local file_path = vim.uri_to_fname(result.path)
       if vim.fn.filereadable(file_path) == 1 then
-        local content = vim.fn.readfile(file_path)
-        local text = table.concat(content, '\n')
+        local f = assert(io.open(file_path, 'r'))
+        local text = f:read('*a')
+        f:close()
+
         return text, nil
       end
       return nil, nil


lsp/gh_actions_ls.lua
```diff
@@ -30,7 +30,20 @@ return {
...
-
+  handlers = {
+    ['actions/readFile'] = function(_, result)
+      if type(result.path) ~= 'string' then
+        return nil, nil
+      end
+      local file_path = vim.uri_to_fname(result.path)
+      if vim.fn.filereadable(file_path) == 1 then
+        local content = vim.fn.readfile(file_path)
+        local text = table.concat(content, '\n')
```
justinmk:
```suggestion
        local f = assert(io.open(file_path, 'r'))
        local text = f:read('*a')
        f:close()
```

justinmk: we should probably add `vim.fs.read()` in core 😅

justinmk:
> But in order for the lsp to work properly, it should be populated with this "shape", i dont know whats your opinion on providing a default implementation for this

if nvim-lspconfig can provide a reasonable default, then it should 👍 

should probably leave that for a separate PR though.

lsp/gh_actions_ls.lua
jugarpeupv: modified!

jugarpeupv:
> > But in order for the lsp to work properly, it should be populated with this "shape", i dont know whats your opinion on providing a default implementation for this
> 
> if nvim-lspconfig can provide a reasonable default, then it should 👍
> 
> should probably leave that for a separate PR though.

Agree, i currently have this implementation for populating init_options, which is working well. However i noticed that when i change to a new worktree, then i get warnings just like the language server was not properly configured. I realized it is because, each time i change worktree (change directory) the lsp is not reinitialized with the new init_options, concretely the 'workspace_path' pointing to the new worktree dir

I am unsure about how to deal with this, i am not quite familiar with lua/lsp, but i am keen on contributing, even if it's with little PR's like this one



```lua
M.get_gh_actions_init_options = function(org, workspace_path, session_token)
	org = org or "my-org"
	workspace_path = workspace_path or vim.fn.getcwd()
	session_token = session_token or os.getenv("GH_TOKEN")

	local function get_repo_name()
		local handle = io.popen("git remote get-url origin 2>/dev/null")
		if not handle then
			return nil
		end
		local result = handle:read("*a")
		handle:close()
		if not result or result == "" then
			return nil
		end
		-- Remove trailing newline
		result = result:gsub("%s+$", "")
		-- Extract repo name from URL
		local repo = result:match("([^/:]+)%.git$")
		return repo
	end
	local repo_name = get_repo_name()

	local repo_info = fetch_github_repo(repo_name, session_token, org, workspace_path)
	return {
		sessionToken = session_token,
		repos = {
			repo_info,
		},
	}
end


```

justinmk:
> each time i change worktree (change directory) the lsp is not reinitialized with the new init_options

`init_options` is sent only when the LSP server starts, right? The mechanism in `vim.lsp` for deciding when to start a new server is `reuse_client`, see `:help vim.lsp.Config`.

Example: https://github.com/neovim/nvim-lspconfig/blob/5f550bbb4d58bb2bfbbed4633fc69c3f0eccaf3a/lsp/ast_grep.lua#L15-L18

PR #4047, opened 2025-08-31
fix(expert): reorder the markers

heywhy:
Due to the current order, multiple instances of the LSP are created, even though they both belong to the same root directory. As seen in the screenshot, the project name is proxy, but when I use the 'go to definition' action, the Phoenix module gets opened, and a new LSP is spun up for it.

Screenshot 2025-08-31 at 17 36 15

So, the change in this PR is to mitigate the problem by making sure the .git has higher priority over the mixfile.

### DIFF
--- a/lsp/expert.lua
+++ b/lsp/expert.lua
@@ -7,6 +7,6 @@
 ---@type vim.lsp.Config
 return {
   filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
-  root_markers = { 'mix.exs', '.git' },
+  root_markers = { '.git', 'mix.exs' },
   cmd = { 'expert' },
 }


justinmk:
> they both belong to the same root directory.

 is the mix.exs file at a differnt level than .git ?



> the Phoenix module gets opened

don't know what that means or why it's relevant.

with the new ordering in this PR, any `.git` that is "closer" than `mix.exs` will be treated as the root. that will result in the same (reverse) situation, right?

heywhy:
Let me highlight a few things below;

1. I have a project `proxy`, inside it is the git repository (`.git`)
2. The phoenix module I'm referring to here is a dependency in the `deps` folder of the same project directory.

Basically, when I open one of the project modules and I use the 'go to definition' on a dependency module, I expect that the same language server instance that was spawned for the project itself was reused; instead, a new language server instance gets spawned for the dependency because inside its folder is a `mix.exs` file  (as seen in the screenshot I shared).

But when I flip the `root_markers` order, this issue does not happen.

heywhy:
> with the new ordering in this PR, any .git that is "closer" than mix.exs will be treated as the root. that will result in the same (reverse) situation, right?

To be sincere, I do not expect the order to be an issue because `/path/to/project` and `/path/to/project/deps/phoenix` share the same root (`/path/to/project`). But the current matching strategy seems to be that it finds a mixfile in `/path/to/project/deps/phoenix` and uses that as the root, then spawns a new language instance.

And when I invert the `root_markers` order, the problem disappears.

lsp/expert.lua
```diff
@@ -7,6 +7,6 @@
 ---@type vim.lsp.Config
 return {
   filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
-  root_markers = { 'mix.exs', '.git' },
+  root_markers = { '.git', 'mix.exs' },
```
justinmk, line 10: what version of Nvim do you have? if you have at least Nvim 0.11.2, then you can specify "equal priority" so that at each directory level, both of these are matched at each level.

does this fix your issue:

```suggestion
  root_markers = { { '.git', 'mix.exs' } },
```

if it does, then change this to:

```suggestion
  root_markers = vim.fn.has('nvim-0.11.3') == 1 and { { '.git', 'mix.exs' } } or { '.git', 'mix.exs' },
```

heywhy: I'm on nvim-0.11.3. Trying your suggestion didn't solve the issue, regardless of the order (`root_markers = { { '.git', 'mix.exs' } }` or `root_markers = { { 'mix.exs', '.git' } }`).

PR #4048, opened 2025-08-31
fix(vtsls): inform the LS it's client is neovim

heywhy:
This fix is important for the language to determine the file uri it generates when 'go to definition' action is invoked.


justinmk:
I wonder if `vim.lsp` core should do this by default. Though it seems like `initializationOptions` does not have any [standard fields](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize).

heywhy:
I wish the vim.lsp core did it too, but like you said, there's no standard at the moment, so till then.

PR #4050, opened 2025-09-02
feat(roslyn_ls): support override completion and refresh diagnostics

khoido2003:
### Problem
- Diagnostics were only refreshed for the current buffer, not all buffers of the client.
- Roslyn did not handle override method completion correctly.

### Solution
- Added `roslyn.client.completionComplexEdit` command handler to apply edits for override completions.
- Added `refresh_diagnostics(client)` function and call it on project initialization and relevant buffer events.

### Testing
- Verified diagnostics update across multiple buffers.
- Verified override method completion inserts the correct method body.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -22,7 +22,7 @@
 local uv = vim.uv
 local fs = vim.fs
 
-local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = false })
+local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = true })
 
 ---@param client vim.lsp.Client
 ---@param target string
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -22,6 +22,8 @@
 local uv = vim.uv
 local fs = vim.fs
 
+local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = false })
+
 ---@param client vim.lsp.Client
 ---@param target string
 local function on_init_sln(client, target)
@@ -44,19 +46,30 @@ local function on_init_project(client, project_files)
   })
 end
 
+---@param client vim.lsp.Client
+local function refresh_diagnostics(client)
+  local buffers = vim.lsp.get_buffers_by_client_id(client.id)
+  for _, buf in ipairs(buffers) do
+    if vim.api.nvim_buf_is_loaded(buf) then
+      client:request(
+        vim.lsp.protocol.Methods.textDocument_diagnostic,
+        { textDocument = vim.lsp.util.make_text_document_params(buf) },
+        nil,
+        buf
+      )
+    end
+  end
+end
+
 local function roslyn_handlers()
   return {
     ['workspace/projectInitializationComplete'] = function(_, _, ctx)
       vim.notify('Roslyn project initialization complete', vim.log.levels.INFO, { title = 'roslyn_ls' })
-
-      local buffers = vim.lsp.get_buffers_by_client_id(ctx.client_id)
       local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
-      for _, buf in ipairs(buffers) do
-        client:request(vim.lsp.protocol.Methods.textDocument_diagnostic, {
-          textDocument = vim.lsp.util.make_text_document_params(buf),
-        }, nil, buf)
-      end
+      refresh_diagnostics(client)
+      return vim.NIL
     end,
+
     ['workspace/_roslyn_projectHasUnresolvedDependencies'] = function()
       vim.notify('Detected missing dependencies. Run `dotnet restore` command.', vim.log.levels.ERROR, {
         title = 'roslyn_ls',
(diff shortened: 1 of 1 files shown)


lsp/roslyn_ls.lua
```diff
@@ -147,6 +183,20 @@ return {
       end
     end,
   },
+
+  on_attach = function(client, bufnr)
+    local group = vim.api.nvim_create_augroup('roslyn_ls_' .. client.id, { clear = false })
+
+    vim.api.nvim_create_autocmd({ 'BufWritePost', 'InsertLeave' }, {
+      group = group,
+      buffer = bufnr,
```
justinmk: creating client-specific augroups/autocmds is sus. is this intended to avoid creating redundant autocmds for the same buffer? if so, that can be done by skipping `nvim_create_autocmd` if `#nvim_get_autocmds({...}) > 0`.

so, create the group once time at the top of this file.

then do something like this here:

```suggestion
    if vim.api.nvim_get_autocmds({buffer= bufnr,group=augroup})[1]
      return
    end

    vim.api.nvim_create_autocmd({ 'BufWritePost', 'InsertLeave' }, {
      group = group,
      buffer = bufnr,
```

lsp/roslyn_ls.lua
```diff
@@ -147,6 +183,20 @@ return {
       end
     end,
   },
+
+  on_attach = function(client, bufnr)
+    local group = vim.api.nvim_create_augroup('roslyn_ls_' .. client.id, { clear = false })
```
justinmk:
```suggestion
    local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = false })
```

lsp/roslyn_ls.lua
khoido2003: Thanks for the feedback 🙏. I updated the code:
- Use a single augroup lspconfig.roslyn_ls.
- Guard against duplicate autocmds with nvim_get_autocmds.

Let me know if you’d like further adjustments.

lsp/roslyn_ls.lua
```diff
@@ -147,6 +183,25 @@ return {
       end
     end,
   },
+
+  on_attach = function(client, bufnr)
+    local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = false })
```
justinmk: `clear=false` works, but i think it's a bit more obvious if we just define this once, near the top of this file. this file isn't loaded unless the user actually uses this config, so there's little concern about defining an unnecessary augroup.

lsp/roslyn_ls.lua
```diff
@@ -147,6 +183,25 @@ return {
...
+    -- avoid duplicate autocmds for same buffer
+    if vim.api.nvim_get_autocmds({ buffer = bufnr, group = group })[1] then
+      return
+    end
+
+    vim.api.nvim_create_autocmd({ 'BufWritePost', 'InsertLeave' }, {
+      group = group,
+      buffer = bufnr,
+      callback = function()
+        refresh_diagnostics(client)
```
justinmk, line 199: is this basically implementing https://github.com/neovim/neovim/issues/29792 ? why doesn't the advice there work?

possible also related:

- https://github.com/neovim/neovim/issues/32247
- https://github.com/neovim/neovim/issues/30385

khoido2003: Hi @justinmk,
Thanks for the review and for pointing out the related issues! I’ll address your question about #29792 and the possible relation to #32247 and #30385.

My changes aim to address two specific issues with Roslyn LSP in nvim-lspconfig:

- Diagnostics not updating across all open buffers: Roslyn primarily uses pull-based diagnostics, and Neovim’s default behavior only refreshes diagnostics for the active buffer on text changes. This leaves diagnostics in non-active buffers stale, especially in multi-file C# projects where changes in one file can affect others. The refresh_diagnostics(client) function, triggered on BufWritePost and InsertLeave, ensures diagnostics are pulled for all buffers in the client/workspace, providing a consistent view without waiting for buffer switches.
- Override method completion: Roslyn’s override completions weren’t being applied correctly, so I added the roslyn.client.completionComplexEdit handler to properly apply these edits.

Regarding #29792, I agree it’s related, as it calls for a core Neovim flag like pull_diagnostics_on_save to handle pull-based LSPs more efficiently. However, Roslyn’s reliance on pull diagnostics and project-wide context makes the default debounce_text_changes insufficient, as it’s still tied to didChange events and only updates the current buffer. The autocmds I added are a server-specific workaround to mimic an “on-save” or “on-idle” pull model, ensuring all buffers reflect the latest diagnostics. Without these, users would need to manually trigger refreshes or switch buffers to see updates, which is suboptimal.

justinmk: ok please add that info to the commit message

khoido2003: Thanks @justinmk 🙏
I’ve moved the augroup definition to the top of the file, and I updated the commit message with the explanation about why Roslyn needs explicit diagnostic refreshes (pull-based model, multi-buffer consistency)

lsp/roslyn_ls.lua
```diff
@@ -22,6 +22,8 @@
 local uv = vim.uv
 local fs = vim.fs
 
+local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = false })
```
justinmk:
```suggestion
local group = vim.api.nvim_create_augroup('lspconfig.roslyn_ls', { clear = true })
```

PR #4051, opened 2025-09-03
feat(biome,eslint,svelte,ts_ls,tsgo,vtsls): add deno.lock root marker

kmoschcau:
Problem:
`deno.lock` is not recognized as a root marker in JavaScript related servers.

Solution:
Add `deno.lock` as a root marker.


kmoschcau:
I had this problem with the `svelte` server, which would then go up as far as the file system root and then crash once it found circular symlinks in `/run/udev/watch`. See here: https://github.com/sveltejs/language-tools/issues/2836

PR #4054, opened 2025-09-03
fix(expert): implement root_dir instead

### LINKED ISSUE #4052: eXpert LSP root markers change breaks monorepos
### Description

The change provided here https://github.com/neovim/nvim-lspconfig/pull/4047 breaks monorepo structures when there's no mix.exs files in the same level of the .git folder. 

The recommended approach in the eXpert LSP is to have the [mix.exs](https://github.com/elixir-lang/expert/blob/e510182b90fc931cc0fe9e65d4a241521d990312/pages/installation.md?plain=1#L187) as a lead priority just as in lexical

heywhy:
Copy the approach from `lsp/elixirls.lua` to support monorepos.

fix #4052

### DIFF
--- a/lsp/expert.lua
+++ b/lsp/expert.lua
@@ -3,10 +3,19 @@
 --- https://github.com/elixir-lang/expert
 ---
 --- Expert is the official language server implementation for the Elixir programming language.
+---
+--- 'root_dir' is chosen like this: if two or more directories containing `mix.exs` were found when searching directories upward, the second one (higher up) is chosen, with the assumption that it is the root of an umbrella app. Otherwise the directory containing the single mix.exs that was found is chosen.
 
 ---@type vim.lsp.Config
 return {
   filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
-  root_markers = { '.git', 'mix.exs' },
   cmd = { 'expert' },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
+    local child_or_root_path, maybe_umbrella_path = unpack(matches)
+    local root_dir = vim.fs.dirname(maybe_umbrella_path or child_or_root_path)
+
+    on_dir(root_dir)
+  end,
 }
--- a/lsp/expert.lua
+++ b/lsp/expert.lua
@@ -15,6 +15,8 @@ return {
   cmd = { 'expert' },
   root_dir = function(bufnr, on_dir)
     local fname = vim.api.nvim_buf_get_name(bufnr)
+    --- Elixir workspaces may have `mix.exs` files, e.g. for an "umbrella" layout or monorepo. So we
+    --- specify `limit=2` and treat the highest one (if any) as the root of an umbrella app.
     local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
     local child_or_root_path, maybe_umbrella_path = unpack(matches)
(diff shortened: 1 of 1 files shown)


lsp/expert.lua
```diff
@@ -7,6 +7,13 @@
 ---@type vim.lsp.Config
 return {
   filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
-  root_markers = { '.git', 'mix.exs' },
   cmd = { 'expert' },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
```
justinmk, line 20: did you intentionally omit `.git` ?

`limit=2` seems random. oh, is it because a workspace might have `mix.exs` at multiple levels? interesting. a concise source-code comment here is worth having.

i didn't read the discussion in #4052 , mind saying 1-2 sentences in the PR description here ?

justinmk: oh i see that you copied this from https://github.com/neovim/nvim-lspconfig/blob/master/lsp/elixirls.lua (this would be another case where the PR description can save time :)

can you confirm my guess about limit=2 ? and the `.git` question.

heywhy: I apologize for the dense description, as I assumed you were following the discussion in #4052.

> did you intentionally omit .git 

Yes, this was omitted because a .git folder is redundant in identifying an Elixir project.

> can you confirm my guess about limit=2 

Yes, a workspace can have a `mix.exs` file at multiple levels, especially for Elixir Umbrella structure and the monorepo layout that the author of the issue seems to utilise. In Elixir, you can point to a directory containing a `mix.exs` file as a dependency regardless of the levels.

And just as you've noticed, this is merely a copy of ElixirLS config because both language servers behave the same way. In addition, I was trying to add more lines than needed, hence the no comment, but I would be adding some in a moment.

lsp/expert.lua
```diff
@@ -3,10 +3,19 @@
 --- https://github.com/elixir-lang/expert
 ---
 --- Expert is the official language server implementation for the Elixir programming language.
+---
+--- 'root_dir' is chosen like this: if two or more directories containing `mix.exs` were found when searching directories upward, the second one (higher up) is chosen, with the assumption that it is the root of an umbrella app. Otherwise the directory containing the single mix.exs that was found is chosen.
```
justinmk:
```suggestion
--- 'root_dir' is chosen like this: if two or more directories containing `mix.exs` were found when
--- searching directories upward, the second one (higher up) is chosen, with the assumption that it
--- is the root of an umbrella app. Otherwise the directory containing the single mix.exs that was
--- found is chosen.
```

lsp/expert.lua
```diff
@@ -3,10 +3,22 @@
...
+--- found is chosen.
 
 ---@type vim.lsp.Config
 return {
   filetypes = { 'elixir', 'eelixir', 'heex', 'surface' },
-  root_markers = { '.git', 'mix.exs' },
   cmd = { 'expert' },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
```
justinmk, line 20:
```suggestion
    --- Elixir workspaces may have `mix.exs` files, e.g. for an "umbrella" layout or monorepo. So we
    --- specify `limit=2` and treat the highest one (if any) as the root of an umbrella app.
    local matches = vim.fs.find({ 'mix.exs' }, { upward = true, limit = 2, path = fname })
```

PR #4056, opened 2025-09-04
feat: zuban

YodaEmbedding:
Recently released language server / typechecker by the author of Jedi.

- https://zubanls.com/
- https://github.com/zubanls/zuban


lua/lspconfig/configs/zuban.lua
justinmk, line 1: not adding new configs to `lua/lspconfig/` , see readme

justinmk:
> Zuban is 20–200× faster than Mypy, while using roughly half the memory and CPU compared to Ty and Pyrefly. It offers both a PyRight-like mode and a Mypy-compatible mode, which behaves just like Mypy; supporting the same config files, command-line flags, and error messages.

Sounds cool 👀

PR #4058, opened 2025-09-04
fix: Allowing cmd to be customized for angularls

fjtluz:
I changed so cmd can be easily overriden using angularls, in case the user is not using ngserver, as is my situation


justinmk:
you did not read the readme

fjtluz:
@justinmk really did not, only after lint failure I noticed that things had changed, my bad

PR #4060, opened 2025-09-05
feat(zk): implement the zk.tag.list command

ateoi:


lsp/zk.lua
```diff
@@ -1,74 +1,121 @@
...
   filetypes = { 'markdown' },
-  root_markers = { '.zk' },
-  ---@param bufnr integer
-  ---@param client vim.lsp.Client
+  root_dir = function(bufnr, on_dir)
+    local root = vim.fs.root(bufnr, '.zk')
+    if root ~= nil then
+      on_dir(root)
+    end
+  end,
```
justinmk: what does this improve, vs root_markers? 

if you don't call `on_dir()` when `root_dir` is nil, that will disable single-file support. can set `workspace_required=true` instead. then leave root_markers as it was.

ateoi: Yes, the idea was to disable single-file support. ZK is not a generic markdown language server, it only makes sense to enable it for documents managed by zk.

lsp/zk.lua
```diff
@@ -1,74 +1,117 @@
...
-    vim.api.nvim_buf_create_user_command(bufnr, 'LspZkList', function()
-      local bufpath = vim.api.nvim_buf_get_name(bufnr)
-      local root = find_zk_root(bufpath)
+    vim.api.nvim_buf_create_user_command(bufnr, 'LspZkList', function(_)
+      zk_list(client, bufnr, {}, function(path, _)
+        vim.cmd('edit ' .. path)
+      end)
+    end, { desc = 'ZkList' })
 
+    vim.api.nvim_buf_create_user_command(bufnr, 'LspZkTagList', function(_)
```
justinmk: strange to put `_` for a parameter. why not just omit it?

when a parameter is unused, but the one after it is used, then it makes sense to name the unused one something like `_foo`. but i'm not aware of any reason to name a parameter `_` if it is not followed by *used* parameters.

same for all other cases in this PR

lsp/zk.lua
```diff
@@ -1,74 +1,117 @@
...
+
+    local paths = vim.tbl_map(function(item)
+      return item.path
+    end, result)
+    local titles = vim.tbl_map(function(item)
+      return item.title
+    end, result)
+    vim.ui.select(titles, {}, function(item, idx)
+      if item ~= nil then
+        action(paths[idx], item)
```
justinmk: `tbl_map` uses `pairs()` which is not guaranteed to have stable order, even on the same data. Could do this instead, which avoids needing tbl_map at all:

```lua
vim.ui.select(result, {
  format_item = function(item)
    return item.title
  end,
}, function(item)
  if item ~= nil then
    action(item.path, item.title)
  end
end)
```

lsp/zk.lua
```diff
@@ -1,74 +1,117 @@
...
         end
-        local paths = vim.tbl_map(function(item)
-          return item.path
+
+        local tags = vim.tbl_map(function(item)
+          return item.name
         end, result)
-        vim.ui.select(paths, {}, function(choice)
-          vim.cmd('edit ' .. choice)
+        vim.ui.select(tags, {}, function(item)
```
justinmk: use `{format_item=...}` to avoid the `tbl_map`, as shown in my other comment.

the reason this matters is because people copy what other configs are doing, so doing it right in this config helps avoid problems in the future.

PR #4061, opened 2025-09-06
fix(roslyn_ls): remove unused handler

seblyng:
This handler was removed a long time ago and was replaced with another handler that is already implemented here: https://github.com/dotnet/vscode-csharp/commit/9cc41f6d2535a5cb5533dc5ae8f3a10ade16e7b5


justinmk:
Nice, thanks for providing a reference.

PR #4067, opened 2025-09-07
fix: remove extra table wrapper in root_markers

laryXP:
fix: remove extra table wrapper in root_markers

The previous code wrapped root_markers in an extra table, causing vim.fs.root to fail in Neovim 0.11.3+. This fixes the type error and ensures ts_ls works correctly.

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -59,8 +59,6 @@ return {
     -- We select then from the project root, which is identified by the presence of a package
     -- manager lock file.
     local root_markers = { 'package-lock.json', 'yarn.lock', 'pnpm-lock.yaml', 'bun.lockb', 'bun.lock', 'deno.lock' }
-    -- Give the root markers equal priority by wrapping them in a table
-    root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers } or root_markers
     local project_root = vim.fs.root(bufnr, root_markers)
     if not project_root then
       return


justinmk:
> wrapped root_markers in an extra table, causing vim.fs.root to fail in Neovim 0.11.3+

What is the exact version of Nvim you are using? Nvim 0.11.3 definitely supports this, it signals "equal priority". 

From `:help vim.fs.root()`:

      • {marker}  (`(string|string[]|fun(name: string, path: string): boolean)[]|string|fun(name: string, path: string): boolean`)
                  Filename, function, or list thereof, that decides how to
                  find the root. To indicate "equal priority", specify items
                  in a nested list `{ { 'a.txt', 'b.lua' }, … }`.

justinmk:
Also. to update a PR please just push to your existing branch. Don't create new PRs.

laryXP:
Thanks for pointing that out.  
I’m currently running Neovim v0.12.0-dev (built with LuaJIT 2.1).  
It seems that in this dev build, wrapping `root_markers` in an extra list causes `vim.fs.root()` to fail, while in 0.11.3+ it should still work.  

So the issue might be related to changes between 0.11.3 stable and the 0.12.0 development branch.

lithammer:
> Thanks for pointing that out.  
> 
> I’m currently running Neovim v0.12.0-dev (built with LuaJIT 2.1).  
> 
> It seems that in this dev build, wrapping `root_markers` in an extra list causes `vim.fs.root()` to fail, while in 0.11.3+ it should still work.  
> 
> 
> 
> So the issue might be related to changes between 0.11.3 stable and the 0.12.0 development branch.

The issue is that your nightly version is too old.

PR #4068, opened 2025-09-10
fix(jdtls): remove nonstandard progress handler

rmuir:
Problem:
The nonstandard progress handler was added in #2153 but it can cause some undesirable side-effects out of box ("press enter"). The original bug (mfussenegger/nvim-jdtls#327) was fixed upstream in the language server three years ago (eclipse-jdtls/eclipse.jdt.ls#2258).

Solution:
Remove the workaround. Standard progress works.


justinmk:
Thanks! The explanation and references are helpful.

PR #4069, opened 2025-09-10
feat(harper): add shell filetype

wiserfz:
This PR is add shell filetype to `harper_ls` configuration.


justinmk:
Should harper just be enabled in all filetypes or what?

wiserfz:
> Should harper just be enabled in all filetypes or what?

harper support several languages, docs [here](https://writewithharper.com/docs/integrations/language-server#Supported-Languages), but I think it's can't be enabled in all filetypes.

PR #4073, opened 2025-09-14
feat: StyLua Lua formatter

PolyMeilex:
StyLua [now supports LSP](https://github.com/JohnnyMorganz/StyLua/pull/970) out of the box.


dmitrii-galantsev:
in case others are also confused _**you need stylua 2.2.0 or above**_.
My stylua from nixpkg was 2.1.0 and confused me with `error: Found argument '--lsp' which wasn't expected, or isn't valid in this context`.

PR #4076, opened 2025-09-17
fix(biome,eslint,svelte,ts_ls,tsgo,vtsls): remove deno.lock root marker

### LINKED ISSUE #4074: Unexpectedly ts_ls stating at deno workspace
### Description

Deno typically uses denols. This is because Deno's import mechanism is special, and it won't function properly even if ts_ls is running.

This issue was introduced in https://github.com/neovim/nvim-lspconfig/pull/4051. Since this seems like an unusual situation, would it be okay if I submit a PR to revert it for now?

kmoschcau:
fix(biome,eslint,svelte,ts_ls,tsgo,vtsls): remove deno.lock root marker

Problem:
Adding `deno.lock` caused other servers to start in deno projects where
they should not.

Solution:
Remove `deno.lock` root marker from these other server configurations.

This reverts commit 7d5cf94d70b003385f4410746614b0c8e4d6b63d.

This closes #4074.

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -44,7 +44,7 @@ return {
     -- As stated in the documentation above, this LSP supports monorepos and simple projects.
     -- We select then from the project root, which is identified by the presence of a package
     -- manager lock file.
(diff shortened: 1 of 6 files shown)


justinmk:
pr diff has no changes

kmoschcau:
... I am not sure how the original commit got into this branch again. I am certain it wasn't there yesterday.

PR #4079, opened 2025-09-18
fix(angularls): look for node_modules in subfolders and find correct path for mason Angular LS

### LINKED ISSUE #3859: AngularLs does not properly detect current @angular/core version
### Description

Hey, 

At the moment the angular language server in my project starts with an empty string as parameter for --angularCoreVersion
This leads to features not being enabled in the angular language server. 

The reason is in my opinion this line:
``` lua
local node_modules_dir = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
```
in the angularls.lua file. 
The upward = true setting seems to be wrong. The node_modules folder should be a subfolder of the root directory.
Locally changing it to 
``` lua
local node_modules_dir = vim.fs.find('node_modules', { path = root_dir })[1]
```
fixed the issue of a missing version number.

Also in cases where the version can not be determined, the `--angularCoreVersion` parameter should not be added to the program arguments. right now the parameter just exists with an empty value, which prevents the default handling of the Language server.

I suggest changing the return statement like this: 
``` lua
    local tbl = {
        cmd = {
            'ngserver',
            '--stdio',
            '--tsProbeLocations',
            ts_probe_dirs,
            '--ngProbeLocations',
            ng_probe_dirs,
        },
        filetypes = { 'typescript', 'html', 'typescriptreact', 'typescript.tsx', 'htmlangular' },
        root_markers = { 'angular.json', 'nx.json' },
    }
    if default_angular_core_version ~= "" then
        table.insert(tbl["cmd"], '--angularCoreVersion')
        table.insert(tbl["cmd"], default_angular_core_version)
    end
    return tbl
```

Best regards

erikson84:
First commit fixes #3859 by removing an option to look for node_modules folder upwards in the dir hierarchy from the root folder.

Second commit fixes uncorrect node_modules path in extension path when the language server is installed via mason and falls back to the current strategy if not.

### DIFF
--- a/lsp/angularls.lua
+++ b/lsp/angularls.lua
@@ -57,7 +57,14 @@ local default_angular_core_version = get_angular_core_version()
 --   - typescript
 local ngserver_exe = vim.fn.exepath('ngserver')
 local ngserver_path = #(ngserver_exe or '') > 0 and vim.fs.dirname(vim.uv.fs_realpath(ngserver_exe)) or '?'
-local extension_path = vim.fs.normalize(vim.fs.joinpath(ngserver_path, '../../../'))
+
+local extension_path = ''
+
+if ngserver_path:find('mason', 1, true) then
+  extension_path = vim.fs.joinpath(vim.fn.stdpath('data'), 'mason/packages/angular-language-server/node_modules')
+else
+  extension_path = vim.fs.normalize(vim.fs.joinpath(ngserver_path, '../../../'))
+end
 
 -- angularls will get module by `require.resolve(PROBE_PATH, MODULE_NAME)` of nodejs
 local ts_probe_dirs = vim.iter({ extension_path, default_probe_dir }):join(',')
--- a/lsp/angularls.lua
+++ b/lsp/angularls.lua
@@ -15,7 +15,7 @@
 -- Angular requires a node_modules directory to probe for @angular/language-service and typescript
 -- in order to use your projects configured versions.
 local root_dir = vim.fn.getcwd()
-local node_modules_dir = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
+local node_modules_dir = vim.fs.find('node_modules', { path = root_dir })[1]
 local project_root = node_modules_dir and vim.fs.dirname(node_modules_dir) or '?'
 
 local function get_probe_dir()


erikson84:
I have commited a fix for the stylua issue - removed double quotes (edit: and corrected whitespace!)

lsp/angularls.lua
```diff
@@ -57,7 +57,14 @@ local default_angular_core_version = get_angular_core_version()
 --   - typescript
 local ngserver_exe = vim.fn.exepath('ngserver')
 local ngserver_path = #(ngserver_exe or '') > 0 and vim.fs.dirname(vim.uv.fs_realpath(ngserver_exe)) or '?'
-local extension_path = vim.fs.normalize(vim.fs.joinpath(ngserver_path, '../../../'))
+
+local extension_path = ''
+
+if ngserver_path:find('mason', 1, true) then
+  extension_path = vim.fs.joinpath(vim.fn.stdpath('data'), 'mason/packages/angular-language-server/node_modules')
```
justinmk, line 64: no, we aren't going to add special support for mason. that is not sustainable. there are 1000s of configs. mason should be no different than any other package manager.

lsp/angularls.lua
```diff
@@ -15,7 +15,7 @@
 -- Angular requires a node_modules directory to probe for @angular/language-service and typescript
 -- in order to use your projects configured versions.
 local root_dir = vim.fn.getcwd()
-local node_modules_dir = vim.fs.find('node_modules', { path = root_dir, upward = true })[1]
+local node_modules_dir = vim.fs.find('node_modules', { path = root_dir })[1]
```
justinmk, line 18: I think I mentioned when this config was first created, that getcwd() in the module is only going to accidentally work if the user starts Nvim in the right directory. That's a rather flimsy assumption. It's not surprising that this config has bugs.

The correct approach would be to move this logic to a function, that is called by `reuse_client` or maybe change `cmd` to a function.

lsp/angularls.lua
erikson84: It's understandable. This is a specific mason issue related to how it organize its binary files. I thought about a generic approach to search inside Neovim's data folder, but I dno't think that brute focing in the general LSP config is the right way - the user should know the path and pass it correctly in the config, if it's not the original project folder.

lsp/angularls.lua
erikson84: Thanks for the tip! I'm changing the code so cmd is a function and properly resolve the root dir at runtime. I will cancel this pull request and submit another when I'm done.

erikson84:
I'm closing this pull request to work on the feedback and give a better solution for root path resolution, probably by making cmd a function.

PR #4080, opened 2025-09-18
feat(efm): add some default settings

jspricke:
The user still needs to enable them by setting the filetpyes.
Also make efm less noisy by passing -q.

### DIFF
--- a/lsp/efm.lua
+++ b/lsp/efm.lua
@@ -20,6 +20,80 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'efm-langserver' },
+  cmd = { 'efm-langserver', '-q'},
   root_markers = { '.git' },
+  settings = {
+    languages = {
+      cmake = {
+        {
+          formatCommand = "cmake-format -",
+          formatStdin = true
+        },
+        {
+          lintCommand = "cmake-lint",
+          lintFormats = {
+            "%f:%l,%c: %m"
+          }
+        }
+      },
+      json = {
+        {
+          formatCommand = "python3 -m json.tool",
+          formatStdin = true
+        },
+        {
+          lintCommand = "python3 -m json.tool",
+          lintStdin = true,
+          lintIgnoreExitCode = true,
+          lintFormats = {
+            "%m: line %l column %c (char %r)"
+          }
+        }
+      },
+      markdown = {
+        {
+          formatCommand = "pandoc -f markdown -t gfm -sp --tab-stop=2",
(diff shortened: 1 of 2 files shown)


lsp/efm.lua
```diff
@@ -20,6 +20,80 @@
...
   root_markers = { '.git' },
+  settings = {
+    languages = {
+      cmake = {
+        {
+          formatCommand = "cmake-format -",
+          formatStdin = true
+        },
+        {
+          lintCommand = "cmake-lint",
```
justinmk, line 33: uh... what's the plan here? why aren't these maintained in https://github.com/mattn/efm-langserver itself?

It's easy to imagine this config becoming a "configs collection" inside of nvim-lspconfig (which is itself a "configs collection"), reinventing null-ls or [none-ls](https://github.com/nvimtools/none-ls.nvim).

There are 100s of formatters and linters out there.

justinmk:
This looks like a useful sample for people to use as a starting point. So it's fine. But don't think it's scalable.

Will merge after the lint issues are fixed.

lsp/efm.lua
polyzen: There's already https://github.com/creativenull/efmls-configs-nvim

jspricke: I was not aware of that, thanks!

(Closing as that is better.)

polyzen: Also maybe check out https://github.com/mfussenegger/nvim-lint

jspricke: Thanks but that one does not do formatters from what I see.

polyzen: For formatting: https://github.com/stevearc/conform.nvim

There's also https://github.com/chrisgrieser/nvim-rulebook which is an "all-around helper" that I've been meaning to try.

jspricke: Thanks for all the pointer but currently it feels still like efm (with the efmls-configs-nvim) is the better option. Or do you have other reasons?

polyzen: https://github.com/nvimtools/none-ls.nvim#motivation

justinmk: Perhaps those are worth mentioning in the docstring for this config.

jspricke: Mentioning at least https://github.com/creativenull/efmls-configs-nvim sound good.

@polyzen I just switched to nvim-lint and conform :). (somehow efm broke..)

PR #4081, opened 2025-09-19
feat(apex_ls): add vim.lsp.config support

andrewderekjackson:
Adds lsp config for apex language service, essentially just converting the existing legacy config to the new format.

https://github.com/neovim/nvim-lspconfig/issues/3705

### DIFF
--- a/lsp/apex_ls.lua
+++ b/lsp/apex_ls.lua
@@ -1,0 +1,59 @@
+--- @brief
+---
+--- https://github.com/forcedotcom/salesforcedx-vscode
+---
+--- Language server for Apex.
+---
+--- For manual installation, download the JAR file from the [VSCode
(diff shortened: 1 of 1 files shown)


lsp/apex_ls.lua
```diff
@@ -0,0 +1,59 @@
...
+--- })
+--- ```
+
+---@type vim.lsp.Config
+return {
+  cmd = function(dispatchers, config)
+    local local_cmd = {
+      vim.env.JAVA_HOME and (vim.env.JAVA_HOME .. '/bin/java') or 'java',
+      '-cp',
+      config.apex_jar_path,
```
justinmk, line 42: Interesting approach. And since `cmd()` gained the `config` arg, this is much more ergonomic. We should keep this is mind for other configs.

andrewderekjackson: I can't take any credit for that I'm afraid - the approach was shamefully pillaged from the biome config.

Issue #3850, opened 2025-05-13
Sorbet needs cmd_cwd


doits:
### Description

This is about the sorbet ls: https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lsp/sorbet.lua

If I open a file while my current working directory isn't the root dir of the project, I get this error:

```
Sorbet's language server requires a single input directory. However, 0 are configured: []
```

This happens launching the language server manually, too (`srb tc --lsp` not in project root).

To fix it, `cmd_cwd` can be set, for example like this:

```lua
vim.lsp.config('sorbet', {
  cmd_cwd = vim.fs.root(0, 'Gemfile')
})
```

Though I am not sure why this is needed, since it should be set automatically to the `root_dir` if I am correct (which should be found by `root_markers`?):

https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lua/lspconfig/manager.lua#L118-L120

And I don't see any other ls config make use of `cmd_cwd` (can't imagine sorbet is the only ls with this requirement, or is it?).

So is `cmd_cwd` needed here? Or is there some other issue?

NVIM v0.11.1
Build type: Release
LuaJIT 2.1.1744318430

justinmk:
> Though I am not sure why this is needed, since it should be set automatically to the `root_dir` if I am correct (which should be found by `root_markers`?):

from `:help lsp` : 

```
      • {cmd_cwd}?             (`string`, default: cwd) Directory to launch
                               the `cmd` process. Not related to `root_dir`.
```

`nvim-lspconfig/lua/lspconfig/manager.lua` is irrelevant, not used for `vim.lsp.config`.

regarding this:

```
vim.lsp.config('sorbet', {
  cmd_cwd = vim.fs.root(0, 'Gemfile')
})
```

that `cmd_cwd` will be evaluated *once*, and depends on whatever random buffer might be open. it's not evaluated for each buffer.

powerman:
@justinmk There is similar issue with `ast_grep`: it either should be run inside project dir (which won't happens when we run nvim in one project's dir and later open a file from another project) or get extra `--config /another/project/sgconfig.yml` args.

How this can be fixed? This won't work (probably because it's too late and `cmd` is already executed):
```lua
        before_init = function(_, config)
            config.cmd_cwd = config.root_dir
        end,
```

[...]

PR #4082, opened 2025-09-19
fix(stylua): set cmd_cwd to root_dir for config lookup

rmuir:
PROBLEM

Stylua LSP server does not make use of rootUri in any way. If nvim is launched from a directory where cwd != rootDir, stylua will apply the incorrect configuration (internal defaults), and reformat code with an unexpected style.

SOLUTION

Set the cmd_cwd via reuse_client(), similar to ast_grep. See #3850 for more details.

### DIFF
--- a/lsp/stylua.lua
+++ b/lsp/stylua.lua
@@ -7,6 +7,11 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'stylua', '--lsp' },
+  -- server looks for configuration in CWD only, does not use rootUri
+  reuse_client = function(client, config)
+    config.cmd_cwd = config.root_dir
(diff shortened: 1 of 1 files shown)


justinmk:
isn't this a server bug ? it should respect the root_dir that is passed to it, CWD shouldn't matter

was this reported to stylua LS?

rmuir:
I will take the issue over there: honestly didn't think it was a bug (the spec is super-vague), just another oddity of how stylua processes configuration files, and that we were "holding it wrong".

justinmk:
i'm not objecting to this since it looks pretty low-risk and clever. but also think it's worth reporting and discussing. please link to the upstream issue here

PR #4083, opened 2025-09-19
refactor(angularls): better root path resolution

erikson84:
Expanding on #4079, I refactored the whole Angular LS config so that it employs Neovim's LSP root resolution and created a generic method to collect possible node_modules paths, including if ngserver is installed somewhere inside Neovim's data folder.

Problem:
The current config for Angular LS make strong assumptions when trying to find the root dir, which can lead to unexpected LSP crashes.

Solution:
By defining the 'cmd' field as a function, the config employs Neovim's LSP root resolution to identify the correct path and find the relevant node_modules folder.


josephemorgan:
For what it's worth, this fixed `angularls` on my setup (Windows 11)

justinmk:
Yeah this looks better, thanks. The old version was doing stuff at module-load time, which is just wrong.

BBboy01:
This slowed down the speed of first enter a project (almost 2 seconds), while in a monorepo project, every time you enter a new sub repo, 2 seconds will be hanging…

justinmk:
@BBboy01 do you realize the old config would not have sent the correct `--tsProbeLocations` in the case you are describing? It would have only checked once, when Nvim starts.

There are several places in the config where `node_modules` is searched, can you add logging to see which one is slow

BBboy01:
> @BBboy01 do you realize the old config would not have sent the correct `--tsProbeLocations` in the case you are describing? It would have only checked once, when Nvim starts.

Sorry, I didn't realize this because it works fine on my local project.

> There are several places in the config where node_modules is searched, can you add logging to see which one is slow

looks like `globpath` takes too long

nvim-profile-flamegraph

BBboy01:
> looks like globpath takes too long

```bash
cd $XDG_PATH/nvim
fd "node_modules"
```

CleanShot 2025-11-04 at 14 16 51@2x

BBboy01:
https://github.com/helix-editor/helix/pull/10166

Do we really need to get all the paths? Seems like angular language server will handle this.

justinmk:
> [helix-editor/helix#10166](https://github.com/helix-editor/helix/pull/10166)
> 
> Do we really need to get all the paths? Seems like angular language server will handle this.

That makes a lot of sense to me, PR welcome...

erikson84:
helix solution seems to assume the user has installed an is running the AngularLS and its depenencies from the current project (it points to '.').

The collect_node_modules should not recover all node_modules inside Neovim's data folder, only those where the AngularLS is installed. If the user installed the AngularLS via some package manager inside Neovim (like Mason), it should find the adequate LS and TS library paths. The glob strategy can indeed be slow if the user has many plugins installed.

justinmk:
@erikson84 can you review https://github.com/neovim/nvim-lspconfig/pull/4182

BBboy01:
> helix solution seems to assume the user has installed an is running the AngularLS and its depenencies from the current project (it points to '.').

Yes indeed, since initialize an angular project with `npx @angular/cli new` not include `@angular/language-service` and `@angular/language-service` already is a dependency of `@angular/language-server`, helix solution may not a good choice.

Logic from official Angular VS Code extension:

https://github.com/angular/vscode-ng-language-service/blob/774ef052df2a8c04c4247c2579c883c67ca0ae7a/client/src/client.ts#L367
https://github.com/angular/vscode-ng-language-service/blob/774ef052df2a8c04c4247c2579c883c67ca0ae7a/client/src/client.ts#L446

https://github.com/angular/vscode-ng-language-service/blob/774ef052df2a8c04c4247c2579c883c67ca0ae7a/client/src/client.ts#L617-L627

Check priority order: 
  1. Angular VS Code extension path
  2. all of opened workspace folders.

In my opinion, from `vim.fn.exepath('ngserver')` we already get the "Angular VS Code extension path", so we don't need to search all `$XDG_DATA_PATH/nvim`

erikson84:
> In my opinion, from `vim.fn.exepath('ngserver')` we already get the "Angular VS Code extension path", so we don't need to search all `$XDG_DATA_PATH/nvim`

VSCode extension can get away with it because it can refer to its own install path; nvim-lspconfig does not install any language server so it must make some assumptions about where to find the executable. If the AngularLS doesn't have a global shim (like if it is installed in Neovim's data folder via Masor or another manager), it won't be enough. But many LS configs assume there is some globally available executable or shim for the language server.

@justinmk What do you think? Maybe we can work as other configs and pass the responsability of providing a shim for the server to the user, so we can do away with the glob inside Neovim's data foler.

erikson84:
> @justinmk What do you think? Maybe we can work as other configs and pass the responsability of providing a shim for the server to the user, so we can do away with the glob inside Neovim's data foler.

I just experimented a little more and the glob search strategy (or something similar, like `vim.fs.find`) still will be needed in Windows because the exposed executable is wrapped in a `.cmd` file, both when installed in the project or via Mason.

BBboy01:
> If the AngularLS doesn't have a global shim (like if it is installed in Neovim's data folder via Masor or another manager), it won't be enough. But many LS configs assume there is some globally available executable or shim for the language server.

If a user has not globally installed ngserver nor added it to Neovim's runtimepath, then it is the user's responsibility to specify where the plugin is installed.

justinmk:
> Maybe we can work as other configs and pass the responsability of providing a shim for the server to the user, so we can do away with the glob inside Neovim's data foler.

That is the usual approach. I don't get why angular is special (please educate me).

As a compromise, how about putting the glob logic in a `LspAngularFind` (or similar) command? Then users can try that on-demand to get the path.

erikson84:
@BBboy01 @justinmk You are right about the user's responsibility in providing a proper globally available executable or shim! My assumption was based on my local setup on Windows and using Mason, but it was wrong. Mason does put the executable in the runtime path; the problem is how Windows wraps the server in a `.cmd` file. I'm working on a version that does not require the costy glob by processing the `.cmd` file when the user is on Windows.

erikson84:
And it's done! The PR is here: #4190

PR #4084, opened 2025-09-20
fix(leanls): migrate to vim.lsp.config

StamesJames:
I have no experience with the lspconfigs, but I saw that the `leanls.lua` was the only one missing in `lsp/` compared to `lua/lspconfig/configs/`. I simply copied the old settings and migrated them.
This is related to the Issue [427](https://github.com/Julian/lean.nvim/issues/427) the lean.nvim repo. I will also write a PR that changes the use of the old `require('lspconfig').leanls` to `vim.lsp.enable` and `vim.lsp.config` but this will only work once the `leanls.lua` is in `lsp/`
I tested this PR with an updated `lean.nvim` plugin already using `vim.lsp.enable` and `vim.lsp.congig` and it seems to work.

### DIFF
--- a/lsp/leanls.lua
+++ b/lsp/leanls.lua
@@ -1,0 +1,45 @@
+---@brief
+--- https://github.com/leanprover/lean4
+---
+--- Lean installation instructions can be found
+--- [here](https://leanprover-community.github.io/get_started.html#regular-install).
+---
+--- The Lean language server is included in any Lean installation and
+--- does not require any additional packages.
+---
+--- Note: that if you're using [lean.nvim](https://github.com/Julian/lean.nvim),
+--- that plugin fully handles the setup of the Lean language server,
+--- and you shouldn't set up `leanls` both with it and `lspconfig`.
+
+---@type vim.lsp.Config
+return {
+  cmd = function(dispatchers, config)
+    local local_cmd = { 'lake', 'serve', '--', config.root_dir }
+    return vim.lsp.rpc.start(local_cmd, dispatchers)
+  end,
+  filetypes = { 'lean' },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    fname = vim.fs.normalize(fname)
+    -- check if inside lean stdlib
+    local stdlib_dir
+    do
+      local _, endpos = fname:find '/lean/library'
+      if endpos then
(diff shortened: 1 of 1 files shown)


StamesJames:
I am not sure if I understand the linter message for `single_file_support = true,` right. As I understood it, this is now true by default, so I discarded it completely from the config.

lsp/leanls.lua
```diff
@@ -0,0 +1,46 @@
...
+--- and you shouldn't set up `leanls` both with it and `lspconfig`.
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'lake', 'serve', '--' },
+  filetypes = { 'lean' },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    fname = vim.fs.normalize(fname)
+    -- check if inside elan stdlib
```
justinmk: i assume this is a typo ?

```suggestion
    -- check if inside lean stdlib
```

lsp/leanls.lua
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+
+    on_dir(
+      vim.fs.root(fname, { 'lakefile.toml', 'lakefile.lean', 'lean-toolchain' })
+        or stdlib_dir
+        or vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1])
+    )
+  end,
+  on_new_config = function(config, root_dir)
```
justinmk: this is not supported by `vim.lsp.config`

the top of https://github.com/neovim/nvim-lspconfig/issues/3705 mentions a potential way instead: see [here](https://github.com/neovim/nvim-lspconfig/pull/4081/files#r2364565897).

in this case of lean, it looks like it doesn't even need a custom field on the config, it just needs to define `cmd` as a function, and then use `config.root_dir` from the `cmd()` function.

lsp/leanls.lua
StamesJames: Oh, yes, this is probably true. I thought they mean `elan` the lean version manager in the other files.

lsp/leanls.lua
StamesJames: Oh yes I will look into this. I even have read this Issue but than still copied the old config without thinking :sweat_smile:

StamesJames:
Like this, it should now call the cmd with the root_dir as before.

justinmk:
Thanks!

PR #4085, opened 2025-09-21
Fallback root directories for `foam_ls`

FoamScience:
Although not common, the associated LSP (`foam_ls`) can handle single files just fine.
And Neovim's LSP client errors out if a `nil` root folder is returned; so we look for the git root, or just use the file's folder.

I only made changes to `lsp/foam_ls.lua`. Let me know if changes are also required for the one in `lua/lspconfig/configs`.

### DIFF
--- a/doc/configs.md
+++ b/doc/configs.md
@@ -2400,11 +2400,11 @@ Default config:
   {
     editorInfo = {
       name = "Neovim",
-      version = "0.12.0-dev+gc2136e3590"
+      version = "0.12.0-dev+gb756a6165a"
     },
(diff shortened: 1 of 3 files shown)


justinmk:
> or just use the file's folder.

Note that will result in a new client for each file.

FoamScience:
> > or just use the file's folder.
> 
> Note that will result in a new client for each file.

Thanks for the quick merge! I didn't think about that! but I think it's alright, as for most users the root dir will be resolved correctly; dangling single files are not useful in this language.

PR #4087, opened 2025-09-21
feat(eslint): align legacy eslint config with new lsp/eslint.lua config

kapral18:
copy paste lsp solution parts (`root_dir` + `before_init` bodies) into legacy eslint.lua counterparts


kapral18:
lint failed apparently no changes are expected in legacy configs

justinmk:
That is correct. Please read the readme, and the pinned issue.

No one should be using `lua/lspconfig/configs/` .

PR #4090, opened 2025-09-21
Adjust jdtls root_marker list.

dareni:
Alter the order for .git so it becomes a catch all rather than an identifier for an explicit type of java project. Appending .git to the marker list will alleviate root identification problems where the java project directory is a subdirectory of the directory containing the .git marker.

### DIFF
--- a/lsp/jdtls.lua
+++ b/lsp/jdtls.lua
@@ -63,6 +63,20 @@ local function get_jdtls_jvm_args()
   return unpack(args)
 end
 
+local root_markers1 = {
+  -- Multi-module projects
+  'build.gradle',
+  'build.gradle.kts',
+  -- Single-module projects
+  'build.xml', -- Ant
+  'pom.xml', -- Maven
+  'settings.gradle', -- Gradle
+  'settings.gradle.kts', -- Gradle
+}
+local root_markers2 = {
+  '.git',
+}
+
 ---@type vim.lsp.Config
 return {
   cmd = {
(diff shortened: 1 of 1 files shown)


justinmk:
Read about "equal priority" (requires Nvim 0.11.3+) : https://neovim.io/doc/user/lsp.html#lsp-root_markers

To support Nvim before 0.11.3 (which did not have the "equal priority" feature), need to do something like this:

```lua
local root_markers1 = {
  -- Multi-module projects
  'build.gradle',
  'build.gradle.kts',
  -- Single-module projects
  'build.xml', -- Ant
  'pom.xml', -- Maven
  'settings.gradle', -- Gradle
  'settings.gradle.kts', -- Gradle
}
local root_markers2 = {
  '.git'
}

root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers1, root_markers2 } or vim.list_extend(root_markers1, root_markers2)
```

justinmk:
p.s. please do not create new PRs to update a PR. just force-push to your existing branch instead.

dareni:
Ok. Great, thanks.

deathbeam:
@dareni I made this pull request that basically reverts this: https://github.com/neovim/nvim-lspconfig/pull/4104

So just wanted to ask here if there was any specific scenario where nested maven project actually did break jdtls? Because this change causes the opposite issue where it basically breaks multi module maven projects instead (a lot more common scenario than nested maven project that do not even really breaks usually, at least from my experience and works fine regardless)

PR #4091, opened 2025-09-23
fix(commands): Support ids and names in LspStop. Add bang forcing

pete-coleman:
The 0.11.2 version of `LspStop` did not support using id to stop client.

This PR adds it as well as adding bang support and updating the help to be more descriptive/accurate.


justinmk:
Do not create new PRs to update a PR. Just force-push to your existing branch instead.

pete-coleman:
Yeah sorry, I created a PR from master and I didn't know how active it was here and so moved it to a feature branch.

PR #4092, opened 2025-09-23
fix(commands): support ids and names in lspstop and add bang forcing

pete-coleman:
The 0.11.2 version of LspStop did not support using id to stop client.

This PR adds it as well as adding bang support and updating the help to be more descriptive/accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,27 @@ Most of the time, the reason for failure is present in the logs.
 
 * `:LspInfo` (alias to `:checkhealth vim.lsp`) shows the status of active and configured language servers.
 * `:LspStart <config_name>` Start the requested server name. Will only successfully start if the command detects a root directory matching the current config.
-* `:LspStop [<client_id_or_name>]` Stops the given server. Defaults to stopping all servers active on the current buffer. To force stop add `++force`
+* `:LspStop[!] [<client_id_or_name>] ...`
+  * Stops one or more clients by client id or config name.
+  * If no arguments are provided, all clients are stopped.
+  * If [!] is used, forcefully stops clients even if they are still initializing.
+
+This command adapts its behaviour based on the Neovim version:
+
+### Neovim >= 0.11.2
+
+* Supports [!] (bang) to forcefully stop clients.
+* Example:
+   `:LspStop 1 tsserver`
+   `:LspStop! pyright`
+
+### Neovim < 0.11.2
+
+* Does not support [!] (bang).
+* Use `++force` to forcefully stop clients.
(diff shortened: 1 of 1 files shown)


justinmk:
No, not going to complicate these just to support people that can't be bothered to update to the latest *stable* Nvim version. 

Also, 

> Do not create new PRs to update a PR. Just force-push to your existing branch instead.

pete-coleman:
Maybe it's my mistake but this updates the version of the command that is used on the stable version of neovim. The legacy version is untouched and the help file (should) explain that.

justinmk:
The latest stable Nvim is 0.11.4. There is no reason to support users on older versions. This command is not that important, and it's not that difficult for users to upgrade. Or if it is, then they can either live without this command or copy it to their configs.

When documentation starts saying things like:

> This command adapts its behaviour based on the Neovim version:

that's a signal to ask if this is really needed. That kind of things makes docs horrible to read and maintain.

pete-coleman:
@justinmk this is the 0.11.2 and above version of the command. That's what I'm saying, the command I'm updating is the new one, not the old one.

pete-coleman:
If you're happy for me to remove the 0.11.2 check and remove the "legacy" commands altogether I'll happily add that to this branch?

pete-coleman:
> When documentation starts saying things like:
> 
> This command adapts its behaviour based on the Neovim version:
> that's a signal to ask if this is really needed. That kind of things makes docs horrible to read and maintain.

Whilst I agree, this was already being done, all I've done is document it so it's at least clear what it's doing rather than just silently providing a different command to users

justinmk:
oh, by "the 0.11.2 version of LspStop", you mean 0.11.2+, not 0.11.2 or older 👍

pete-coleman:
> oh, by "the 0.11.2 version of LspStop", you mean 0.11.2+, not 0.11.2 or older 👍

I knew it wasn't clear, my bad!

justinmk:
If this is only adding support for `:LspStop <client_id>`, then it still isn't wanted. This was mentioned in the related PR that implemented the current form.

Users have no reason to be passing client-ids to an interactive command. 

And *plugins* absolutely should be using the programmatic interface, which is `vim.lsp.stop_client()`. 

Thus, this is a solution that does not solve a problem.

Again, when the docs become complicated, that is a signal to ask why we are doing something. Right now the docs are simple: pass a name. Why are we bothering people by explaining "oh you can pass a client-id too... even though that is already possible with `vim.lsp.stop_client`"

PR #4098, opened 2025-09-24
feat(lsp/ocamllsp.lua): use root_markers instead of root_dir

edwintorok:
This drops the dependency on `lspconfig.util`, which appears deprecated.

This also makes it possible to use just the config file without the rest of the framework.

The root pattern grouping and prioritisation follows the one proposed here: https://github.com/ocaml/ocaml.org/pull/3322/files#diff-6721080d4f565e09348cab4e8fe6d6393b9e4449bf9c7e2b3f054a0b986809d6R224

Tested with NeoVim v0.11.4

### DIFF
--- a/lsp/ocamllsp.lua
+++ b/lsp/ocamllsp.lua
@@ -9,8 +9,6 @@
 --- opam install ocaml-lsp-server
 --- ```
 
-local util = require 'lspconfig.util'
-
 local language_id_of = {
   menhir = 'ocaml.menhir',
   ocaml = 'ocaml',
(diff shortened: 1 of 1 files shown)


lsp/ocamllsp.lua
```diff
@@ -20,17 +18,30 @@ local language_id_of = {
...
   filetypes = { 'ocaml', 'menhir', 'ocamlinterface', 'ocamllex', 'reason', 'dune' },
-  root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern('*.opam', 'esy.json', 'package.json', '.git', 'dune-project', 'dune-workspace')(fname))
-  end,
+  root_markers = {
+    { 'dune-project', 'dune-workspace' },
+    { '*.opam', 'esy.json', 'package.json' },
+    '.git',
+  },
```
justinmk: The "nested" format requires Nvim 0.11.3, hopefully we don't have any ocaml users on older Nvim :)

Other configs check `has('nvim-0.11.3') == 1` to decide the format

lsp/ocamllsp.lua
```diff
@@ -20,17 +18,30 @@ local language_id_of = {
...
 return {
   cmd = { 'ocamllsp' },
   filetypes = { 'ocaml', 'menhir', 'ocamlinterface', 'ocamllex', 'reason', 'dune' },
-  root_dir = function(bufnr, on_dir)
-    local fname = vim.api.nvim_buf_get_name(bufnr)
-    on_dir(util.root_pattern('*.opam', 'esy.json', 'package.json', '.git', 'dune-project', 'dune-workspace')(fname))
-  end,
+  root_markers = {
+    { 'dune-project', 'dune-workspace' },
+    { '*.opam', 'esy.json', 'package.json' },
```
justinmk: I don't think `root_markers` supports wildcard/glob. Depends on https://github.com/neovim/neovim/pull/33771

Is `'*.opam'` an important case?

lsp/ocamllsp.lua
edwintorok: Thanks, I'll update.

edwintorok: Updated, probably best to keep backwards compat with older NeoVim releases for now.

lsp/ocamllsp.lua
edwintorok: Most of the time I'd expect to find a `dune-project` or `.git`, it'd be very rare for a project to have neither.
`*.opam` should be future-proof though, once neovim supports it then it'll start working (until then I assume it'd match just a literal `*.opam` file).

However I've added `opam`, some projects have either an `opam` file at their root (very simple ones that only define one `opam` project), or an `opam` directory (containing all the `*.opam` files), and this one should work without wildcards.

PR #4100, opened 2025-09-27
feat(lsp/vale-ls): Add asciidoc filetype

benomahony:
Adding Asciidoc filetype to vale-ls

Question: Would you like me to add other doc formats vale-ls supports?


justinmk:
> Question: Would you like me to add other doc formats vale-ls supports?

Why not? We need users (like you) to make those judgements. I don't know anything about this LS.

justinmk:
> Question: Would you like me to add other doc formats vale-ls supports?

do you plan to do that here?

benomahony:
Apologies for the delay, I've added a couple more but I think it's good now!

PR #4104, opened 2025-09-29
fix(jdtls): increase priority of .git marker and add mvnw and gradlew

### LINKED ISSUE #4102: default lsp rooter jdtls config is incorrect for multi module maven projects
### Description

When in multi-module maven project the structure looks like this:

```
module-a/pom.xml
module-b/pom.xml
pom.xml
```

where pom.xml is parent that includes all submodules. With the default rooter config using pom.xml and navigating to module-a it will only start the lsp in module-a workdir, which usually breaks in multi-module projects that depend on eachother (example is impl depending on dto with parent including both)

There is no easy solution to this as its hard to distinguish between child and parent pom.xml.

One solution is to just have .git in rooter config as well, but this was recently removed because of nested java projects, but this use-case is arguably *way less* common than multimodule maven project starting from root git dir, so at minimum this should be reverted (and it do not even breaks jdtls if for example cwd is root dir and the java project is in lets say my-project/pom.xml etc, so i am not sure why it was even removed in first place).

Another improvement is to add mvnw to root files, its similar to gradlew but used less commonly (but projects do use it sometimes).

deathbeam:
Using git directory as last resort before defaulting to pom.xml is important for multi module maven projects. Example of structure:

pom.xml
submodule-a/pom.xml
submodule-b/pom.xml

If sumbodule-b depends on submodule-a (common pattern for dto/impl), the parent usually has info about linking the 2 projects (and without that go to definition etc will jump to compiled source that might not be present/up t date).

This structure is also fine and do not breaks jdtls:

my-project/pom.xml

e.g its fine if working directory is git repo root (and this pattern is also much less common than multi module maven projects that without this change break always).

Closes #4102

### DIFF
--- a/lsp/jdtls.lua
+++ b/lsp/jdtls.lua
@@ -65,16 +65,22 @@ end
 
 local root_markers1 = {
   -- Multi-module projects
+  'mvnw',
+  'gradlew',
   'build.gradle',
   'build.gradle.kts',
+  -- Use git directory as last resort for multi-module maven projects
+  -- In multi-module maven projects it is not really possible to determine what is the parent directory
+  -- and what is submodule directory. And jdtls do not breaks if the parent directory is at higher level than
+  -- actual parent pom.xml so propagating all the way to root git directory is fine
+  '.git',
+}
+local root_markers2 = {
   -- Single-module projects
   'build.xml', -- Ant
   'pom.xml', -- Maven
   'settings.gradle', -- Gradle
   'settings.gradle.kts', -- Gradle
-}
-local root_markers2 = {
-  '.git',
 }
 
 ---@type vim.lsp.Config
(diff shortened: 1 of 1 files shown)


lsp/jdtls.lua
```diff
@@ -65,17 +65,23 @@ end
 
 local root_markers1 = {
   -- Multi-module projects
+  'mvnw',
+  'gradlew',
   'build.gradle',
   'build.gradle.kts',
+  -- Use git directory as last resort for multi-module maven projects
+  -- In multi-module maven projects it is not really possible to determine what is the parent directory
+  -- and what is submodule directory. And jdtls do not breaks if the parent directory is at higher level than
```
justinmk:
```suggestion
  -- and what is submodule directory. And jdtls does not break if the parent directory is at higher level than
```

PR #4132, opened 2025-10-17
fix(ruby_lsp): prevent duplicate clients on second buffer

stask:
Problem: When opening multiple Ruby buffers, a second LSP client is spawned instead of reusing the first client, even though they share the same root directory.

Root cause: The `reuse_client` function sets `cmd_cwd` as a side effect during reuse checks. This means the first client is created without `cmd_cwd` set, causing the reuse check to fail for the second buffer.

Solution: Use `before_init` to ensure `cmd_cwd` is set for every client before initialization. Remove the custom `reuse_client` function to use Neovim's default reuse logic (compare name + root_dir).

Testing: Tested with Neovim 0.11.4 by opening 3 Ruby files. All buffers now correctly reuse the first client.

Note: `lsp/ast_grep.lua` has the identical pattern and likely has the same issue.

### DIFF
--- a/lsp/ruby_lsp.lua
+++ b/lsp/ruby_lsp.lua
@@ -20,8 +20,7 @@ return {
   init_options = {
     formatter = 'auto',
   },
-  reuse_client = function(client, config)
+  before_init = function(_, config)
     config.cmd_cwd = config.root_dir
-    return client.config.cmd_cwd == config.cmd_cwd
   end,
 }


lsp/ruby_lsp.lua
```diff
@@ -20,8 +20,12 @@ return {
   init_options = {
     formatter = 'auto',
   },
+  on_new_config = function(config, root_dir)
```
justinmk: `on_new_config` is nothing a thing in vim.lsp.config. did you actually test this? 

https://github.com/neovim/nvim-lspconfig/pull/4081 shows an alternative approach for ensuring that a `cmd` has the correct CWD for LSes that (wrongly) care about that.

lsp/ruby_lsp.lua
```diff
@@ -20,8 +20,12 @@ return {
...
+    config.cmd_cwd = root_dir
+  end,
   reuse_client = function(client, config)
-    config.cmd_cwd = config.root_dir
-    return client.config.cmd_cwd == config.cmd_cwd
+    local client_cwd = client.config.cmd_cwd or client.root_dir
+    local config_cwd = config.cmd_cwd or config.root_dir
+    return client_cwd == config_cwd
```
justinmk: why is this thing forcing root_dir to always be equal to CWD? 

i would think this whole `reuse_client` can be dropped, assuming the other comment is addressed.

lsp/ruby_lsp.lua
stask: You're right, this part is unnecessary, i was trying different things and the change in `reuse_client` is enough. 
removing it.

lsp/ruby_lsp.lua
stask: The custom `reuse_client` was added in commit 903045b for monorepo support, but I don't fully understand why cmd_cwd comparison is necessary.
I don't have any monorepos that use ruby. This worked for me and i think it doesn't break the original monorepo support.

stask:
Updated the fix based on feedback:

1. `on_new_config` doesn't exist in vim.lsp.config - switched to using `before_init` to set `cmd_cwd` instead
2. Removed the custom `reuse_client` function entirely - once `cmd_cwd` is properly set via `before_init`, the default Neovim reuse logic (compare name + root_dir) works correctly

This approach:
- Fixes the duplicate client bug (tested by opening 3 Ruby files)
- Preserves monorepo support (cmd_cwd is set to root_dir for every client)
- Simplifies the code by using default behavior

The commit has been squashed and the message updated.

lsp/ruby_lsp.lua
```diff
@@ -20,8 +20,7 @@ return {
   init_options = {
     formatter = 'auto',
   },
-  reuse_client = function(client, config)
+  before_init = function(_, config)
     config.cmd_cwd = config.root_dir
```
justinmk, line 24: sadly, `before_init` [happens](https://github.com/neovim/neovim/blob/07915941014307b054cbbb131369377647b7fb06/runtime/lua/vim/lsp.lua#L183-L190) *after* the `cmd` is invoked, which means setting the `cmd_cwd` won't have the effect that was intended here.

justinmk:
this will need a rebase on https://github.com/neovim/nvim-lspconfig/pull/4408

Issue #4136, opened 2025-10-18
`:LspStop ++force`: Invalid server name '++force'


superatomic:
### Description

The feature to force stop a language server with `++force` (https://github.com/neovim/nvim-lspconfig/pull/2294) appears to have been removed (I assume in the transition from `require('lspconfig')` to `vim.lsp`).

To reproduce:
- Start a language server
- `:LspStop ++force`

superatomic:
Not relevant to the bug, but related: As a subjective change, I would support using `:LspStop!` instead of `:LspStop ++force` for when this feature is reimplemented. The bang isn't being used for anything, and it's shorter than adding `++force`.

justinmk:
Removed from docs. Not clear what its purpose was but it's no longer supported.

superatomic:
I would assume it used to do the equivalent running `kill -9` on the LSP server. It's relatively important as a feature, as `lua-language-server` will refuse to handle SIGINT and SIGTERM signals (and is not stopped by `:LspStop`).

justinmk:
`:LspStop` should do that automatically after waiting X seconds, similar to how `jobstop()` works. Don't need a `++force` for that.

If that feature is missing, PR welcome.

justinmk:
Copying from https://github.com/neovim/nvim-lspconfig/pull/4140#issuecomment-3426270993 for reference: 

> `enable(false)` already calls `:stop()`: https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp.lua#L590-L591
> 
> oh, but I see this PR adds a `true` arg to "force" it.
> 
> but you can already get "force" behavior [by calling `LspStop` twice](https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp/client.lua#L864-L866)? 
> 
> > --- By default, it will just request the server to shutdown without force. If
> > --- you request to stop a client which has previously been requested to
> > --- shutdown, it will automatically escalate and force shutdown.
> 
> 
> If that's not good enough, I suggest sending a PR to Nvim core (and we can backport it to Nvim 0.11) which will automatically "escalate" after X seconds, similar to [what we already do on Nvim exit](https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp.lua#L1169-L1172).

PR #4140, opened 2025-10-20
feat: resupport force stop language server

superatomic:
Reintroduces behavior added in #2294. See #4136 for discussion.

Differences from the old implementation:
- Uses `:LspStop!` instead of `:LspStop ++force`. If this is an undesirable change, I can implement it the original way.
- Supports the same behavior for `:LspRestart`.

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -128,33 +128,31 @@ if vim.fn.has('nvim-0.11.2') == 1 then
 
     -- Default to restarting all active servers
     if #clients == 0 then
-      clients = vim
-        .iter(vim.lsp.get_clients())
-        :map(function(client)
-          return client.name
-        end)
-        :totable()
+      clients = vim.lsp.get_clients()
     end
 
-    for _, name in ipairs(clients) do
+    for client in vim.iter(clients) do
+      local name = client.name
       if vim.lsp.config[name] == nil then
         vim.notify(("Invalid server name '%s'"):format(name))
       else
         vim.lsp.enable(name, false)
+        if info.bang then
+          client:stop(true)
+        end
       end
     end
 
     local timer = assert(vim.uv.new_timer())
     timer:start(500, 0, function()
-      for _, name in ipairs(clients) do
-        vim.schedule_wrap(function(x)
-          vim.lsp.enable(x)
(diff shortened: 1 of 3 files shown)


justinmk:
`enable(false)` already calls `:stop()`: https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp.lua#L590-L591

oh, but I see this PR adds a `true` arg to "force" it.

but you can already get "force" behavior [by calling `LspStop` twice](https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp/client.lua#L864-L866)? 

> --- By default, it will just request the server to shutdown without force. If
> --- you request to stop a client which has previously been requested to
> --- shutdown, it will automatically escalate and force shutdown.


If that's not good enough, I suggest sending a PR to Nvim core (and we can backport it to Nvim 0.11) which will automatically "escalate" after X seconds, similar to [what we already do on Nvim exit](https://github.com/neovim/neovim/blob/b67eff38fe19876ab228007897224ec04b58aa40/runtime/lua/vim/lsp.lua#L1169-L1172).

superatomic:
It is true that you can call `LspStop` twice to achieve this force behavior. The issue is that some servers require being forced-stopped, and having to run `:LspStop` twice every single time in these instances can be annoying.

I'll take a look at adding escalation behavior to neovim core.

plugin/lspconfig.lua
```diff
@@ -128,33 +128,31 @@ if vim.fn.has('nvim-0.11.2') == 1 then
 
     -- Default to restarting all active servers
     if #clients == 0 then
-      clients = vim
-        .iter(vim.lsp.get_clients())
-        :map(function(client)
-          return client.name
-        end)
-        :totable()
+      clients = vim.lsp.get_clients()
```
justinmk, line 131: good idea 👍

justinmk:
The code here is actually a nice improvement, and the bang variant is quite reasonable. So let's merge this.



> I'll take a look at adding escalation behavior to neovim core.

That will be appreciated, it's the standard pattern that all such "teardown"/"stop" features should follow.

Issue #4137, opened 2025-10-20
JDTLS config creates the wrong root folder for a multi project gradle setup


NickJAllen:
### Description

When using lsp-config/lsp/jdtls.lua to configure Java projects that are multi project it detects the wrong project root folder. For each sub project it uses the folder with a build.gradle file instead of the root folder with the settings.gradle and gradlew file. This is with neovim 0.11.4. Here is the relevant info that I see with LspInfo command after opening a Java file in one of the sub projects:

  - cmd: <function @/Users/nick/.local/share/nvim-kickstart/lazy/nvim-lspconfig/lsp/jdtls.lua:77>
  - filetypes: java
  - init_options: {
      extendedClientCapabilities = {
        advancedExtractRefactoringSupport = true,
        advancedOrganizeImportsSupport = true,
        classFileContentsSupport = true,
        executeClientCommandSupport = true,
        generateConstructorsPromptSupport = true,
        generateDelegateMethodsPromptSupport = true,
        generateToStringPromptSupport = true,
        hashCodeEqualsPromptSupport = true,
        inferSelectionSupport = { "extractMethod", "extractVariable", "extractConstant", "extractVariableAllOccurrence" },
        moveRefactoringSupport = true,
        overrideMethodsPromptSupport = true
      }
    }
  - root_markers: { { "mvnw", "gradlew", "build.gradle", "build.gradle.kts", ".git" }, { "build.xml", "pom.xml", "settings.gradle", "settings.gradle.kts" } }

Because build.gradle is defined as a root marker it detects the build.gradle file in the sub project and uses that as the root.

NickJAllen:
To me it looks like the project root markers defined are the wrong way round. settings.gradle is for multi module projects not single module projects as the comments suggest.

justinmk:
Please review these recent changes and see if they make sense to you:

- https://github.com/neovim/nvim-lspconfig/pull/4107
- https://github.com/neovim/nvim-lspconfig/pull/4105
- https://github.com/neovim/nvim-lspconfig/pull/4104

cc @deathbeam @dareni

deathbeam:
I havent used gradle in long time so I was not double checking the original gradle related stuff that was already there but yes @NickJAllen is correct, the gradle markers should be other way around: https://docs.gradle.org/current/userguide/multi_project_builds.html

PR #4142, opened 2025-10-22
fix(jdtls): swap settings.gradle and build.gradle in root markers

deathbeam:
settings.gradle is marker for multi-module Gradle projects.

See: https://docs.gradle.org/current/userguide/multi_project_builds.html
Closes: #4137

### DIFF
--- a/lsp/jdtls.lua
+++ b/lsp/jdtls.lua
@@ -52,10 +52,10 @@ end
 
 local root_markers1 = {
   -- Multi-module projects
-  'mvnw',
-  'gradlew',
-  'build.gradle',
-  'build.gradle.kts',
+  'mvnw', -- Maven
+  'gradlew', -- Gradle
+  'settings.gradle', -- Gradle
+  'settings.gradle.kts', -- Gradle
   -- Use git directory as last resort for multi-module maven projects
   -- In multi-module maven projects it is not really possible to determine what is the parent directory
   -- and what is submodule directory. And jdtls does not break if the parent directory is at higher level than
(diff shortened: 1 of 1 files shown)


d2fn:
@deathbeam I'm using nvim-lspconfig on a multimodule Java gradle project and my lsp integration using nvim-jdtls stopped working. After a round with `git bisect` I tracked this commit `daeb18c3a646f93bebafffe350b6b0cf9659d784` down as the root cause. Any advice?

deathbeam:
> @deathbeam I'm using nvim-lspconfig on a multimodule Java gradle project and my lsp integration using nvim-jdtls stopped working. After a round with `git bisect` I tracked this commit `daeb18c3a646f93bebafffe350b6b0cf9659d784` down as the root cause. Any advice?

Check `:LspInfo` and check if Root directory under `jdtls` is correct. Should be matching settings.gradle location. The implementation should be compliant with the official docs so that would be the expected behaviour: https://docs.gradle.org/current/userguide/multi_project_builds.html

Or check if your structure differs from the linked one

d2fn:
The structure of my project is correct per the official docs and exactly matches what you linked to. The root is correct in `:LspInfo`. `:LspStart` was working fine. The only change is this commit.

d2fn:
Looking closer at the change, it looks like this commit broke the ability to handle a build.gradle at the project root, which is quite common for multimodule projects and correct per the official gradle docs.

deathbeam:
> Looking closer at the change, it looks like this commit broke the ability to handle a build.gradle at the project root, which is quite common for multimodule projects and correct per the official gradle docs.

Its probably better to just open issue for this but this can break only if settings.gradle is not present at all and you are not in git repo so if you start in submodule with build.gradle, and parent dir also has build.gradle it will get resolved to the submodule.

d2fn:
I must offer an apology -- While this commit did technically break my setup, it did so in a way that drew some attention to ways I could improve my multi module setup and make things more correct. This is a large Netflix codebase with its own quirks. After ensuring that different modules were all attaching to the same lsp with the same root, things are actually working better now. Thanks.

PR #4146, opened 2025-10-23
Add 'terraform' to supported filetypes

deade1e:
Does it make sense to specify "terraform" as filetype when using tofu_ls? By default .tf files get identified as "terraform", even if they are opentofu.

### DIFF
--- a/lsp/tofu_ls.lua
+++ b/lsp/tofu_ls.lua
@@ -6,6 +6,6 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'tofu-ls', 'serve' },
-  filetypes = { 'opentofu', 'opentofu-vars' },
+  filetypes = { 'opentofu', 'opentofu-vars', 'terraform' },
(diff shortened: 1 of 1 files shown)


justinmk:
> By default .tf files get identified as "terraform", even if they are opentofu.

Are they different formats? Does terraform have a different LS? Does opentofu LS support non-opentofu terraform files?

deade1e:
> > By default .tf files get identified as "terraform", even if they are opentofu.
> 
> Are they different formats? Does terraform have a different LS? Does opentofu LS support non-opentofu terraform files?

I have no extensive experience in tofu/terraform, but from what I know they should be mostly interchangeable as tofu is derived from terraform.

To be quick, Terraform changed its license to a closed one and then a fork appeared: OpenTofu.

Terraform appears to have its LS. It's even documented/added in nvim-lspconfig.

Neovim does recognize .tf files as "terraform" by default.

PR #4151, opened 2025-10-24
feat(rust-analyzer): out-of-the-box support for codelens

Atan-D-RP4:
Codelens requires a bit of special configuration to work on rust-analyzer as described here: neovim/neovim#34353 (comment)


justinmk:
This looks useful but should it be mentioned in the  docstring (which will update the generated docs)?

PR #4152, opened 2025-10-24
fix: Lsp[Stop/Restart] fix

### LINKED ISSUE #4153: Passing arguments to LspRestart causes it to fail
### Description

Running LspRestart with any LSP name (e.g. `LspRestart lua_ls`) fails with the following error:
```
Error executing Lua callback: /usr/share/nvim/runtime/lua/vim/lsp.lua:388: name: expected non
-wildcard string, got nil                                                                    
stack traceback:                                                                             
        [C]: in function 'error'                                                             
        vim/shared.lua: in function 'validate'                                               
        /usr/share/nvim/runtime/lua/vim/lsp.lua:388: in function 'validate_config_name'      
        /usr/share/nvim/runtime/lua/vim/lsp.lua:408: in function '__index'                   
        ...ocal/share/nvim/lazy/nvim-lspconfig/plugin/lspconfig.lua:136: in function <...ocal
/share/nvim/lazy/nvim-lspconfig/plugin/lspconfig.lua:126> 
```

Poking around, it seems to only happen since this commit - https://github.com/neovim/nvim-lspconfig/commit/0d90bb9bfb8e8b6b64f4a19b7dd3619f5d2788b5

From a quick look, each client is now expected to be a table, but when passing names into the function, each client will be a string. This breaks everywhere that uses `client.name`, which is now nil.

Atan-D-RP4:
Problem:
0d90bb9 broke Lsp[Stop/Restart]  due accessing vim.lsp.config with potentially nil/invalid name with vim.lsp.config[name] in the case of #clients == 0 being false.

Fix:
Just added an `else` clause that makes the clients variable a vim.lsp.Clients[] instead list/table of str's.

fix #4153

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -129,6 +129,15 @@ if vim.fn.has('nvim-0.11.2') == 1 then
     -- Default to restarting all active servers
     if #clients == 0 then
       clients = vim.lsp.get_clients()
+    else
+      clients = vim
+        .iter(clients)
+        :map(function(client_name)
+          return vim.lsp.get_clients(function(client)
+            return client.name == client_name
+          end)
+        end)
+        :to_table()
     end
 
     for client in vim.iter(clients) do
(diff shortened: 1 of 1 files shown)


justinmk:
@superatomic do you see a way to address this without 20 lines of new code ?

superatomic:
Oops! I completely forgot to test the modified commands with arguments. Sorry for breaking `master`.

> @superatomic do you see a way to address this without 20 lines of new code ?

There's one way to do this: revert most of the changes I made in #4140, but replace `client:stop(true)` with `vim.iter(vim.lsp.get_clients({ name = name })):each(function(client) client:stop(true) end)`. This allows for stopping a client by name, instead of changing the code to require a client. I can open a PR with these changes if you would prefer that to this current one.

justinmk:
fixed in https://github.com/neovim/nvim-lspconfig/pull/4159

PR #4157, opened 2025-10-27
Add Gitlab Duo LSP configuration

tachyons:
# Add GitLab Duo Language Server configuration

## Summary

This PR adds support for the [GitLab Duo Language Server](https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp), enabling AI-powered code suggestions through the Language Server Protocol.

## What is GitLab Duo?

GitLab Duo is GitLab's AI-powered coding assistant that provides intelligent code completion and suggestions directly in your editor. The LSP implementation allows any LSP-compatible editor to integrate with GitLab Duo.

## Key Features

- **OAuth Device Flow Authentication**: Secure authentication using GitLab's OAuth device flow
- **Automatic Token Refresh**: Handles token expiration and refresh automatically
- **Inline Code Completion**: Native support for Neovim's inline completion API (requires Neovim 0.12+)
- **Multi-language Support**: Works with 25+ programming languages including Python, JavaScript, TypeScript, Go, Rust, and more
- **User Commands**: Convenient commands for authentication management:
  - `:LspGitLabDuoSignIn` - Authenticate with GitLab
  - `:LspGitLabDuoSignOut` - Sign out and remove tokens
  - `:LspGitLabDuoStatus` - View authentication status

## Prerequisites

- Node.js and npm installed
- GitLab account with Duo Pro license
- Neovim 0.12+ (for inline completion support)

## Current Limitations

**This initial implementation only supports GitLab.com.** Support for self-managed GitLab instances will be added in a future iteration. The configuration is structured to make this addition straightforward when the OAuth application setup is documented for self-managed instances.

### Basic Setup

```lua
vim.lsp.enable("gitlab_duo")
```

### Enable Inline Completion

Add this to your config to enable inline completions with Tab acceptance:

```lua
vim.api.nvim_create_autocmd('LspAttach', {
  callback = function(args)
    local client = vim.lsp.get_client_by_id(args.data.client_id)
    
    if vim.lsp.inline_completion then
      vim.lsp.inline_completion.enable(true, { bufnr = args.buf })
      
      -- Tab to accept
      vim.keymap.set('i', '<Tab>', function()
        if vim.lsp.inline_completion.is_visible() then
          return vim.lsp.inline_completion.accept()
        else
          return '<Tab>'
        end
      end, { expr = true, buffer = args.buf })
    end
  end
})
```

### Authentication Flow

1. Run `:LspGitLabDuoSignIn` in any buffer with the LSP attached
2. Your browser will open to GitLab's authorization page
3. Authorize the application
4. The LSP will automatically authenticate and start providing suggestions

Tokens are stored locally and refreshed automatically.

## Implementation Details

- Uses `npx` to run the GitLab LSP package from GitLab's npm registry
- Implements OAuth device flow for secure authentication
- Token storage in Neovim's data directory
- Automatic token refresh with 60-second expiration buffer
- Error handling for token validation and feature state changes
- Telemetry disabled by default
- Hardcoded to gitlab.com in this iteration

## Testing

Tested on:
- Neovim v0.12+
- Multiple file types (Lua, Python, JavaScript, TypeScript, Go, Rust)
- Token refresh flow
- Authentication error handling

## Future Enhancements

- Support for self-managed GitLab instances with custom URLs
- Configurable OAuth client credentials for self-managed setups

## Related Links

- [GitLab LSP Repository](https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp)
- [GitLab Duo Documentation](https://docs.gitlab.com/ee/user/project/repository/code_suggestions/)

Screenshot 2025-11-08 at 7 55 25 PM
Screenshot 2025-11-08 at 7 55 46 PM

### DIFF
--- a/lsp/gitlab_duo.lua
+++ b/lsp/gitlab_duo.lua
@@ -370,7 +370,6 @@ return {
     },
   },
   settings = {
-    token = get_valid_token() or '',
     baseUrl = config.gitlab_url,
     logLevel = 'info',
     codeCompletion = {


lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,426 @@
...
+---
+---       -- Alt/Option+] for next suggestion
+---       vim.keymap.set('i', '<M-]>', vim.lsp.inline_completion.select_next,
+---         { buffer = bufnr, desc = 'GitLab Duo: Next suggestion' })
+---     end
+---   end
+--- })
+--- ```
+
+local curl = require('plenary.curl')
```
justinmk: Nvim 0.12 has `vim.net.request` but currently does not support POST :(

We have gone to a lot of trouble to keep all the configs in this repo totally "isolated" (self-contained), so would like to avoid a dependency on `plenary`.

Writing a `curl` wrapper for the purposes of this config using `vim.system` should not be much code.

Alternatively, we are interested in gaining POST support for `vim.net.request` if you want to work on that.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,426 @@
...
+--- 3. Enable inline completion in LspAttach event (see example below)
+---
+--- **Inline Completion Example:**
+--- ```lua
+--- vim.api.nvim_create_autocmd('LspAttach', {
+---   callback = function(args)
+---     local bufnr = args.buf
+---     local client = assert(vim.lsp.get_client_by_id(args.data.client_id))
+---
+---     if client.name == 'gitlab_duo' and
```
justinmk: can we drop this condition? similar to https://github.com/neovim/nvim-lspconfig/blob/master/lsp/copilot.lua , there is nothing service-specific about inline_completion. The `desc` field should be generic like in the copilot config.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,426 @@
...
+---     end
+---   end
+--- })
+--- ```
+
+local curl = require('plenary.curl')
+
+-- Configuration
+local config = {
+  gitlab_url = vim.env.GITLAB_URL or 'https://gitlab.com',
```
justinmk: Just mentioning: users can provide arbitrary fields using `vim.lsp.config()`:

```lua
vim.lsp.config('gitlab_duo', {
  gitlab_duo = {
    gitlab_url = '...',
  },
})
```

Then in callbacks such as `cmd()` in this config, which are passed a `config` object, you can access those fields:

```lua
cmd = function(..., config)
  local foo = config.gitlab_duo
  ...
end)
```

lsp/gitlab_duo.lua
tachyons: I removed plenary dependency, thanks to Claude. I'm new to lua, So I may not be able to make substantial contributions for now

lsp/gitlab_duo.lua
tachyons: @justinmk  It is now removed, I was bit worried if it can have unexpected behaviours if we haven't added allow list of client names.

lsp/gitlab_duo.lua
tachyons: @justinmk I couldn't get this working, cmd expects a table, not a function, right ?

justinmk: `cmd` can be a function. Example https://github.com/neovim/nvim-lspconfig/blob/cbd1e91296189a765094810de1b71a04e2597579/lsp/leanls.lua#L16-L19

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
+---@brief
+---
+--- GitLab Duo Language Server Configuration for Neovim
+---
+--- https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp
+---
+--- The GitLab LSP enables any editor or IDE to integrate with GitLab Duo
+--- for AI-powered code suggestions via the Language Server Protocol.
+---
+--- **Prerequisites:**
```
justinmk: avoid `*` since the doc gen might not be smart enough to remove it, and `:help` will flag it as an invalid tag
```suggestion
--- Prerequisites:
```

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
...
+---         { buffer = bufnr, desc = 'GitLab Duo: Next suggestion' })
+---     end
+---   end
+--- })
+--- ```
+
+-- Configuration
+local config = {
+  gitlab_url = vim.env.GITLAB_URL or 'https://gitlab.com',
+  client_id = '5f1f9933c9bff0a3e908007703f260bf1ff87bcdb91d38279a9f0d0ddecceadf',
```
justinmk: maybe a code comment that gives a reference about where this id came from. e.g. is it a personally generated thing or part of official gitlab docs.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
...
+    return true
+  end
+  return false
+end
+
+local function load_token()
+  local file = io.open(config.token_file, 'r')
+  if file then
+    local content = file:read('*all')
+    file:close()
```
justinmk: can save code here by using `vim.fn.readblob()`

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
...
+  if response.status ~= 200 then
+    vim.notify('Device authorization failed: ' .. response.status, vim.log.levels.ERROR)
+    return nil
+  end
+
+  local ok, data = pcall(vim.json.decode, response.body)
+  if not ok then
+    vim.notify('Failed to parse authorization response', vim.log.levels.ERROR)
+    return nil
+  end
```
justinmk: might be better just to drop the `pcall` so that an actual error is thrown with a stacktrace.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
...
+  local auth_data = device_authorization()
+  if not auth_data then
+    return
+  end
+
+  local message = string.format(
+    'GitLab Duo Authorization\n\n'
+      .. '1. Open: %s\n'
+      .. '2. Enter code: %s\n\n'
+      .. 'Code copied to clipboard. Waiting for authorization...',
```
justinmk: nit: can just a single `[[...]]` string with `vim.text.indent(0,...)` to dedent the text.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,473 @@
...
+  local choice = vim.fn.confirm(message, '&Open Browser\n&Continue')
+
+  if choice == 1 then
+    -- Try to open browser
+    local open_cmd = vim.fn.has('mac') == 1 and 'open'
+      or vim.fn.has('unix') == 1 and 'xdg-open'
+      or vim.fn.has('win32') == 1 and 'start'
+
+    if open_cmd then
+      vim.fn.system(open_cmd .. ' ' .. auth_data.verification_uri)
```
justinmk: use `vim.ui.open()`

justinmk:
This is a lot of code which is normally discouraged in this repo, but it looks pretty good and points to some common use-cases that we need to start thinking about in the stdlib. And for supporting LSP/AI providers that require sign-in, it's good to have self-contained examples (another is `copilot.lua`) of what is required to complete that full process. 

So I'm fine with this.

lsp/gitlab_duo.lua
tachyons: That is a neat API, thanks

lsp/gitlab_duo.lua
tachyons: I removed this and redirected to the browser by default.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,440 @@
...
+---
+--- The GitLab LSP enables any editor or IDE to integrate with GitLab Duo
+--- for AI-powered code suggestions via the Language Server Protocol.
+---
+--- Prerequisites:
+--- - Node.js and npm installed
+--- - GitLab account with Duo Pro license
+--- - Internet connection for OAuth device flow
+---
+--- **Setup:**
```
justinmk: and other cases...
```suggestion
--- Setup:
```

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,440 @@
...
+      version = tostring(vim.version()),
+      vendor = 'Neovim',
+    },
+    extension = {
+      name = 'Neovim LSP Client',
+      version = tostring(vim.version()),
+    },
+  },
+  settings = {
+    token = get_valid_token() or '',
```
justinmk: this only happens once, at module-load time. if the token expires what then? i think `settings` can be modified in an `on_init` handler.

tachyons: @justinmk We already do that on `on_int` by checking the `/gitlab/token/check` call. We use `client.notify` to update this value in LSP

justinmk: Then just set this to `''` ? It is misleading to just sprinkle a `get_valid_token()` call here, which again will have side effects like network calls at module-load time. It's better to do things in one place.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,440 @@
...
+--- - Node.js and npm installed
+--- - GitLab account with Duo Pro license
+--- - Internet connection for OAuth device flow
+---
+--- Setup:
+--- 1. Run :LspGitLabDuoSignIn to start OAuth authentication
+--- 2. Follow the browser prompts to authorize
+--- 3. Enable inline completion in LspAttach event (see example below)
+---
+--- **Inline Completion Example:**
```
justinmk: I guess it wasn't clear but I was hoping that all cases would be looked at. Not just the ones I mentioned.

tachyons: @justinmk No, I was planning to do it with the other things including making gitlab_url and application_id configurable.

tachyons: @justinmk I couldn't get configuration support working properly with config block, So I dropped  support for self managed instances from this iteration. Could you review rest of the changes ?

lsp/gitlab_duo.lua
tachyons: @justinmk Thanks, I dropped method call from settings

lsp/gitlab_duo.lua
tachyons: @justinmk Thanks, cmd does not requires gitlab_uri, it is required in the `workspace/didChangeConfiguration` call.  I couldn't find an easy way to pass this value around, we also need to make client_id configurable for self managed GitLab instances as the client_id will also be different. So I thought doing it in a separate iteration.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,441 @@
...
+---     end
+---   end
+--- })
+--- ```
+
+-- Configuration
+local config = {
+  gitlab_url = 'https://gitlab.com',
+  -- This is a oauth application created from tachyons-gitlab account with `api` scope
+  client_id = '00bb391f527d2e77b3467b0b6b900151cc6a28dcfb18fa1249871e43bc3e5832',
```
justinmk, line 56: That seems kind of strange. Do users need to generate their own client_id? Is this config dependent on your account?

lsp/gitlab_duo.lua
justinmk: It's still there 👀

tachyons: My bad, it was re-added on git conflict, it is removed again

lsp/gitlab_duo.lua
tachyons: @justinmk There is no need to create a client_id per user. Alternatively one of the maintainer of this repo can create a [GitLab application ](https://docs.gitlab.com/integration/oauth_provider/)and add the client_id here.

justinmk: Makes no difference to me, just trying to understand, and ensure that we document stuff and can answer questions in the future.

lsp/gitlab_duo.lua
```diff
@@ -0,0 +1,440 @@
...
+--- Inline Completion Example:
+--- ```lua
+--- vim.api.nvim_create_autocmd('LspAttach', {
+---   callback = function(args)
+---     local bufnr = args.buf
+---     local client = assert(vim.lsp.get_client_by_id(args.data.client_id))
+---
+---     if vim.lsp.inline_completion and
+---        client:supports_method(vim.lsp.protocol.Methods.textDocument_inlineCompletion, bufnr) then
+---       vim.lsp.inline_completion.enable(true, { bufnr = bufnr })
```
justinmk, line 29: I should have asked this for Copilot too, but ... why don't we just do this when this config is enabled? Why should the user need to manually add this.

tachyons: @justinmk I used copilot.lua as the reference implication,  that is why I followed the same pattern

justinmk: I think we should change both imo

lsp/gitlab_duo.lua
justinmk: > I couldn't find an easy way to pass this value around

You can set arbitrary properties on the `client` or `config`.

PR #4164, opened 2025-10-28
fix(jdtls): guard root_dir lookup

### LINKED ISSUE #4162: jdtls `config` parameter doesn't exist
### Description

The config parameter is not valid for the table of type `vim.lsp.Config` [here](https://github.com/neovim/nvim-lspconfig/blob/513fc3e2d1a041f2cf0a6379de2670ec586f4469/lsp/jdtls.lua#L77), and it errors with the appropriate message
```
WARN ...user/.local/share/nvim/lazy/nvim-lspconfig/lsp/jdtls.lua:81: attempt to index local 'config' (a nil value)
```

> [!NOTE]
> This issue likely affects the following files as well, but I am much too lazy to test it
> - apex_ls.lua
> - leanls.lua
> - glint.lua
> - biome.lua

Neovim version: v0.11.2
OS: Gentoo

cnaples79:
## Summary
- ensure the jdtls cmd helper falls back to the default root_dir logic when user config is nil
- reuse default_config root detection when user-provided config is absent
- add a regression test covering the nil-config scenario

## Testing
- regression added to test the nil-config scenario

Fixes #4162

### DIFF
--- a/lsp/jdtls.lua
+++ b/lsp/jdtls.lua
@@ -32,12 +32,30 @@
 ---       vim.lsp.config('jdtls', { cmd = { 'jdtls' } })
 ---     ```
 
+local default_jdtls_config = require('lspconfig.configs.jdtls').default_config
+
 local function get_jdtls_cache_dir()
   return vim.fn.stdpath('cache') .. '/jdtls'
 end
 
 local function get_jdtls_workspace_dir()
   return get_jdtls_cache_dir() .. '/workspace'
+end
+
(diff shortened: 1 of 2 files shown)


justinmk:
thanks for attempting to fix, but please review code before posting AI generated stuff. and also mention when AI was used.

besides the fact that `lua/lspconfig/*` is deprecated (clearly mentioned in the README), #4162 is a wontfix for reason mentioned there.

Issue #3849, opened 2025-05-13
`csharp_ls` not correctly using root_dir using 0.11 setup


jjiangweilan:
### Description

I have switched to the new 0.11 setup, both clangd and lua_ls work find. After inspecting lsp log, it seems csharp_ls is not looking for .sln and .csproj in root dir but in the current working directory. If I manually call the old `lspconfig.csharp_ls.setup{}`, It correctly use the root dir.

lithammer:
From what I can tell they have more or less identical root dir logic. So that's surprising.

https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lsp/csharp_ls.lua#L15-L18

https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lua/lspconfig/configs/csharp_ls.lua#L6-L8

jjiangweilan:
> From what I can tell they have more or less identical root dir logic. So that's surprising.
> 
> [nvim-lspconfig/lsp/csharp_ls.lua](https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lsp/csharp_ls.lua#L15-L18)
> 
> Lines 15 to 18 in [61e5109](/neovim/nvim-lspconfig/commit/61e5109c8cf24807e4ae29813a3a82b31821dd45)
> 
>  root_dir = function(bufnr, on_dir) 
>    local fname = vim.api.nvim_buf_get_name(bufnr) 
>    on_dir(util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname)) 
>  end, 
> [nvim-lspconfig/lua/lspconfig/configs/csharp_ls.lua](https://github.com/neovim/nvim-lspconfig/blob/61e5109c8cf24807e4ae29813a3a82b31821dd45/lua/lspconfig/configs/csharp_ls.lua#L6-L8)
> 
> Lines 6 to 8 in [61e5109](/neovim/nvim-lspconfig/commit/61e5109c8cf24807e4ae29813a3a82b31821dd45)
> 
>  root_dir = function(fname) 
>    return util.root_pattern '*.sln'(fname) or util.root_pattern '*.csproj'(fname) 
>  end,

[...]

PR #4166, opened 2025-10-29
feat(csharp_ls): runs csharp-ls where sln, slnx or csproj file exists

pogyomo:
Currently csharp-language-server starts to find solution file or project file from current directly, and just ignore rootPath and rootUri as metioned in razzmatazz/csharp-language-server#272.

Also, legacy style of configuration[ set cmd_cwd to root_dir](https://github.com/neovim/nvim-lspconfig/blob/2b52bc2190c8efde2e4de02d829a138666774c7c/lua/lspconfig/manager.lua#L118-L120), but not currently.
So, as mentioned in #3849, the server cannot find solution file or project file as the server is launched in cwd not in the directory root_dir find out.

This problem will be solved in future version of language server, but I think it would be better to set cmd_cwd the place same as root_dir, as currently it's one solution for this problem, and in future, it will help users who using old version of server.

### DIFF
--- a/lsp/csharp_ls.lua
+++ b/lsp/csharp_ls.lua
@@ -13,6 +13,14 @@ local util = require 'lspconfig.util'
 ---@type vim.lsp.Config
 return {
   cmd = { 'csharp-ls' },
+  cmd_cwd = vim.fs.root(0, {
+    function(name, _)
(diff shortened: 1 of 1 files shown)


justinmk:
This only works if you are working on one project/solution. Look at other configs in this repo to see a better approach.

PR #4171, opened 2025-10-30
fix(pyright): prioritize "pyrightconfig.json" root marker

trevorhauter:
**Problem**: When working in a python monorepo with a base `pyproject.toml`, and `pyrightconfig.json` files that extend this base configuration in subdirectories, `pyright` will not recognize these `pyrightconfig.json` and identify these subdirectories (project1 | project2) as root directories because `pyproject.toml` takes precedence. 

```sh
# Example structure
monorepo
├── project1
│   ├── pyrightconfig.json
│   └── src
│       └── main.py
├── project2
│   ├── pyrightconfig.json
│   └── src
│       └── main.py
└── pyproject.toml
```

From the pyright docs [1]

> A “pyrightconfig.json” file always takes precedent over “pyproject.toml” if both are present

**Solution**: Update the order of the `root_markers` to prefer `pyrightconfig.json` so priority is in line with pyrights documentation.

**Testing**: When creating a simple monorepo structure with the following command
```sh
mkdir monorepo && touch monorepo/pyproject.toml && mkdir monorepo/project1 && touch monorepo/project1/pyrightconfig.json && mkdir monorepo/project1/src && touch monorepo/project1/src/main.py && cp -r monorepo/project1 monorepo/project2
```
You should be able to open one of the subdirectories `nvim monorepo/project1/src/main.py` and check your LSP root dir with `:LspInfo`, and it should be attached to the subdirectory.

1. https://github.com/microsoft/pyright/blob/main/docs/configuration.md#pyright-configuration

### DIFF
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -25,12 +25,12 @@ return {
   cmd = { 'basedpyright-langserver', '--stdio' },
   filetypes = { 'python' },
   root_markers = {
+    'pyrightconfig.json',
     'pyproject.toml',
     'setup.py',
     'setup.cfg',
     'requirements.txt',
     'Pipfile',
-    'pyrightconfig.json',
     '.git',
   },
   settings = {


justinmk:
Thanks! Would you mind doing this for https://github.com/neovim/nvim-lspconfig/blob/master/lsp/basedpyright.lua too, in this pr

justinmk:
look for the one in `lsp/`, ignore the one in `lua/lspconfig/` :)

trevorhauter:
@justinmk Fixed! thanks for the help

PR #4176, opened 2025-11-05
feat: vsrocq, rename "vscoqtop" config

simon-dima:
Starting with version 2.3.0, VsCoq (called `vscoqtop` here) has been renamed to [VsRocq](https://github.com/rocq-prover/vsrocq).


justinmk:
Ok, please also update https://github.com/neovim/nvim-lspconfig/blob/master/lsp/vscoqtop.lua so that it throws an error or something. Otherwise we are maintaining duplicate configs, and we get confused PRs updating one but not the other, etc.

    error('This config was renamed to ...')

PR #4177, opened 2025-11-05
fix(leanls) : fix compatibility issues

yd278:
**The old file has compatibility issues with the newest version of the `lean.nvim` plugin.**

The `lean.nvim` plugin now handles language server attachment internally and no longer requires `nvim-lspconfig`.  (https://github.com/Julian/lean.nvim/commit/4f62922d43d87e682727a4d584c6e1fc39cace62) 
However, for users who still need `nvim-lspconfig` installed (e.g., for other languages), the existing file would shadow the newer one, leading to conflicts. This update directly copies the relevant file from `lean.nvim` to resolve this.


lsp/leanls.lua
```diff
@@ -42,4 +190,31 @@ return {
...
   end,
+
+  capabilities = {
+    lean = {
+      silentDiagnosticSupport = true,
+    },
+  },
+  handlers = {
+    ['$/lean/fileProgress'] = file_progress_handler,
+    [vim.lsp.protocol.Methods.textDocument_publishDiagnostics] = on_publish_diagnostics,
```
lithammer: Why a custom diagnostics handler?

lsp/leanls.lua
```diff
@@ -1,23 +1,171 @@
...
---- [here](https://leanprover-community.github.io/get_started.html#regular-install).
----
---- The Lean language server is included in any Lean installation and
---- does not require any additional packages.
+local Buffer = require 'std.nvim.buffer'
+local std = require 'std.lsp'
+
+local diagnostic = require 'lean.diagnostic'
+local log = require 'lean.log'
+local lsp = require 'lean.lsp'
```
lithammer: This doesn't work. All these modules are from lean.nvim.

lsp/leanls.lua
```diff
@@ -42,4 +190,31 @@ return {
...
+    lean = {
+      silentDiagnosticSupport = true,
+    },
+  },
+  handlers = {
+    ['$/lean/fileProgress'] = file_progress_handler,
+    [vim.lsp.protocol.Methods.textDocument_publishDiagnostics] = on_publish_diagnostics,
+  },
+  init_options = {
+    editDelay = 10, -- see #289
```
lithammer: This comment is referencing an issue on the lean.nvim repo, so it must be absolute.

lithammer (changes requested):
I don't think this is a good idea since it's depending on external libraries (i.e. lean.nvim). To fix the shadowing, lean.nvim should probably move from `lsp/leanls.lua` to `after/lsp/leanls.lua` 🤔

yd278:
That makes sense, Thank you!

 Or maybe it's safe to just remove this lua script to prevent shadowing? I didn't do the research but I suspect nobody will write lean 4 script with only lspconfig but without lean.nvim as it's recommended by the official website.

justinmk:
IIUC from https://github.com/neovim/nvim-lspconfig/pull/4084 , it sounds like @StamesJames doesn't use leanls so we can probably drop this then. 

Although, to prevent going in circles, perhaps at least leave a comment in this file that says "go install X instead".

StamesJames:
@justinmk sorry, I have not followed the conversation lately. Yes, I do not use leanls directly in my config just the lean.nvim plugin, if that is what you mean.

PR #4179, opened 2025-11-06
fix(buf_ls): reuse client across workspaces

emcfarlane:
This updates the buf LSP config to always reuse the client. The `buf lsp serve` is meant to be used with multiple workspaces. This fixes goto definition with cached modules.

### DIFF
--- a/lsp/buf_ls.lua
+++ b/lsp/buf_ls.lua
@@ -11,6 +11,7 @@ return {
   filetypes = { 'proto' },
   root_markers = { 'buf.yaml', '.git' },
   reuse_client = function()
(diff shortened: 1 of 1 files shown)


lsp/buf_ls.lua
```diff
@@ -10,4 +10,7 @@ return {
   cmd = { 'buf', 'lsp', 'serve', '--timeout=0', '--log-format=text' },
   filetypes = { 'proto' },
   root_markers = { 'buf.yaml', '.git' },
+  reuse_client = function()
+    return true
```
justinmk, line 15:
```suggestion
    -- `buf lsp serve` is meant to be used with multiple workspaces.
    return true
```

Issue #4175, opened 2025-11-04
angularls spawing error without being in angular project


DocAmaroo:
### Description

Hi everyone,

I have recently updated the angularls package, and since then, the error below is always spawning when on a non-angular project. In my case, it spawns only once when opening a typescript file and can be triggered using `LspRestart`.

```
Client angularls quit with exit code 1 and signal 0. Check log for errors ...
```

Here's `LspLog` lines that could help

```log
[WARN][2025-11-04 14:20:01] ...m/lsp/client.lua:875	"The language server eslint triggers a registerCapability handler for workspace/didChangeWorkspaceFolders despite dynamicRegistration set to false. Report upstream, this warning is harmless"
[ERROR][2025-11-04 14:20:02] ...p/_transport.lua:36	"rpc"	"ngserver"	"stderr"	"home/.../.local/share/nvim/mason/packages/angular-language-server/node_modules/@angular/language-server/index.js:220289\n      throw new Error(`Failed to resolve '${packageName}' with minimum version '${minVersion}' from ` + JSON.stringify(probeLocations, null, 2));\n      ^\n\nError: Failed to resolve '@angular/language-service' with minimum version '15.0' from [\n  \"/home/.../node_modules/@angular/language-server/node_modules\"\n]\n    at resolveWithMinVersion (/home/.../.local/share/nvim/mason/packages/angular-language-server/node_modules/@angular/language-server/index.js:220289:13)\n    at resolveNgLangSvc (/home/.../.local/share/nvim/mason/packages/angular-language-server/node_modules/@angular/language-server/index.js:220340:14)\n    at main (/home/.../.local/share/nvim/mason/packages/angular-language-server/node_modules/@angular/language-server/index.js:220366:54)\n    at Object.<anonymous> (/home/.../.local/share/nvim/mason/packages/angular-language-server/node_modules/@angular/language-server/index.js:220398:1)\n    at Module._compile (node:internal/modules/cjs/loader:1706:14)\n    at Object..js (node:internal/modules/cjs/loader:1839:10)\n    at Module.load (node:internal/modules/cjs/loader:1441:32)\n    at Function._load (node:internal/modules/cjs/loader:1263:12)\n    at TracingChannel.traceSync (node:diagnostics_channel:322:14)\n    at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)\n\nNode.js v22.20.0\n"
```

[...]

PR #4182, opened 2025-11-12
fix(angularls): spawning error on non-angular project #4175

DocAmaroo:
Problem:
angularls spawing errors on non angular-project

Solution:
Updating path resolution to internal servers (e.g., Mason)

### DIFF
--- a/lsp/angularls.lua
+++ b/lsp/angularls.lua
@@ -36,7 +36,7 @@ local function collect_node_modules(root_dir)
   local internal_servers = fn.globpath(fn.stdpath('data'), '**/node_modules/.bin/ngserver', true, true)
   for _, exe in ipairs(internal_servers) do
     local realpath = uv.fs_realpath(exe) or exe
-    local candidate = fs.normalize(fs.joinpath(fs.dirname(realpath), '../../node_modules'))
+    local candidate = fs.normalize(fs.joinpath(fs.dirname(realpath), '../../../'))
     if uv.fs_stat(candidate) then
       table.insert(results, candidate)
     end


DocAmaroo:
It looks like the errors was spawned because it fails to find the @angular/language-server installed by Mason. I guess its the same with others.

DocAmaroo:
Sorry for the forced-pushes, I used the wrong git config.

As mentioned in the issue #4175, I do have a quick question about the use of `config.root` here:

https://github.com/neovim/nvim-lspconfig/blob/c8503e63c6afab3ed34b49865a4a4edbb1ebf4a8/lsp/angularls.lua#L69

Shouldn't that be `config.root_dir` ?

radsaq:
> Sorry for the forced-pushes, I used the wrong git config.
> 
> As mentioned in the issue #4175, I do have a quick question about the use of `config.root` here:
> 
> https://github.com/neovim/nvim-lspconfig/blob/c8503e63c6afab3ed34b49865a4a4edbb1ebf4a8/lsp/angularls.lua#L69
> 
> Shouldn't that be `config.root_dir` ?

I'm not an expert, but it seems like it. In a monorepo the current code fails for me because it defaults to cwd. Changing it to root_dir makes it properly use the lsp root instead.

erikson84:
@DocAmaroo, you are right, the field should be `root_dir`, not `root`. Relevant [documentation](https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig) excerpt:

> {root_dir}? (string) Directory where the LSP server will base its workspaceFolders, rootUri, and rootPath on initialization.

About the correct path for the node_modules dir: I needed it to work on my Windows machine (were I first developed the config), but now I tested on a Linux workstation and also got the error you mentioned. I will check your solution on Windows and report back.

DocAmaroo:
> @DocAmaroo, you are right, the field should be `root_dir`, not `root`. Relevant [documentation](https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig) excerpt:
> 
> > {root_dir}? (string) Directory where the LSP server will base its workspaceFolders, rootUri, and rootPath on initialization.
> 
> About the correct path for the node_modules dir: I needed it to work on my Windows machine (were I first developed the config), but now I tested on a Linux workstation and also got the error you mentioned. I will check your solution on Windows and report back.

Should I make that change here, or in a separate PR, since it doesn’t affect the issue this one was opened for?

erikson84:
Just checked the code on Windows, and it breaks the configuration there. It happens because, on Windows, ngserver is wrapped on a `.cmd` file, which is not in the expected directory, i.e., if we go up three levels with `../../..`, we won't end up in a node_modules folder.

It does fix the problem on Linux (which has a symbolic link to the original executable that is correctly recovered by Neovim's helper functions).

I'm thinking on reviewing this and the problems mentioned on #4083. I believe we can branch the `collect_node_modules` to use the glob with the original path resolution on Windows and keep your solution on the previous block of code, which recovers the `node_module` related to the `ngserver` executable.

DocAmaroo:
Oh okay, thanks for the review then. If I have time, I may try to install it on Windows too and try things.

For the meantime, I will open a new PR for the `config.root_dir` to split this up

lsp/angularls.lua
```diff
@@ -36,7 +36,7 @@ local function collect_node_modules(root_dir)
   local internal_servers = fn.globpath(fn.stdpath('data'), '**/node_modules/.bin/ngserver', true, true)
```
justinmk, line 36: > on Windows, and it breaks the configuration there. It happens because, on Windows, ngserver is wrapped on a `.cmd` file, which is not in the expected directory

Does this solve that issue?


```suggestion
  local internal_servers = fn.globpath(fn.stdpath('data'), '**/node_modules/.bin/ngserver{,.cmd}', true, true)
```

erikson84: The glob itself works both on Windows and Linux to find the executable path. But it's costly! I'm working on a solution that use different strategies for each OS, but using only the `vim.fn.exepath` function to speed up the process. For the sake of completion, I'm testing it both on Windows and Linux. I will open a new PR soon!

erikson84:
I opened a new PR (#4190) which uses a new strategy to find the `ngserver` `node_modules` folder. With it, @DocAmaroo solution works both on Windows and Linux.

PR #4184, opened 2025-11-13
fix(angularls): replace non-existant config.root with config.root_dir

DocAmaroo:
Problem:
`config.root` is not defined in the documentation

Solution
Replace it by `config.root_dir`

### DIFF
--- a/lsp/angularls.lua
+++ b/lsp/angularls.lua
@@ -66,7 +66,7 @@ end
 ---@type vim.lsp.Config
 return {
   cmd = function(dispatchers, config)
-    local root_dir = (config and config.root) or fn.getcwd()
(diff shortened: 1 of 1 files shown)


lsp/angularls.lua
```diff
@@ -66,7 +66,7 @@ end
 ---@type vim.lsp.Config
 return {
   cmd = function(dispatchers, config)
-    local root_dir = (config and config.root) or fn.getcwd()
+    local root_dir = (config and config.root_dir) or fn.getcwd()
```
justinmk, line 69: Thanks 👍 btw, this is correctly flagged by LuaLS, we just haven't enabled it in CI yet (for this repo):

Image

PR #4193, opened 2025-11-16
fix(lsp): improve ts_ls root detection for deno projects

yzs249175311:
Refactor project root detection logic to properly exclude ts_ls when deno.json/deno.lock are at the same or higher level than project markers.

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -62,16 +62,14 @@ return {
     -- Give the root markers equal priority by wrapping them in a table
     root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers, { '.git' } }
       or vim.list_extend(root_markers, { '.git' })
-
     -- exclude deno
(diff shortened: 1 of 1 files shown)


lsp/ts_ls.lua
```diff
@@ -62,16 +62,18 @@ return {
...
-
     -- We fallback to the current working directory if no project root is found
-    local project_root = vim.fs.root(bufnr, root_markers) or vim.fn.getcwd()
-
-    on_dir(project_root)
+    if project_root then
+      on_dir(project_root)
+    else
+      on_dir(vim.fn.getcwd())
+    end
```
justinmk:
```suggestion
    on_dir(project_root or vim.fn.getcwd())
```

dimfeld:
See #4196, this totally breaks things in repositories that don't use Deno. Was about to post a fix but someone beat me to it: #4197 :)

justinmk:
sigh

PR #4199, opened 2025-11-17
feat: rumdl 

raviqqe:
This PR adds support for the [`rumdl`](https://github.com/rvben/rumdl) language server for Markdown files.

As its MDX support is [questionable](https://github.com/rvben/rumdl/issues/123), the configuration is applied to only pure Markdown files.

### DIFF
--- a/lsp/rumdl.lua
+++ b/lsp/rumdl.lua
@@ -1,0 +1,12 @@
+---@brief
+---
+--- https://github.com/rvben/rumdl
+---
+--- Markdown Linter and Formatter written in Rust.
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'rumdl', 'server' },
+  filetypes = { 'markdown' },
+  root_markers = { '.git' },
+}


barbedo:
I'm getting duplicated diagnostics when using this configuration. It seems that `rumdl` supports both `publishDiagnostics` and `diagnostics`.

I don't know if there is a way to tell `rumdl` to avoid pushing diagnostics, but I silenced it by stubbing the `publishDiagnostics` handler:

```lua
handlers = {
  ["textDocument/publishDiagnostics"] = function() end,
},
```

raviqqe:
https://github.com/neovim/nvim-lspconfig/pull/4199#issuecomment-3616193609

Does anyone know if this is the editor's responsibility to make the diagnostic items unique, or servers should not mix them up? I couldn't find any mention in the LSP documentation...

- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic
- [Chat GPT's claim: Do not mix them up](https://chatgpt.com/share/693300a5-95b0-800c-bb32-0ef839fe6f7c)

barbedo:
I also can't find anything explicit in the specifications, which implies that both pushing and pulling should be supported independently by the client.

From the discussion in https://github.com/microsoft/language-server-protocol/issues/1743, it seems that the [original intent](https://github.com/microsoft/language-server-protocol/issues/1743#issuecomment-1582417329) was to specify that the server pushing diagnostics is not needed if the client communicates that it supports pulling them, but that someone [forgot to clarify this in the specs](https://github.com/microsoft/language-server-protocol/issues/1743#issuecomment-2525582480).

justinmk:
Can improve it, meanwhile it sounds like people are using this so we can merge it.

PR #4211, opened 2025-11-22
feat(lua_ls): enable codelens and inlay hints

mikejmcguirk:
Problem: Current lua_ls config does not enable codelens and inlay hints

Solution: Add enable for both of them


justinmk:
Confirmed that this works with `lua-language-server 3.15.0` . Thanks!

PR #4214, opened 2025-11-24
feat(gh_actions_ls): enable advanced functionality

rmuir:
Problem

gh_actions_ls requires repository metadata and access token for full functionality.

Solution

Use the `gh` CLI, if available, to populate the necessary metadata for the language server to provide completion and validation of "uses ... with" parameters, repository/organization variables and secrets.

Notes:
* More info: https://github.com/actions/languageservices/tree/main/languageserver#providing-advanced-functionality
* `gh repo view` is used, as opposed to alternatives such as `gh api repos/{owner}/{repo}` because it contains all the data needed and "just works".
* You don't need to have configured a repo with `gh`, just have `gh` available and authed. For example, it still works if you havent `gh repo set-default`.
* Tested with org repo, personal repo, and GHES with separate server. It is currently not possible to provide complete functionality for the GHES case with this language server... 
 
Questions:
* Is there a way to populate init_options like this without blocking the editor?

cc: @jugarpeupv who raised this on #4046

### DIFF
--- a/lsp/gh_actions_ls.lua
+++ b/lsp/gh_actions_ls.lua
@@ -14,6 +14,32 @@
 --- npm install -g gh-actions-language-server
 --- ```
 
+--- Session token used to perform authenticated API requests
+---@return string? token
+local function token()
+  local result = vim.system({ 'gh', 'auth', 'token', '-h', 'github.com' }):wait()
+  return result.code == 0 and vim.trim(assert(result.stdout)) or nil
+end
+
+--- Repository metadata needed for full features
+---@param root_dir string
+---@return table? metadata about the repository
+local function repos(root_dir)
+  local result = vim.system({ 'gh', 'repo', 'view', '--json', 'owner,name,isInOrganization' }):wait()
+  if result.code == 0 then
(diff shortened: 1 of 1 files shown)


lsp/gh_actions_ls.lua
```diff
@@ -47,6 +73,14 @@ return {
     end,
   },
   init_options = {}, -- needs to be present https://github.com/neovim/nvim-lspconfig/pull/3713#issuecomment-2857394868
+  before_init = function(params, config)
+    if config.root_dir and vim.fn.executable('gh') == 1 then
+      params.initializationOptions = {
+        sessionToken = token(),
```
justinmk, line 79: should probably ad a `:LspGhLogin` or something like that, rather than calling this as soon as the config is resolved.

rmuir: will have to think thru how to plumb this. unfortunately the server requires this information as `initializationOptions` up-front, and doesn't appear to allow it to be set via `workspace/didChangeConfiguration` or similar...

PR #4215, opened 2025-11-25
fix(hls): typo in the provided formatter makes the lsp crash when trying to format

talarys:
As stated in the title there is a small typo in the config that makes the lsp crash the moment it tries to format.
https://github.com/phadej/cabal-fmt


justinmk:
cc @mizlan

PR #4217, opened 2025-11-26
feat(lsp): use smarter detection of typescript project

AsPulse:
**Problem:** TypeScript projects face an issue where the appropriate LSP to launch differs depending on whether it's a Deno project or a Node/Bun project. Furthermore, as Deno started supporting workspaces and coexisting with package.json, problems arise such as LSPs failing to launch or two LSPs launching simultaneously. This has led to the codebase for resolving these issues becoming complex and prone to omissions.

**Solution:** I introduced a function `detect_project` into `lspconfig.typescript` to determine the TypeScript project type (either `node`, `bun`, or `deno`) and its root directory. I then rewrote the logic in various TypeScript-related LSPs (`biome`, `denols`, `eslint`, `ts_ls`, `tsgo`, `vtsls`), which was previously written individually, to use the determination made by this function.

**Testing:** I created and ran a test runner to confirm that `detect_project` works as expected in [AsPulse/lspconfig-detection-test](https://github.com/AsPulse/lspconfig-detection-test/tree/7a99a07). In `fixtures/`, we describe situations that should be tested, and in `tests/`, we describe which runtime and `root_dir` should be detected. I've confirmed that the code in the PR passes these tests. I also confirmed that a neovim environment that was installed nvim-lspconfig, included in the PR, functioned as intended.

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -12,6 +12,7 @@
 --- `biome` supports monorepos by default. It will automatically find the `biome.json` corresponding to the package you are working on, as described in the [documentation](https://biomejs.dev/guides/big-projects/#monorepo). This works without the need of spawning multiple instances of `biome`, saving memory.
 
 local util = require 'lspconfig.util'
+local typescript = require 'lspconfig.typescript'
 
 ---@type vim.lsp.Config
 return {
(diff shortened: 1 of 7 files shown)


justinmk:
Assuming that https://github.com/neovim/nvim-lspconfig/pull/4304 resolved this.

PR #4225, opened 2025-12-03
feat(ts_ls): add LspTypescriptGoToSourceDefinition command

icholy:


lsp/ts_ls.lua
```diff
@@ -128,5 +130,26 @@ return {
...
+      local win = vim.api.nvim_get_current_win()
+      local params = vim.lsp.util.make_position_params(win, client.offset_encoding)
+      client:exec_cmd({
+        command = '_typescript.goToSourceDefinition',
+        title = 'Go to source definition',
+        arguments = { params.textDocument.uri, params.position },
+      }, { bufnr = bufnr }, function(err, result)
+        if err then
+          vim.notify('Go to source definition failed: ' .. err.message, vim.log.levels.ERROR)
+          return
```
justinmk, line 145: We really should get in the habit of using `error()` for errors.

PR #4232, opened 2025-12-11
feat(nextls): add default value for cmd

GaetanLepage:
[No matter the installation method](https://www.elixir-tools.dev/docs/next-ls/installation/), it seems like `nextls` should be run with the `nextls --stdio` command.
I suggest to set this as a default value for `cmd`.

This is what [mason-lspconfig.nvim does FWIW](https://github.com/mason-org/mason-lspconfig.nvim/blob/c55bd8a8fb191e24176c206a7af1dd51ce7276a5/lua/mason-lspconfig/lsp/nextls.lua).

### DIFF
--- a/lsp/nextls.lua
+++ b/lsp/nextls.lua
@@ -2,10 +2,11 @@
 ---
 --- https://github.com/elixir-tools/next-ls
 ---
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Thanks for the quick merges @justinmk :)

PR #4233, opened 2025-12-12
feat(flow): use flow executable for cmd if available in PATH

GaetanLepage:
**Note: I'm not sure if this is overkilled and/or overreaching. Feel free to close the PR if irrelevant.**

This PR dynamically sets `cmd` for [flow](https://flow.org/) depending on `flow`'s presence on the `$PATH`.

### DIFF
--- a/lsp/flow.lua
+++ b/lsp/flow.lua
@@ -14,7 +14,16 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'npx', '--no-install', 'flow', 'lsp' },
+  cmd = function(dispatchers)
+    local cmd = nil
+    if vim.fn.executable('flow') then
+      cmd = { 'flow', 'lsp' }
+    else
+      cmd = { 'npx', '--no-install', 'flow', 'lsp' }
+    end
+
+    return vim.lsp.rpc.start(cmd, dispatchers)
+  end,
   filetypes = { 'javascript', 'javascriptreact', 'javascript.jsx' },
   root_markers = { '.flowconfig' },
 }


justinmk:
If this kind of "search all cmd possiblities" pattern becomes common it points to a pattern that we should try to lift into `vim.lsp` or somehow make less painful. Meanwhile, I think it's fine to start visualizing the pain, in this form, to see how common it is :)

MattSturgeon (approved):
LGTM, I don't see why this wouldn't work. Is it just a draft until you (or someone) is able to test?

mrcjkb:
> If this kind of "search all cmd possiblities" pattern becomes common it points to a pattern that we should try to lift into `vim.lsp` or somehow make less painful. Meanwhile, I think it's fine to start visualizing the pain, in this form, to see how common it is :)

I've used this pattern a few times.

- On some systems, there's a `haskell-language-server-wrapper` executable, which selects the correct compiler version. On others it's aliased to `haskell-language-server` or there's no wrapper.

- `jdt-language-server` can be packaged as `jdtls` on some systems (I don't remember which ones, but I have that check in my config because it broke at some point).

GaetanLepage:
If anyone test this patch, please share your observations. I never used `flow`, so I'm not too sure.

Otherwise, this should be ready to be merged :)

GaetanLepage:
> If anyone test this patch, please share your observations. I never used `flow`, so I'm not too sure.
> 
> Otherwise, this should be ready to be merged :)

EDIT: I was able to test the patch and it works.

@justinmk feel free to merge if you don't see anything else to address :)

teto:
merging before the nix mafia (for good) mobs justin :p

PR #4234, opened 2025-12-12
feat: remove `:Lsp*` commands for nvim 0.12

superatomic:
To be merged when https://github.com/neovim/neovim/pull/36925 is merged.

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -3,6 +3,10 @@ if vim.g.lspconfig ~= nil then
 end
 vim.g.lspconfig = 1
 
+if vim.fn.exists(':lsp') == 2 then
+  return
+end
+
 if vim.fn.has('nvim-0.11') == 0 then
   vim.deprecate('nvim-lspconfig support for Nvim 0.10 or older', 'Nvim 0.11+', 'v3.0.0', 'nvim-lspconfig', false)
 end


plugin/lspconfig.lua
```diff
@@ -3,6 +3,10 @@ if vim.g.lspconfig ~= nil then
 end
 vim.g.lspconfig = 1
 
+if vim.fn.has('nvim-0.12') == 1 then
```
justinmk: could check `2 == vim.fn.exists(':lsp')` and then can merge this without waiting :)

superatomic: Done! Out of curiosity, why does `exists()` return 2? `:h exists()` says it only returns 0 or 1.

justinmk: see `:cmdname` note in `:h exists()`. probably should clarify the docs

superatomic: I see. Just scrolled down to the return value and didn't read the whole thing.

seblyng:
What about `:LspLog`? I know `:LspInfo` is just a wrapper around `:che vim.lsp` so that is not so much more to type

superatomic:
There's a possible argument for `:lsp` by itself being a synonym for `:checkhealth vim.lsp`. I decided against it in the PR because it felt wrong to me. `:che vim.lsp` is pretty short, especially with completion.

You can also easily put `vim.api.nvim_create_user_command('LspLog', 'che vim.lsp', {})` in your own config. I don't think it really needs to be upstreamed. `:LspLog` only exists because it came before `:che vim.lsp`.

justinmk:
LspLog != LspInfo

superatomic:
Oh my bad. I misunderstood.

bhuynhdev:
I believe logs are stored in `~/.local/state/nvim/lsp.log`, so `LspLog` - though inconvenient that it was removed -  can be replaced with `:e ~/local/state/nvim/lsp.log`.

Native `:lsp log` command would be very convenient and useful

superatomic:
~~I believe the plan is to introduce a `:log` command that can be implemented by vim modules and plugins, like how `:checkhealth` works. So the command would be `:log vim.lsp`.~~

(Edit: My memory is inaccurate. The plan is much simpler; just have an `:editlog` command that opens files in `stdpath('log')`)

In the meantime you can implement a simple custom user command that reimplements `:LspLog`, as the implementation is rather simple:

```lua
vim.api.nvim_create_user_command('LspLog', function() 
  vim.cmd.tabnew { vim.lsp.log.get_filename() }
end, { 
  desc = 'Opens the Nvim LSP client log.', 
})
```

superatomic:
`:log lsp` is now available in master.

PR #4235, opened 2025-12-12
Update installation instructions for astro-ls

AndrijaSkontra:
Added note about TypeScript installation requirement.

### DIFF
--- a/lsp/astro.lua
+++ b/lsp/astro.lua
@@ -5,6 +5,7 @@
 --- `astro-ls` can be installed via `npm`:
 --- ```sh
 --- npm install -g @astrojs/language-server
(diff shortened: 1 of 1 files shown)


lsp/astro.lua
```diff
@@ -5,6 +5,7 @@
 --- `astro-ls` can be installed via `npm`:
 --- ```sh
 --- npm install -g @astrojs/language-server
+--- sometimes installing typescript is also required, so just run: npm install --save-dev typescript
```
justinmk, line 8: this is in the middle of a codeblock...

justinmk:
"sometimes" just means we don't actually know if this is true.

PR #4242, opened 2025-12-16
feat(oxlint/oxfmt): adopt and fully integrate the recommended language servers

typedrifter:
Hello,

The goal of this PR is to replace `oxc_language_server` with the two “new” language servers provided by `oxlint` and `oxfmt`, as recommended in:
https://github.com/oxc-project/oxc/issues/15119

The motivation is to align with the upstream architecture and rely on the officially recommended entry points instead of using `oxc_language_server` directly.

> Change in architecture:
>     oxc_language_server should accept different tools
>     oxlint --lsp and oxfmt --lsp uses oxc_language_server with only the required tool
>     IDE is calling oxlint --lsp and oxfmt --lsp
>     oxc_language_server should be generic to support other tools too (vitest, ...)

In addition to the language server changes, this PR:
- Enables formatting through `vim.lsp.buf.format` in oxfmt
- Adds an `:LspOxlintFixAll` command to apply automatic fixes, following the official documentation:
  https://oxc.rs/docs/guide/usage/linter/automatic-fixes.html
-- Adds default configuration (commented) for Oxlint LS

I am currently using this configuration in my own projects, and it has been working very well so far.

This PR updates the oxlint configuration and adds oxfmt support in a single change. I grouped them together because the split of `oxc_language_server` made it feel more logical to handle both at once. However, I am happy to split this into two separate PRs if that is preferred.

### DIFF
--- a/lsp/oxlint.lua
+++ b/lsp/oxlint.lua
@@ -3,7 +3,8 @@
 --- https://github.com/oxc-project/oxc
 --- https://oxc.rs/docs/guide/usage/linter.html
 ---
---- `oxlint` is a linter for JavaScript / Typescript supporting over 500 rules from ESLint and its popular plugins
+--- `oxlint` is a linter for JavaScript / TypeScript supporting over 500 rules from ESLint and its popular plugins.
+--- It also supports linting framework files (Vue, Svelte, Astro) by analyzing their <script> blocks.
 --- It can be installed via `npm`:
 ---
 --- ```sh
@@ -22,6 +23,9 @@ return {
     'typescript',
     'typescriptreact',
     'typescript.tsx',
+    'vue',
+    'svelte',
+    'astro',
   },
   workspace_required = true,
   on_attach = function(client, bufnr)


lsp/oxfmt.lua
```diff
@@ -0,0 +1,45 @@
...
+
+    -- Oxfmt resolves configuration by walking upward and using the nearest config file
+    -- to the file being processed. We therefore compute the root directory by locating
+    -- the closest `.oxfmtrc.json` (or `package.json` fallback) above the buffer.
+    local root_markers = util.insert_package_json({ '.oxfmtrc.json' }, 'oxfmt', fname)[1]
+    on_dir(vim.fs.dirname(vim.fs.find(root_markers, { path = fname, upward = true })[1]))
+  end,
+  init_options = {
+    {
+      workspaceUri = 'file://' .. vim.fs.dirname(vim.fs.find('.git', { path = fname, upward = true })[1]),
```
lithammer: `fname` is undefined here. And what does this do, and why is it different from `root_dir`?

lsp/oxfmt.lua
```diff
@@ -0,0 +1,45 @@
+--- @brief
+---
+--- https://github.com/oxc-project/oxc
+--- https://oxc.rs/docs/guide/usage/formatter.html
+---
+--- `oxfmt` is a is a Prettier-compatible code formatter for JavaScript / Typescript.
```
TimShilov:
```suggestion
--- `oxfmt` is a Prettier-compatible code formatter for JavaScript / Typescript.
```

lsp/oxlint.lua
```diff
@@ -1,8 +1,9 @@
 --- @brief
 ---
 --- https://github.com/oxc-project/oxc
+--- https://oxc.rs/docs/guide/usage/linter.html
 ---
---- `oxc` is a linter / formatter for JavaScript / Typescript supporting over 500 rules from ESLint and its popular plugins
+--- `oxclint` is a linter for JavaScript / Typescript supporting over 500 rules from ESLint and its popular plugins
```
TimShilov:
```suggestion
--- `oxlint` is a linter for JavaScript / Typescript supporting over 500 rules from ESLint and its popular plugins
```

lsp/oxfmt.lua
typedrifter: Thanks for catching this. `fname` being undefined is an error on my side, and I’m surprised it slipped past linting.

I had to explicitly set `workspaceUri` because the language server requires it to function correctly. I’m not very familiar with LSP internals and had to dig a bit to get this working, but there’s no reason for it to differ from root_dir, that’s a mistake, and I’ll fix it.

lsp/oxlint.lua
```diff
@@ -23,9 +24,40 @@ return {
...
     local fname = vim.api.nvim_buf_get_name(bufnr)
-    local root_markers = util.insert_package_json({ '.oxlintrc.json' }, 'oxlint', fname)
+
+    -- Oxlint resolves configuration by walking upward and using the nearest config file
+    -- to the file being processed. We therefore compute the root directory by locating
+    -- the closest `.oxlintrc.json` (or `package.json` fallback) above the buffer.
+    local root_markers = util.insert_package_json({ '.oxlintrc.json' }, 'oxlint', fname)[1]
     on_dir(vim.fs.dirname(vim.fs.find(root_markers, { path = fname, upward = true })[1]))
   end,
+  before_init = function(_, config)
```
typedrifter: I’m looking for advice here. It seems to me that this could make user customization harder, but I don’t see a good alternative: without correctly setting `workspaceUri`, the options won’t work.

Would it make more sense to comment this out and document it instead, so users who want to customize the options can provide a `before_init` themselves?

typedrifter: Mhh and it doesn't seem to work, I'm not able to pass option anywhere else but `init_options`...

typedrifter: Ok now it is not needed.

typedrifter:
Thanks to recent updates, specifying a workspaceUri in the init_options is not mandatory anymore.

Customizing the LS settings can now simply be done.
For example, to enable all fixes applied with `LspOxlintFixAll`'
```lua
vim.lsp.config("oxlint", {
	init_options = {
		settings = {
			["fixKind"] = "all",
		},
	},
})
```

lsp/oxfmt.lua
```diff
@@ -0,0 +1,36 @@
...
+  workspace_required = true,
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+
+    -- Oxfmt resolves configuration by walking upward and using the nearest config file
+    -- to the file being processed. We therefore compute the root directory by locating
+    -- the closest `.oxfmtrc.json` (or `package.json` fallback) above the buffer.
+    local root_markers = util.insert_package_json({ '.oxfmtrc.json' }, 'oxfmt', fname)[1]
+    on_dir(vim.fs.dirname(vim.fs.find(root_markers, { path = fname, upward = true })[1]))
+  end,
```
typedrifter, line 51: `fmt.experimental` is deprecated, it is always on.
https://github.com/oxc-project/oxc/pull/16919

lsp/oxfmt.lua
```diff
@@ -0,0 +1,36 @@
...
+--- ```sh
+--- npm i -g oxfmt
+--- ```
+
+local util = require 'lspconfig.util'
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'oxfmt', '--lsp' },
+  filetypes = {
```
SebastienWae, line 19: oxfmt support more filetypes, maybe they should also be added
https://oxc.rs/docs/guide/usage/formatter.html#supported-languages

typedrifter: You are right, let's add them.

lsp/oxfmt.lua
```diff
@@ -0,0 +1,52 @@
...
+    'html',
+    'vue',
+    'handlebars',
+    'hbs',
+    'css',
+    'scss',
+    'less',
+    'graphql',
+    'markdown',
+    'mdx',
```
raviqqe, line 40: Isn't this `markdown.mdx`? At least, my local Neovim `v0.12.0-dev-1977+gdd6ed20272` shows it in the status bar.

typedrifter: Mhh pretty sure it depends on configuration, with v0.11.5 vanilla I don't get any filetype on mdx files. Not sure about how to manage this case.

lsp/oxlint.lua
```diff
@@ -13,19 +15,48 @@ local util = require 'lspconfig.util'
...
+    })
+  end,
   root_dir = function(bufnr, on_dir)
     local fname = vim.api.nvim_buf_get_name(bufnr)
-    local root_markers = util.insert_package_json({ '.oxlintrc.json' }, 'oxlint', fname)
+
+    -- Oxlint resolves configuration by walking upward and using the nearest config file
+    -- to the file being processed. We therefore compute the root directory by locating
+    -- the closest `.oxlintrc.json` (or `package.json` fallback) above the buffer.
+    local root_markers = util.insert_package_json({ '.oxlintrc.json' }, 'oxlint', fname)[1]
```
guillenotfound, line 48: Maybe add the `.oxlintrc.jsonc` as well?

lsp/oxfmt.lua
```diff
@@ -0,0 +1,52 @@
...
+  },
+  workspace_required = true,
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+
+    -- Oxfmt resolves configuration by walking upward and using the nearest config file
+    -- to the file being processed. We therefore compute the root directory by locating
+    -- the closest `.oxfmtrc.json` (or `package.json` fallback) above the buffer.
+    local root_markers = util.insert_package_json({ '.oxfmtrc.json' }, 'oxfmt', fname)[1]
```
guillenotfound, line 49: Maybe add the `.oxfmtrc.jsonc` as well?

lsp/oxlint.lua
typedrifter: According to the documentation it doesn't look for `.jsonc` files (but accept jsonc _format_).

> Only .json config files are supported, but oxlint configuration files support comments (like jsonc).

[https://oxc.rs/docs/guide/usage/linter/config.html#create-a-config-file](https://oxc.rs/docs/guide/usage/linter/config.html#create-a-config-file)

PR #4243, opened 2025-12-16
docs: ":lsp ..." in Nvim 0.12

justinmk:


doc/lspconfig.txt
justinmk, line 79: We probably need `:lsp info` , would also like `:che vim.lsp` to better visualize the "current buffer" properties, somehow.

PR #4247, opened 2025-12-19
fix(golangci_lint_ls): disable all output formats except JSON

timofurrer:
This change set enables all possible output formats that exist for golangci-lint. This is necessary because if a user configures output formats in their `.golangci.yml` configuration (for example for CI) then this is always in addition to the ones enabled on the command line. Therefore, we manually disable all possible output formats.

Currently, there doesn't exist a `golangci-lint` option to exclusively enable an output format or disable all from the config.

### DIFF
--- a/lsp/golangci_lint_ls.lua
+++ b/lsp/golangci_lint_ls.lua
@@ -18,7 +18,22 @@ return {
   cmd = { 'golangci-lint-langserver' },
   filetypes = { 'go', 'gomod' },
   init_options = {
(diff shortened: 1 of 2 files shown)


timofurrer:
I've put this back to a draft to explore other options, like https://github.com/golangci/golangci-lint/pull/6273.

timofurrer:
The "other options" I was exploring turn out to take rather more time than expected and it's unclear to where to go from there. In favor of preventing others to run into issue it might be fine to go with this iteration.

PR #4250, opened 2025-12-19
fix(buf_ls): remove beta in comments

stefanvanburen:
Also removes the timeout option, which is unnecessary with the latest version of the LSP.


stefanvanburen:
Fixed version of #4249.

PR #4254, opened 2025-12-22
fix(roslyn_ls): fix lsp attachment on decompiled code

mbwilding:
This fix allows us to gd into decompiled code with the lsp attached, and allows nested decompilation

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -159,6 +159,11 @@ return {
       if root_dir then
         cb(root_dir)
       end
+    else
+      local existing_client = vim.lsp.get_clients({ name = 'roslyn_ls' })[1]
+      if existing_client and existing_client.config.root_dir then
+        cb(existing_client.config.root_dir)
+      end
     end
   end,
   on_init = {


lsp/roslyn_ls.lua
```diff
@@ -159,6 +159,11 @@ return {
       if root_dir then
         cb(root_dir)
       end
+    else
+      local existing_client = vim.lsp.get_clients({ name = 'roslyn_ls' })[1]
+      if existing_client and existing_client.config.root_dir then
```
justinmk, line 164: In a multi-workspace session, this will just choose a random client with a random workspace (root dir)?

mbwilding: Pretty much, but not sure how to detect what instance is in use when it goes into the decompiled code buffer, then reference that server. For now this works fine, got any suggestions?

justinmk: Depends on the "filepath" of the decompiled code, right? What does it look like?

mbwilding: On Linux here is an example: `/tmp/MetadataAsSource/f2bfb96543ea401db636aee47d3d9bab/DecompilationMetadataAsSourceFileProvider/d5782a14d4c844c584628f7ff48a3ef4/Console.cs`

Can see the original code is checking that `/tmp/MetadataAsSource` doesn't match, my changes are in the else of that.

```lua
if not bufname:match('^' .. fs.joinpath('/tmp/MetadataAsSource/')) then
...
else
  my changes
```

It is outside of whatever project/root you are working within.

I will try returning a root based on current path of the decompiled code as a test.

EDIT: Looks like it spins up a new roslyn_ls instance when we do that, which is inefficient. But I found a work around, we can use `vim.fn.bufnr('#')` and get the correct server. Will create another PR after I test extensively.

mbwilding: https://github.com/neovim/nvim-lspconfig/pull/4257

Resolved in this PR.

PR #4257, opened 2025-12-23
fix(roslyn_ls): detect correct lsp client for decompiled code

mbwilding:
Follow up from https://github.com/neovim/nvim-lspconfig/pull/4254 @justinmk

This change resolves the LSP root_dir more reliably by first checking the previous buffer’s attached client and reusing its root_dir when available. If no matching client is found, it falls back to any existing client’s root_dir. As a final fall-back, it sets the root_dir to the decompilation root, which intentionally forces a new LSP client to be created.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -160,7 +160,7 @@ return {
         cb(root_dir)
       end
     else
-      -- decompiled code
+      -- Decompiled code (example: "/tmp/MetadataAsSource/f2bfba/DecompilationMetadataAsSourceFileProvider/d5782a/Console.cs")
       local prev_buf = vim.fn.bufnr('#')
       local client = vim.lsp.get_clients({
         name = 'roslyn_ls',


justinmk:
The `vim.fn.bufnr('#')` attempt seems ok, but drop the rest of the "fallback" logic. That should shrink this to about 6 lines instead of 36. This config is getting too complex and the other fallbacks aren't worth it.

mbwilding:
Sorted it out, wanted it to be robust. Updated the code.
Could make it just this if you want but I'd rather what is currently committed, as if someone goes into `/tmp/MetadataAsSource/` it'll throw an error.

```lua
local client = vim.lsp.get_clients({ name = 'roslyn_ls', bufnr = vim.fn.bufnr('#') })[1]
cb(client.config.root_dir)
```

mbwilding:
Turns out we need more guards, if we browse to files manually in those folders.

mbwilding:
Ready for review now, squashed my commits to reduce noise.

lsp/roslyn_ls.lua
```diff
@@ -160,9 +160,14 @@ return {
         cb(root_dir)
       end
     else
-      local existing_client = vim.lsp.get_clients({ name = 'roslyn_ls' })[1]
-      if existing_client and existing_client.config.root_dir then
-        cb(existing_client.config.root_dir)
+      -- decompiled code
```
justinmk: An example is always very high-bandwidth and worthwhile.
```suggestion
      -- Decompiled code (example: "/tmp/MetadataAsSource/f2bfba/DecompilationMetadataAsSourceFileProvider/d5782a/Console.cs")
```

PR #4260, opened 2025-12-24
fix(tblgen_lsp_server): also search build dir for db

sweiglbosker:
cmake will generate `tablegen_compile_commands.yml` in the build dir by default, so the current config will miss it in pretty much every case.
I guess a better approach might be to scan recursively from the root dir, but i think this will work for most cases.


lsp/tblgen_lsp_server.lua
justinmk, line 25: Would be better to define `cmd` as a function (see other configs for example) instead of doing this at module-load time (which depends on Nvim's CWD, which is arbitrary).

Issue #4134, opened 2025-10-17
Consider change of current language server implementation for unit files


JFryy:
### Description

Hey, I'd like to add this to nvim-lspconfig for "systemd type" files but I see there is already one in this repository that I think [might be no longer maintained](https://github.com/psacawa/systemd-language-server). Would this be a change appropriate? Wanted to open an issue before opening an unsolicited PR considering. Relates to https://github.com/JFryy/systemd-lsp/issues/18

justinmk:
https://github.com/neovim/nvim-lspconfig/pull/3514 @cwrau are you still using this LS ?

https://github.com/JFryy/systemd-lsp looks active , i would be fine with just updating the current config and mentioning a URL to the old one in the docstring.

cwrau:
@JFryy what makes you think that the current lsp is not maintained anymore? 🤔

@justinmk I'm currently using the current one, works just great, but if this one works the same as the old one ([and is listed in mason!!](https://github.com/mason-org/mason-registry/blob/main/packages/systemd-language-server/package.yaml)) then I don't have a strong opinion 😁

JFryy:
Hey @cwrau, full disclaimer: I am pretty biased since I am the maintainer of the alternate `jfryy/systemd-lsp` suggested in this issue. But I think it might be on account of last commit being around a year ago and some longstanding PRs Issues in the repo. Maybe was premature to have opened this, sorry just following up on an issue in my project with due diligence more than anything, that LS will chug along fine for a while even without updates.

JFryy:
I think this one has a few more features in fact, getting it into Mason is odd because they request that it's included here first IIRC 😛

cwrau:
> I think this one has a few more features in fact, getting it into Mason is odd because they request that it's included here first IIRC 😛 

I mean as long as it's in both it's OK for me for this to be swapped out, even if for a short period of time it's a little finicky. 

Not that I have any authority to decide this anyways 😅

magnuslarsen:
> [@justinmk](https://github.com/justinmk) I'm currently using the current one, works just great, but if this one works the same as the old one ([and is listed in mason!!](https://github.com/mason-org/mason-registry/blob/main/packages/systemd-language-server/package.yaml)) then I don't have a strong opinion 😁

@JFryy's LSP is now also in Mason :-) https://github.com/mason-org/mason-registry/pull/12914

justinmk:
[...]

PR #4262, opened 2025-12-30
feat: deprecate systemd_ls in favor of systemd_lsp

magnuslarsen:
I think I did this right - otherwise let me know :-)

closes https://github.com/neovim/nvim-lspconfig/issues/4134

### DIFF
--- a/lsp/systemd_ls.lua
+++ b/lsp/systemd_ls.lua
@@ -1,17 +1,8 @@
 ---@brief
 ---
---- https://github.com/psacawa/systemd-language-server
----
---- `systemd-language-server` can be installed via `pip`:
---- ```sh
---- pip install systemd-language-server
---- ```
----
---- Language Server for Systemd unit files
+--- Renamed to [systemd_lsp](#systemd_lsp)
+
+vim.deprecate('systemd_ls', 'systemd_lsp', '2.0.0', 'nvim-lspconfig', false)
 
 ---@type vim.lsp.Config
-return {
-  cmd = { 'systemd-language-server' },
-  filetypes = { 'systemd' },
-  root_markers = { '.git' },
-}
+return vim.lsp.config.systemd_ls


lsp/systemd_ls.lua
```diff
@@ -1,17 +1,8 @@
...
----
---- `systemd-language-server` can be installed via `pip`:
---- ```sh
---- pip install systemd-language-server
---- ```
----
---- Language Server for Systemd unit files
+--- Renamed to [systemd_lsp](#systemd_lsp)
+
+vim.deprecate('systemd_ls', 'systemd_lsp', '2.0.0', 'nvim-lspconfig', false)
```
Arnie97, line 5: It does not make sense to declare that `the feature will be removed in nvim-lspconfig 2.0.0`, because [v2.0.0](https://github.com/neovim/nvim-lspconfig/releases/tag/v2.0.0) has been released on April 11, 2025.

lsp/systemd_ls.lua
```diff
@@ -1,17 +1,8 @@
...
 
 ---@type vim.lsp.Config
-return {
-  cmd = { 'systemd-language-server' },
-  filetypes = { 'systemd' },
-  root_markers = { '.git' },
-}
+return vim.lsp.config.systemd_ls
```
Arnie97, line 8: This is an inﬁnite recursion. I guess you meant to `return vim.lsp.config.systemd_lsp`, not `systemd_ls` itself. This caused stack overflows in my neovim config, just like what LazyVim/LazyVim#6913 has reported.

magnuslarsen: Yep - it was fixed in https://github.com/neovim/nvim-lspconfig/commit/9506ae35c9061d7a0f4a91caf8244b764a2a8fba

lsp/systemd_ls.lua
magnuslarsen: Any guidelines for when it should be deprecated? Just the next major version?

PR #4264, opened 2026-01-01
feat: pony_language_server

mfelsche:
I left out generating the docs, as there seems to be a ci job doing just that.

I can totally generate docs, if required.

### DIFF
--- a/lsp/pony_language_server.lua
+++ b/lsp/pony_language_server.lua
@@ -1,7 +1,7 @@
 ---@brief
 ---
 --- https://github.com/ponylang/pony-language-server
-
+---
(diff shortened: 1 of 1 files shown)


lsp/pony_language_server.lua
```diff
@@ -0,0 +1,12 @@
+---@brief
+---
+--- https://github.com/ponylang/pony-language-server
+
+--- Language server for the Pony programming language
```
justinmk, line 5:
```suggestion
--- https://github.com/ponylang/pony-language-server
---
--- Language server for the Pony programming language
```

PR #4265, opened 2026-01-03
fix(tsgo): prevent attaching to non file buffers

igorlfs:
Hey,

`tsgo` is still a rather experimental server. While testing, I ran into numerous crashes related to attaching to non file buffers (from plugins, etc).

I believe this will eventually get fixed on the server side. In the meantime, we could have this...

### DIFF
--- a/lsp/tsgo.lua
+++ b/lsp/tsgo.lua
@@ -47,6 +47,12 @@ return {
       return
     end
 
+    -- prevent from attaching to non file buffers
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    if not vim.uv.fs_stat(fname) then
+      return
+    end
+
     -- We fallback to the current working directory if no project root is found
     local project_root = vim.fs.root(bufnr, root_markers) or vim.fn.getcwd()
 


igorlfs:
Marking as draft because I'm unsure if others are also having this issue. And also, the server _does not explicitly state_ that is lacks supports for non file URIs, AFAIK. Although, if we are to take my experience as reference, this feature is clearly not supported.

igorlfs:
> the server does not explicitly state that is lacks supports for non file URIs

Unmarking as draft because the server isn't able to resolve imports for these files (which might be unintended), making for a poor experience (particularly when using plugins for diffing files). 

I'll leave it up for the maintainers to decide if it's best to have this as the bundled behavior.

Note: I'm no longer running into crashes with these files

justinmk:
"non file" in Vim is `buftype=nofile`, but i guess what is meant here is a new, unsaved file?

and thus, `workspace_required=true` doesn't do what you want ? https://github.com/neovim/nvim-lspconfig/blob/66fd02ad1c7ea31616d3ca678fa04e6d0b360824/lsp/ccls.lua#L46-L47

igorlfs:
> but i guess what is meant here is a new, unsaved file?

Not really. I'm mostly concerned with buffers from a specific plugin (octo.nvim). It uses the empty buftype for its special diff buffers (`octo://...`). Most language servers seem to not care about that, and `tsgo` handles it a lot better now, but it's still unable to resolve imports, which ends up creating many fake diagnostics.

Back when I opened this PR, `tsgo` had a lot more trouble dealing with these buffers, frequently crashing. But that's solved now, the only missing bit is the import resolution. To me, that's a strong indicator that this situation will eventually be fully resolved on the server side (as is the case with the original `ts_ls`), so I'm closing this.

> workspace_required=true doesn't do what you want ? 

That's not the case here, but I'd always avoid setting `workspace_required` (if it's not truly required), because someone out there is relying on this behavior...

PR #4266, opened 2026-01-03
fix: cleanup unknown filetypes

igorlfs:
Hey,

This cleans up a bunch of configs, removing unknown filetypes.

Q: **Why some configs and not the others?** 
A: I cleaned all the ones I use. Well, most of them. `yamlls` seems to have some fair use cases.

New configs could require a sample `vim.filetype.add` if they use a non-standard filetypes;


lsp/html.lua
```diff
@@ -25,7 +25,7 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'vscode-html-language-server', '--stdio' },
-  filetypes = { 'html', 'templ' },
+  filetypes = { 'html' },
```
justinmk, line 28: isn't "templ" for go?

justinmk:
Somewhat hesitant about this until we are certain which ones aren't provided by third party ftplugins. If they are, then we shouldn't remove them.

lsp/html.lua
lithammer: I think that's `gotmpl`?

https://github.com/ngalaiko/tree-sitter-go-template?tab=readme-ov-file#neovim-integration-using-nvim-treesitter

igorlfs:
Maybe we could reduce the scope to not include `tailwindcss` for now? The other ones are way simpler

justinmk:
SGTM, but I thought `typescript.tsx` was a Vim convention, something to do with "sub types"

igorlfs:
> I thought typescript.tsx was a Vim convention

I'm not aware of such convention, though I'm not a React guy. AFAIK that's a mistake that's been copy-pasted ad infinitum.

PR #4269, opened 2026-01-04
fix: show deprecation messages on_init

superatomic:
Fixes:
- ~~`vscoqtop` was marked to be removed in version 2.0.0, so remove it.~~ Looks like this was a mistake. Bumping the version to 3.0.0.
- Instead of displaying the deprecation notice when the config is loaded, which causes it to incorrectly appear when using functions like `vim.lsp.get_configs()` (see https://github.com/neovim/neovim/pull/37237) or the command `:lsp enable`, only display the deprecation notice when the config is actually used.


perryrh0dan:
I was consistently seeing a deprecation warning related to systemd_ls on startup. In my case, the issue was caused by outdated versions of mason-org/mason.nvim and mason-org/mason-lspconfig.nvim. After updating both plugins to their latest versions, the deprecation message stopped appearing.

superatomic:
Of note: this is currently an issue when using `:lsp enable` with no arguments, and is not a problem limited to a hypothetical future `vim.lsp.get_configs()`.

justinmk:
This is a thoughtful idea. The amount of code is unfortunate, but it's still fairly easy to copy-paste from the docs 👍 .

PR #4288, opened 2026-01-20
fix: include deno in biome

emonadeo:
#4130 prevents a bunch of lsps to start in a deno project, including biome. However I think the biome lsp should indeed be active in deno projects with a `biome.json`.

I could of course just supply my own lsp config in my neovim config, but I would like to challenge the "defaults" here.


justinmk:
@xbwwj do you see any problem with this

PR #4293, opened 2026-01-26
fix(ruby_lsp): specify cmd_cwd to ruby-lsp spawn at root directory

hasunpark:
Problem:
Ruby-lsp spawn in cwd even lsp detect root directory correctly

Solution:
Specify cmd_cwd to spawn cmd at root directory

### DIFF
--- a/lsp/ruby_lsp.lua
+++ b/lsp/ruby_lsp.lua
@@ -14,7 +14,13 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'ruby-lsp' },
+  cmd = function(dispatchers, config)
+    return vim.lsp.rpc.start(
+      { 'ruby-lsp' },
+      dispatchers,
+      config.root_dir and { cwd = config.cmd_cwd or config.root_dir }
+    )
+  end,
   filetypes = { 'ruby', 'eruby' },
   root_markers = { 'Gemfile', '.git' },
   init_options = {


lsp/ruby_lsp.lua
```diff
@@ -13,13 +13,15 @@
...
+local ROOT_MARKERS = { 'Gemfile', '.git' }
 return {
   cmd = { 'ruby-lsp' },
   filetypes = { 'ruby', 'eruby' },
-  root_markers = { 'Gemfile', '.git' },
+  root_markers = ROOT_MARKERS,
   init_options = {
     formatter = 'auto',
   },
+  cmd_cwd = vim.fs.root(0, ROOT_MARKERS),
```
justinmk: this is called as soon as the module is loaded, which is arbitrary and depends on user config, CWD, etc. 

to resolve CWD dynamically, define `cmd` as a function. 

example:

https://github.com/neovim/nvim-lspconfig/blob/f7e7504bf1792bcfcdf72b52eceacdb3fcce1e32/lsp/tsgo.lua#L19-L26

hasunpark: Hi @justinmk thanks for letting know there are better way to solve this issue!

I've rework to using cmd as a function. now it spawn with root_dir or cmd_cwd(case when user specified) as cwd

justinmk:
With Nvim 0.13 https://github.com/neovim/neovim/issues/40326 , the `cmd` CWD is derived from `root_dir` by default.

PR #4295, opened 2026-01-27
fix(roslyn_ls): make test if file is decompiled or not os agnostic

Async10:
At the moment the path which is used to test if a file is decompiled or not is hardcoded and incorrect (at least for windows users). It doesn't respect `\` as path separator and makes wrong assumptions about the temp path.  Hence why `vim.lsp.buf.definition` is not working in decompiled files.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -108,6 +108,16 @@ local function roslyn_handlers()
   }
 end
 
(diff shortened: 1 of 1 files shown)


justinmk:
in the future please just force push to your existing branch, don't open multiple PRs https://github.com/neovim/nvim-lspconfig/pull/4296

PR #4296, opened 2026-01-27
fix(roslyn_ls): make test if decompiled os agnostic

Async10:
At the moment hard coded assumptions about decompiled path are made. That is why `vim.lsp.buf.definition` is not working in decompiled files on windows.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -98,6 +98,16 @@ local function roslyn_handlers()
   }
 end
 
+---@param bufname string
+---@return boolean
+local function is_decompiled(bufname)
+  local _, endpos = bufname:find('[/\\]MetadataAsSource[/\\]')
+  if endpos == nil then
+    return false
+  end
+  return vim.fn.finddir(bufname:sub(1, endpos), vim.env.TMP or vim.env.TEMP) ~= ''
+end
+
 ---@type vim.lsp.Config
 return {
   name = 'roslyn_ls',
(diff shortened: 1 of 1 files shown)


lsp/roslyn_ls.lua
```diff
@@ -143,7 +153,7 @@ return {
     local bufname = vim.api.nvim_buf_get_name(bufnr)
     -- don't try to find sln or csproj for files from libraries
     -- outside of the project
-    if not bufname:match('^' .. fs.joinpath('/tmp/MetadataAsSource/')) then
```
justinmk, line 146: why not just change this to something like

    if not bufname:match('^' .. fs.joinpath([=[[/\]tmp/MetadataAsSource[/\]]=])) then

Async10: Your suggestion doesn't work on windows. It has the same problems as the current implementation. It doesn't respect backslash as a path separator nor does it respect temporay directories that aren't named tmp like C:\Users\xxx\AppData\Local\Temp.

justinmk: by "it" i guess you mean joinpath(). I don't actually know why joinpath() is being used here, with 1 arg...

Async10: By it I meant your approach as a whole. I agree the use of joinpath is useless here. From your reply I understand that you want to stick to `string.match` but the pattern you provided is not working and I can't think of an easy one that covers all the windows edge cases like multiple temporary directories and [short file names](https://en.wikipedia.org/wiki/8.3_filename). Look i am telling you that this config has a problem. I shared my working solution and I am sorry to say that but your feedback is not helpful to me.

justinmk: My suggestion missed the `/` in the middle. It was just to show the idea, not literal/tested code. 

Anyway, I see that your code is avoiding hardcoding `/tmp` which makes sense.

PR #4299, opened 2026-01-30
Fix some issues with the oxlint config

kmaglione:
This branch originally had a couple more fixes, but they conflicted with other fixes on rebase.

The current version:

- Adds help text for the `:LspOxlintFixAll` command
- Automatically enables type-aware linting when appropriate
- Copies the `settings` table to initialization options, if it exists
- Stops checking `package.json` for an `oxlint` section when searching for a root directory, since Oxlint has never supported that type of configuration

### DIFF
--- a/lsp/oxlint.lua
+++ b/lsp/oxlint.lua
@@ -34,10 +34,8 @@ return {
   filetypes = {
     'javascript',
     'javascriptreact',
-    'javascript.jsx',
     'typescript',
     'typescriptreact',
-    'typescript.tsx',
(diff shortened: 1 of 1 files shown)


igorlfs:
While you're at it, can you remove the filetypes `typescript.tsx` and `javascript.jsx`? They are not valid, see #4266

kmaglione:
> While you're at it, can you remove the filetypes `typescript.tsx` and `javascript.jsx`? They are not valid, see #4266

Done

PR #4301, opened 2026-01-30
feat(clice): add clice lsp config

tsurumi-yizhou:


lsp/clice.lua
```diff
@@ -0,0 +1,19 @@
+---@brief
+--- https://github.com/clice-io/clice
+---
+--- Clice is a next generation C++ language server
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'clice', '--mode=pipe' },
+  filetypes = { 'cpp', 'cc', 'cxx', 'hpp', 'hxx', 'h', 'ixx', 'cppm' },
```
justinmk: Are all of these valid Vim *filetypes*, or are some of these file extensions? `filetypes` does not support file extensions yet (and when it does, it will require a "glob" expression, thus extensions such as `h` here are invalid).

The filetype for a `.h` file is `cpp`, not `h`.

lsp/clice.lua
```diff
@@ -0,0 +1,19 @@
+---@brief
+--- https://github.com/clice-io/clice
+---
+--- Clice is a next generation C++ language server
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'clice', '--mode=pipe' },
+  filetypes = { 'cpp' },
```
justinmk, line 9: may also want to add `c` here, since a c++ project may have both cpp and c files

lsp/clice.lua
```diff
@@ -0,0 +1,19 @@
...
+  root_markers = {
+    '.git/',
```
igorlfs, line 11:
```suggestion
    '.git',
```

Issue #4300, opened 2026-01-30
denols and ts_ls conflict, neither activate in many typical scenarios


isaacs:
### Description

Out of the box, it is impossible to attach the LSP client to most deno projects.

The `denols` configuration will not attach if there is a `package.json` file at a parent path equal to or greater than the nearest `deno.lock` file.

The `ts_ls` configuration will not attach if there is a `deno.lock` file at a parent path equal to or greater than the nearest `package-lock.json` file. (It does not respect `package.json`, though it ought to.)

This means that if you have a `package.json` and `deno.lock` file in the same directory, neither lsp client will attach.

Consider the following fairly typical case case:

```
my-project-workspace
+-- package-lock.json
+-- deno.lock
+-- packages
    +-- deno-package
    |    +-- deno.json
    |    +-- package.json
    |    +-- src
    |        +-- index.ts
    +-- node-package
        +-- package.json
        +-- src
            +-- index.ts
```

The file at `packages/deno-package/src/index.ts` will not
attach the `denols` client, because there is a `package.json`
file present closer than the nearest `deno.lock` file.

The file at `packages/node-package/src/index.ts` will not attach
the `ts_ls` client, because there is a `deno.lock` file present
parallel to the nearest `package-lock.json` file.

There is no LSP attachment for any typescript files anywhere in the project!

## Proposed solution

If both lsps are enabled, then at least one should apply for any given TS file, no matter what.

### `denols` behavior

- Look for a `deno.lock`, `deno.json`, or `deno.jsonc` file.
- Attach as long as the deno root is *at least as close* as the
  non-deno lockfile or project marker

```lua
local deno_markers = { 'deno.lock', 'deno.json', 'deno.jsonc' }
local project_markers = {
  'package-lock.json',
  'yarn.lock',
  'pnpm-lock.yaml',
  'bun.lockb',
  'bun.lock',
  'package.json',
  'node_modules',
}
local deno_root = vim.fs.root(bufnr, deno_markers)
local project_root = vim.fs.root(bufnr, project_markers)

if project_root and (not deno_root or #project_root > #deno_root) then
    return
end
on_dir(deno_root or vim.fn.getcwd())
```

### `ts_ls` behavior

This should be the exact inverse of the `denols` attachment
behavior.

- Look for a `deno.lock`, `deno.json`, or `deno.jsonc` file.
- Attach as long as the deno root is further away than any
  non-deno project root.

```lua
[...]

PR #4304, opened 2026-02-02
fix(deno, ts_ls): denols/ts_ls inference

isaacs:
Update the logic to infer the type properly when a Deno file is part of a non-Deno monorepo project, enabling using the denols and ts_ls (or tsgo) LSPs side by side in the same monorepo without conflict.

Fix: #4300


isaacs:
Removed the virtualTextDocument handler, because I had missed that it was added to the lsp config on the master branch already 😅

KingJordan152:
Would it be possible to also add this logic to `vtsls`?

isaacs:
> Would it be possible to also add this logic to `vtsls`?

Sure, I'll push another commit in just a moment.

justinmk:
Very nice. Thank you for looking into this!

PR #4306, opened 2026-02-04
ci: generate annotations for Lua types

ofseed:
Automatically generate Lua types from the LSP server settings JSON schema
screenshot

Modified from [neoconf.nvim](https://github.com/folke/neoconf.nvim)

### DIFF
--- a/lsp/perlpls.lua
+++ b/lsp/perlpls.lua
@@ -10,6 +10,7 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'pls' },
+  ---@type lspconfig.settings.perlpls
   settings = {
     perl = {
       perlcritic = { enabled = false },
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Run luals
         run: |
           export VIMRUNTIME="$(nvim --headless --cmd 'echo $VIMRUNTIME' +q 2>&1)"
-          ./build/luals/bin/lua-language-server --configpath=$(pwd)/.luarc.json --check=lsp/ --checklevel=Hint
+          ./build/luals/bin/lua-language-server --configpath=$(pwd)/.luarc.json --check=lsp/ --check=lua/lspconfig/types/lsp/ --checklevel=Hint
 
 
   stylua:
--- a/lsp/basedpyright.lua
+++ b/lsp/basedpyright.lua
@@ -34,6 +34,7 @@ return {
     'Pipfile',
     '.git',
   },
+  ---@type lspconfig.settings.basedpyright
   settings = {
     basedpyright = {
       analysis = {
--- a/lsp/bashls.lua
+++ b/lsp/bashls.lua
@@ -12,6 +12,7 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'bash-language-server', 'start' },
+  ---@type lspconfig.settings.bashls
   settings = {
     bashIde = {
       -- Glob pattern for finding and parsing shell script files in the workspace.
--- a/lsp/cssls.lua
+++ b/lsp/cssls.lua
@@ -26,6 +26,7 @@ return {
   filetypes = { 'css', 'scss', 'less' },
   init_options = { provideFormatter = true }, -- needed to enable formatting capabilities
   root_markers = { 'package.json', '.git' },
+  ---@type lspconfig.settings.cssls
   settings = {
     css = { validate = true },
     scss = { validate = true },
--- a/lsp/dartls.lua
+++ b/lsp/dartls.lua
@@ -16,6 +16,7 @@ return {
     outline = true,
     flutterOutline = true,
   },
+  ---@type lspconfig.settings.dartls
   settings = {
     dart = {
       completeFunctionCalls = true,
--- a/lsp/denols.lua
+++ b/lsp/denols.lua
@@ -124,6 +124,7 @@ return {
       on_dir(project_root or deno_lock_root or deno_root)
     end
   end,
+  ---@type lspconfig.settings.denols
   settings = {
     deno = {
       enable = true,
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -142,6 +142,7 @@ return {
     on_dir(project_root)
   end,
   -- Refer to https://github.com/Microsoft/vscode-eslint#settings-options for documentation.
+  ---@type lspconfig.settings.eslint
   settings = {
     validate = 'on',
     ---@diagnostic disable-next-line: assign-type-mismatch
--- a/lsp/fsautocomplete.lua
+++ b/lsp/fsautocomplete.lua
@@ -31,6 +31,7 @@ return {
     AutomaticWorkspaceInit = true,
   },
   -- this recommended settings values taken from  https://github.com/ionide/FsAutoComplete?tab=readme-ov-file#settings
+  ---@type lspconfig.settings.fsautocomplete
   settings = {
     FSharp = {
       keywordsAutocomplete = true,
--- a/lsp/html.lua
+++ b/lsp/html.lua
@@ -27,6 +27,7 @@ return {
   cmd = { 'vscode-html-language-server', '--stdio' },
   filetypes = { 'html' },
   root_markers = { 'package.json', '.git' },
+  ---@type lspconfig.settings.html
   settings = {},
   init_options = {
     provideFormatter = true,
--- a/lsp/intelephense.lua
+++ b/lsp/intelephense.lua
@@ -30,6 +30,7 @@ return {
   cmd = { 'intelephense', '--stdio' },
   filetypes = { 'php' },
   root_markers = { '.git', 'composer.json' },
+  ---@type lspconfig.settings.intelephense
   settings = {
     intelephense = {
       telemetry = {
--- a/lsp/java_language_server.lua
+++ b/lsp/java_language_server.lua
@@ -11,5 +11,6 @@ return {
   cmd = { 'java-language-server' },
   filetypes = { 'java' },
   root_markers = { 'build.gradle', 'build.gradle.kts', 'pom.xml', '.git' },
+  ---@type lspconfig.settings.java_language_server
   settings = {},
 }
--- a/lsp/ltex.lua
+++ b/lsp/ltex.lua
@@ -77,6 +77,7 @@ return {
   filetypes = filetypes,
   root_markers = { '.git' },
   get_language_id = get_language_id,
+  ---@type lspconfig.settings.ltex
   settings = {
     ltex = {
       enabled = enabled_ids,
--- a/lsp/lua_ls.lua
+++ b/lsp/lua_ls.lua
@@ -85,6 +85,7 @@ return {
   filetypes = { 'lua' },
   root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers1, root_markers2, { '.git' } }
     or vim.list_extend(vim.list_extend(root_markers1, root_markers2), { '.git' }),
+  ---@type lspconfig.settings.lua_ls
   settings = {
     Lua = {
       codeLens = { enable = true },
--- a/lsp/omnisharp.lua
+++ b/lsp/omnisharp.lua
@@ -43,6 +43,7 @@ return {
       workspaceFolders = false, -- https://github.com/OmniSharp/omnisharp-roslyn/issues/909
     },
   },
+  ---@type lspconfig.settings.omnisharp
   settings = {
     FormattingOptions = {
       -- Enables support for reading code style, naming convention and analyzer
--- a/lsp/perlls.lua
+++ b/lsp/perlls.lua
@@ -17,6 +17,7 @@ return {
     '--port 13603',
     '--nostdio 0',
   },
+  ---@type lspconfig.settings.perlls
   settings = {
     perl = {
       perlCmd = 'perl',
--- a/lsp/pyright.lua
+++ b/lsp/pyright.lua
@@ -34,6 +34,7 @@ return {
     'Pipfile',
     '.git',
   },
+  ---@type lspconfig.settings.pyright
   settings = {
     python = {
       analysis = {
--- a/lsp/rescriptls.lua
+++ b/lsp/rescriptls.lua
@@ -39,6 +39,7 @@ return {
   cmd = { 'rescript-language-server', '--stdio' },
   filetypes = { 'rescript' },
   root_markers = { 'bsconfig.json', 'rescript.json', '.git' },
+  ---@type lspconfig.settings.rescriptls
   settings = {},
   init_options = {
     extensionConfiguration = {
--- a/lsp/ruff_lsp.lua
+++ b/lsp/ruff_lsp.lua
@@ -26,5 +26,6 @@ return {
   cmd = { 'ruff-lsp' },
   filetypes = { 'python' },
   root_markers = { 'pyproject.toml', 'ruff.toml', '.git' },
+  ---@type lspconfig.settings.ruff_lsp
   settings = {},
 }
--- a/lsp/rust_analyzer.lua
+++ b/lsp/rust_analyzer.lua
@@ -143,6 +143,7 @@ return {
       },
     },
   },
+  ---@type lspconfig.settings.rust_analyzer
   settings = {
     ['rust-analyzer'] = {
       lens = {
--- a/lsp/solargraph.lua
+++ b/lsp/solargraph.lua
@@ -13,6 +13,7 @@
 ---@type vim.lsp.Config
 return {
   cmd = { 'solargraph', 'stdio' },
+  ---@type lspconfig.settings.solargraph
   settings = {
     solargraph = {
       diagnostics = true,
--- a/lsp/spectral.lua
+++ b/lsp/spectral.lua
@@ -14,6 +14,7 @@ return {
   cmd = { 'spectral-language-server', '--stdio' },
   filetypes = { 'yaml', 'json', 'yml' },
   root_markers = { '.spectral.yaml', '.spectral.yml', '.spectral.json', '.spectral.js' },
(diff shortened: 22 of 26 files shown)


igorlfs:
Hey, thanks for looking into this. 

With #4010 I had a similar goal, although with a much more limited scope. 

When this PR is finished, the changes, from a user's perspective, are:
- Built-in configurations will now use custom types (e.g., instead of generic `vim.lsp.Config`, would use `lspconfig.tsgo` for `tsgo`, etc)
- These custom types will be exposed, so they can also be used elsewhere (i.e., user's config)

Are these assumptions correct?

ofseed:
> * Built-in configurations will now use custom types (e.g., instead of generic `vim.lsp.Config`, would use `lspconfig.tsgo` for `tsgo`, etc)


Since the previous way to enable lspconfig followed the `lspconfig.tsgo.setup` pattern, neoconf was able to configure the type for keys like `tsgo`. This allowed users to avoid manual annotations once the generated types were loaded. However, the current `vim.lsp.config` chooses to pass the server name as a positional argument rather than a key, so this method might not be effective—it might still work via `---@overload`, but I haven't tried that yet.

I don't intend to deprecate `@type vim.lsp.Config`, but rather to provide compatibility. My current preliminary plan is to provide a series of separate `@class lspconfig.tsgo.settings`, so that manual annotation is at least possible.

> * These custom types will be exposed, so they can also be used elsewhere (i.e., user's config)

Yes, this is a very important objective. In fact, neoconf itself doesn't need these type definitions; it generates them purely as a feature to make it convenient for users to use these types within their own configurations.

ofseed:
I just noticed an issue: since the JSON schemas are all JSON with comments rather than standard JSON, `vim.json` cannot be used here. Are there any plans to have `vim.json` support jsonc? Otherwise, a jsonc parser must be added here. cc @justinmk

justinmk:
> Are there any plans to have `vim.json` support jsonc?

I'm favor of that but we don't have a tracking issue for it, nor in https://github.com/openresty/lua-cjson . 

If it's not too horrible, this may even justify forking cjson if it won't be accepted upstream. 

cc @skewb1k

arkriny:
> Are there any plans to have `vim.json` support jsonc?

I think it's reasonable to have a flag for that.

> I'm favor of that but we don't have a tracking issue for it, nor in https://github.com/openresty/lua-cjson .
> 
> If it's not too horrible, this may even justify forking cjson if it won't be accepted upstream.

We can take a similar approach to https://github.com/neovim/neovim/pull/35574 and https://github.com/neovim/neovim/pull/35424 by opening PRs to both upstream and neovim, and being okay with merging into neovim first. If upstream is not interested, we can certainly diverge. Additionally, I recall some discussions on Matrix regarding hard-forking due to API differences and other factors, so I don't think that's a problem.

arkriny:
> I just noticed an issue: since the JSON schemas are all JSON with comments rather than standard JSON, `vim.json` cannot be used here.

While it's cool to support comments in `vim.json` and I'm working on the patch, I don't quite understand why its needed for JSON schemas? It seems that all schemas are valid JSONs, and the specification even introduces the [$comment](https://json-schema.org/understanding-json-schema/reference/comments#comments) field, which is intended to be ignored semantically. Probably it wouldn't be necessary if the comment syntax was allowed. 

```lua
local function readfile(path)
  local file = io.open(path, "r")
  if not file then
    return nil
  end
  local content = file:read("*a")
  file:close()
  return content
end

local dir = "./schemas"
local files = vim.fn.readdir(dir)
for _, file in ipairs(files) do
  local file_path = vim.fs.joinpath(dir, file)
  local content = readfile(file_path)
  vim.json.decode(content)
end
```

Running this in the root of https://github.com/folke/neoconf.nvim succeeds (introducing comment manually will cause it to fail).
I was thinking that in neoconf, jsonc is necessary to support comments in the user's config file. Am I missing something?

ofseed:
You are correct. I noticed that neoconf uses a JSONC decoder for JSON schemas, which led to my misunderstanding that JSON schemas could be JSONC; they should strictly be standard JSON.

The reason neoconf uses JSONC parsing here should because this wrapped function is also used to parse settings.json, which is allowed to be JSONC. Therefore, for this PR, using vim.json is sufficient.

scripts/gen_types/util.lua
```diff
@@ -0,0 +1,84 @@
...
+  M.write_file(tmp, vim.json.encode(obj))
+  local fd = io.popen('jq -S < ' .. tmp)
+  if not fd then
+    error('Could not format json')
+  end
+  local ret = fd:read('*a')
+  if ret == '' then
+    error('Could not format json')
+  end
+  fd:close()
```
yochem: I'm not sure what the state of this PR is but can't this be `vim.json.encode(obj, { indent = '  ', sort_keys = true })`?

scripts/gen_types/util.lua
```diff
@@ -0,0 +1,84 @@
...
+end
+
+function M.fetch(url)
+  local fd = io.popen(string.format('curl -s -k %q', url))
+  if not fd then
+    error(('Could not download %s'):format(url))
+  end
+  local ret = fd:read('*a')
+  fd:close()
+  return ret
```
yochem: `vim.net.request`?

ofseed: It looks like this method can only be asynchronous, but a synchronous call is needed here. It may require a major refactor to make that possible.

justinmk: `vim.wait` ?

ofseed:
I have completed the essential refactoring; this PR is now in a usable state.

lsp/perlpls.lua
```diff
@@ -1,4 +1,5 @@
 ---@brief
+---@schema https://raw.githubusercontent.com/FractalBoy/perl-language-server/master/client/package.json
```
justinmk: is `@schema` a LuaLS/emmylua feature? should we mention it in the readme?

README.md
```diff
@@ -170,6 +170,102 @@ Most of the time, the reason for failure is present in the logs.
...
+    },
+  },
+}
+
+vim.lsp.config('lua_ls', config)
+```
+
+Following LSP server settings types are available for annotations:
+
+<!-- GENERATED:START -->
```
justinmk: do we need this?

README.md
```diff
@@ -170,6 +170,102 @@ Most of the time, the reason for failure is present in the logs.
 * `:lsp disable [<config_name>]` (`:LspStop` for Nvim 0.11 or older) Stops the given server. Defaults to stopping all servers active on the current buffer. To force stop use `:LspStop!`
 * `:lsp restart [<client_name>]` (`:LspRestart` for Nvim 0.11 or older) Restarts the given client, and attempts to reattach to all previously attached buffers. Defaults to restarting all active servers.
 
+## LSP Settings Type Annotations
+
+`nvim-lspconfig` generates Lua type definitions for each supported LSP server.
+By manually adding annotations (e.g., `---@type lspconfig.settings.server_name`),
```
justinmk, line 176: This is cool. Let's also document this in the help doc (perhaps in a new section `lspconfig-settings-types`, and can mention/link it in the USAGE section).

.github/workflows/gen-annotations.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Commit generated files
+        env:
+          COMMIT_MSG: |
+            chore: update generated types
+            skip-checks: true
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add README.md
+          git add -A schemas lua/lspconfig/types
```
justinmk: should we stick with the `_meta` convention from the core repo?

scripts/gen_types/schemas.lua
```diff
@@ -0,0 +1,193 @@
...
+      package_url = package_json,
+      settings_file = util.path('schemas/' .. server .. '.json'),
+    }
+  end
+  ret = vim.tbl_deep_extend('force', ret, M.overrides)
+
+  return ret
+end
+
+function M.translate(props, nls_url)
```
justinmk: docstring with a brief input/output example will help here

scripts/gen_types/schemas.lua
```diff
@@ -0,0 +1,193 @@
...
+        end
+        fixdoc(v)
+      end
+    end
+  end
+  fixdoc(props)
+end
+
+---@param schema LspSchema
+function M.fetch_schema(schema)
```
justinmk:
```suggestion
--- Gets ... from ...
---
---@param schema LspSchema
function M.fetch_schema(schema)
```

scripts/gen_types/schemas.lua
```diff
@@ -0,0 +1,193 @@
...
+end
+
+function M.clean()
+  ---@diagnostic disable-next-line: param-type-mismatch
+  for _, f in pairs(vim.fn.expand(util.path('schemas/*.json'), false, true)) do
+    vim.uv.fs_unlink(f)
+  end
+end
+
+function M.update_schemas()
```
justinmk:
```suggestion
--- Updates files in-place at /foo/bar
function M.update_schemas()
```

scripts/gen_types/util.lua
```diff
@@ -0,0 +1,33 @@
+local M = {}
+
+function M.path(str)
+  local f = debug.getinfo(1, 'S').source:sub(2)
+  f = vim.fn.fnamemodify(f, ':p')
+  local fname = vim.fn.fnamemodify(f, ':h:h:h') .. '/' .. (str or '')
+  return vim.uv.fs_realpath(fname) or fname
+end
+
+function M.read_file(file)
```
justinmk: `vim.fn.readblob()` ?

scripts/gen_types/util.lua
```diff
@@ -0,0 +1,33 @@
+local M = {}
+
+function M.path(str)
```
justinmk: need docstring with brief example input/output

scripts/gen_types/annotations.lua
```diff
@@ -0,0 +1,259 @@
...
+  local schema = Settings.new()
+  for key, prop in pairs(json.properties) do
+    prop.leaf = true
+    schema:set(key, prop)
+  end
+
+  M.process_object(M.class_name, M.fix_props(schema:get()))
+end
+
+function M.build()
```
justinmk: docstring with brief example input/output

lsp/perlpls.lua
ofseed: This is just something I defined myself so that we can annotate the JSON schema while adding a server. I may need to mention it in the contributing guide or somewhere similar.

However, [emmylua-analyzer-rust](https://github.com/EmmyLuaLs/emmylua-analyzer-rust) does have a `@schema` （https://github.com/EmmyLuaLs/emmylua-analyzer-rust/pull/952）field that functions like `@type`, using a remote JSON schema as the `@type`. But it seems to support a limited version of JSON schema. I haven't tried it yet.

README.md
ofseed: Its purpose is to list which LSP server settings are generated by this script; it is not strictly necessary, and maintaining it for that reason might be a bit of a hassle. I think I should delete it.

.github/workflows/gen-annotations.yml
ofseed: Since these generated types are unrelated to the repository itself, it might be better to give it a more meaningful name.

ofseed:
> The `lua/lspconfig/types/lsp.lua` file is big and will keep growing. Does it avoid a perf issue if we split it into a bunch of files per-config?

I haven't noticed any performance issues during my own use, but the problem is that the default `preloadFileSize` field in the Lua LS setting is `500` KB, while the `lsp.lua` type file generated by this script is around 750 KB, causing the file to be ignored by default. It might be worth splitting it up or updating the `preloadFileSize` setting in `lsp/lua_ls.lua`.

I was just about to apply `@type lspconfig.settings` to the settings under `lsp/`, but then I noticed that some of the generated results seem to be invalid, maybe because the provided link is invalid. I probably need to double-check them before requesting a review again.

lua/lspconfig/types/lsp/yamlls.lua
```diff
@@ -0,0 +1,105 @@
...
+-- Enable/disable validation feature
+-- 
+-- ```lua
+-- default = true
+-- ```
+---@field validate boolean
+
+---@class lspconfig.settings.yamlls
+---@field redhat _.lspconfig.settings.yamlls.Redhat
+---@field yaml _.lspconfig.settings.yamlls.Yaml
```
justinmk: can we ensure EOL at EOF? it's a minor thing but avoids unnecessary differences vs the rest of the codebase.

doc/lspconfig.txt
```diff
@@ -47,6 +47,8 @@ example to activate the "clangd" config: >lua
     })
     -- Enable filetype-based activation of the config.
     vim.lsp.enable('clangd')
+
+lspconfig provides type annotations for LSP settings, see |lspconfig-settings-types|.
 <
```
justinmk:
```suggestion
<

lspconfig provides type annotations for LSP settings, see |lspconfig-settings-types|.

```

.github/workflows/gen-annotations.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: rhysd/action-setup-vim@v1
+        with:
+          neovim: true
+          version: nightly
+      - name: Generate JSON schemas
+        run: |
+          nvim --headless -u NONE -i NONE -l scripts/gen_json_schemas.lua
+      - name: Generate annotations
+        run: |
+          nvim --headless -u NONE -i NONE -l scripts/gen_annotations.lua
```
justinmk: nit: `--headless -u NONE -i NONE` shouldn't be needed with `-l`

justinmk:
the codespell CI should ignore types/ since we don't control the spelling there. 
the other failing lints are presumably legitimate

.styluaignore
```diff
@@ -0,0 +1 @@
+lua/lspconfig/types/lsp
```
justinmk, line 1: why ignore for stylua? I guess because it might require running `stylua` on the generated output, though maybe not, if the generated output can conform to stylua expectations.

pro: this will help catch bugs in the generated output
con: if it complicates the generation logic then maybe not worth it

ofseed: Since the code generated by this script consists of comments, stylua has very little effect (it doesn't understand LuaCATS). The CI errors are caused by trailing spaces within these comments. While there is indeed some logic in the script that causes trailing spaces, I discovered after fixing it that the descriptions in the JSON schemas provided by these LSP servers contain trailing spaces themselves. I think it might be better to just ignore them?

justinmk: Fair enough. Related but different, it looks like LuaLS is flagging `Duplicate defined fields` ... The diff does indeed appear to have duplicate `_.lspconfig.settings.vtsls.Format` definitions for example. So is this CI check useful ?

```
+---@class _.lspconfig.settings.vtsls.Format
+---@field baseIndentSize? number
+---@field convertTabsToSpaces? boolean
+---@field indentSize? number
+---0: None 1: Block 2: Smart
+---@field indentStyle? number
+---@field newLineCharacter? string
+---@field tabSize? number
+---@field trimTrailingWhitespace? boolean
+
+---@class _.lspconfig.settings.vtsls.Javascript
+---@field format? _.lspconfig.settings.vtsls.Format
+
+---@class _.lspconfig.settings.vtsls.Tsserver
+---TypeScript plugins that are not locally avaiable in the workspace. Usually the plugin configuration can be found in the `contributes.typescriptServerPlugins` field of `package.json` of the corresponding VSCode extension.
+---
+---```lua
+---default = {}
+---```
+---@field globalPlugins? object[]
+
+---@class _.lspconfig.settings.vtsls.Format
+---@field baseIndentSize? number
+---@field convertTabsToSpaces? boolean
+---@field indentSize? number
+---0: None 1: Block 2: Smart
+---@field indentStyle? number
+---@field newLineCharacter? string
+---@field tabSize? number
+---@field trimTrailingWhitespace? boolean
+
+---@class _.lspconfig.settings.vtsls.Typescript
+---@field format? _.lspconfig.settings.vtsls.Format
+---@field globalTsdk? string
```

ofseed: Yes, Lua LS identified several issues already present in the original script (for example, the `object[]` in the code you showed, which clearly doesn't exist). I am currently fixing them. The Lua LS checks are valuable.

scripts/gen_json_schemas.lua
```diff
@@ -0,0 +1,247 @@
...
+  for _, name in ipairs(names) do
+    local schema_config = schemas[name]
+    print(('Generating schema for %s'):format(name))
+
+    if not vim.uv.fs_stat(schema_config.settings_file) then
+      local ok, schema_json = pcall(generate_server_schema, schema_config)
+      if ok then
+        vim.fn.mkdir(vim.fn.fnamemodify(schema_config.settings_file, ':h'), 'p')
+        vim.fn.writefile(
+          vim.split(vim.json.encode(schema_json, { indent = '  ', sort_keys = true }), '\n', { plain = true }),
```
justinmk, line 238: this reminds me we need https://github.com/neovim/neovim/pull/31930 to make `writefile()` less annoying :)

scripts/gen_json_schemas.lua
```diff
@@ -0,0 +1,247 @@
...
+    translate_schema_docs(schema_json, schema.package_url:gsub('package%.json$', 'package.nls.json'))
+  end
+
+  return schema_json
+end
+
+---Regenerates all schema files under the local `schemas/` directory.
+local function generate_all_schemas()
+  ---@diagnostic disable-next-line: param-type-mismatch
+  vim.fn.delete(vim.fs.joinpath(vim.uv.cwd(), 'schemas'), 'rf')
```
justinmk, line 224: just fyi, we have `vim.fs.rm()` now too

ofseed:
I just noticed that I lost the `Co-authored-by` in the commit during the final rebase; thanks to @folke for providing the original version!

PR #4308, opened 2026-02-05
docs: fix the vala_ls link

cnjhb:
vala_ls is now at https://github.com/vala-lang/vala-language-server


justinmk:
the docs are generated.

edit needs to be made here: https://github.com/neovim/nvim-lspconfig/blob/0480b120318ec8bab27b530ffab6ee76a8c4d806/lsp/vala_ls.lua#L3

PR #4327, opened 2026-02-26
fix: drop invalid filetypes from legacy configs

barrettruth:
## Problem

78596b61 removed `javascript.jsx` and `typescript.tsx` from the `lsp/` configs but these same invalid filetypes remain in the legacy `lua/lspconfig/configs/` files. Annoyingly these still surface in `checkhealth vim.lsp` warnings when _not_ using the old configs:

image

## Solution

Remove `javascript.jsx` and `typescript.tsx` from all legacy config filetype lists from all affected LSPs.

***NOTE***: the CI will NOT pass because it checks that old LSP configs are NOT modified. Of course, this PR will not pass that check.

I argue that these changes _should_ be affected because default opt-out functionality (using old configs) ought _not_ affected your "correct" setup. Waiting for the old configs to be removed will take too long and needlessly pollute nearly everyone's `:checkhealth` in the meantime.

### DIFF
--- a/lua/lspconfig/configs/angularls.lua
+++ b/lua/lspconfig/configs/angularls.lua
@@ -56,7 +56,7 @@ return {
       '--angularCoreVersion',
       default_angular_core_version,
     },
-    filetypes = { 'typescript', 'html', 'typescriptreact', 'typescript.tsx', 'htmlangular' },
+    filetypes = { 'typescript', 'html', 'typescriptreact', 'htmlangular' },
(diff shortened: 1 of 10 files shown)


justinmk:
> Annoyingly these still surface in `checkhealth vim.lsp` warnings when _not_ using the old configs:

um... that is either a bug in the healthcheck, or something in your config is touching/loading the configs somehow.

We can merge this meanwhile though.

PR #4329, opened 2026-02-27
fix(eslint): don't implicitly enable `experimental.useFlatConfig`

### LINKED ISSUE #4318: vscode-eslint-language-server breaks with Eslint 10 (with `useFlatConfig` setting)
### Description

Hi

```
NVIM v0.12.0-nightly+16495e6
Build type: Release
LuaJIT 2.1.1741730670
``` 

And in https://github.com/microsoft/vscode-eslint/blob/6d61063d2f7361e728a1968d7a3fd4edb41acd16/server/src/eslint.ts#L925

```typescript
const eslintPath = settings.experimental?.useFlatConfig ? 'eslint/use-at-your-own-risk' : 'eslint';
```
In https://github.com/neovim/nvim-lspconfig/blob/44acfe887d4056f704ccc4f17513ed41c9e2b2e6/lsp/eslint.lua#L196

We always set the config to true if there is a flat config file (like `eslint.config.js`)

In [Flat config rollout plans - ESLint - Pluggable JavaScript Linter](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0) (emphasis mine):

> For API users, you’ll need to stop using the classes from the /use-at-your-own-risk entrypoint. The API in v10.0.0 will change in the following ways:
>
>    The ESLint class will now be the only way to interact with the core.
>    The LegacyESLint class will be removed.
>    **The FlatESLint class will be removed.**
>    The shouldUseFlatConfig() method will always return true.
>    The Linter class configType constructor option will be invalid and throw an error.

See [fix: remove fake `FlatESLint` and `LegacyESLint` exports by fasttime · Pull Request #20460 · eslint/eslint](https://github.com/eslint/eslint/pull/20460)

The following error occurs in the `nvim/lsp.log`:

```log
[START][2026-02-18 13:18:09] LSP logging initiated
[ERROR][2026-02-18 13:18:09] ...unwrapped-16495e6/share/nvim/runtime/lua/vim/lsp/log.lua:151	"The eslint library loaded from /home/neuromante/code/js/eslint10-test/node_modules/eslint/lib/unsupported-api.js doesn't export a FlatESLint class."
```

I think removing the setting is not enough, this will break eslint < 10 users using the flat config.

Can we somehow check the eslint version before toggling the flag or should we handle this in our own configs?

Thanks

faergeek:
Fixes #4318

Problem: `experimental.useFlatConfig` forcefully turns on the logic to use a separate programmatic interface for flat config usage in ESLint. That interface was removed in ESLint v10 where flat config format is the only supported config format. But when the setting is turned on the language server still looks for that interface and fails when it can't find it. The end result is that language server turns off it's validation/diagnostics when used with ESLint v10.

Solution: do not turn on `experimental.useFlatConfig` setting based on the presence of flat config files. Instead use the default behavior of the language server which matches the ESLint's default behavior and let the user explicitly override it via config when necessary.

### DIFF
--- a/lsp/eslint.lua
+++ b/lsp/eslint.lua
@@ -38,6 +38,24 @@
 --- You can use a different version of ESLint in each package, but it is recommended to use the same version of ESLint in all packages. The location of the ESLint binary will be determined automatically.
 ---
 --- /!\ When using flat config files, you need to use them across all your packages in your monorepo, as it's a global setting for the server.
+---
+--- ### Flat config in ESLint versions prior to 10.0
+---
+--- If you're using a ESLint version that supports both flat config and eslintrc (>= 8.21, < 10.0) and want to change
+--- the [default behavior](https://eslint.org/blog/2023/10/flat-config-rollout-plans/), you'll need to set
+--- `experimental.useFlatConfig` accordingly:
+--- ```lua
+--- vim.lsp.config("eslint", {
+---   settings = {
+---     experimental = {
+---       -- If you want to use flat config on >= 8.21, < 9.0
+---       useFlatConfig = true,
+---       -- Or if you want to use eslintrc on 9.*
+---       -- useFlatConfig = false,
+---     }
+---   }
+--- })
+--- ```
 
 local util = require 'lspconfig.util'
 local lsp = vim.lsp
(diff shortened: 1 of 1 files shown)


lsp/eslint.lua
```diff
@@ -129,9 +147,7 @@ return {
     ---@diagnostic disable-next-line: assign-type-mismatch
     packageManager = nil,
     useESLintClass = false,
-    experimental = {
-      useFlatConfig = false,
-    },
+    experimental = {},
```
faergeek, line 150: This still needs to be present as the current published version (~2 years ago) of `vscode-langservers-extracted` assumes it exists and fails otherwise while trying to get the value inside it. Current version of `vscode-eslint` doesn't seem to have this issue.

PR #4334, opened 2026-03-04
feat(roslyn_ls): use official roslyn binary name as fallback

616b2f:
As mentioned here (https://github.com/mason-org/mason-registry/pull/6330#issuecomment-3971724747), there is a change (https://github.com/dotnet/roslyn/pull/82201/changes) that the official language server can be easily installed via `dotnet` cli.

This change uses it's binary name per default.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -113,7 +113,8 @@ return {
   name = 'roslyn_ls',
   offset_encoding = 'utf-8',
   cmd = {
-    vim.fn.executable('Microsoft.CodeAnalysis.LanguageServer') == 1 and 'Microsoft.CodeAnalysis.LanguageServer' or 'roslyn-language-server',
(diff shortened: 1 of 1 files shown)


lsp/roslyn_ls.lua
```diff
@@ -113,7 +113,8 @@ return {
   name = 'roslyn_ls',
   offset_encoding = 'utf-8',
   cmd = {
-    'Microsoft.CodeAnalysis.LanguageServer',
+    vim.fn.executable('Microsoft.CodeAnalysis.LanguageServer') == 1 and 'Microsoft.CodeAnalysis.LanguageServer'
+      or 'roslyn-language-server',
```
justinmk, line 117: should we prefer `roslyn-language-server` ? this change currently will prefer `Microsoft.CodeAnalysis.LanguageServer`

616b2f: My thoughts where this should behave non breaking, should anyone already using it, for all other it should not matter much in both cases, this is my personal assumption

PR #4338, opened 2026-03-08
fix(rust_analyzer): recognize std lib sources in sysroot

### LINKED ISSUE #4337: rust_analyzer: going to type definition in standard library results in failure
### Description

Erroneously first reported in rust analyzer: https://github.com/rust-lang/rust-analyzer/issues/21737 but it turned to be an issue with nvim-lspconfig instead.

Using system installed Rust and trying to navigate to some standard type like `String` results in following error:
```
[rust_analyzer] cmd failed with code 101: table: 0x4485228447d8                                                                                                                                                                                                
error: failed to parse manifest at `/usr/lib/rustlib/src/rust/library/Cargo.toml`                                                                                                                                                                              
                                                                                                                                                                                                                                                               
Caused by:                                                                                                                                                                                                                                                     
  the cargo feature `profile-rustflags` requires a nightly version of Cargo, but this is the `stable` channel                                                                                                                                                  
  See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.                                                                                                                                           
  See https://doc.rust-lang.org/cargo/reference/unstable.html#profile-rustflags-option for more information about using this feature. 
```
and library sources are not parsed. That's because [is_library()](https://github.com/neovim/nvim-lspconfig/blob/2163c54bb6cfec53e3e555665ada945b8c8331b9/lsp/rust_analyzer.lua#L38-L53) does not consider `/usr/lib/rustlib/src/rust/library` as directory with library sources and attempt to determine workspace root:
https://github.com/neovim/nvim-lspconfig/blob/2163c54bb6cfec53e3e555665ada945b8c8331b9/lsp/rust_analyzer.lua#L78-L100
fails because even stable Rust versions use unstable cargo features in library Cargo.toml files

jpalus:
Problem:
System installed sources of Rust standard library are not detected as library sources.

Solution:
Add sources location to paths used for library check. See also: https://rust-analyzer.github.io/book/configuration.html#cargo.sysrootSrc

Fixes #4337

### DIFF
--- a/lsp/rust_analyzer.lua
+++ b/lsp/rust_analyzer.lua
@@ -35,6 +35,33 @@ local function reload_workspace(bufnr)
   end
 end
 
+local function user_sysroot_src()
+  return vim.tbl_get(vim.lsp.config['rust_analyzer'], 'settings', 'rust-analyzer', 'cargo', 'sysrootSrc')
+end
+
+local function default_sysroot_src()
+  local sysroot = vim.tbl_get(vim.lsp.config['rust_analyzer'], 'settings', 'rust-analyzer', 'cargo', 'sysroot')
+  if not sysroot then
+    local rustc = os.getenv 'RUSTC' or 'rustc'
+    local result = vim.system({ rustc, '--print', 'sysroot' }, { text = true }):wait()
+
+    local stdout = result.stdout
+    if result.code == 0 and stdout then
+      if string.sub(stdout, #stdout) == '\n' then
+        if #stdout > 1 then
+          sysroot = string.sub(stdout, 1, #stdout - 1)
+        else
+          sysroot = ''
+        end
+      else
+        sysroot = stdout
+      end
+    end
+  end
+
+  return sysroot and vim.fs.joinpath(sysroot, 'lib/rustlib/src/rust/library') or nil
+end
+
 local function is_library(fname)
   local user_home = vim.fs.normalize(vim.env.HOME)
   local cargo_home = os.getenv 'CARGO_HOME' or user_home .. '/.cargo'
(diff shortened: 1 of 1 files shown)


lsp/rust_analyzer.lua
```diff
@@ -35,6 +35,33 @@ local function reload_workspace(bufnr)
   end
 end
 
+local function user_sysroot_src()
+  return vim.tbl_get(vim.lsp.config['rust_analyzer'], 'settings', 'rust-analyzer', 'cargo', 'sysrootSrc')
+end
+
+local function default_sysroot_src()
+  local sysroot = vim.tbl_get(vim.lsp.config['rust_analyzer'], 'settings', 'rust-analyzer', 'cargo', 'sysroot')
```
justinmk, line 43: does this depend on the user's config `settings` item? that seems rather niche.

this PR really needs a clearer explanation, and possibly doc updates if relevant.

it almost seems like this is reinventing things that cargo or rust should be doing for us. cc @mikejmcguirk @igorlfs

justinmk:
from https://github.com/rust-lang/rust-analyzer/issues/21737 

>  It's probably not a good idea to try to 'automatically' detect the workspace without user input like Nvim seems to do.

so... does this PR actually address that?

lsp/rust_analyzer.lua
jpalus: rust-analyzer decides where to look for std lib sources either from:
1. User configured location passed in initialize request (InitializeParams.initializationOptions) which originates from `settings`
2. Uses default location `{cargo.sysroot}/lib/rustlib/src/rust/library` but same as above `cargo.sysroot` might be passed by user based on `settings`

So yes it might depend on user's config `settings` item.

jpalus:
> from [rust-lang/rust-analyzer#21737](https://github.com/rust-lang/rust-analyzer/issues/21737)
> 
> > It's probably not a good idea to try to 'automatically' detect the workspace without user input like Nvim seems to do.
> 
> so... does this PR actually address that?

I'm not going to argue with the decision made for rust-analyzer in lspconfig to find workspace root for user's projects. I actually think it's a good call. What it shouldn't do is to treat library sources as independent project. It should treat it as library and reuse existing lsp client. In order to do so detection logic needs to be fixed to recognize library sources in sysroot as library and that's what this PR tries to address.

mikejmcguirk:
> from [rust-lang/rust-analyzer#21737](https://github.com/rust-lang/rust-analyzer/issues/21737)
> 
> > It's probably not a good idea to try to 'automatically' detect the workspace without user input like Nvim seems to do.
> 
> so... does this PR actually address that?

It does. By adding sysroot to the library detection, it effectively makes Neovim understand that the user is trying to go to definition.

roslyn_ls uses a similar pattern:

https://github.com/neovim/nvim-lspconfig/blob/faa403f95fd82cc24ebd0cee6af41b0de2dae00c/lsp/roslyn_ls.lua#L157

PR #4341, opened 2026-03-11
feat: mpls

tahv:
This PR adds the [mpls](https://github.com/mhersson/mpls) (Markdown Preview Language Server) config.
The project meet the criteria for a new config with 180+ GitHub stars.

Regarding the config itself:

- The `LspMplsOpenPreview` command open the Markdown preview in the default browser.
- The autocmd ensures the preview focus the correct buffer when entering a new buffer.

### DIFF
--- a/lsp/mpls.lua
+++ b/lsp/mpls.lua
@@ -21,6 +21,7 @@ return {
       pattern = { '*.md' },
       group = vim.api.nvim_create_augroup('lspconfig.mpls.focus', { clear = true }),
       callback = function(ctx)
+        ---@diagnostic disable-next-line:param-type-mismatch
         client:notify('mpls/editorDidChangeFocus', { uri = ctx.match })
       end,
       desc = 'mpls: notify buffer focus changed',


lsp/mpls.lua
```diff
@@ -0,0 +1,36 @@
...
+  },
+  root_markers = { '.marksman.toml', '.git' },
+  filetypes = { 'markdown' },
+  on_attach = function(client, bufnr)
+    vim.api.nvim_create_autocmd('BufEnter', {
+      pattern = { '*.md' },
+      group = vim.api.nvim_create_augroup('lspconfig.mpls.focus', { clear = true }),
+      callback = function(ctx)
+        ---@diagnostic disable-next-line:param-type-mismatch
+        client:notify('mpls/editorDidChangeFocus', { uri = ctx.match })
```
justinmk, line 25: seems strange that this is needed. Doesn't LSP send enough events already, so the server knows what document is focused?

tahv: I *think* this has to do with this `mpls` feature:

> By default, all files update in the same browser window (single-page mode). With `--tabs`, each file opens in its own browser tab with a unique URL. In single-page mode, link clicks update the preview; in multi-tab mode, they open new tabs.

That's not something i have used myself, but i can raise it with the project owner.

PR #4342, opened 2026-03-11
fix(roslyn_ls): navigation in macos decompiled files

LukeElrod:
Fixes LSP navigation in decompiled files for systems with symlinked TMPDIR (macOS)


justinmk:
ci is failing

also, this seems like a bug in the LS ... though i guess it's harmless...

seblyng:
https://github.com/seblyng/roslyn.nvim/issues/297

Should check if TMPDIR is set first.

> also, this seems like a bug in the LS ...

Yeah it's not the first/only one😕

PR #4343, opened 2026-03-13
docs: json schema for "settings"

justinmk:
Document how to maintain the "settings" schemas.

@ofseed can you double check this


ofseed:
I eventually reverted the approach of using `@schema` to label the JSON Schema address in each individual configuration because some JSON Schemas are not standalone files; instead, they are written directly within `package.json` (such as [zls](https://raw.githubusercontent.com/microsoft/vscode/main/extensions/typescript-language-features/package.json)). This `package.json` follows the standard Node.js format, and VS Code uses a custom `contributes.configuration` field to define the JSON Schema for `settings.json`. I suspect this nuance might be a bit difficult to describe.

justinmk:
> This `package.json` follows the standard Node.js format, and VS Code uses a custom `contributes.configuration` field to define the JSON Schema for `settings.json`. I suspect this nuance might be a bit difficult to describe.

If the script has these expectations currently, we need to mention it somehow in the docs.

We need to update `index` for new configs, right?

ofseed:
Yes, that is how it currently works. There might be more natural approaches, such as a type annotation that is more appropriate than the previously envisioned `@schema`,  because it might be more natural to place each URL within each LSP server's respective file.

PR #4344, opened 2026-03-14
feat(svelte): enable inlay hints

igorlfs:


igorlfs:
Not sure how the generated annotations slipped in, but I guess that's not a big deal

PR #4345, opened 2026-03-14
fix: skip buffers with URI schemes in root_dir

niklaas:
## Problem

Some LSP configs implement a custom `root_dir` function that unconditionally
calls `on_dir(project_root or vim.fn.getcwd())`. When no project root is found,
this falls back to `getcwd()`, causing the LSP to attach to _any_ buffer of a
matching filetype including buffers with non-file URI schemes such as
`fugitive://` (vim-fugitive).

Once attached, the LSP sends requests using the `fugitive://` URI, which is not
a valid `file://` URI and causes errors.

The affected configs are `ts_ls`, `tsgo`, `vtsls`, and `matlab_ls`.

Configs that use `root_markers` only (e.g. `lua_ls`, `ruby_lsp`) are naturally
safe because `vim.fs.root()` returns `nil` for non-file buffers, so `on_dir` is
never called.

## Solution

Add a guard at the top of each affected `root_dir` function that returns early
when the buffer name starts with a URI scheme:

```lua
if vim.api.nvim_buf_get_name(bufnr):match('^%a+://') then
    return
end
```

From my understanding, neovim buffer names are always plain paths (e.g.
/home/user/file.ts) for real files, never file:// URIs, so this guard only
skips genuinely non-file buffers.

I don't think the early returns trigger false positives but for sure let me know if they do.

### DIFF
--- a/lsp/matlab_ls.lua
+++ b/lsp/matlab_ls.lua
@@ -20,6 +20,9 @@ return {
   cmd = { 'matlab-language-server', '--stdio' },
   filetypes = { 'matlab' },
   root_dir = function(bufnr, on_dir)
+    if vim.api.nvim_buf_get_name(bufnr):match('^%a+://') then
+      return
+    end
     local root_dir = vim.fs.root(bufnr, '.git')
     on_dir(root_dir or vim.fn.getcwd())
   end,
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -84,6 +84,9 @@ return {
     'typescriptreact',
   },
   root_dir = function(bufnr, on_dir)
+    if vim.api.nvim_buf_get_name(bufnr):match('^%a+://') then
+      return
+    end
     -- The project root is where the LSP can be started from
     -- As stated in the documentation above, this LSP supports monorepos and simple projects.
     -- We select then from the project root, which is identified by the presence of a package
--- a/lsp/tsgo.lua
+++ b/lsp/tsgo.lua
@@ -61,6 +61,9 @@ return {
     'typescriptreact',
   },
   root_dir = function(bufnr, on_dir)
+    if vim.api.nvim_buf_get_name(bufnr):match('^%a+://') then
+      return
+    end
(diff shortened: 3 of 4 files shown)


lsp/matlab_ls.lua
```diff
@@ -20,6 +20,9 @@ return {
   cmd = { 'matlab-language-server', '--stdio' },
   filetypes = { 'matlab' },
   root_dir = function(bufnr, on_dir)
+    if vim.api.nvim_buf_get_name(bufnr):match('^%a+://') then
+      return
+    end
     local root_dir = vim.fs.root(bufnr, '.git')
     on_dir(root_dir or vim.fn.getcwd())
```
justinmk, line 27: Thanks for the clear explanation. But I don't quite see why the solution was chosen. 

I think it's more like a hint that `getcwd()` should not be used as a fallback. People really seem to love it though ...

justinmk:
> From my understanding, neovim buffer names are always plain paths (e.g.
> /home/user/file.ts) for real files, never file:// URIs

That may change in the future though.

I guess you could future-proof your current PR by only skipping if `is_uri and root() == nil`

lsp/matlab_ls.lua
niklaas: > I think it's more like a hint that `getcwd()` should not be used as a fallback.

Good catch, I agree. I guess I subconsciously decided to use a guard to avoid introducing a breaking change.

However, I'm happy to remove calling `getcwd()`. As you indicate, arguably, the issue I ran into demonstrates the function isn't the proper approach to choose there because it's brittle.

Let me know if you'd be willing to introduce a breaking change.

igorlfs:
Using `vim.uv.fs_stat` might be preferable here? See https://github.com/neovim/nvim-lspconfig/pull/3899

lsp/matlab_ls.lua
niklaas: @justinmk just saw you closed this one. I was waiting for your input regarding the breaking change. I'm happy to pick it up again.

PR #4349, opened 2026-03-17
New versions of nextflow_ls moving from nextflow-language-server-all.jar to nextflow-language-server

toniher:
They recently moved to use a `nextflow-language-server` (actually a wrapper that runs a jar). This way, it ensures lsp continues with minimal user intervention. /cc https://github.com/nextflow-io/language-server/issues/56


toniher:
Following checks, I didn't change neither `configs` files nor  `lua/lspconfig/configs/nextflow_ls.lua`

PR #4352, opened 2026-03-19
Change tab width from 8 to 2 spaces

cxwx:
for lsp/*.lua the tab should be set to 2
if I use indent highlight plugin

set tabstop = 8 cause

截屏2026-03-19 18 19 27


justinmk:
the current setting is correct. tab_width != indent_size. and there are no tabs in this codebase anyway.

PR #4355, opened 2026-03-20
fix: prevent file leak in `root_markers_with_field()`

uss-stargazer:
Problem:
Files related to language servers would stay locked (for me, `package.json` would be locked by the `tailwindcss` language server until it was killed).

Solution:
I'm pretty sure there is a file leak in the `root_markers_with_field()` function because the `io:lines` loop is broken if the field is found and the file is not closed properly. Explicitly closing the file fixed it for me...


justinmk:
Yikes, good catch. There is also `vim.fn.readblob()` in recentish Nvim versions.

PR #4356, opened 2026-03-25
fix: Set biome root for projects without package managers

Nukesor:
Heyo,

First up, thanks for lspconfig :). Much appreciated!

The current biome project root logic uses JS package manager lockfiles to get the project root in mono repos.

If one is not using any JS (wasm) and raw CSS, that doesn't work out. In those cases, only a biome.json exists as a marker on where the frontend's project root inside the mono repo may be located.

Cheers

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -50,6 +50,8 @@ return {
       'bun.lockb',
       'bun.lock',
       'deno.lock',
+      'biome.json',
+      'biome.jsonc',
     }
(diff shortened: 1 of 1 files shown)


igorlfs:
Hello,

Could these be assigned a lower priority? i.e., by wrapping them into their own table.

As is, this change ends up eagerly finding the root dir inside a monorepo, leading to multiple instances being spawned (one for each package)

justinmk:
sgtm

PR #4358, opened 2026-03-27
fix(roslyn_ls): add roslyn.client.fixAllCodeAction command handler

khoido2003:
## Problem

When a `roslyn.client.fixAllCodeAction` command is sent by the Roslyn
language server, Neovim would display a warning:

    Language server `roslyn_ls` does not support command
    `roslyn.client.fixAllCodeAction`. This command may require a
    client extension.

This is because nvim-lspconfig had no handler registered for this command.

## Solution

Add a `roslyn.client.fixAllCodeAction` command handler that:

1. Presents the user with a `vim.ui.select` scope picker populated from
   the `FixAllFlavors` field sent by the server (`Document`, `Project`,
   `Solution`, `ContainingMember`, `ContainingType`).
2. Calls Roslyn's custom `codeAction/resolveFixAll` endpoint (not the
   standard `codeAction/resolve`) with the original action data untouched
   and the chosen scope as a top-level `scope` field, which is what the
   server's `CodeActionFixAllResolveHandler` expects.
3. Applies the resulting `workspace/edit` via `vim.lsp.util.apply_workspace_edit`.

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -108,6 +108,59 @@ local function is_decompiled(bufname)
   return vim.fn.finddir(bufname:sub(1, endpos), uv.os_tmpdir()) ~= ''
 end
 
+---@param client vim.lsp.Client
+---@param action table
+local function apply_action(client, action)
+  if action.edit then
+    vim.lsp.util.apply_workspace_edit(action.edit, client.offset_encoding)
+  end
+  if action.command then
+    client:exec_cmd(action.command)
+  end
+end
+
+---@param client vim.lsp.Client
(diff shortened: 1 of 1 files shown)


justinmk:
can the code be deduplicated? i assume these handlers in this config are all doing similar stuff, and the config is getting big.

khoido2003:
Hi, thanks for the feedback. I have addressed the duplication concern in the
follow-up commit. Not sure if I can do anything better than this. I will make further adjustments if needed.

justinmk:
I think it could be further deduplicated, but it's a start. Thanks

PR #4332, opened 2026-03-01
feat: phpantom_lsp

calebdw:
Hello!

This adds support for phpantom_lsp, the new rust lsp for PHP

See: https://github.com/AJenbo/phpantom_lsp

Thanks!

CC: @ajenbo

PR #4362, opened 2026-03-27
feat: phpantom_lsp

calebdw:
Supercedes #4332

This adds support for phpantom_lsp, the new rust lsp for PHP

See: https://github.com/AJenbo/phpantom_lsp

Thanks!

CC: @ajenbo


AJenbo:
image

PR #4363, opened 2026-03-27
fix(biome): set lower priority for config files as root

igorlfs:
Addresses https://github.com/neovim/nvim-lspconfig/pull/4356#issuecomment-4139528898

### DIFF
--- a/lsp/biome.lua
+++ b/lsp/biome.lua
@@ -50,12 +50,12 @@ return {
       'bun.lockb',
       'bun.lock',
       'deno.lock',
-      'biome.json',
-      'biome.jsonc',
     }
+    -- Set a lower priority to avoid spawning multiple servers on monorepos
+    local biome_config_files = { 'biome.json', 'biome.jsonc' }
     -- Give the root markers equal priority by wrapping them in a table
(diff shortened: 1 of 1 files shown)


lsp/biome.lua
```diff
@@ -50,20 +50,18 @@ return {
...
       'deno.lock',
-      'biome.json',
-      'biome.jsonc',
     }
+    local biome_config_files = { 'biome.json', 'biome.jsonc' }
     -- Give the root markers equal priority by wrapping them in a table
-    root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers, { '.git' } }
-      or vim.list_extend(root_markers, { '.git' })
+    root_markers = vim.fn.has('nvim-0.11.3') == 1 and { root_markers, biome_config_files, { '.git' } }
+      or vim.list_extend(root_markers, vim.list_extend(biome_config_files, { '.git' }))
```
igorlfs, line 58: This feels dumb, is there a more clever approach?

justinmk: probably flatten() would be more clever. we're going to drop 0.11.x support eventually, so it doesn't matter much.

PR #4364, opened 2026-03-30
docs: document `vim.pack` as preferred installation method

arkriny:
* docs: document `vim.pack` as preferred installation method
  Nvim 0.12 is not nightly anymore.

* docs: replace deprecated `set_log_level` with `vim.lsp.log.set_level`

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,15 @@ These configs are **best-effort and supported by the community (you).** See [con
 
 * Requires Nvim 0.11.3+.
     * Support for Nvim 0.10 [will be removed](https://github.com/neovim/nvim-lspconfig/issues/3693). Upgrade Nvim and nvim-lspconfig before reporting an issue.
-* Install nvim-lspconfig using Vim's "packages" feature:
-  ```
-  git clone https://github.com/neovim/nvim-lspconfig ~/.config/nvim/pack/nvim/start/nvim-lspconfig
-  ```
-* Or with Nvim 0.12 (nightly), you can use the builtin `vim.pack` plugin manager:
+* With Nvim 0.12+, you can use the builtin `vim.pack` plugin manager:
   ```lua
   vim.pack.add{
     { src = 'https://github.com/neovim/nvim-lspconfig' },
   }
+  ```
+* Or install nvim-lspconfig using Vim's "packages" feature:
+  ```
+  git clone https://github.com/neovim/nvim-lspconfig ~/.config/nvim/pack/nvim/start/nvim-lspconfig
   ```
 * Or use a 3rd-party plugin manager.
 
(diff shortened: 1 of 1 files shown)


debeetle:
I meet a problem that using `vim.pack` will put lspconfig under `~/.local/share/nvim/site/pack/core/opt` by default. It seems that the lazy directory will prevent the `vim.lsp.enable()` form working well.

justinmk:
@debeetle that is obviously not true since I and many others use vim.pack to install nvim-lspconfig.

debeetle:
After extensive use and testing, I believe I have gained a certain understanding of this issue, and therefore I have decided to provide some clarification:

1. First, `vim.pack.add` does indeed install packages to the `opt` path by default; this is by design. In `init.lua`, I use `vim.pack.add()` to install the `nvim-lspconfig` plugin and configure `vim.lsp.enable()` for various file formats separately in the `ftplugin` path because I don't want to enable all LSP configurations indiscriminately for all file formats. The problem I encountered was that `vim.lsp.enable()` was ineffective. Using the command `:checkhealth vim.lsp`, I checked and found no Active Clients. However, when I placed the `nvim-lspconfig` package in the `start` path, the LSP started normally, and the defined LSPs appeared in Active Clients.

2. I carefully read Evgeni Chasnovski's [article](https://echasnovski.com/blog/2026-03-13-a-guide-to-vim-pack) on `vim.pack` usage examples and found that the configuration for `nvim-lspconfig` always follows `vim.lsp.enable()`. I tried adding the `vim.pack.add` command to the configuration under the ftplugin path (like in the article, but not in `init.lua`), and the `lsp` was listed in Active Clients and worked correctly.

I don't understand why separating it in my current configuration would cause `vim.lsp.enable()` to be ineffective. Am I doing something wrong? Any suggestions will be appreciated.

arkriny:
> and configure `vim.lsp.enable()` for various file formats separately in the `ftplugin` path because I don't want to enable all LSP configurations indiscriminately for all file formats. The problem I encountered was that `vim.lsp.enable()` was ineffective.

I suspect that placing the vim.lsp.enable() call inside ftplugin prevents the LSP from attaching because it uses a FileType autocmd internally. That autocmd has to be created before the file is opened, but in your case it is only created after the file is already open and ftplugin has been executed.

It is fine to call vim.lsp.enable() for all servers in init.lua, because it only enables automatic attachment based on filetype - it does not start all servers indiscriminately.

debeetle:
> I suspect that placing the vim.lsp.enable() call inside ftplugin prevents the LSP from attaching because it uses a FileType autocmd internally. That autocmd has to be created before the file is opened, but in your case it is only created after the file is already open and ftplugin has been executed.
> 
I'm aware of this situation, but when I use `vim.lsp.enable()` in `init.lua`, no lsp is listed on `Active Clients`.

justinmk:
sounds like an ordering issue. `vim.pack.add()` adds the `opt/` plugins to the runtimepath. if you are calling `vim.lsp.enable()` before that, it won't find them

PR #4367, opened 2026-04-01
feat(robotcode): add virtual env support

Methapon2001:
This will allow `robotcode` to solve library inside python virtual env.

### DIFF
--- a/lsp/robotcode.lua
+++ b/lsp/robotcode.lua
@@ -11,7 +11,7 @@ return {
   cmd = { 'robotcode', 'language-server' },
   filetypes = { 'robot', 'resource' },
   root_markers = { 'robot.toml', 'pyproject.toml', 'Pipfile', '.git' },
-  cmd_env = venv and { PYTHONPATH = vim.fn.glob(venv .. '/lib/python*/site-packages') } or nil,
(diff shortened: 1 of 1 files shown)


lsp/robotcode.lua
```diff
@@ -4,11 +4,14 @@
...
 --- RobotCode - Language Server Protocol implementation for Robot Framework.
 
+local venv = os.getenv('VIRTUAL_ENV')
+
 ---@type vim.lsp.Config
 return {
   cmd = { 'robotcode', 'language-server' },
   filetypes = { 'robot', 'resource' },
   root_markers = { 'robot.toml', 'pyproject.toml', 'Pipfile', '.git' },
+  cmd_env = venv and { PYTHONPATH = vim.fn.glob(venv .. '/lib/python*/site-packages') } or nil,
```
justinmk: is glob() guaranteed to return a single filename ?

Methapon2001: I've checked and it does not return only single entry but it return multiple entries separated by newline.
I'll push a fix to replace new line with colon so that it can resolve library in side all path found.

justinmk:
okie dokie

Issue #4370, opened 2026-04-03
tailwindcss relies on single quotes


besserwisser:
### Description

The tailwindcss configuration relies on a string that has a single quote: 
`local target = "@import 'tailwindcss';"`

Due to this restriction files with the strings `@import "tailwindcss";` are not found and you get errors like: 

```
tailwindcss: -32603 Request textDocument/completion failed with message: Cannot read properties of undefined (reading 'length')
```

Is there an easy fix to search for single and double quotes? 

Thank you!

https://github.com/neovim/nvim-lspconfig/blob/master/lsp/tailwindcss.lua

besserwisser:
I opened a PR to fix it
https://github.com/neovim/nvim-lspconfig/pull/4371

PR #4371, opened 2026-04-03
Update tailwindcss.lua to support multiple import patterns

besserwisser:
fix https://github.com/neovim/nvim-lspconfig/issues/4370

### DIFF
--- a/lsp/tailwindcss.lua
+++ b/lsp/tailwindcss.lua
@@ -7,8 +7,7 @@
 local util = require 'lspconfig.util'
 
 local function find_tailwind_global_css()
-  local target = "@import 'tailwindcss';"
-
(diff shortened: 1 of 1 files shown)


lsp/tailwindcss.lua
```diff
@@ -7,7 +7,11 @@
 local util = require 'lspconfig.util'
 
 local function find_tailwind_global_css()
-  local target = "@import 'tailwindcss';"
```
justinmk, line 10: find() takes a Lua pattern, so just change this to a pattern?
 
    local target = [[%@import ['"]tailwindcss['"]%;]]

besserwisser: Thank you, good point. I changed it.

Issue #4373, opened 2026-04-03
tailwindcss causes multi second UI freeze


besserwisser:
**Problem**

Opening any file handled by the tailwindcss LSP (e.g. TypeScript, TSX, HTML) causes a multi-second UI freeze on first attach. The freeze scales with project size.

**Root cause**

tailwindcss.lua defines find_tailwind_global_css() which is called synchronously inside before_init. It uses vim.fs.find with limit = math.huge to recursively scan the entire project tree for CSS/SCSS/PCSS files, reading each one to find an @import "tailwindcss" statement:

```lua
 local files = vim.fs.find(function(name)
   return name:match("%.css$") or name:match("%.scss$") or name:match("%.pcss$")
 end, {
   path = root,
   type = "file",
   limit = math.huge, -- search full tree  ← the problem
 })
```

Because root_dir falls back to .git for Tailwind v4 projects, the scan root is the git repository root, which typically includes node_modules and other large directories.

This blocks Neovim's main loop for the entire duration of the scan.

**Expected behavior**

File open should not freeze. We should use a non-blocking approach or skip folders like node_modules. 

**Possible fix**

Replace the blocking file scan with async scan or don't scan folders that are listed in the .gitignore file. Not sure if using `git`  cli is wanted. But `git ls-files` could help a lot.

justinmk:
> Not sure if using `git` cli is wanted. But `git ls-files` could help a lot.

sure, worth a try. this feature is only a few days old, and it definitely should be less intrusive. 

p.s. also consider the nit here: https://github.com/neovim/nvim-lspconfig/pull/4222#discussion_r3018499628


cc @CaoKha

besserwisser:
@justinmk  I created a PR here. Checking if git is installed just to be sure and fallback to the slow solution. 
https://github.com/neovim/nvim-lspconfig/pull/4374
BUT I did not add your nit. Because it would mean that `find_tailwind_global_css()` would run even if you provide a explicit `configFile`. At least as far as I understand, because the function is run and the result is passed to the `tbl_deep_extend` as an argument.

oedotme:
I just created another PR that removes the custom detection logic entirely. I've tested and shared the correct syntax to use `configFile` directly without any layer.

cc @justinmk @besserwisser @CaoKha

besserwisser:
@oedotme ohh I see. The native lsp detection does according to official docs:

[...]

PR #4374, opened 2026-04-03
fix(tailwindcss): multi second freeze

besserwisser:
https://github.com/neovim/nvim-lspconfig/issues/4373

### DIFF
--- a/lsp/tailwindcss.lua
+++ b/lsp/tailwindcss.lua
@@ -6,48 +6,78 @@
 --- npm install -g @tailwindcss/language-server
 local util = require 'lspconfig.util'
 
+--- Find the first CSS/SCSS/PCSS file containing a Tailwind v4 import directive.
+---@return string? path absolute path to the file, or nil if not found
 local function find_tailwind_global_css()
-  local target = [[%@import ['"]tailwindcss['"]%;]]
-  -- Find project root using `.git`
+  local targets = {
+    "@import 'tailwindcss'",
+    '@import "tailwindcss"',
+  }
+
   local buf = vim.api.nvim_get_current_buf()
   local root = vim.fs.root(buf, function(name)
     return name == '.git'
   end)
 
   if not root then
-    return nil -- no project root found
+    return nil
   end
 
-  local files ---@type string[]
-
-  -- Use `git ls-files` to respect .gitignore and avoid scanning node_modules etc.
+  -- Fast path: use git grep to search only tracked + untracked (but not ignored) files
   if vim.fn.executable('git') == 1 then
-    local ls = vim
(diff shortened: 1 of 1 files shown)


CaoKha:
why not doing that like this:
```lua
local function find_tailwind_global_css()
  local target = "@import 'tailwindcss'"

  local buf = vim.api.nvim_get_current_buf()
  local root = vim.fs.root(buf, function(name)
    return name == '.git'
  end)

  if not root then
    return nil
  end

  -- Fast path: use git grep to search only tracked files
  if vim.fn.executable('git') == 1 then
    local result = vim
      .system({
        'git',
        'grep',
        '-l',
        target,
        '--',
        '*.css',
        '*.scss',
        '*.pcss',
      }, { cwd = root, text = true })
      :wait()

    if result.code == 0 and result.stdout and result.stdout ~= '' then
      local first_match = vim.split(result.stdout, '\n')[1]
      if first_match and first_match ~= '' then
        return root .. '/' .. first_match
      end
    end
  end

  -- Fallback: scan filesystem with vim.fs.dir, skipping heavy directories
  local skip_dirs = {
    node_modules = true,
    ['.git'] = true,
    dist = true,
  }

  for name, typ in
    vim.fs.dir(root, {
      depth = math.huge,
      skip = function(dirname)
        return not skip_dirs[vim.fs.basename(dirname)]
      end,
    })
  do
    if typ == 'file' and (name:match('%.css$') or name:match('%.scss$') or name:match('%.pcss$')) then
      local path = root .. '/' .. name
      local ok, content = pcall(vim.fn.readblob, path)
      if ok and content:find(target, 1, true) then
        return path
      end
    end
  end

  return nil
end
```
I think maybe we should avoid using regex for better readability ?
For the fallback, we should also ignore folders frequently found in projects that use tailwind.

besserwisser:
Thank you @CaoKha  for the input. I pushed some changes based on `git grep`. But with some differences to your solution: 

1. I match both single and double quoted imports (`@import 'tailwindcss'` and `@import "tailwindcss"`) via a `targets` list, which also makes it easy to add new patterns later. This is something that I already did in a previous (merged) commit. But we cannot use the lua pattern here for the `git grep`, so I needed to update it. 
 2. I added `--untracked` to git grep so newly created but not yet staged CSS files are also found.
 3. If git grep runs successfully but finds nothing (exit code ≤ 1), I return `nil` immediately instead of falling back to the filesystem scan. No match means it's simply not a tailwind project I would say
 4. I cast `math.huge` with `--[[@as integer]]` to fix a LuaLS type mismatch on the depth parameter.

I appreciate some check, since I am not a regular contributor or lua expert.

besserwisser:
Closed in favor of the better solution to use the lsps native project detection: 
https://github.com/neovim/nvim-lspconfig/pull/4376

PR #4382, opened 2026-04-08
feat: zizmor

peytondmurray:
This PR adds support for [zizmor](https://github.com/zizmorcore/zizmor), a static analysis tool for GitHub Actions.

There have been a number of supply chain attacks recently that have me thinking about security, and some colleagues have recommended zizmor for catching some of the more common things (like pinning action versions). I saw that it has an experimental LSP mode, so I've added it here.

This is basically a copy of what is already in https://github.com/neovim/nvim-lspconfig/blob/master/lsp/gh_actions_ls.lua, except the docs and cmd have been updated to refer to zizmor rather than the github actions language server. That means that only github actions files are targeted, using the same machinery as `gh_actions_ls` already uses. For clarity, github actions LS provides completion, definitions, etc for github actions files while zizmor [focuses on security-related concerns](https://docs.zizmor.sh/audits/), so they cover different use cases.

Right now, the project has 4.1k stars and is very active, so it meets the criteria for being added as a supported language server here, but I will note that zizmor's LSP mode is experimental. There's a [meta-issue tracking LSP work here](https://github.com/zizmorcore/zizmor/issues/986) for anyone interested.

### DIFF
--- a/lsp/zizmor.lua
+++ b/lsp/zizmor.lua
@@ -1,0 +1,53 @@
+---@brief
+---
+--- https://github.com/zizmorcore/zizmor
+---
+--- Zizmor language server.
+---
+--- `zizmor` can be installed by following the instructions [here](https://docs.zizmor.sh/installation/).
+---
+--- The default `cmd` assumes that the `zizmor` binary can be found in `$PATH`.
+---
+--- See `zizmor`'s [documentation](https://docs.zizmor.sh/) for additional documentation.
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'zizmor', '--lsp' },
+  filetypes = { 'yaml' },
+
+  -- `root_dir` ensures that the LSP does not attach to all yaml files
+  root_dir = function(bufnr, on_dir)
+    local parent = vim.fs.dirname(vim.api.nvim_buf_get_name(bufnr))
+    if
+      vim.endswith(parent, '/.github/workflows')
+      or vim.endswith(parent, '/.forgejo/workflows')
+      or vim.endswith(parent, '/.gitea/workflows')
+    then
+      on_dir(parent)
+    end
+  end,
+  handlers = {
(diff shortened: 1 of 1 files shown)


lsp/zizmor.lua
```diff
@@ -0,0 +1,53 @@
...
+  handlers = {
+    ['actions/readFile'] = function(_, result)
+      if type(result.path) ~= 'string' then
+        return nil, nil
+      end
+      local file_path = vim.uri_to_fname(result.path)
+      if vim.fn.filereadable(file_path) == 1 then
+        local f = assert(io.open(file_path, 'r'))
+        local text = f:read('*a')
+        f:close()
```
justinmk: `vim.fn.readblob()`

lsp/zizmor.lua
```diff
@@ -0,0 +1,53 @@
...
+  cmd = { 'zizmor', '--lsp' },
+  filetypes = { 'yaml' },
+
+  -- `root_dir` ensures that the LSP does not attach to all yaml files
+  root_dir = function(bufnr, on_dir)
+    local parent = vim.fs.dirname(vim.api.nvim_buf_get_name(bufnr))
+    if
+      vim.endswith(parent, '/.github/workflows')
+      or vim.endswith(parent, '/.forgejo/workflows')
+      or vim.endswith(parent, '/.gitea/workflows')
```
rmuir, line 25: this language server works for composite actions as well, and also for the dependabot configuration files.

lsp/zizmor.lua
rmuir: you can remove this entire handler. zizmor does not call any `actions/readFile`

lsp/zizmor.lua
peytondmurray: Nice, I had no idea! Added `/.github/dependabot.{yml,yaml}` and `action.yml` as valid buffer names to check for.

PR #4384, opened 2026-04-10
feat(buf_ls): add filetype for buf config files

stefanvanburen:
The next release of `buf` will include LSP features that include support for buf config files. This follows the approach that other LSPs follow for specifying custom filetypes (a comment + adding that custom filetype to filetypes).

Ref: https://github.com/bufbuild/buf/pull/4438

### DIFF
--- a/lsp/buf_ls.lua
+++ b/lsp/buf_ls.lua
@@ -4,11 +4,31 @@
 --- buf lsp included in the cli itself
 ---
 --- buf lsp is a Protobuf language server compatible with Buf modules and workspaces
+---
+--- buf lsp also supports Buf configuration files. The `buf-config` filetype is not
+--- detected automatically; register it manually (see below) or override the filetypes:
+---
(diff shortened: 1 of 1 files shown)


lsp/buf_ls.lua
```diff
@@ -4,11 +4,31 @@
...
+---     ['buf.policy.yaml'] = 'buf-config',
+---     ['buf.lock'] = 'buf-config',
+---   },
+--- })
+--- ```
+---
+--- Optionally, tell treesitter to treat buf config files as YAML for syntax highlighting:
+---
+--- ```lua
+--- vim.treesitter.language.register('yaml', 'buf-config')
```
stefanvanburen, line 25: this might be moot with https://github.com/vim/vim/pull/19935 / https://github.com/neovim/neovim/pull/38897, since all of the extensions / names are taken care of upstream...

PR #4386, opened 2026-04-11
feat: use node_modules local executable if available

thinety:
In a similar vein to #3957, this PR allows the use of local `node_modules` executables for some of the most common tools in the JS ecosystem.

Some of the tools already had support for this (`biome`,  `glint`, `oxfmt`, `oxlint` and `tsgo`), so for these the code was just slightly refactored to use the new `lspconfig.utils` function `get_node_modules_executable`.

### DIFF
--- a/lsp/astro.lua
+++ b/lsp/astro.lua
@@ -47,7 +47,16 @@ local util = require 'lspconfig.util'
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'astro-ls', '--stdio' },
+  cmd = function(dispatchers, config)
+    local cmd = 'astro-ls'
+    if (config or {}).root_dir then
+      local local_cmd = vim.fs.joinpath(config.root_dir, 'node_modules/.bin', cmd)
+      if vim.fn.executable(local_cmd) == 1 then
+        cmd = local_cmd
+      end
+    end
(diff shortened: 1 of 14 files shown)


lua/lspconfig/util.lua
```diff
@@ -101,6 +101,14 @@ function M.get_typescript_server_path(root_dir)
   return ''
 end
 
+function M.get_node_modules_executable(cmd, root_dir)
+  local local_cmd = root_dir and root_dir .. '/node_modules/.bin/' .. cmd
+  if local_cmd and vim.fn.executable(local_cmd) == 1 then
+    return local_cmd
```
justinmk: thanks for the code re-use effort, but configs in nvim-lspconfig are intended to be fully self-contained, so we do not want to agument `util.lua`. and it looks like this code could be simplified anyway. 

note that `vim.fs.joinpath()` could be used and it accepts `nil` args

thinety: I'm not sure if my latest changes are any more readable than before... but I believe the `nil` check for `config.root_dir` is important, otherwise we are always checking for `node_modules`, even in a subdirectory (in other words, always checking if `vim.fs.joinpath((config or {}).root_dir, 'node_modules/.bin', cmd)` is executable).

PR #4392, opened 2026-04-15
feat(groovyls): use groovy-language-server from $PATH

briandipalma:
I'd say most people will be using mason to install this. Anyway I'm not sure what was there before would work as the path would require the jar to be on your/neovim's path, I think, which won't be the case for something you download manually/or build locally so you'd have to move it there in which case creating a little shell script like mason does is hardly much extra effort.

image

### DIFF
--- a/lsp/groovyls.lua
+++ b/lsp/groovyls.lua
@@ -20,11 +20,14 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = {
-    'java',
-    '-jar',
-    'groovy-language-server-all.jar',
-  },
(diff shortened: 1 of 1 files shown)


justinmk:
> most people will be using mason to install this

no, "mason" is never a valid rationale. but "find a program in the global $PATH" is.

 the current config expects `groovy-language-server-all.jar` to be in CWD which is flimsy, so i'm just going to merge this because it's not any worse.

justinmk:
the stylua issue should be fixed though

PR #4394, opened 2026-04-16
fix(flow): invalid condition

justinmk:


mickro:
mouths now that I have a local patch that I removed and re-add, before and after every update of lspconfig.

Thanks a lot.

... btw, as flow has some much instability between version, I believe that using a global flow-bin isn't a good practice.

so my "patch" was to always use npx which is able to catch to correct version of flow from the project.

```
cmd = { 'npx', '--no-install', 'flow', 'lsp' }

Issue #4391, opened 2026-04-15
denols root_dir logic contains incorrect logic


bodograumann:
### Description

I haven't read all the previous issues regarding the denols config, but currently the logic contains redundant statements.
In particular `"deno.lock"` is contained both in `project_root` and in `deno_lock_root`. The condition for starting the LS i

```lua
    if
      (deno_lock_root and (not project_root or #deno_lock_root > #project_root))
      or (deno_root and (not project_root or #deno_root >= #project_root))
    then
```

However, when `deno_lock_root` is true, `project_root` will also always be true.
Similarly `#deno_lock_root > #project_root` can never be true, because if the lock file is closer than all other project files, it is still determining `project_root`.

Maybe `"deno.lock"` should be removed from `root_markers`?

justinmk:
hmm, WDYT @isaacs

isaacs:
It's possible that this could be simplified, sure. I sort of just hacked at it until it behaved properly, and there were a few other edge cases that were brought up at the time.

The intention is that if a `deno.lock` or `deno.json` file is found closer than other root markers (or at the same level), then it's a deno file. If there's a deno.lock file that's further away than other root markers, then it might not be.

For example, you might have something like this in a complicated monorepo:

```
./.git/
./deno.lock
./deno.json
./src/foo.ts   <-- deno file
./npmproject/ <-- nested non-deno project
./npmproject/package-lock.json
./npmproject/package.json
./npmproject/src/foo.ts  <-- not a deno file
./npmproject/packages/denoproject/ <-- sub-workspace in the non-deno project
./npmproject/packages/denoproject/deno.json
./npmproject/packages/denoproject/src/foo.ts <-- this is a deno file
```

I ran into this on the getsentry/sentry-javascript project, which is _mostly_ traditional typescript that should be handled by `ts_ls`, but has a deno SDK in `packages/deno` which is code that runs on Deno.

bodograumann:
Interesting example. From reading the code I don't think the current config works that way.
The only non-deno root marker is `.git`. So the result of `root_dir` will be the same for `./src/foo.ts` and `./npmproject/src/foo.ts`. Both will be marked as deno files, because `deno_root` and `project_root` are the same.

So we should check for `package.json`, `package-lock.json` and the likes.
A similar thing happens in the vtsls and eslint configs. So wouldn't it make sense to have shared logic?
Then at least the various lsps would be attached consistently.
[...]

PR #4395, opened 2026-04-17
feat: update oxfmt and oxlint to detect vite plus

V1OL3TF0X:
Since oxfmt and oxlint have the ability to use vite plus config (as mentioned here: https://github.com/voidzero-dev/vite-plus/issues/629), it would be great to update lspconfig to reflect that (terribly sorry it took me 4 commits to format message correctly)


justinmk:
I don't have a sense for how common or urgent this is. Need to hear from other oxlint users.


At some point I would expect users to just... use `vim.lsp.config('oxlint', {...}` to override the default config, to say where their root file is.



These js projects are out of control, there seems to be 17 different config files in every js workspace. I wonder how other editors deal with this. Related to that, there is a cost to having complex logic: 

- https://github.com/neovim/nvim-lspconfig/issues/4391

V1OL3TF0X:
> I don't have a sense for how common or urgent this is. Need to hear from other oxlint users.

Npmx says 136k downloads/week for vite-plus, 4m/week for oxlint / oxfmt. The main problem for me was this feature was not well documented, so I had to dig pretty deep to find out how it's done (came from an issue about supporting per-file formatting on vp and then 4 layers of linked issues). If you think it's not worth changing the config, I'd leave a comment at least to guide someone who wants to implement it. I imagine vite-plus getting more popular with time tho.
 
> These js projects are out of control, there seems to be 17 different config files in every js workspace.

Kinda lol
Here it's a case of vite-plus being a unified toolchain built on top of oxc, which means it integrates both binaries into its config - still quite annoying to handle, yeah

justinmk:
https://github.com/neovim/nvim-lspconfig/issues/4391 is what worries me. 

A comment in the docstring is a good start, giving your insight.

I'm not opposed to this change but hoping to hear from other oxlint users. And your `root_markers_with_field` change looks useful

TheAlexLichter:
A bit of background: Oxlint and Oxfmt are being used in two typical scenarios:

1. standalone with an .oxlintrc/oxlint.config.ts (and equivalent for oxfmt)
2. As part of Vite Plus, which acts as a "wrapper CLI" to avoid having tons of manual tools installed, honoring the vite.config.ts

So both cases are relevant

V1OL3TF0X:
> A bit of background: Oxlint and Oxfmt are being used in two typical scenarios:
> 
> 1. standalone with an .oxlintrc/oxlint.config.ts (and equivalent for oxfmt)
> 2. As part of Vite Plus, which acts as a "wrapper CLI" to avoid having tons of manual tools installed, honoring the vite.config.ts
> 
> So both cases are relevant

And @TheAlexLichter @justinmk correct me if I'm wrong, but the settings here act like a rough heuristic - oxlint/oxfmt will detect their proper configs anyway, according to their internal logic (I know oxlint has some config inheritance for example)

justinmk:
> A comment in the docstring is a good start, giving your insight.

was waiting for this. is there anything that should be mentioned in the doc? if so, please include it in this pr

V1OL3TF0X:
@justinmk Sorry, I must've missed it - added a short explanation and linked a PR where it was implemented - lmk if that's what you had in mind

PR #4396, opened 2026-04-21
fix(lua_ls): add lspconfig runtime to lua config

flatplum:
Problem:
The commented config for lua_ls doesn't make type definitions aware to the server

Fix:
Add runtime to library

Checked on Nvim v0.12.1

### DIFF
--- a/lsp/lua_ls.lua
+++ b/lsp/lua_ls.lua
@@ -42,6 +42,8 @@
 ---         checkThirdParty = false,
 ---         library = {
 ---           vim.env.VIMRUNTIME,
+---           -- Necessary for LSP Settings Type Annotations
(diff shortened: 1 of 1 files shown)


lsp/lua_ls.lua
```diff
@@ -42,6 +42,8 @@
 ---         checkThirdParty = false,
 ---         library = {
 ---           vim.env.VIMRUNTIME,
+---           -- Necessary for LSP Settings Type Annotations
```
justinmk:
```suggestion
---           -- For LSP Settings Type Annotations: https://github.com/neovim/nvim-lspconfig#lsp-settings-type-annotations
```

lsp/lua_ls.lua
justinmk, line 56: This also would do it

PR #4405, opened 2026-04-27
fix(zls): use up to date config schema for LSP settings types

Techatrix:
The LSP settings types for the [ZLS](https://github.com/zigtools/zls) language server were based of https://github.com/zigtools/zls-vscode which has been deprecated and contains mostly outdated config options. It has since been replaced by https://codeberg.org/ziglang/vscode-zig. ZLS offers a [schema.json](https://github.com/zigtools/zls/blob/master/schema.json) file to describe it's own `zls.json` config file that also matches LSP based configuration. This is beneficial over the `package.json` because it avoids the inclusion of config options that are specific the VS Code extension.

Because the schema file is primarily intended for standalone a `zls.json` config file, the properties are not prefixed with `zls.` which is why I had to add the `append_config_section` option.

### DIFF
--- a/scripts/gen_json_schemas.lua
+++ b/scripts/gen_json_schemas.lua
@@ -61,7 +61,7 @@ local index = {
   vtsls = 'https://raw.githubusercontent.com/yioneko/vtsls/main/packages/service/configuration.schema.json',
   vue_ls = 'https://raw.githubusercontent.com/vuejs/vetur/master/package.json',
   yamlls = 'https://raw.githubusercontent.com/redhat-developer/vscode-yaml/master/package.json',
-  zls = 'https://raw.githubusercontent.com/zigtools/zls-vscode/master/package.json',
+  zls = 'https://raw.githubusercontent.com/zigtools/zls/master/schema.json',
 }
 
 ---@param url string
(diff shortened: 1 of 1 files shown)


scripts/gen_json_schemas.lua
```diff
@@ -213,6 +209,16 @@ local function generate_server_schema(schema)
     properties = config_schema.properties
   end
 
+  if schema.append_config_section then
```
justinmk: can you add a comment that gives a hint showing an example
```suggestion
  -- "..." => "..."
  if schema.append_config_section then
```

Techatrix: Something like this?

```suggestion
  -- `properties[config_name]` => `properties[config_section .. '.' .. config_name]`
  -- `properties["enable_snippets"]` => `properties["zls.enable_snippets"]`
  if schema.append_config_section then
```

justinmk: yes, but only the 2nd line. also, that looks like "prepend" not "append"

scripts/gen_json_schemas.lua
```diff
@@ -104,6 +96,7 @@ end
 ---@field package_url string url of the package.json of the LSP server
 ---@field settings_file string file of the settings json schema of the LSP server
 ---@field translate? boolean
+---@field append_config_section? string add a LSP config section to each schema property
```
justinmk:
```suggestion
---@field prefix? string Prepend this string to each schema property.
```

PR #4414, opened 2026-05-01
feat: add make_ls

Waitzz:
I found a make lsp that runs well and is simple to use: https://github.com/owenrumney/make-ls. I hope it can be added to nvim-lspconfig. Although it is not known by many people at present, it is really useful .

### DIFF
--- a/lsp/make_ls.lua
+++ b/lsp/make_ls.lua
@@ -1,0 +1,17 @@
+---@brief
+---
+--- https://github.com/owenrumney/make-ls
+---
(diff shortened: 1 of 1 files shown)


justinmk:
👀  there is also https://github.com/lewis6991/make-ls @lewis6991 

we could add both, but what shall we name them... 🤔 

do they have different `cmd` names?

lewis6991:
> do they have different cmd names?

No 😬

lewis6991:
I guess we'll need to include the org in the name to distinguish them.

PR #4422, opened 2026-05-12
feat(ts_ls): support node_modules not in workspace root directory

Tsukina-7mochi:
This pull request enhances the TypeScript LSP (`ts_ls.lua`) configuration to better support projects with non-standard `node_modules` layouts, such as monorepos or pnpm workspaces. The main improvement is the introduction of a helper function to locate packages within `node_modules` directories, ensuring the correct TypeScript server and binaries are used regardless of project structure.

**Improvements for locating TypeScript packages and binaries:**

* Added a new `find_node_package` function that searches for a specified package within `node_modules` directories up the directory tree, improving compatibility with monorepos and pnpm setups.
* Updated `init_options.tsserver.path` to use the new helper, ensuring the TypeScript server uses the correct `lib` directory for the current buffer, even if `node_modules` is not at the project root.
* Modified the `cmd` function to use `find_node_package` when resolving the local `typescript-language-server` binary, making sure the correct binary is used in complex project structures.

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -73,15 +73,42 @@
 --- If DENO ROOT is found, and it's longer than or equal to PROJECT ROOT, then this is a Deno file, and we abort.
 --- Otherwise, attach at PROJECT ROOT, or the cwd if not found.
 
+--- Searches for a package within `node_modules` directories starting from the given root.
+--- Walks up the directory tree via `vim.fs.find` to locate `node_modules` folders,
+--- then checks each one for the named package.
+---
+---@param root string The starting directory path for the search.
+---@param name string The package name (or relative path) to look for inside `node_modules`.
+---@return string|nil path The absolute path to the package if found and readable, otherwise nil.
+local function find_node_package(root, name)
+  for _, dir in ipairs(vim.fs.find('node_modules', { path = root })) do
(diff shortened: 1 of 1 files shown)


justinmk:
does this match what the tsgo and other js configs are doing? they all generally end up having the same problems, so want to make sure they are following the same patterns.

Tsukina-7mochi:
@justinmk  I only knew about this in the context of Node.js, so I looked into it. Since this is about `ts_ls`, researching TypeScript and tsgo should be sufficient.

- Node.js: https://nodejs.org/api/modules.html#loading-from-node_modules-folders
- TypeScript: https://www.typescriptlang.org/docs/handbook/modules/reference.html#node_modules-package-lookups
  - There were other module resolution methods before TS 6.0, but in my experience, I have never seen any projects using that version that also use this option.
    - 6.0 release note: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---moduleresolution-classic
  - It was thought that tsgo would not change because it is a line-to-line port, but I am providing the source just in case.
    - https://github.com/microsoft/typescript-go/blob/0a7c6b47e6163892880e0e5814e519c435445f11/internal/module/resolver.go#L1003-L1016

PR #4425, opened 2026-05-14
feat: add support for hare-lsp

WhyNotHugo:


WhyNotHugo:
If I understand CI's output correctly, I shouldn't commit the changes to `doc/configs.md` and this is done by some other process? Reverted that file.

WhyNotHugo:
All green now

WhyNotHugo:
Our of curiosity, why does the final commit lack the author's full name, present in the commit in this PR?

GitHub's UI only shows the author in the raw patch view, but nowhere else (apparently).

Commit in PR has full name: https://github.com/neovim/nvim-lspconfig/commit/e6c2e4319917edb11a4915f79ef5256773857280.patch

Commit in master has only first name: https://github.com/neovim/nvim-lspconfig/commit/57f7d75090a99ff5f7fb71a8c13a0773af32bc2f.patch

Is this something GitHub is enforcing?

justinmk:
That's likely based on your profile settings. When I do a squash-merge it apparently uses your profile info instead of the commit data (which somewhat makes sense because otherwise it would need to guess the committer from potentially multiple squashed commits)

PR #4433, opened 2026-05-20
fix(ruby-lsp): `reuse_client` issue in ruby lsp to avoid spawning extra LSP instances

dgmora:
## AI Disclosure

This change was generated with Codex. I am not familiar with the codebase. But the change is working for me, so I figured I could share it. Happy to get extra edits or a different PR with the same changes, personally what I needed is to avoid spawning 2 LSPs

## Problem

When `vim.lsp.start` is called twice for the same buffer/root_dir, `ruby_lsp` fails to dedup and a second LSP process is spawned.

I hit this with lazy.nvim: any `ft = "ruby"` plugin causes `vim.lsp.start` to fire twice. Once on the initial `FileType` event, and again after lazy.nvim loads the plugin and replays `FileType`. End result: two `ruby_lsp` processes attached to the same buffer.

```lua
-- lazy.nvim spec that triggers it
{
  "neovim/nvim-lspconfig",
  opts = { servers = { ruby_lsp = { cmd = { "mise", "x", "--", "ruby-lsp" } } } },
},
{ "tpope/vim-rails", ft = "ruby" },
```

But it's not lazy.nvim-specific. Any double call to `vim.lsp.start` for the same root reproduces it:

```lua
vim.lsp.start({ name = "ruby_lsp", cmd = { "ruby-lsp" }, root_dir = "/tmp/x" })
vim.lsp.start({ name = "ruby_lsp", cmd = { "ruby-lsp" }, root_dir = "/tmp/x" })
-- :LspInfo → two ruby_lsp clients, two processes
```

## Cause

The current `reuse_client` mutates the *incoming* config and then compares it against the *existing* client:

```lua
  config.cmd_cwd = config.root_dir
  return client.name == config.name and client.config.cmd_cwd == config.cmd_cwd
end,
```

`cmd_cwd` is never set on the existing client. It's not in the default spec, and mutating the new `config` doesn't propagate back to `client.config`. So `client.config.cmd_cwd` is `nil`, the comparison fails, and dedup never happens.

## Fix

Compare `root_dir` directly. Both sides have it.

```lua
  config.cmd_cwd = config.root_dir
  return client.name == config.name and client.config.root_dir == config.root_dir
end,
```

The `config.cmd_cwd = config.root_dir` line is kept to preserve existing behavior, but it appears to be dead code in this file: `cmd` is defined as a function that already does `cwd = config.cmd_cwd or config.root_dir` on every spawn, so the process cwd is correctly set whether or not the mutation runs. It could likely be removed in a follow-up.

## Note on `ast_grep`

`lsp/ast_grep.lua` has the same `reuse_client` pattern (introduced together in #4099 / #4408). It's intentionally left out of this PR because, unlike `ruby_lsp`, `ast_grep`'s `cmd` is a static array and relies on `cmd_cwd` for process cwd. Removing the mutation there has a tangible (if narrow) behavioral effect, so it deserves its own discussion.


justinmk:
The change looks good, but as you'll notice, there are no tests in this repo and we don't plan to write tests for 1000 different configs. In the future, please also minimize verbosity from what your AI produces. Two pages of explanation is excessive for a 1-line change.

PR #4434, opened 2026-05-21
fix(oxc): fix oxlint and oxfmt vite plus config search

### LINKED ISSUE #4432: oxlint is not found in monorepo
### Description

Because of the last changes, `oxlint` is not found in a monorepo, where each app has it's own `vite.config.ts`, but `oxlint` is installed only in the root project.
```
.
├── apps/
│   ├── app1/
│   │   └── vite.config.ts
│   └── app2/
│       └── vite.config.ts
├── node_modules/
├── package.json
└── vite.config.ts
└── oxlint.config.ts
```

Because the `root_dir` function looks for the first `vite.config.ts`, it finds it in for example `app1`, but the `node_modules` is not there.

V1OL3TF0X:
Adds additional check for `fmt`/`lint` fields in `vite.config.ts`, fix #4432

### DIFF
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -100,7 +100,7 @@ function M.root_markers_with_fields(root_files, new_names, fields, fname)
       to_find = vim
         .iter(to_find)
         :filter(function(s)
-          return line:find(s)
+          return !line:find(s)
         end)
         :totable()
       if #to_find == 0 then
--- a/lua/lspconfig/util.lua
+++ b/lua/lspconfig/util.lua
@@ -76,7 +76,11 @@ function M.root_markers_with_field(root_files, new_names, field, fname, match_mo
           return not line:find(s)
         end)
         :totable()
-      return #to_find == 0
+      if #to_find == 0 then
+        to_find = vim.deepcopy(files)
+        return true
+      end
+      return false
     end
   for _, f in ipairs(found or {}) do
     -- Match the given `field`.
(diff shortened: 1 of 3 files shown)


lua/lspconfig/util.lua
```diff
@@ -79,6 +79,41 @@ function M.root_markers_with_field(root_files, new_names, field, fname)
...
+--- Appends `new_names` to `root_files` if all `fields` are found in any such file in any ancestor of `fname`.
+---
+--- NOTE: this does a "breadth-first" search, so is broken for multi-project workspaces:
+--- https://github.com/neovim/nvim-lspconfig/issues/3818#issuecomment-2848836794
+---
+--- @param root_files string[] List of root-marker files to append to.
+--- @param new_names string[] Potential root-marker filenames (e.g. `{ 'package.json', 'package.json5' }`) to inspect for the given `field`.
+--- @param fields string[] Fields to search for in the given `new_names` files.
+--- @param fname string Full path of the current buffer name to start searching upwards from.
+function M.root_markers_with_fields(root_files, new_names, fields, fname)
```
justinmk: `root_markers_with_field` already accepts `field string | string[]` , why can't we use it? or enhance it

V1OL3TF0X: It's a case of match any (root_marker_with_field) vs match all (root_markers_with_fields). I didn't want to enhance it, to not introduce too much ifology and obfuscate what the fn does

justinmk: 2 nearly-same functions is worse than a `any` or `all` parameter

V1OL3TF0X: Okay, updated to reflect that

PR #4436, opened 2026-05-22
fix(roslyn_ls): remove unnecessary code and recommend maintained plugin for razor

seblyng:
I am one of the maintainers of [roslyn.nvim](https://github.com/seblyng/roslyn.nvim). The roslyn server has made some changes that makes some of the code here not needed anymore:

- The arguments to cmd are no longer required
- The `workspace/_roslyn_projectNeedsRestore` handler is removed internally in roslyn

Also, I changed the recommendation to use roslyn.nvim over rzls.nvim for razor support since rzls.nvim has been deprecated for quite some time, and the effort for razor is moved into roslyn.nvim, with the maintainer of that plugin joining me in maintaining roslyn.nvim

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -68,7 +68,7 @@ local function roslyn_handlers()
     end,
     ['razor/provideDynamicFileInfo'] = function(_, _, _)
       vim.notify(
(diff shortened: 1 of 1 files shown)


justinmk:
thank you!

cc @tris203

PR #4445, opened 2026-06-01
fix(golangci_lint_ls): fixed creation of separate clients for files in GOMODCACHE and GOROOT/src

krukroman:
## Problem:
`neovim` will spawn separate language server clients for files in `GOMODCACHE` and `GOROOT/src` directories if user goes there through `go to definition`.

## Solution:
Use already implemented solution from [gopls config](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/gopls.lua)

### DIFF
--- a/lsp/golangci_lint_ls.lua
+++ b/lsp/golangci_lint_ls.lua
@@ -13,6 +13,56 @@
 --- go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
 --- ```
 
+--- @class go_dir_args
+---
+--- @field envvar_id string
+---
+--- @field subdir string?
+
+---@param dir_args go_dir_args
+---@return string?
+local function get_go_dir(dir_args)
+  local cmd = { 'go', 'env', dir_args.envvar_id }
+  local ok, sys_obj = pcall(vim.system, cmd, { text = true })
+
+  -- handles case when go is not installed on user machine
+  if not ok then
+    vim.schedule(function()
+      vim.notify(
+        ('[golangci_lint_ls] cmd %s failed with error %s'):format(vim.inspect(cmd), vim.inspect(sys_obj)),
+        vim.log.levels.ERROR
+      )
+    end)
+
+    return nil
+  end
+
+  local output = sys_obj:wait()
+  local res = vim.trim(output.stdout or '')
+
+  -- handles case when go env cmd failed or res is an empty string
+  if output.code ~= 0 or res == '' then
+    vim.schedule(function()
+      vim.notify(
+        ('[golangci_lint_ls] get ' .. dir_args.envvar_id .. ' dir cmd failed with code %d: %s\n%s'):format(
+          output.code,
+          vim.inspect(cmd),
+          output.stderr
+        ),
+        vim.log.levels.WARN
+      )
+    end)
+
+    return nil
+  end
+
+  if dir_args.subdir and dir_args.subdir ~= '' then
+    res = vim.fs.joinpath(res, dir_args.subdir)
+  end
+
+  return res
+end
+
 ---@type vim.lsp.Config
 return {
   cmd = { 'golangci-lint-langserver' },
@@ -35,6 +85,21 @@ return {
       '--output.json.path=stdout',
     },
   },
+  reuse_client = function(client, config)
+    if client.name ~= config.name or client:is_stopped() then
+      return false
+    end
+
+    if client.root_dir ~= config.root_dir then
+      local mod_cache = get_go_dir({ envvar_id = 'GOMODCACHE' })
+      local std_lib = get_go_dir({ envvar_id = 'GOROOT', subdir = 'src' })
+
+      -- reuse client if config root dir is equal to go mod cache dir, or go standard library dir
+      return config.root_dir == mod_cache or config.root_dir == std_lib
+    end
+
+    return true
+  end,
   root_markers = {
     '.golangci.yml',
     '.golangci.yaml',


lsp/golangci_lint_ls.lua
```diff
@@ -35,15 +133,13 @@ return {
...
-    '.golangci.toml',
-    '.golangci.json',
-    'go.work',
-    'go.mod',
-    '.git',
-  },
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    get_mod_cache_dir()
+    get_std_lib_dir()
```
justinmk: why don't these return values, instead of setting shared state, so the control flow is clearer

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,104 @@
...
+---@return string?
+local function get_root_dir(fname)
+  if mod_cache and fname:sub(1, #mod_cache) == mod_cache then
+    local clients = vim.lsp.get_clients({ name = 'golangci_lint_ls' })
+    if #clients > 0 then
+      return clients[#clients].config.root_dir
+    end
+  end
+  if std_lib and fname:sub(1, #std_lib) == std_lib then
+    local clients = vim.lsp.get_clients({ name = 'golangci_lint_ls' })
```
justinmk: not sure what's going on here. seems like `reuse_client` needs to be defined, instead of poking around and looking for existing clients.

lsp/golangci_lint_ls.lua
krukroman: Because these functions are async and use `identify_go_dir` function, which uses `vim.system` call to obtain go env variables

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,104 @@
...
+--- @field envvar_id string
+---
+--- @field custom_subdir string?
+
+local mod_cache = nil
+local std_lib = nil
+
+---@param custom_args go_dir_custom_args
+---@param on_complete fun(dir: string | nil)
+local function identify_go_dir(custom_args, on_complete)
```
justinmk:
```suggestion
local function get_go_dir(custom_args, on_done)
```

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,104 @@
...
+    local res = vim.trim(output.stdout or '')
+    if output.code == 0 and res ~= '' then
+      if custom_args.custom_subdir and custom_args.custom_subdir ~= '' then
+        res = res .. custom_args.custom_subdir
+      end
+      on_complete(res)
+    else
+      vim.schedule(function()
+        vim.notify(
+          ('[golangci_lint_ls] identify ' .. custom_args.envvar_id .. ' dir cmd failed with code %d: %s\n%s'):format(
```
justinmk: just `error()`

lsp/golangci_lint_ls.lua
justinmk: then that introduces a race. just call the functions from `identify_go_dir` ...

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,78 @@
...
+---@return string?
+local function get_go_dir(custom_args)
+  local cmd = { 'go', 'env', custom_args.envvar_id }
+  local ok, sys_obj = pcall(vim.system, cmd, { text = true })
+
+  -- handles case when go is not installed on user machine
+  if not ok then
+    vim.schedule(function()
+      vim.notify(
+        ('[golangci_lint_ls] cmd %s failed with error %s'):format(vim.inspect(cmd), vim.inspect(sys_obj)),
```
justinmk, line 32: just `error()` ...

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,78 @@
...
+  end
+
+  local output = sys_obj:wait()
+  local res = vim.trim(output.stdout or '')
+
+  -- handles case when go env cmd failed or res is an empty string
+  if output.code ~= 0 or res == '' then
+    vim.schedule(function()
+      vim.notify(
+        ('[golangci_lint_ls] get ' .. custom_args.envvar_id .. ' dir cmd failed with code %d: %s\n%s'):format(
```
justinmk: this is not merely a "warning" situation. just `error()`

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,78 @@
 --- go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
 --- ```
 
+--- @class go_dir_custom_args
+---
+--- @field envvar_id string
+---
+--- @field custom_subdir string?
+
+---@param custom_args go_dir_custom_args
```
justinmk: why is this called "custom args" ?

lsp/golangci_lint_ls.lua
```diff
@@ -13,6 +13,78 @@
...
+  }
+
+  ---@type string[]
+  local dir_paths = {}
+
+  for _, args in ipairs(custom_args) do
+    local path = get_go_dir(args)
+
+    if path and path ~= '' then
+      table.insert(dir_paths, path)
```
justinmk: this is over-engineered and thus unncessarily verbose. just call `get_go_dir` explicitly, twice, with the appropriate args.

lsp/golangci_lint_ls.lua
```diff
@@ -35,6 +107,19 @@ return {
...
   },
+  reuse_client = function(client, config)
+    if client.name ~= config.name then
+      return false
+    end
+
+    if client.root_dir ~= config.root_dir then
+      local go_dirs = get_go_dirs()
+
+      return vim.list_contains(go_dirs, config.root_dir)
```
justinmk: a 1-line comment sure would help people reading this config in the future

lsp/golangci_lint_ls.lua
krukroman: `error()` terminates functions, so ot will will break [lsp.start](https://github.com/neovim/neovim/blob/fecc151d0306da95628a05203fd7273eb95580ee/runtime/lua/vim/lsp.lua#L761-L772) for clients that use `reuse_client` 

```lua
  for _, client in pairs(lsp.client._all) do
    if reuse_client(client, config) then
      if opts.attach == false then
        return client.id
      end


      if lsp.buf_attach_client(bufnr, client.id) then
        return client.id
      end
      return
    end
  end
```

lsp/golangci_lint_ls.lua
krukroman: [reason](https://github.com/neovim/nvim-lspconfig/pull/4445#discussion_r3355681420)

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
krukroman: done

lsp/golangci_lint_ls.lua
krukroman: done

krukroman:
converted to draft, because after looking at [`reuse_client_default`](https://github.com/neovim/neovim/blob/fecc151d0306da95628a05203fd7273eb95580ee/runtime/lua/vim/lsp.lua#L120-L145), I realized that `reuse_client` logic is deeper than comparison between `client.root_dir` and `config.root_dir`

krukroman:
closed, because proper implementation requires to copy-paste [`reuse_client_default` implementation](https://github.com/neovim/neovim/blob/fecc151d0306da95628a05203fd7273eb95580ee/runtime/lua/vim/lsp.lua#L120-L145) and add custom condition to it.

<details>
<summary>Possible workaround</summary>

```lua
--- @class go_dir_args
---
--- @field envvar_id string
---
--- @field subdir string?

---@param dir_args go_dir_args
---@return string?
local function get_go_dir(dir_args)
  local cmd = { 'go', 'env', dir_args.envvar_id }
  local ok, sys_obj = pcall(vim.system, cmd, { text = true })

  -- handles case when go is not installed on user machine
  if not ok then
    vim.schedule(function()
      vim.notify(
        ('[golangci_lint_ls] cmd %s failed with error %s'):format(vim.inspect(cmd), vim.inspect(sys_obj)),
        vim.log.levels.ERROR
      )
    end)

    return nil
  end

  local output = sys_obj:wait()
  local res = vim.trim(output.stdout or '')

  -- handles case when go env cmd failed or res is an empty string
  if output.code ~= 0 or res == '' then
    vim.schedule(function()
      vim.notify(
        ('[golangci_lint_ls] get ' .. dir_args.envvar_id .. ' dir cmd failed with code %d: %s\n%s'):format(
          output.code,
          vim.inspect(cmd),
          output.stderr
        ),
        vim.log.levels.WARN
      )
    end)

    return nil
  end

  if dir_args.subdir and dir_args.subdir ~= '' then
    res = vim.fs.joinpath(res, dir_args.subdir)
  end

  return res
end

---@type vim.lsp.Config
return {
-- config
    if client.name ~= config.name or client:is_stopped() then
      return false
    end

    local config_folders = vim.lsp._get_workspace_folders(config.workspace_folders or config.root_dir)

    if not config_folders or not next(config_folders) then
      -- Reuse if the client was configured with no workspace folders
      local client_config_folders =
        vim.lsp._get_workspace_folders(client.config.workspace_folders or client.config.root_dir)
      return not client_config_folders or not next(client_config_folders)
    end

    for _, config_folder in ipairs(config_folders) do
      local found = false
      for _, client_folder in ipairs(client.workspace_folders or {}) do
        if config_folder.uri == client_folder.uri then
          found = true
          break
        end
      end

      local mod_cache = get_go_dir({ envvar_id = 'GOMODCACHE' })
      local std_lib = get_go_dir({ envvar_id = 'GOROOT', subdir = 'src' })
      local config_folder_name = vim.uri_to_fname(config_folder.uri)

      -- Reuse the client if config folder belongs to go mod cache, or to go std lib
      if
        (mod_cache and vim.startswith(config_folder_name, mod_cache))
        or (std_lib and vim.startswith(config_folder_name, std_lib))
      then
        found = true
      end

      if not found then
        return false
      end
    end

    return true
  end,
-- config
}
```

</details>

justinmk:
> proper implementation requires to copy-paste [`reuse_client_default` implementation](https://github.com/neovim/neovim/blob/fecc151d0306da95628a05203fd7273eb95580ee/runtime/lua/vim/lsp.lua#L120-L145) and add custom condition to it.

@lewis6991 any advice here?

lewis6991:
Haven't looked into the fine detail but I don't think configs should have this much amount of code. I can't imagine other editors have configs this complex.

PR #4447, opened 2026-06-03
fix(emmylua_ls): move settings under emmylua section

marcuscaisey:
# Problem
EmmyLua requests configuration from the "emmylua" section, but the current configuration doesn't expose the settings under that section.

Example request and response:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "workspace/configuration",
  "params": {
    "items": [
      {
        "scopeUri": "file:///Users/marcus/.dotfiles",
        "section": "emmylua"
      }
    ]
  }
}
```
```json
{
  "result": [
    null
  ],
  "id": 1,
  "jsonrpc": "2.0"
}
```

# Solution
Move the settings under the "emmylua" section.

After these changes:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "workspace/configuration",
  "params": {
    "items": [
      {
        "scopeUri": "file:///Users/marcus/.dotfiles",
        "section": "emmylua"
      }
    ]
  }
}
```
```json
{
  "result": [
    {
      "hint": {
        "enable": true
      },
      "codeLens": {
        "enable": true
      }
    }
  ],
  "id": 1,
  "jsonrpc": "2.0"
}
```

### DIFF
--- a/lsp/emmylua_ls.lua
+++ b/lsp/emmylua_ls.lua
@@ -73,7 +73,9 @@ return {
     or vim.list_extend(vim.list_extend(root_markers1, root_markers2), { '.git' }),
   workspace_required = false,
   settings = {
-    codeLens = { enable = true },
-    hint = { enable = true },
+    emmylua = {
+      codeLens = { enable = true },
+      hint = { enable = true },
+    },
   },
 }


justinmk:
oh, this was missed in https://github.com/neovim/nvim-lspconfig/pull/4426 😅 

thanks!

sharpchen:
Looks like this breaks lazydev.nvim as it uses `Lua = ...`. See https://github.com/folke/lazydev.nvim/blob/ff2cbcba459b637ec3fd165a2be59b7bbaeedf0d/lua/lazydev/workspace.lua#L189-L200

This is a slightly breaking for users used to have `Lua = ...` in their setting.

Is it new design of emmylua? I am using `0.22.0`, maybe we should use `Lua = ...` if it's still compatible with emmylua.

EDIT: ah I saw there's already a pr for this: https://github.com/folke/lazydev.nvim/pull/141
EDIT: for anyone want to get lazydev work with current lspconfig preset, use `vim.NIL` to unset `emmylua` section:
```lua
vim.lsp.config('emmylua_ls', {
  -- see: https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/main/docs/config/emmyrc_json_EN.md
  settings = {
    emmylua = vim.NIL,
    Lua = {
      codeLens = { enable = true },
      hint = { enable = true },
    },
  },
})
```

PR #4450, opened 2026-06-08
feat: fallow lsp

igorlfs:
cc @tris203

note: [repo](https://github.com/fallow-rs/fallow) has 3k+ stars

### DIFF
--- a/lsp/fallow.lua
+++ b/lsp/fallow.lua
@@ -1,0 +1,28 @@
+---@brief
+---
+--- https://github.com/fallow-rs/fallow
+---
+--- Codebase intelligence for TypeScript and JavaScript.
+
+---@type vim.lsp.Config
+return {
+  cmd = function(dispatchers, config)
+    local cmd = 'fallow-lsp'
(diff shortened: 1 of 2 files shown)


lsp/fallow.lua
```diff
@@ -0,0 +1,19 @@
...
+---
+--- https://github.com/fallow-rs/fallow
+---
+--- Codebase intelligence for TypeScript and JavaScript.
+
+---@type vim.lsp.Config
+return {
+  cmd = { 'fallow-lsp' },
+  filetypes = { 'javascript', 'typescript', 'javascriptreact', 'typescriptreact' },
+  root_markers = { '.fallowrc.json', '.git' },
```
igorlfs, line 20: Intentionally left out the `package.json` from the [standard configuration](https://github.com/fallow-rs/fallow/tree/main/editors/nvim) to be more "monorepo friendly"

igorlfs:
~Might as well check if the binary is available locally~ Done

PR #4453, opened 2026-06-18
feat: terraformls enable codelens

GustavEikaas:
I noticed that terraform_ls does support codelens for references but its opt-in. I'm not sure if it should be default enabled or commented out for reference but it would be nice for it to be discoverable from this repo rather than having to dig through docs in terraform repo

image

Thoughts?

Here is the reference https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md?plain=1#L162

### DIFF
--- a/lsp/terraformls.lua
+++ b/lsp/terraformls.lua
@@ -37,4 +37,12 @@ return {
   cmd = { 'terraform-ls', 'serve' },
   filetypes = { 'terraform', 'terraform-vars' },
   root_markers = { '.terraform', '.git' },
+  capabilities = {
+    experimental = {
+      showReferencesCommandId = 'client.showReferences',
+    },
+  },
+  on_attach = function(_, bufnr)
+    vim.lsp.codelens.enable(true, { bufnr = bufnr })
+  end,
 }


justinmk:
cool, thanks. these little bits of knowledge are the exact purpose of having a configs-collection like nvim-lspconfig :)

GustavEikaas:
Yeah I absolutely love this knowledge base

sleepdeprecation:
this sucks and i hate that i had to adjust my config just for terraformls to ensure it respects my preference to keep codelens disabled. this should not have been made a default without respecting already existing codelens settings.

justinmk:
you shouldn't need to make a terraform-specific change, your global `vim.lsp.config('*', {...})` should be able to set this key to `vim.NIL` (I didn't test it though)

what are your existing codelens settings ?

sleepdeprecation:
my existing codelens settings were none because i wanted none. the first thing i did was add `vim.lsp.codelens.enable(false)` to my base config, which wasn't respected, so i added this to my lazy config to force it off.

```
{
  "neovim/nvim-lspconfig",
  config = function()
    vim.lsp.enable("terraformls")
    -- ADDED BELOW
    vim.lsp.config("terraformls", {
      on_attach = function() end,
      capabilities = {
        experimental = {
          showReferencesCommandId = '',
        },
      },
    })
    -- ADDED ABOVE
  end,
},
```

justinmk:
I agree configs should respect `vim.lsp.codelens.is_enabled()`. Should be a simple change to this config, would appreciate if someone can send a PR.

fwiw, you can globally configure vim.lsp defaults like this (but in this particular case, the "capability" is terraform-specific so probably doesn't feel like what you want):

```lua
    vim.lsp.config('*', {
      capabilities = {
        experimental = {
          showReferencesCommandId = '',
        },
      },
    })
```

PR #4456, opened 2026-06-19
fix(hhvm): run server in project root

antonk52:
hhvm may fail if neovim is started outside of directory containing `.hhconfig`. This PR fixes it.

Added --from for logging purposes. Useful when multiple editors are opened side by side.

### DIFF
--- a/lsp/hhvm.lua
+++ b/lsp/hhvm.lua
@@ -8,7 +8,13 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'hh_client', 'lsp' },
+  cmd = function(dispatchers, config)
+    return vim.lsp.rpc.start({ 'hh_client', 'lsp', '--from', 'neovim' }, dispatchers, {
+      -- cwd should be the same as project root, not neovim's process cwd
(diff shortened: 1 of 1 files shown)


lsp/hhvm.lua
```diff
@@ -8,7 +8,13 @@
 
 ---@type vim.lsp.Config
 return {
-  cmd = { 'hh_client', 'lsp' },
+  cmd = function(dispatchers, config)
+    return vim.lsp.rpc.start({ 'hh_client', 'lsp', '--from', 'neovim' }, dispatchers, {
+      -- cwd should be the same as project root, not neovim's process cwd
+      cwd = config.root_dir,
```
justinmk, line 14: hmm. This should probably be the default behavior of the `cmd: string[]` case 🤔

antonk52: Agreed, this would probably be a breaking change at this point though

justinmk: With Nvim 0.13 https://github.com/neovim/neovim/issues/40326 , the `cmd` CWD is derived from `root_dir` by default.

PR #4462, opened 2026-06-30
fix(svelte): support svelteserver installed in node_modules

bltavares:
Problem:
Currently the `svelte` configuration expects a globally installed `svelteserver` binary. Meanwhile, other LSP configs support binaries installed as dev dependencies on `node_modules/.bin`.
A global language server makes it harder to have project-scoped version management.

Solution:
Implement a `cmd` finder similar to `oxlint` and `oxfmt` which considers `node_modules/bin` as an additional lookup path as well as the global install

Workaround:
Meanwhile, users can define the `cmd` function on their own
`vim.lsp.config('svelte')` as needed.

### DIFF
--- a/lsp/svelte.lua
+++ b/lsp/svelte.lua
@@ -6,12 +6,21 @@
 ---
 --- `svelte-language-server` can be installed via `npm`:
 --- ```sh
---- npm install -g svelte-language-server
+--- npm install [-g] svelte-language-server
 --- ```
 
 ---@type vim.lsp.Config
 return {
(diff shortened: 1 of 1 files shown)


justinmk:
thanks. i'm going to merge this without checking that it matches the same patterns as the oxlint config, and hope that this considered all the same edge cases etc. if not i'll be sad.

bltavares:
Thanks! That is a lot of trust haha I double checked other configs using `node_modules/.bin` and they are all matching this PR, just to be safe!

PR #4463, opened 2026-07-01
fix(ruff_lsp): deprecate `ruff_lsp` in favour of `ruff`

### LINKED ISSUE #4461: [ruff-lsp]: project is deprecated in favour of `ruff server`
### Description

There are currently 2 configs in the project, the linter and formatter, `ruff` and the LSP designed around it, `ruff-lsp` but ever since v0.5.1, `ruff` has included the lsp implementation, which can be accessed via `ruff server`.

Recently, the `ruff-lsp` project was archived on GitHub, noting that it is deprecated along with a migration guide to `ruff`. The `ruff` config does mention that it is the new implementation but there is no indication in the `ruff-lsp` config that it should not be used.

In this situation, I only see 2 options:
1. Either `ruff-lsp` should be removed, or
2. There should be a warning in the `ruff-lsp` config documentation that it is deprecated and that it is suggested to migrate to `ruff`

code-IM-perfect:
The `ruff-lsp` project was archived a while back, the lsp is now built into the `ruff` project itself. This PR does the following:

1. Add a warning comment into the docs of `ruff_lsp` config
2. Add a `vim.deprecate` call to the `ruff_lsp` config (suggesting that it will be dropped in `v3.0.0` release)
3. Modify `ruff` config documentation to reflect that `ruff_lsp` isn't maintained anymore

Closes #4461

### DIFF
--- a/lsp/ruff_lsp.lua
+++ b/lsp/ruff_lsp.lua
@@ -2,7 +2,7 @@
 ---
 --- https://github.com/astral-sh/ruff-lsp
 ---
---- WARN: `ruff-lsp` is now deprecated, LSP functionality is now built into `ruff` itslef, consider setting up `ruff` instead.
+--- WARN: `ruff-lsp` is now deprecated, LSP functionality is now built into `ruff` itself, consider setting up `ruff` instead.
 ---
 --- A Language Server Protocol implementation for Ruff, an extremely fast Python linter and code transformation tool, written in Rust. It can be installed via pip.
 ---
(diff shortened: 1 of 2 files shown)


lsp/ruff_lsp.lua
```diff
@@ -2,6 +2,8 @@
 ---
 --- https://github.com/astral-sh/ruff-lsp
 ---
+--- WARN: `ruff-lsp` is now deprecated, LSP functionality is now built into `ruff` itself, consider setting up `ruff` instead.
```
justinmk: i think WARNING gets highlighted automatically in vimdoc. or NOTE.

```suggestion
--- WARNING: `ruff-lsp` is now deprecated, LSP functionality is now built into `ruff` itself, consider setting up `ruff` instead.
```

PR #4464, opened 2026-07-02
fix(rust_analyzer): error thrown if rustc is not found

cornservant:
This commit prevents the propagation of an error due to rustc not being found, which can mess up e.g. syntax highlighting via treesitter.


justinmk:
silently ignoring an error is not acceptable.

if the user enables the config but the config just silently doesn't work, the user will be confused. what is your advice in that case?

cornservant:
A vim.notify() will do. With my patch, as it is right now, it currently notifies the user that a workspace could not be found. Of course this is misleading when the actual problem is the absence of a rust installation.

image

PR #4470, opened 2026-07-11
docs: qmlls6

kindlyway:
This adds qmlls6 as an option in nvim-lspconfig, this is different from normal qmlls as some distros like arch have a seperate package with a cmd of qmlls6 while other distros like nixos have a cmd of qmlls


justinmk:
? we aren't going to add configs for every random program alias. Users can override the default cmd via:

    vim.config('qmlls', { cmd = ... }

a doc update to the existing `qmlls.lua` would be acceptable

kindlyway:
sorry i'll fix it

lsp/qmlls.lua
```diff
@@ -5,6 +5,8 @@
 --- > QML Language Server is a tool shipped with Qt that helps you write code in your favorite (LSP-supporting) editor.
 ---
 --- Source in the [QtDeclarative repository](https://code.qt.io/cgit/qt/qtdeclarative.git/)
+---
+--- Note: On some distros, the cmd of qmlls is `qmlls6`. **You can fix this by changing `cmd` to `qmlls6` in your call to `vim.lsp.config`**
```
justinmk: the fix is a 1 liner, so let's mention it. that actually is shorter than trying to describe it in english...

please confirm that the fix works for you

    vim.config('qmlls', { cmd = { 'qmlls6' } }

kindlyway: i have added this and yes the fix works for me on arch

lsp/qmlls.lua
```diff
@@ -5,6 +5,8 @@
 --- > QML Language Server is a tool shipped with Qt that helps you write code in your favorite (LSP-supporting) editor.
 ---
 --- Source in the [QtDeclarative repository](https://code.qt.io/cgit/qt/qtdeclarative.git/)
+---
+--- Note: On some distros, the cmd of qmlls is `qmlls6`. **You can fix this by adding `vim.lsp.config('qmlls', { cmd = 'qmlls6' })`**
```
justinmk:
```suggestion
--- Note: On some distros, the cmd of qmlls is `qmlls6`. You can override `cmd`:
--- ```lua
--- vim.lsp.config('qmlls', { cmd = { 'qmlls6' } })
--- ```
```

PR #4480, opened 2026-07-20
fix(ast_grep): stop reuse_client from spawning duplicate lsp client

grscheller:
Problem:
Old config set config.cmd_cwd = config.root_dir inside reuse_client itself. On the first client's initial start, reuse_client is never invoked (nothing exists yet to compare against), so cmd_cwd is never set on it. On the second start attempt, the comparison client.config.cmd_cwd == config.cmd_cwd evaluates nil == <string>, which fails, and a second client spawns.

Solution:
Change reuse_client key-value pair to

    return client.name == config.name
      and client.config.root_dir == config.root_dir
  end,

Now the first spawned client is reused. Only one lua_ls client was ever spawned. Only the `reuse_client` section looked fishy to me.

Testing:
Tested only with my Lua based Neovim configuration. I was unable to figure out what the old config was trying to do. So more testing might be needed with other languages.

### DIFF
--- a/lsp/ast_grep.lua
+++ b/lsp/ast_grep.lua
@@ -13,8 +13,8 @@ return {
   cmd = { 'ast-grep', 'lsp' },
   workspace_required = true,
   reuse_client = function(client, config)
-    config.cmd_cwd = config.root_dir
-    return client.name == config.name and client.config.cmd_cwd == config.cmd_cwd
+    return client.name == config.name
+      and client.config.root_dir == config.root_dir
   end,
   filetypes = { -- https://ast-grep.github.io/reference/languages.html
     'bash',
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks. Would you mind fixing the lint issue?

grscheller:
No problem. I fixed the lint issue.

lsp/ast_grep.lua
```diff
@@ -13,8 +13,7 @@ return {
   cmd = { 'ast-grep', 'lsp' },
   workspace_required = true,
   reuse_client = function(client, config)
-    config.cmd_cwd = config.root_dir
```
justinmk, line 16: btw, i think whatever this line was intending to do, should be unnecessary since https://github.com/neovim/neovim/pull/40331 (Nvim 0.13)

PR #4481, opened 2026-07-21
refactor!: drop `lsp/rls.lua`

Its-Just-Nans:
Updated rls description to indicate it has been replaced by rust-analyzer.


justinmk:
https://github.com/rust-lang/rls was archived in 2022 ? why don't we just delete `lsp/rls.lua` ?

Its-Just-Nans:
> why don't we just delete `lsp/rls.lua` ?

Yup that was the other solution. Should I do that?

justinmk:
yes. it's confusing if there are multiple configs for the same purpose. rls was deprecated long ago.

PR #4487, opened 2026-07-28
corrected roslyn_ls.lua call to .iter to take a table as expected

chepeniv:
neovim kept giving me an annoying error every time i edited a .cs file. reading the traceback i saw wherein the call was improperly invoked. hope this is helpful

### DIFF
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -52,7 +52,7 @@ end
 ---@param client vim.lsp.Client
 local function refresh_diagnostics(client)
   local capabilities = vim
-    .iter({client.dynamic_capabilities.capabilities.diagnosticProvider})
+    .iter(client.dynamic_capabilities.capabilities.diagnosticProvider or {})
     :map(function(cap)
       return cap.registerOptions.identifier
     end)
--- a/lsp/roslyn_ls.lua
+++ b/lsp/roslyn_ls.lua
@@ -52,7 +52,7 @@ end
 ---@param client vim.lsp.Client
 local function refresh_diagnostics(client)
   local capabilities = vim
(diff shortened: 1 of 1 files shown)


lsp/roslyn_ls.lua
```diff
@@ -52,7 +52,7 @@ end
 ---@param client vim.lsp.Client
 local function refresh_diagnostics(client)
   local capabilities = vim
-    .iter(client.dynamic_capabilities.capabilities.diagnosticProvider)
+    .iter({client.dynamic_capabilities.capabilities.diagnosticProvider})
```
justinmk: are you sure this doesn't accidentally work because `diagnosticProvider` happens to be nil for you? maybe this is better:


```suggestion
    .iter(client.dynamic_capabilities.capabilities.diagnosticProvider or {})
```

justinmk:
@kotanys

kotanys:
Sorry, I didn't expect the thing to be nil. The `or {}` definitely should be there (all client.dynamic_capabilities.capabilities fields are tables as far as I can tell)

However, I did try
1. neovim 0.12.4 and nightly (commit e3c5974)
2. roslyn ls from Mason (5.8.0-1.26266.2) and from nuget (5.9.0-1.26303.1)
3. my config and bare `vim.pack.add({"https://github.com/neovim/nvim-lspconfig.git"})  vim.lsp.enable("roslyn_ls")`

and couldn't reproduce.

lsp/roslyn_ls.lua
chepeniv: agreed! it's better with `or {}`

chepeniv:
@kotanys i might have gotten that error since i'm not on windows, and things work quite a bit differently on linux?? idk

kotanys:
@chepeniv 
Do you mind sharing your roslyn ls version?

PR #4491, opened 2026-08-03
fix(ols): reuse existing client when opening libraries

mikejmcguirk:
Problem: When going to the definition of a library function, a new ols client is created.

Solution: Detect if the file is a library. If so, re-use a pre-existing client.

----------------

Similar to the current rust-analyzer logic: https://github.com/neovim/nvim-lspconfig/blob/1c0d8f70dbc8827263eedc3cf7021ceba0f68689/lsp/rust_analyzer.lua#L65

Comments have links to odin's documentation.

### DIFF
--- a/lsp/ols.lua
+++ b/lsp/ols.lua
@@ -6,12 +6,51 @@
 
 local util = require 'lspconfig.util'
 
+---@return string?
+local function default_odin_root()
+  -- https://odin-lang.org/docs/install/#release-requirements--notes
+  local odin_root = os.getenv('ODIN_ROOT')
+  if odin_root ~= nil and odin_root ~= '' then
+    return vim.fs.normalize(odin_root)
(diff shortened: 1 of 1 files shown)


justinmk:
I assume there's a reason that `reuse_client` predicate doesn't fit? (I don't remember the details for the Rust case)

mikejmcguirk:
The reuse_client predicate doesn't have a param for the specific bufnr being attached.

It might be possible to figure out based on the config's root dir, but that creates an implicit dependency on `.odin` files being one of the root patterns the config looks for. It would also require re-implementing the default workspace detection logic, rather than leveraging it as done here.

justinmk:
> The reuse_client predicate doesn't have a param for the specific bufnr being attached.

hmm... that seems like a big miss :/

PR #4493, opened 2026-08-10
feat: tsc

### LINKED ISSUE #4467: Migrate from tsgo to TypeScript 7.0 (tsc)
### Description

There are multiple LSP commands for TypeScript, such as `vtsls` and `tsgo`. nvim-lspconfig supports `tsgo` from the NPM package [`@typescript/native-preview`](https://www.npmjs.com/package/@typescript/native-preview). This package represents the beta rewrite of the TypeScript compiler to Go. It is meant to be used by users who want to try out beta versions of TypeScript 7.0 before it is released.

This week, TypeScript 7.0 was released. Version 7.0 was rewritten to Go. It ships with LSP support (run `tsc --lsp --stdio`). Install it using the [NPM package `typescript`](https://www.npmjs.com/package/typescript). Now that TypeScript 7.0 has been released, I imagine most users will want to migrate from the beta version (`tsgo`) to the intended stable version (`tsc`).

I suggest that nvim-lspconfig support `tsc --lsp --stdio`.

rliebz:
Problem:
`tsgo` was a temporary executable available as a preview version for the Typescript 7.0 Go rewrite. Now that Typescript 7.0 has been released, the `tsgo` executable name and project are being rolled back into the main Typescript project. As of that release, the standard executable is installed from `npm` via the name `typescript`, and executable itself is named `tsc`.

Solution:
Deprecate `tsgo` as a language server name, and instead configure a more appropriately named language server that supports the permanent `tsc` executable name in addition to the `tsgo` executable name to keep backward compatibility.

The only changes are names, docs, and within the `cmd` for `tsc`, support for both the `tsc` and `tsgo` executable names, with a preference for `tsc`.

Closes #4467


justinmk:
shouldn't we also deprecate `ts_ls.lua` ? 

normally i'd say we should just update it, but since the cmd name changed maybe it's less confusing to start a new config (as done here).

sharpchen:
Although tsgo is newer, it doesn't support es5, so i think we should keep ts_ls just in case someone still need es5 support

justinmk:
ok let's mention that hint on the ts_ls docs. and let's delete the rest of its doc, or at least parts of it that duplicate the "tsc" docs.

rliebz:
Added a note on `ts_ls` about tsc/es5 support. It looks like the only duplicated section in the docs is "Monorepo support", so I deleted that section and linked to the tsc docs instead.

The `root_dir` section of each config is also the same. I left them for now, but if we wanted to reduce the duplication and treat `tsc` as the sort of standard/centralized configuration, we could turn that into a reference.

justinmk:
Thanks. Will need some feedback on when we should deprecate `lsp/ts_ls.lua`

PR #4500, opened 2026-08-15
notify the user if rustc/cargo are missing instead of throwing an error (v2)

### LINKED ISSUE #4465: rust_analyzer: lsp config throws error when rust is not installed
### Description

I am using a minimal config mentioned below and do not have rust installed.
In particular the commands `rustc` and `cargo` are not in the PATH.

Under these conditions, the following error is displayed when opening a rust file:
```
Error in BufReadPost Autokommandos für "*":
Lua callback: ...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:28: BufReadPost Autokommandos für "*"..FileType Autokommandos für "*": Vim(append)
:Lua callback: vim/_core/system:324: ENOENT: no such file or directory (cmd): 'rustc'
stack traceback:
        [C]: in function 'error'
        vim/_core/system:324: in function 'spawn'
        vim/_core/system:432: in function 'system'
        .../site/pack/core/opt/nvim-lspconfig/lsp/rust_analyzer.lua:46: in function 'default_sysroot_src'
        .../site/pack/core/opt/nvim-lspconfig/lsp/rust_analyzer.lua:74: in function 'is_library'
        .../site/pack/core/opt/nvim-lspconfig/lsp/rust_analyzer.lua:90: in function 'root_dir'
        ...ovim-unwrapped-0.12.3/share/nvim/runtime/lua/vim/lsp.lua:561: in function 'lsp_enable_callback'
        ...ovim-unwrapped-0.12.3/share/nvim/runtime/lua/vim/lsp.lua:649: in function <...ovim-unwrapped-0.12.3/share/nvim/runtime/lua/vim/lsp.lua:648>
        [C]: in function 'nvim_cmd'
        ...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:28: in function <...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:27>
        [C]: in function 'pcall'
        vim/_core/shared:1648: in function <vim/_core/shared:1628>
        [C]: in function '_with'
        ...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:27: in function <...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:10>
stack traceback:
        [C]: in function '_with'
        ...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:27: in function <...-neovim-unwrapped-0.12.3/share/nvim/runtime/filetype.lua:10>
```
Furthermore, there is no syntax highlighting.

Image

Image

Syntax highlighting works however, when commenting out the `vim.lsp.enable({ "rust_analyzer" })` line.

Image

I think this bug affects primarily devenv or nix flake users who keep per project rust installs or users that use the same vim config across devices and wish to read and edit rust source files even without the language server present.

I previously made a PR (https://github.com/neovim/nvim-lspconfig/pull/4464) that "works for me", but it would need more work to be up to standard.

## Config
```lua
-- init.lua
vim.pack.add({ "https://github.com/neovim/nvim-lspconfig" })
vim.lsp.enable({ "rust_analyzer" })
```

## Version
```
NVIM v0.12.3
Build type: Release
LuaJIT 2.1.1774638290
```

## OS
```
NixOS 26.05 (Yarara)
```

cornservant:
Fixes https://github.com/neovim/nvim-lspconfig/issues/4465

This is a revised version of https://github.com/neovim/nvim-lspconfig/pull/4464
- we define `get_rustc` and `get_cargo` functions to check if the rust compiler and the rust package manager cargo are available
- checks are performed on lsp initialization and also before spawning the respective processes
- it displays a warning  if rustc or cargo cannot be found, so the user stays informed
- it only displays the warning once to not be annoying in case the user intentionally does not have rustc/cargo installed (but rust_analyzer exists in the PATH) (thanks to @idbrii)

### DIFF
--- a/lsp/rust_analyzer.lua
+++ b/lsp/rust_analyzer.lua
@@ -44,8 +44,9 @@ local function get_rustc(opt)
     return command
   elseif opt.notify_on_error then
     vim.notify_once(
-      "[rust_analyzer] `rustc` (Rust compiler) not found. Please make sure it is installed and your PATH or RUSTC environment variables are set up correctly.",
-      vim.log.levels.WARN)
+      '[rust_analyzer] `rustc` (Rust compiler) not found. Please make sure it is installed and your PATH or RUSTC environment variables are set up correctly.',
+      vim.log.levels.WARN
+    )
   end
 end
 
@@ -57,8 +58,10 @@ local function get_cargo(opt)
   if ok then
     return command
   elseif opt.notify_on_error then
-    vim.notify_once("[rust_analyzer] `cargo` (Rust package manager) not found. Please make sure it is installed.",
-      vim.log.levels.WARN)
+    vim.notify_once(
+      '[rust_analyzer] `cargo` (Rust package manager) not found. Please make sure it is installed.',
+      vim.log.levels.WARN
+    )
   end
 end
 
--- a/lsp/rust_analyzer.lua
+++ b/lsp/rust_analyzer.lua
@@ -35,34 +35,32 @@ local function reload_workspace(bufnr)
   end
 end
 
----@param opt { notify_on_error: boolean? }
----@return string|nil
-local function get_rustc(opt)
-  local command = os.getenv 'RUSTC' or 'rustc'
-  local ok, _ = pcall(vim.system, { command })
-  if ok then
-    return command
-  elseif opt.notify_on_error then
-    vim.notify_once(
-      '[rust_analyzer] `rustc` (Rust compiler) not found. Please make sure it is installed and your PATH or RUSTC environment variables are set up correctly.',
-      vim.log.levels.WARN
-    )
-  end
-end
+---@param command string
+---@param opt { env_override: string? }?
+---@return boolean
+local function executable_exists(command, opt)
+  opt = opt or {}
+  local env_command = opt.env_override and os.getenv(opt.env_override)
+  local ok = false
+  local error_msg = ''
 
----@param opt { notify_on_error: boolean? }
----@return string|nil
-local function get_cargo(opt)
-  local command = 'cargo'
-  local ok, _ = pcall(vim.system, { command })
-  if ok then
-    return command
-  elseif opt.notify_on_error then
-    vim.notify_once(
-      '[rust_analyzer] `cargo` (Rust package manager) not found. Please make sure it is installed.',
-      vim.log.levels.WARN
+  if env_command then
+    ok = vim.fn.executable(env_command) == 1
+    error_msg = ('%s not found. Please make sure the environment variable %s points to the %s executable or unset it.'):format(
+      command,
+      opt.env_override,
+      command
     )
+  else
+    ok = vim.fn.executable(command) == 1
+    error_msg = ('%s not found. Please make sure it is installed.'):format(command)
   end
+
+  if not ok then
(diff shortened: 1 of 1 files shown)


justinmk:
please try to deduplicate the code (within the file)

lsp/rust_analyzer.lua
```diff
@@ -35,6 +35,34 @@ local function reload_workspace(bufnr)
...
+---@return boolean
+local function executable_exists(command, opt)
+  opt = opt or {}
+  local env_command = opt.env_override and os.getenv(opt.env_override)
+  local ok = false
+  local error_msg = ''
+
+  if env_command then
+    ok = vim.fn.executable(env_command) == 1
+    error_msg = ('%s not found. Please make sure the environment variable %s points to the %s executable or unset it.'):format(
```
justinmk:
```suggestion
    error_msg = ('%s not found. Ensure the environment variable %s points to the %s executable or unset it.'):format(
```

lsp/rust_analyzer.lua
```diff
@@ -35,6 +35,34 @@ local function reload_workspace(bufnr)
...
+      command,
+      opt.env_override,
+      command
+    )
+  else
+    ok = vim.fn.executable(command) == 1
+    error_msg = ('%s not found. Please make sure it is installed.'):format(command)
```
justinmk:
```suggestion
    error_msg = ('%s not found.'):format(command)
```

lsp/rust_analyzer.lua
```diff
@@ -86,6 +114,10 @@ return {
   cmd = { 'rust-analyzer' },
   filetypes = { 'rust' },
   root_dir = function(bufnr, on_dir)
+    if not executable_exists('rustc', { env_override = 'RUSTC' }) or not executable_exists('cargo') then
```
justinmk: we generally should not try to support env var overrides. it can end up being a big burden, where a simple config is expected to support all the little quirks of a build system. e.g. there might be other magic env vars like `RUSTC_EXTRA_ARGS` or whatever, who knows.

i'm not going to block this change, but just noting that this is usually a bad idea.

justinmk:
why do we need to check both rustc *and* cargo? is it actually reasonable for `cargo` to not be installed?

the simple, reliable way to address things like https://github.com/neovim/nvim-lspconfig/issues/4465 is to let `cargo` give the right error message. surely it fails if `rustc` is missing? we shouldn't need to handle env vars like RUSTC; `cargo` should be the one to figure that out.

so unless i'm missing something, it should be much simpler to just call `cargo` and `notify_once` whatever error it gives. or if it's missing, then we say so.

lsp/rust_analyzer.lua
cornservant: Currently rustc is called by the `default_sysroot_src` function, which is where this env override takes place.
For this PR I didn't want to change existing behaviours someone else might depend on or have a mismatch with respect to the error message in case someone does set the `RUSTC` env variable.


https://github.com/neovim/nvim-lspconfig/blob/10e9995a8638ad8f72898de3003ef8c686fe6a52/lsp/rust_analyzer.lua#L42-L47

PR #4502, opened 2026-08-17
fix(tsc): unable to find tsgo

BBboy01:
In the code, tsc is searched first, followed by tsgo. For large code repositories that use tsgo but have difficulty upgrading the tsc version, they may match a tsc version that does not support LSP. Therefore, it is necessary to prioritize searching for tsgo and confirm the tsc version

### DIFF
--- a/lsp/tsc.lua
+++ b/lsp/tsc.lua
@@ -73,22 +73,37 @@ return {
     },
   },
   cmd = function(dispatchers, config)
-    local cmd = 'tsc'
-    local bins = { 'tsc', 'tsgo' }
-    for _, bin in ipairs(bins) do
-      if (config or {}).root_dir then
-        local local_cmd = vim.fs.joinpath(config.root_dir, 'node_modules/.bin', bin)
-        if vim.fn.executable(local_cmd) == 1 then
-          cmd = local_cmd
-          break
+    local bins = { 'tsgo', 'tsc' }
+
+    local local_cmds = (config and config.root_dir)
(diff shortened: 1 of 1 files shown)


justinmk:
surely this doesn't require so much new code

BBboy01:
I try keep the minimal change now

nitaicharan:
Ran this branch locally. `local_cmd` is scoped to the `root_dir` block, so the second `validate_tsc_version(local_cmd)` gets nil. In a project with no `node_modules/.bin/tsc` but `tsc` on `$PATH`:

```
vim/_core/system:318: bad argument #1 to 'spawn' (string expected, got nil)
```

and nothing attaches.

Two smaller things. `cmd:match('tsc$')` checks `cmd`, which is still the `'tsc'` default at that point, not `bin` — so it's true on every iteration and the tsgo/tsc distinction doesn't hold. And `break` on a failed version check leaves the loop entirely instead of trying the next candidate, so `cmd` stays plain `'tsc'`; if that one is also pre-7.0 you're back where you started.

I hit the same bug and opened #4505 — it probes in `root_dir` instead, caching the result per project root, so the check runs once per root and `cmd` can't end up with a binary `root_dir` didn't accept. Happy to close mine if you'd rather carry the fix here, just didn't want two PRs on the same file with no pointer between them.

justinmk:
If https://github.com/neovim/nvim-lspconfig/pull/4505 is more robust let's go with that.

PR #4505, opened 2026-08-17
fix(tsc): skip binaries that do not support --lsp

### LINKED ISSUE #4504: `tsc` config picks a binary that does not support `--lsp`, so the server never starts
### Description

Nvim version: `NVIM v0.12.4` (`LuaJIT 2.1.1785763465`, build `RelWithDebInfo`)

Enabled with the stock config shipped in this repo:

```lua
vim.lsp.enable('tsc')
```

`lsp/tsc.lua` resolves its binary in `cmd` using `vim.fn.executable()` only:

```lua
  local cmd = 'tsc'
  local bins = { 'tsc', 'tsgo' }
  for _, bin in ipairs(bins) do
    if (config or {}).root_dir then
      local local_cmd = vim.fs.joinpath(config.root_dir, 'node_modules/.bin', bin)
      if vim.fn.executable(local_cmd) == 1 then
        cmd = local_cmd
        break
      end
    end
    if vim.fn.executable(bin) == 1 then
      cmd = bin
      break
    end
  end
  return vim.lsp.rpc.start({ cmd, '--lsp', '--stdio' }, dispatchers)
end,
```

The `--lsp` flag only exists in the native compiler, TypeScript 7.0 and newer. Any project
that still depends on TypeScript 5.x or 6.x has a `node_modules/.bin/tsc` that is executable
but has no `--lsp` flag at all. Because the project-local path is preferred and only tested
with `executable()`, that binary wins the lookup, gets spawned with `--lsp --stdio`, and
immediately errors out. The client never attaches, and a TypeScript 7.0+ binary that *is*
present on `$PATH` (e.g. installed through Mason) is never considered.

Running the exact command the config builds shows what happens:

```console
$ ./node_modules/.bin/tsc --version
Version 5.9.3

$ ./node_modules/.bin/tsc --lsp --stdio
error TS5023: Unknown compiler option '--lsp'.
error TS5023: Unknown compiler option '--stdio'.
```

The process exits instead of speaking LSP over stdio, so `vim.lsp.rpc.start()` gets a dead
peer. There is no hint in `:checkhealth vim.lsp` that the wrong binary was chosen — the
config looks enabled and the buffer simply has no client.

## Steps to reproduce

1. Have a TypeScript 7.0+ binary on `$PATH` (`tsc --version` → `Version 7.0.2`).
2. Open a project whose `package.json` pins TypeScript 5.x or 6.x, so that
   `node_modules/.bin/tsc` is an older non-native compiler.
3. `nvim src/index.ts`
4. `:lua =vim.lsp.get_clients({ bufnr = 0 })`

## Expected behavior

The config picks a binary that actually supports `--lsp`. If the project-local one is too
old, resolution should fall through to the next candidate (`$PATH` `tsc`, then `tsgo`),
keeping the existing preference order. If no candidate qualifies, the server should not
attach and the user should get a warning explaining why, rather than a silent failure.

## Actual behavior

The project-local pre-7.0 `tsc` is spawned with `--lsp --stdio`, exits with
`error TS5023: Unknown compiler option '--lsp'`, and no client attaches to the buffer.
A working TypeScript 7.0+ binary on `$PATH` is ignored.

## Notes

- The version probe has to happen before `cmd`, otherwise `root_dir` can accept a root for
  which no usable binary exists.
- The probe (`tsc --version`) should be cached per project root so it does not run on every
  client start.

nitaicharan:
Fixes #4504.

`lsp/tsc.lua` picked its binary in `cmd` using `vim.fn.executable()` alone, preferring
`node_modules/.bin`. But `--lsp` only exists in the native compiler, TypeScript 7.0 and
newer, so in any project still pinned to TypeScript 5.x/6.x the project-local binary won the
lookup, got spawned with a flag it does not understand, and died — even when a 7.0+ `tsc`
was sitting on `$PATH`.

Resolution moves to `root_dir`, which probes `--version` and takes the first candidate
reporting major >= 7. The existing preference order is preserved (project-local before
`$PATH`, `tsc` before `tsgo`). The result is cached per project root and `cmd` becomes a
lookup into that cache, so the probe runs once per root rather than on every client start,
and `cmd` cannot disagree with the binary `root_dir` already accepted. When no candidate
qualifies the server does not attach and warns instead of failing to spawn.

### Verification

Nvim v0.12.4, in a project with `node_modules/.bin/tsc` at `Version 5.9.3` and
`tsc` `Version 7.0.2` on `$PATH`.

Before:

```
NOTIFY: Client tsc quit with exit code 1 and signal 0.
clients=0
```

After:

```
clients=1
ATTACHED tsc root=/home/dan/Projects/Personal-Angular
```

`stylua --check .` and `.github/ci/lint.sh` pass.

### DIFF
--- a/lsp/tsc.lua
+++ b/lsp/tsc.lua
@@ -52,6 +52,7 @@
 
 local bin_cache = {} ---@type table<string, string>
 
+--- Checks if the given tsc/tsgo cmd supports the "--lsp" arg.
 ---@param bin string
 ---@return boolean
 local function supports_lsp(bin)
@@ -60,9 +61,9 @@ local function supports_lsp(bin)
   end
 
   local out = vim.system({ bin, '--version' }, { text = true }):wait()
-  local major = tonumber((out.stdout or ''):match('(%d+)%.'))
+  local version = vim.version.parse(out.stdout or '')
 
-  return out.code == 0 and major ~= nil and major >= 7
+  return out.code == 0 and version ~= nil and version.major >= 7
 end
 
 ---@type vim.lsp.Config
--- a/lsp/tsc.lua
+++ b/lsp/tsc.lua
@@ -8,6 +8,10 @@
 ---
 --- `tsc` can be installed via npm `npm install typescript`.
 ---
+--- The language server (`--lsp`) is only available in the native compiler, TypeScript 7.0
+--- and newer. An older binary in `node_modules/.bin` is skipped in favour of one on `$PATH`
+--- that does support it; if no candidate qualifies, the server does not attach.
+---
 --- ### Monorepo support
 ---
 --- `tsc` supports monorepos by default. It will automatically find the `tsconfig.json` or `jsconfig.json` corresponding to the package you are working on.
@@ -46,6 +50,21 @@
 --- If DENO ROOT is found, and it's longer than or equal to PROJECT ROOT, then this is a Deno file, and we abort.
 --- Otherwise, attach at PROJECT ROOT, or the cwd if not found.
 
+local bin_cache = {} ---@type table<string, string>
+
+---@param bin string
+---@return boolean
+local function supports_lsp(bin)
+  if vim.fn.executable(bin) ~= 1 then
+    return false
+  end
+
+  local out = vim.system({ bin, '--version' }, { text = true }):wait()
+  local major = tonumber((out.stdout or ''):match('(%d+)%.'))
+
+  return out.code == 0 and major ~= nil and major >= 7
+end
+
 ---@type vim.lsp.Config
 return {
   settings = {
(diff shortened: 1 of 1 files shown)


lsp/tsc.lua
```diff
@@ -46,6 +50,21 @@
...
+
+---@param bin string
+---@return boolean
+local function supports_lsp(bin)
+  if vim.fn.executable(bin) ~= 1 then
+    return false
+  end
+
+  local out = vim.system({ bin, '--version' }, { text = true }):wait()
+  local major = tonumber((out.stdout or ''):match('(%d+)%.'))
```
justinmk: `vim.version.parse()` ... (and it has a `strict=false` option)

lsp/tsc.lua
```diff
@@ -46,6 +50,21 @@
 --- If DENO ROOT is found, and it's longer than or equal to PROJECT ROOT, then this is a Deno file, and we abort.
 --- Otherwise, attach at PROJECT ROOT, or the cwd if not found.
 
+local bin_cache = {} ---@type table<string, string>
+
+---@param bin string
```
justinmk, line 56:
```suggestion
--- Checks if the given tsc/tsgo cmd supports the "--lsp" arg.
---@param bin string
```

nitaicharan: Applied in f2fd30e8. The snippet in this thread still shows the old version — that's just the pinned snapshot — but the doc comment is on the file at head.

nitaicharan:
@justinmk both comments are addressed — doc comment added, and the version check uses `vim.version.parse` now. Anything else you want changed, or is this good to merge?

PR #4510, opened 2026-08-21
feat: ms_terraform_lsp

tobiassen:
Adds a config for <https://github.com/Azure/ms-terraform-lsp>, Microsoft's language server for the `azapi`, `azurerm` and `msgraph` Terraform providers.

The gap it fills is `azapi_resource`: the provider schema exposes `body` as `dynamic`, so `terraform-ls` has nothing to complete inside it. This server has the Azure resource schemas embedded, so it completes property names and values there, with hover docs and diagnostics. It only covers Microsoft providers, so it runs alongside `terraformls`.

The server repo has few stars. It may be because its only documented client is Microsoft's Azure Terraform VS Code extension, which ships the binary inside the extension package, so users have little reason to visit the repo. The `azapi` provider it targets has 250M+ Terraform Registry downloads.

Tested locally with both servers attached: type and API version completion, completion and hover inside `body`, `diagnostics` on unknown properties.

### DIFF
--- a/lsp/ms_terraform_lsp.lua
+++ b/lsp/ms_terraform_lsp.lua
@@ -1,0 +1,20 @@
+---@brief
+---
+--- https://github.com/Azure/ms-terraform-lsp
+---
+--- Microsoft Terraform Providers Language Server. Provides completion, hover
+--- documentation and schema validation for the `azapi`, `azurerm` and `msgraph`
+--- providers, and for Azure Verified Modules.
+---
+--- Only covers Microsoft providers, so it is intended to run alongside
+--- [terraformls](#terraformls).
+---
+--- Download a released binary from
+--- https://github.com/Azure/ms-terraform-lsp/releases.
+
+---@type vim.lsp.Config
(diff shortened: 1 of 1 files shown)


justinmk:
there is https://github.com/neovim/nvim-lspconfig/blob/master/lsp/azure_pipelines_ls.lua so i wonder if naming this `azure_ms_terraform_lsp` would make sense, to help discoverability

the CLI name doesn't matter too much, they often change anyway...

tobiassen:
I agree it could help with discoverability, but as this server also covers the `msgraph` provider, which is Microsoft Graph rather than Azure, an `azure_` prefix reads a little narrower than what it does.  
I'm also planning on a Mason package, and Mason names packages after upstream, so it would be listed there as `ms-terraform-lsp`.  
It is hosted under `Azure` github org, so I guess it has that going in favour of the prefix as well as your discoverability argument.

Thanks for the feedback! Your call, I will push the rename if you still prefer the prefix.

PR #4511, opened 2026-08-21
fix(ols): add error handling for 'odin root' command

Silzinc:
Handle errors when executing the 'odin root' command. Currently, if odin is not in the path, the configuration may provoke an error when opening an odin file.


justinmk:
> if odin is not in the path, the configuration may provoke an error when opening an odin file.

then check `executable()` , dont just throw `pcall` on something.

also there is a failing CI

Silzinc:
My bad, that will teach me to edit online. Should be good now. (I force pushed to satisfy the commits convention)

PR #4512, opened 2026-08-24
feat: rust_glancer

h-michael:
Adds a config for [rust-glancer](https://github.com/rust-glancer/rust-glancer), a Rust language server that trades completeness for low memory usage and near-instant editor restarts. The repository has 348 stars.

A few non-obvious things about the setup:

- The project currently only distributes `.vsix` packages for VS Code, so the `rust-glancer` binary has to be built from source (`cargo build --release -p rust-glancer`) to use it with Nvim. `rust-src` is also a hard requirement (`rustup component add rust-src`).
- The server only reads configuration from `initializationOptions` sent at startup; there is no `workspace/configuration` support. Settings go through `init_options`, not `settings`.
- Diagnostics (`cargo check`) are disabled by default.
- The default nearest-`Cargo.toml` `root_markers` resolution breaks for a crate that is a member of a Cargo workspace: it resolves to the member's own directory instead of the workspace root. rust-glancer rejects a Cargo workspace root that lies outside the LSP workspace folder it was given, so every request after `initialize` fails with `the open document could not be associated with an analysis engine`, with no indication of the cause unless `RUST_GLANCER_LOG=debug` is set. The custom `root_dir` walks upward from the buffer and picks the outermost `Cargo.toml` that declares `[workspace]`, falling back to the nearest `Cargo.toml` for standalone crates.

### DIFF
--- a/lsp/rust_glancer.lua
+++ b/lsp/rust_glancer.lua
@@ -1,0 +1,66 @@
+---@brief
+---
+--- https://github.com/rust-glancer/rust-glancer
+---
+--- `rust-glancer`, an incomplete-by-design Rust language server optimized for
+--- low memory usage and near-instant editor restarts.
+---
+--- VS Code is currently the only officially supported editor; the project only
+--- publishes `.vsix` packages, so the `rust-glancer` binary must be built from
+--- source for use with Nvim:
+--- ```sh
+--- git clone https://github.com/rust-glancer/rust-glancer
+--- cd rust-glancer
+--- cargo build --release -p rust-glancer
+--- # add target/release/ to $PATH, or point `cmd` at the built binary
+--- ```
+--- `rust-src` is required regardless of editor:
+--- ```sh
+--- rustup component add rust-src
+--- ```
+---
+--- Diagnostics (`cargo check`) are disabled by default; enable them via
+--- `diagnostics.onStartup` / `diagnostics.onSave` below.
+---
+--- The server reads its configuration only from the LSP `initializationOptions`
+--- sent on startup (there is no `workspace/configuration` support), so options
+--- must be set via `init_options`, not `settings`:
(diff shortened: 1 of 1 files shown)


lsp/rust_glancer.lua
```diff
@@ -0,0 +1,68 @@
...
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local manifests = vim.fs.find('Cargo.toml', { path = fname, upward = true, limit = math.huge })
+
+    for i = #manifests, 1, -1 do
+      local manifest = manifests[i]
+      local file = io.open(manifest, 'r')
+      if file then
+        local content = file:read('*a')
+        file:close()
```
justinmk: can be simplified by using `vim.fn.readfile()`

lsp/rust_glancer.lua
```diff
@@ -0,0 +1,68 @@
...
+  -- workspace folder it was given (see `is_workspace_root_allowed` in its
+  -- source). The nearest-`Cargo.toml` default would land on a member crate's
+  -- own manifest instead of the workspace root above it, so walk all the way
+  -- up and prefer the outermost `Cargo.toml` that declares `[workspace]`.
+  root_dir = function(bufnr, on_dir)
+    local fname = vim.api.nvim_buf_get_name(bufnr)
+    local manifests = vim.fs.find('Cargo.toml', { path = fname, upward = true, limit = math.huge })
+
+    for i = #manifests, 1, -1 do
+      local manifest = manifests[i]
```
justinmk: this is an interesting way to iterate a list... :P why not use ipairs()?

justinmk: oh, i missed the "nearest-to-farthest" part

PR #4516, opened 2026-08-30
fix(esbonio): use version 2 CLI

viktorashi:
Problem:
The Python module command bypasses standalone uv and pipx installations, and the configuration examples target Esbonio 0.16.

Solution:
Launch the public Esbonio 2 CLI, recognize conf.py as a project root, and document the current project configuration model.

i guess everybody who respects themselves already got `uv`?

also new esbonio starts with `esbonio serve`


viktorashi:
thnx @justinmk !

PR #4517, opened 2026-08-31
feat: Update TypeScript version in installation instructions

Its-Just-Nans:
Looking at the README from https://www.npmjs.com/package/typescript-language-server

typescript-language-server needs for typescript@6
tsc needs typescript@7

### DIFF
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -33,6 +33,9 @@
 --- }
 --- ```
 ---
+--- If you also have a `deno.json` or a `deno.jsonc` in your project, the `ts_ls` may not start (depending where the
+--- deno.json is)
+---
 --- Use the `:LspTypescriptSourceAction` command to see "whole file" ("source") code-actions such as:
 --- - organize imports
 --- - remove unused code
--- a/lsp/ts_ls.lua
+++ b/lsp/ts_ls.lua
@@ -9,7 +9,8 @@
 ---
 --- `typescript-language-server` depends on `typescript`. Both packages can be installed via `npm`:
 --- ```sh
---- npm install -g typescript typescript-language-server
+--- npm install -g typescript@<version> typescript-language-server
+--- # if you are using `ts_ls` you will need to install a `typescript` version < 7 for now
 --- ```
 ---
 --- To configure typescript language server, add a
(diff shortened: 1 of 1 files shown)


sharpchen:
I think that's not required. ~~The instruction in readme sync with the typescript version it depends(if you check the version history e.g. 5.3.0, the version in the instruction is just 5.3.0)~~
People can install whatever version of typescript lower than 7.0 to get `ts_ls` work.
It'd be better to have an warning or tip in the comment instead of using a fixed version number.

Its-Just-Nans:
> I think that's not required. ~The instruction in readme sync with the typescript version it depends(if you check the version history e.g. 5.3.0, the version in the instruction is just 5.3.0)~ People can install whatever version of typescript lower than 7.0 to get `ts_ls` work. It'd be better to have an warning or tip in the comment instead of using a fixed version number.

Yes but since two month ago, when you do `npm -g install typescript` it will install the latest which is https://www.npmjs.com/package/typescript/v/7.0.2

Should I edit the MR to add a comment or close it?

thanks

sharpchen:
I'd prefer `npm -g install typescript@<version>` and explain why a version lower than 7.0 is required in the comment

Its-Just-Nans:
> I'd prefer `npm -g install typescript@<version>` and explain why a version lower than 7.0 is required in the comment

Okay i've edited the file

lsp/ts_ls.lua
```diff
@@ -9,7 +9,8 @@
 ---
 --- `typescript-language-server` depends on `typescript`. Both packages can be installed via `npm`:
 --- ```sh
---- npm install -g typescript typescript-language-server
+--- npm install -g typescript@<version> typescript-language-server
+--- # if you are using `ts_ls` you will need to install a `typescript` version < 7 for now
```
sharpchen: What does "for now" mean? I think this would be legacy forever?

I suggest rephrasing it to "ts_ls doesn't support TypeScript 7 and higher versions, hence a `<version>` lower than 7 is required. If you're working with TypeScript with higher version, use tsc.lua instead."

justinmk:
```suggestion
--- # if you are using `ts_ls` you will need to install a `typescript` version < 7 for now
--- npm install -g typescript@<version> typescript-language-server
```

PR #4518, opened 2026-09-01
Update ts_ls.lua

quantum-incongruity:
Fix https://github.com/neovim/neovim/discussions/41580#discussioncomment-18224327

Could be add a way to force the `ts_ls` to start even if there is a `deno.json` ? (I'm using a project which is both deno/node/bun)


quantum-incongruity:
So we cannot add a configuration option to do that ?

PR #4521, opened 2026-09-01
perf: defer requiring util until legacy branch

chernetskyi:
Problem:
`require('lspconfig.util')` runs unconditionally on the top of the file, while it is not required for newer versions of Neovim.

Solution:
Run `require('lspconfig.util')` only after the branch for newer versions exits.

### DIFF
--- a/plugin/lspconfig.lua
+++ b/plugin/lspconfig.lua
@@ -12,7 +12,7 @@ if vim.fn.has('nvim-0.11') == 0 then
 end
 
 local api, lsp = vim.api, vim.lsp
-local util = require('lspconfig.util')
+local util
 
 local completion_sort = function(items)
   table.sort(items)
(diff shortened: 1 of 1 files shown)


plugin/lspconfig.lua
justinmk, line 25: is `util` here going to be available after this change?

the flow seems confusing... e.g., if `lsp_complete_configured_servers` isn't used on some version of Nvim that also doesn't need `util`, then why did we define `lsp_complete_configured_servers` at all?

chernetskyi: > is util here going to be available after this change?

Yes, it should be. `util`'s value is resolved when closure runs, not when it is captured.

> if lsp_complete_configured_servers isn't used on some version of Nvim that also doesn't need util, then why did we define lsp_complete_configured_servers at all?

To be honest, I wanted to get rid of an extra file read, but you're right. `lsp_complete_configured_servers` and some other definitions can be moved lower to save more milliseconds on the newer Neovim versions. Should I move those too, or did I misunderstand you and you're suggesting reverting for clarity?

justinmk: i think it would be clearer if all relevant code is moved within/below the same condition.

chernetskyi: https://github.com/neovim/nvim-lspconfig/pull/4523

PR #4533, opened 2026-09-14
fix: use method syntax for client is_stopped

skarzi:
It's required to get rid of the following deprecation warning:

> client.is_stopped is deprecated. Run ":checkhealth vim.deprecated" for more information
> vim.deprecated:                                                           1 ⚠️
> 
>  ~
> - ⚠️ WARNING client.is_stopped is deprecated. Feature will be removed in Nvim 0.13
>   - ADVICE:
>     - use client:is_stopped instead.


skarzi:
Closing because this compatibility path is not used by supported Neovim versions.
nvim>=0.12 uses the native `:lsp` command, which uses proper methods.

